PR #100, opened 2017-03-03
Filter Node Definitions

### LINKED ISSUE #93: Add filter to "get_node_definitions"
Node queries rely on the "get_node_definition" method in DataSource.php, which resolves an object
based on it's global ID, then returns the appropriate GraphQL Type based on the object that is returned. 

We'll need proper filter(s) in place here so Themes/Plugins can register their own Types that can properly resolve from a node query.

jasonbahl:
- This adds filters to the $node and $type in the `get_node_definitions` method, allowing externally defined "node" types to hook in so they can properly resolve for node queries

Resolves #93

### DIFF
--- a/src/Data/DataSource.php
+++ b/src/Data/DataSource.php
@@ -411,6 +411,10 @@ class DataSource {
 								/**
 								 * Add a filter to allow externally registered node types to resolve based on
 								 * the id_components
+								 *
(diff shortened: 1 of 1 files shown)


jasonbahl:
@CodeProKid pushed updates addressing your comments

CodeProKid:
Looks good.

PR #101, opened 2017-03-03
Bug: RootQueryIssue

jasonbahl:
Fixes a bug where the rootObject queries for postObjects and termObjects were resolving to the same type for all post_types and taxonomies

fixes: #102

### DIFF
--- a/src/Type/PostObject/PostObjectQuery.php
+++ b/src/Type/PostObject/PostObjectQuery.php
@@ -33,7 +33,7 @@ class PostObjectQuery {
 			self::$root_query = [];
 		}
 
-		if ( null === self::$root_query[ $post_type_object->name ] ) {
+		if ( ! empty( $post_type_object->name ) && null === self::$root_query[ $post_type_object->name ] ) {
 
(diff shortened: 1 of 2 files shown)


src/Type/PostObject/PostObjectQuery.php
```diff
@@ -46,7 +50,7 @@ public static function root_query( $post_type_object ) {
 
 		}
 
-		return self::$root_query;
+		return ! empty( self::$root_query[ $post_type_object->name ] ) ? self::$root_query[ $post_type_object->name ] : null;
```
CodeProKid: Will returning `null` here make anything go haywire?

CodeProKid:
Would be nice to get tests for this. Not sure if we need them to merge this, but we at least need a specific issue to write them. I'll leave it up to you.

Also, see the one inline comment above.

jasonbahl:
@CodeProKid pushed an update to this, and also created an Issue for Schema Unit Tests: https://github.com/wp-graphql/wp-graphql/issues/104

PR #103, opened 2017-03-03
WPGraphQL#16 - Attachment Queries

jasonbahl:
- This adjusts the shape of the attachment (mediaItem) Type definition providing fields relevant to attachments/mediaItems, such as sizes, caption, altText, etc. . .
- Also adjusts the post object query test to remove the mimeType query as the field was moved to JUST the mediaItem type

======

Not sure I love how this is organized, but was having trouble coming up with the best organization for it.

since mediaItem type should take advantage of many things the postObjectType has, it makes sense to not have a completely different type class for mediaItem, but it also feels weird to have all the field definitions in the postObjectType file. . .

Maybe create a /Types/MediaItem directory, and include the fields somewhere in there? I dunno. . .feedback appreciated on that. . .thanks!

### DIFF
--- a/src/Type/PostObject/PostObjectType.php
+++ b/src/Type/PostObject/PostObjectType.php
@@ -4,7 +4,6 @@ namespace WPGraphQL\Type\PostObject;
 use GraphQL\Type\Definition\ObjectType;
 use GraphQL\Type\Definition\ResolveInfo;
 use GraphQLRelay\Relay;
-use WPGraphQL\Connections;
 use WPGraphQL\Data\DataSource;
 use WPGraphQL\Type\Comment\Connection\CommentConnectionDefinition;
 use WPGraphQL\Type\TermObject\Connection\TermObjectConnectionDefinition;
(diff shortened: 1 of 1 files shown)


CodeProKid:
Yeah, I'm not loving it in the postObjectType file either... I think this should have it's own directory/file, and just extend the postObject from there? Trying to think... once we start on mutations there's probably some other stuff that will be specific to the MediaType also? If that's the case, then I definitely think it should have it's own directory so we can package all of that stuff together. We can talk about it more tomorrow though. Looks good overall.

jasonbahl:
@CodeProKid ok, ya. . .so I think we need to figure out a good pattern for having postTypes overwrite the queries that get setup for each post_type automatically. 

Right now, a custom post_type can simply add the following properties in the post_type registration:

```
$wp_post_types['attachment']->show_in_graphql = true;
$wp_post_types['attachment']->graphql_single_name = 'mediaItem';
$wp_post_types['attachment']->graphql_plural_name = 'mediaItems';
```

And that will setup a root singular query, a root plural (connection) query, and will also automatically add connections to the taxonomies that have registered connection to the postType. 

If we have custom post types register their own type class, and either extend the PostObjectType or WPObjectType, we'll need to figure out a good way for the custom Type class and potentially other custom classes to overwrite the PostObjectType / PostObjectConnection to take advantage of the dynamic connections, etc that are setup automatically. 

In my early versions, I had another property in the post_type global to set a custom "Type" class, so that might be worth exploring again.

jasonbahl:
For reference to where some custom post type stuff is automatically provided for post_types that just register a `single_name`, `plural_name` and `show_in_graphql`: 

https://github.com/wp-graphql/wp-graphql/search?utf8=%E2%9C%93&q=allowed_post_types

PR #119, opened 2017-05-20
Add a section in docs about available extensions

shadyvb:
This adds some info about available extensions.


BE-Webdesign:
Awesome!  Hope to see more of these.

PR #122, opened 2017-05-22
Add basic avatar test.

### LINKED ISSUE #60: Unit Tests for AvatarType

BE-Webdesign:
Closes #60 and #124. Adds a basic avatar test.

### DIFF
--- a/tests/test-avatar-object-queries.php
+++ b/tests/test-avatar-object-queries.php
@@ -1,0 +1,243 @@
+<?php
+/**
+ * WPGraphQL Test Avatar Object Queries
+ * This tests avatar queries, which are only accessible via User edges.
+ * @package WPGraphQL
+ * @since 0.0.5
+ */
+
+class WP_GraphQL_Test_Avatar_Object_Queries extends WP_UnitTestCase {
+	public $admin;
+
+	/**
+	 * This function is run before each method
+	 * @since 0.0.5
+	 */
+	public function setUp() {
+		parent::setUp();
+
+		$this->admin = $this->factory->user->create( [
+			'role'       => 'admin',
+			'user_email' => 'test@test.com'
+		] );
+	}
+
+	/**
+	 * Runs after each method.
+	 * @since 0.0.5
+	 */
+	public function tearDown() {
+		parent::tearDown();
+	}
+
+	/**
+	 * testPostQuery
+	 *
+	 * This tests creating a single post with data and retrieving said post via a GraphQL query
+	 *
+	 * @since 0.0.5
+	 */
+	public function testAvatarQuery() {
+		/**
+		 * Create the global ID based on the post_type and the created $id
+		 */
+		$global_id = \GraphQLRelay\Relay::toGlobalId( 'user', $this->admin );
+
+		// Override avatar url to match $this->avatar_test_url()
+		add_filter( 'get_avatar_url', array( $this, 'avatar_test_url' ), 10, 1 );
+
+		/**
+		 * Create the query string to pass to the $query
+		 */
+		$query = "
+		query {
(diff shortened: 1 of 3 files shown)


BE-Webdesign:
This is a start to fully covering #60.

tests/test-avatar-object-queries.php
```diff
@@ -0,0 +1,95 @@
...
+		 */
+		$expected = [
+			'data' => [
+				'user' => [
+					'avatar' => [
+						'default'      => 'mm',
+						'extraAttr'    => null,
+						'forceDefault' => null,
+						'foundAvatar'  => null,
+						'height'       => 96,
```
jasonbahl: does this mean there's an issue with the size in the query? Looks like you requested a size of 48, but are expecting it to return 96 (which is the default). . .so is the `48` never getting passed through to the method?

BE-Webdesign: Yes.

jasonbahl: ok, we should probably change the test to expect 48 then, and fix the actual issue instead of having the test expect something incorrect, eh?

BE-Webdesign: We will also want to figure out a way to not worry about the url because it changes on gravatar.  I will probably just hook into a filter and spoof the data.

BE-Webdesign: Going to be revising these to make better sense, wanted to see what the CI would do.

BE-Webdesign: > ok, we should probably change the test to expect 48 then, and fix the actual issue instead of having the test expect something incorrect, eh?

Yup 😄, just testing the CI, and coveralls, to get 100% coverage we need to add a case without the avatar being found.  Going to muck around with the filters for avatar data.

jasonbahl: ya, or maybe just assert that the URL is not empty?

We could do 2 separate tests, one for the known values and an `assertEquals` on the full query result, and another test that just returns the url with an `assertNotEmpty`, or `assertTrue( strpos( $url, 'http://0.gravatar.com/avatar/' ) )` on just `$actual['data']['user']['avatar']['url']` or something to that tune?

jasonbahl: Coo. I'll hold off on reviewing for now then. haha

BE-Webdesign: I'm going to add a review label and an in progress label, so that you can know when to check something over.

BE-Webdesign: > ya, or maybe just assert that the URL is not empty?

I think we will want to use filters and verify that the expected data matches, that way we know WP is grabbing the right information.  If we check for non empty on url, as much as that works, it feels a little bit indirect compared to using a filter, where we can be explicit about the expected value.  By using filters  we can condense it all into one clear test case.  Then for handling a not found avatar that would be another test case etc.

BE-Webdesign:
@jasonbahl Okay these are ready to go, let me know if you want me to open a PR that fixes the tests to pass for this as well, or if that will go into this PR, or if you are going to do it.  Before I pushed I verified that when corrected these tests pass and the Avatar works properly.  I left the fixes out and just pushed the tests ,because I was unsure how you wanted to handle that flow.

jasonbahl:
I think for now if you catch existing bugs as you write tests feel free to include the fix in the PR with the test, or create a new issue with details on what the bug is. either works for me

BE-Webdesign:
> I think for now if you catch existing bugs as you write tests feel free to include the fix in the PR with the test, or create a new issue with details on what the bug is. either works for me

Okay, I will patch this and rebase.

jasonbahl:
👍

PR #140, opened 2017-05-23
PostObjectMutations

jasonbahl:
This introduces postObjectMutations, and includes tests for said mutations.

Also includes lots of formatting changes, which I know I should do in separate PRs, but whatever, deal with it 😜  (I'll really try to stop doing that though, because I know it's annoying to try and code review when there's a bunch of misc formatting changes included)

====

Here's an example of trying to create a page but not being allowed to because I don't have permissions. 

So, I add my Authorization headers to the request (using Basic auth in this example), then I repeat the mutation and you can see data created!

Then I do a follow-up `updatePage` mutation to change the title, and you see the same page is returned, just with a new title. 

pagemutations


.travis.yml
```diff
@@ -33,7 +33,7 @@ matrix:
 # Composer package installation
 install:
   # Install composer packages, will also trigger dump-autoload
-  - composer install --no-interaction
```
jasonbahl: Travis tests were bombing because of this

src/AppContext.php
```diff
@@ -34,4 +34,20 @@ class AppContext {
...
+	 * @access public
+	 */
+	public $config;
+
+	/**
+	 * AppContext constructor.
+	 */
+	public function __construct() {
+
+		$this->config = apply_filters( 'graphql_app_context_config', $this->config );
```
jasonbahl: This will allow the AppContext to have data added to it and accessed throughout the resolve tree. . .not used by anything (yet) but will be useful for WPGraphQL Insights / Apollo Optics support

src/Data/Config.php
```diff
@@ -82,9 +82,7 @@ public function graphql_wp_query_cursor_pagination_support( $where, \WP_Query $q
 				} else {
```
jasonbahl: just formatting changes on this file

src/Data/ConnectionResolver.php
```diff
@@ -32,7 +32,7 @@ public static function resolve( $source, $args, AppContext $context, ResolveInfo
 		$query_args  = static::get_query_args( $source, $args, $context, $info );
 		$query       = static::get_query( $query_args );
 		$array_slice = self::get_array_slice( $query, $args );
-		$connection  = static::get_connection( $array_slice, $args, $query );
+		$connection  = static::get_connection( $query, $array_slice, $source, $args, $context, $info );
```
jasonbahl: Adding more params so context can be passed throughout easier

src/Data/ConnectionResolver.php
```diff
@@ -82,32 +85,6 @@ public static function get_connection( array $array, array $args, $query ) {
...
-	 *
-	 * If First is set, select the amount of records asked for from the front of the data set by using the
-	 * specified order.
-	 *
-	 * If Last is set, select the amount of records asked for from the back of the data set by reversing the
-	 * specified order.
-	 *
-	 *
-	 */
-	public static function get_query_order( $args, $query_args ) {
```
jasonbahl: removed because this never gets called now

src/Type/Comment/Connection/CommentConnectionArgs.php
```diff
@@ -189,9 +190,8 @@ private static function fields() {
 	 */
 	private static function comments_orderby_enum() {
 
-		if ( null === self::$comments_orderby_enum ) {
-
-			self::$comments_orderby_enum = new EnumType([
+		if ( null === self::$comments_orderby_enum ) :
+			self::$comments_orderby_enum = new WPEnumType([
```
CodeProKid: Are we using just `EnumType` other places? Should probably get an issue going to convert all `EnumType`'s to `WPEnumType`

src/Type/Comment/Connection/CommentConnectionDefinition.php
```diff
@@ -35,7 +35,7 @@ class CommentConnectionDefinition {
 	 */
 	public static function connection() {
 
-		if ( null === self::$connection ) {
+		if ( null === self::$connection ) :
```
CodeProKid: Would be good to get a developers handbook going for rules on stuff like this. When to use `:` vs `{`. Or at least an issue so we can start documenting these things.

src/Type/Comment/Connection/CommentConnectionArgs.php
jasonbahl: @CodeProKid  I tried to sweep and convert them all to WPEnumType. . .but I can double check to make sure I got em all

src/Type/Comment/Connection/CommentConnectionDefinition.php
jasonbahl: ya, I go back and forth on my usage sometimes. . .I usually prefer brackets, but use colon for longer blocks as it's more explicit where it's ending. . .but ya, we can definitely decide on a standard on when to go to colon vs. brackets. . .like a certain size of code block being wrapped or something

src/Type/PostObject/Mutation/PostObjectDelete.php
```diff
@@ -0,0 +1,153 @@
...
+	 * Defines the delete mutation for PostTypeObjects
+	 *
+	 * @param \WP_Post_Type $post_type_object
+	 *
+	 * @return array|mixed
+	 */
+	public static function mutate( \WP_Post_Type $post_type_object ) {
+
+		if ( null === self::$mutation ) {
+			self::$mutation = [];
```
CodeProKid: Same question as above

src/Type/PostObject/Mutation/PostObjectDelete.php
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		if ( ! empty( $post_type_object->graphql_single_name ) && empty( self::$mutation[ $post_type_object->graphql_single_name ] ) ) :
+
+			/**
+			 * Set the name of the mutation being performed
+			 */
+			$mutation_name = 'delete' . ucwords( $post_type_object->graphql_single_name );
```
CodeProKid: Needs translation

src/Type/PostObject/Mutation/PostObjectDelete.php
```diff
@@ -0,0 +1,153 @@
...
+
+			self::$mutation[ $post_type_object->graphql_single_name ] = Relay::mutationWithClientMutationId( [
+				'name'                => esc_html( $mutation_name ),
```
CodeProKid: `esc_sql()` might be more appropriate here.

src/Type/PostObject/Mutation/PostObjectDelete.php
```diff
@@ -0,0 +1,153 @@
...
+					/**
+					 * If the post was moved to the trash, spoof the object's status before returning it
+					 */
+					$post_before_delete->post_status = ( false !== $deleted && true !== $force_delete ) ? 'trash' : $post_before_delete->post_status;
+
+					/**
+					 * Return the deletedId and the object before it was deleted
+					 */
+					return [
+						'deletedId' => uniqid(),
```
CodeProKid: Shouldn't this be `$post_before_delete->ID`? instead of a random number?

src/Type/PostObject/Mutation/PostObjectMutation.php
```diff
@@ -0,0 +1,318 @@
...
+				],
+				'title'         => [
+					'type'        => Types::string(),
+					'description' => esc_html__( 'The title of the post', 'wp-graphql' ),
+				],
+				'toPing'        => [
+					'type'        => Types::list_of( Types::string() ),
+					'description' => esc_html__( 'URLs queued to be pinged.', 'wp-graphql' ),
+				],
+			];
```
CodeProKid: I wonder if it would be worth it to abstract this out, since there is probably some overlap with this, and where we define this within the postObject

src/Type/PostObject/Mutation/PostObjectMutation.php
```diff
@@ -0,0 +1,318 @@
...
+
+		if ( 0 === $user_id ) {
+			return false;
+		}
+
+		$now  = time();
+		$lock = "$now:$user_id";
+		update_post_meta( $post->ID, '_edit_lock', $lock );
+
+		return array( $now, $user_id );
```
CodeProKid: use `[]` instead of `array()`

src/Type/PostObject/Mutation/PostObjectUpdate.php
```diff
@@ -0,0 +1,172 @@
...
+		if ( null === self::$mutation ) {
+			self::$mutation = [];
+		}
+
+		if ( ! empty( $post_type_object->graphql_single_name ) && empty( self::$mutation[ $post_type_object->graphql_single_name ] ) ) :
+
+			/**
+			 * Set the name of the mutation being performed
+			 */
+			$mutation_name = 'update' . ucwords( $post_type_object->graphql_single_name );
```
CodeProKid: Translation Needed

src/Type/PostObject/Mutation/PostObjectUpdate.php
```diff
@@ -0,0 +1,172 @@
...
+	 * Defines the Update mutation for PostTypeObjects
+	 *
+	 * @param \WP_Post_Type $post_type_object
+	 *
+	 * @return array|mixed
+	 */
+	public static function mutate( \WP_Post_Type $post_type_object ) {
+
+		if ( null === self::$mutation ) {
+			self::$mutation = [];
```
CodeProKid: Same comment as above.

src/Type/PostObject/Mutation/PostObjectUpdate.php
```diff
@@ -0,0 +1,172 @@
...
+						throw new \Exception( sprintf( __( 'The id %1$d is not of the type "%2$s"', 'wp-graphql' ), $id_parts['id'], $post_type_object->name ) );
+					}
+
+					/**
+					 * Stop now if a user isn't allowed to edit posts
+					 */
+					if ( ! current_user_can( $post_type_object->cap->edit_posts ) ) {
+						// translators: the $post_type_object->graphql_single_name placeholder is the name of the object being mutated
+						throw new \Exception( sprintf( __( 'Sorry, you are not allowed to update a %1$s', 'wp-graphql' ), $post_type_object->graphql_single_name ) );
+					}
```
CodeProKid: I wonder if it would make sense to abstract these cap checks into a `postObjectMutateCheck` utility class or something. Just a thought. Not required before merge.

src/Type/PostObject/Mutation/PostObjectUpdate.php
```diff
@@ -0,0 +1,172 @@
...
+					/**
+					 * insert the post object and get the ID
+					 */
+					$post_args = PostObjectMutation::prepare_post_object( $input, $post_type_object, $mutation_name );
+					$post_args['ID'] = absint( $id_parts['id'] );
+
+					/**
+					 * Insert the post and retrieve the ID
+					 */
+					$post_id = wp_update_post( $post_args );
```
CodeProKid: As with `wp_insert_post()` you actually have to pass `true` as the second param. See docs: https://developer.wordpress.org/reference/functions/wp_update_post/

src/Type/PostObject/Mutation/PostObjectUpdate.php
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	private static function input_fields( $post_type_object ) {
+
+		/**
+		 * Update mutations require an ID to be passed
+		 */
+		return array_merge(
+			[
+				'id'      => [
```
CodeProKid: What is your reformator trying to align this with? haha.

tests/test-post-object-mutations.php
```diff
@@ -0,0 +1,360 @@
...
+
+		$this->title = 'some title';
+		$this->content = 'some content';
+		$this->client_mutation_id = 'someUniqueId';
+		$this->admin            = $this->factory->user->create( [
+			'role' => 'administrator',
+		] );
+		$this->subscriber       = $this->factory->user->create( [
+			'role' => 'subscriber',
+		] );
```
CodeProKid: Indentation here is pretty ugly.

src/Type/PostObject/Mutation/PostObjectCreate.php
```diff
@@ -0,0 +1,143 @@
...
+	 * Defines the create mutation for PostTypeObjects
+	 *
+	 * @param \WP_Post_Type $post_type_object
+	 *
+	 * @return array|mixed
+	 */
+	public static function mutate( \WP_Post_Type $post_type_object ) {
+
+		if ( null === self::$mutation ) {
+			self::$mutation = [];
```
jasonbahl: no upside. changed this.

src/Type/PostObject/Mutation/PostObjectCreate.php
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		if ( ! empty( $post_type_object->graphql_single_name ) && empty( self::$mutation[ $post_type_object->graphql_single_name ] ) ) :
```
jasonbahl: so this is actually the definition of the schema (what mutations are allowed to be executed), not the execution of the mutation.

So, this ensures that we only defined this create mutation exactly once for this post_type, and any circular references will use the existing definition instead of creating a new definition for it.

src/Type/PostObject/Mutation/PostObjectCreate.php
```diff
@@ -0,0 +1,143 @@
...
+
+			/**
+			 * Set the name of the mutation being performed
+			 */
+			$mutation_name = 'create' . ucwords( $post_type_object->graphql_single_name );
```
jasonbahl: I don't think we want to translate this. . .that would be similar to translating a REST endpoint, which would cause all sorts of issues.

src/Type/PostObject/Mutation/PostObjectCreate.php
```diff
@@ -0,0 +1,143 @@
...
+
+					/**
+					 * insert the post object and get the ID
+					 */
+					$post_args = PostObjectMutation::prepare_post_object( $input, $post_type_object, $mutation_name );
+
+					/**
+					 * Insert the post and retrieve the ID
+					 */
+					$post_id = wp_insert_post( $post_args );
```
jasonbahl: nice. . .I didn't realize that. Thanks for pointing out!

src/Type/PostObject/Mutation/PostObjectDelete.php
jasonbahl: changed

src/Type/PostObject/Mutation/PostObjectDelete.php
jasonbahl: see previous note. I think this would break things

src/Type/PostObject/Mutation/PostObjectDelete.php
jasonbahl: So this is the name in the context of the Schema, not in the context of executing, so I think esc_html is still appropriate as it gets sent to the client, not the server

src/Type/PostObject/Mutation/PostObjectDelete.php
jasonbahl: yes, it should indeed. I was messing with an idea for returning a unique ID because Graph.Cool also returns a deletedId. . .but upon inspection, their deletedId is just the ID of the object as well. . .I was thinking it was a unique ID for state management purposes, but we already have the clientMutationId for that. . .so I got this updated.

src/Type/PostObject/Mutation/PostObjectMutation.php
jasonbahl: ya, I'll need to play with how input/output fields can be re-used better since there is definitely overlap in the type/descriptions. 

I'll create a story to look into how that might look. 

There won't always be a 1-to-1 match between what input fields there are and what output fields there are though, so we'll just have to keep that into consideration whatever we come up with. . .but I agree, we could clean up some repetitive code in some way for sure.

jasonbahl: I added an issue here: https://github.com/wp-graphql/wp-graphql/pull/140/files/cd8c754112a7a3559c2acbf6fa742a81d4b72ea9#r118588925

src/Type/PostObject/Mutation/PostObjectMutation.php
jasonbahl: updated

src/Type/PostObject/Mutation/PostObjectUpdate.php
jasonbahl: updated

src/Type/PostObject/Mutation/PostObjectUpdate.php
jasonbahl: Created an issue to work on this: https://github.com/wp-graphql/wp-graphql/issues/143

src/Type/PostObject/Mutation/PostObjectUpdate.php
jasonbahl: updated

jasonbahl:
@CodeProKid Updated this PR and replied to comments. Thanks!

Issue #126, opened 2017-05-23
Add menu_location, menus, and nav_menu_item support.


BE-Webdesign:
Menu locations are registered by the active theme, or a plugin.  Menus are a taxonomy/term deal.  Nav menu items are a post type.  By supporting these three types in the `root_query` we will have really nice support for navigation menus.  From a quick look, I did not see this in the root query yet.

jasonbahl:
closed by #446

PR #162, opened 2017-07-03
#126 Menu Support

jasonbahl:
Opening this PR so it's in the open, but it's not quite ready to merge.


shah:
Hi @jasonbahl I was wondering about the status of this PR. We have a desperate need for this functionality and given how close it is, was curious what might be a good way to get to it to the finish line. :-)

jasonbahl:
@shah I estimate getting it merged within a week or so. Keep an eye on it!

shah:
Thanks @jasonbahl -- you're most kind! wp-graphql has been fun to use, keep up the great work.

jasonbahl:
@shah Hey, since you have a feature that will use this, can you provide some insight into what your needs are? 

Like how would you envision querying for a menu and it's items?

I've been playing with a couple ideas on the shape of the schema and where I'm at now looks something like: 

```
menu(location: {enum of registered locations}) {
  items {
    edges { 
      node { 
        id 
        url 
        ...whateverOtherFields
        childItems {
          edges { 
            node {
               id
               url
               ...whateverOtherFields
            }
          }
        }
      }
    }
  }
}
```

shah:
@jasonbahl yes that schema looks perfect. What we're trying to do is to use WP as a "headless CMS" and pull the content dynamically at runtime into an Ionic custom mobile app. We'd like to use the menu infrastructure to define our navigation in WP and then pull that into our mobile app at runtime through Apollo Client. 

Basically, we'll grab a list of menus (which we customize per institution / author) and the menus will become our navigation buttons in the mobile app. For example, if user1 from institution1 logs in we'd show them their specific menu for navigation while if user2 from institution2 logs in we'd give them their menu. So, if WP menus can be tied to users that would be ideal; if not, we'd use either categories or tags or just the menu name to select appropriately.

Then, we render the menu as Ionic views with CSS and as a menu item button is clicked we'd dynamically read content from posts attached to the menus yielding a good native experience on iOS / Android but with fully dynamic content. Your cool wpGraphQL plugin will give us the best of both worlds: allow the "headless" use for the end user UI and a nice admin back-end for editing with the power of WP.

Hopefully this helps, but please do ask me to elaborate if this is unclear. 

BTW, I was curious why all the queries have 'edges' and 'node' in the Schema -- is that due to the Relay cursor specification? If that could be configurable it would be nice. For example, if in the wpGraphQL plugin config we could set a global variable saying it's for Relay then it would incorporate edges / nodes otherwise the hierarchy would be easier to specify :-)

Thanks for all your hard work!

jasonbahl:
Edges and nodes were initially influenced by the Relay spec, but there's a lot of value in that shape of data where "edge" data can be stored and queried against for data that's not a property of any specific object but is a property of the connection between two objects. 

For example, let's say you and I were 2 friends in a social network. There might be a piece of data like the date we became friends. This isn't a property of my user object or your user object but is a property of the relationship, or the edge, between us.

Let's say WordPress were to add term_relationship_meta (or something similar) at some point, where we could store data that exists only in the context of relationships between two objects, they data would live in the edges of the queries and could provide very powerful features without troublesome client side computations. 

One real example of some "edge"
data where I work is that we have the ability for photos to be used in multiple news articles and everything about the image is the same, but editors have he ability to create a custom caption to localize it for the specific article. This custom caption isn't directly a property of the article or the image, but exists only in the context of the relationship between the article and the image. 

Another benefit of the edges/node structure is that it provides a common naming convention for querying collections of data. Whether it's posts, pages, terms, images, or menu items the shape of the queries is predictable and easy to learn as it's not unique based on the type of data being queried.

So, in short, I don't picture officially supporting both the edge/node shape AND another shape officially in the core plugin because I think there's a lot of value that the edge structure will provide long term. 

However, if you still feel that a different shape would make sense for your particular application, you could definitely hook in and provide your own queries that resolve to the types that will be defined already. 


I'll have to turn this into a blog post because I think there's valu in explaining the edge/node shape and why it exists in the plugin, and how to write a quick filter to add your own queries with a different shape when needed.

jasonbahl:
@shah these resources might help elaborate on why the edge/node shape makes sense for GraphQL in general (not specific to WPGraphQL)

https://stackoverflow.com/a/42623708
https://dev-blog.apollodata.com/explaining-graphql-connections-c48b7c3d6976

shah:
Thanks for the explanation, @jasonbahl -- very helpful!

jasonbahl:
Ok, so I think the MVP of this feature is looking pretty good. I think we can probably add a bit more to `menus` but here's some examples of what you can do right now: 

wpgraphql-menu-queries

jasonbahl:
@shah wanted to follow-up since you asked. I just updated this PR and submitted it for review, so it should be released soon. 

Feel free to checkout this feature branch and check things out in the mean time. 

You can see some example queries in the `tests` here: https://github.com/jasonbahl/wp-graphql/blob/%23126-menu-support/tests/test-menu-queries.php

shah:
@jasonbahl the test looks great, thanks for making the changes so quickly and for the PR. I've asked one of my colleagues to pull the branch and take a look while I'm on the road. Great stuff!

jasonbahl:
addresses #126

jasonbahl:
@CodeProKid 

>One thought I had was that it would be nice to have some parity with wp_nav_menu() and have an "levels" argument like that function has to restrict the levels of navigation for output. I guess it could also be handled on the client too though.

So, the levels are defined by the query. If you want just top level items, you query: 

```
menu(location:"..."){
   menuItems{
      edges{
         node{
            id
            title
            link
            ...whateverOtherFields
         }
      }
   }
}
```

if you want another level: 

```
menu(location:"..."){
   menuItems{
      edges{
         node{
            id
            title
            link
            ...whateverOtherFields
            childItems{
              edges{
                node{
                  id
                  title
                  link
                  ...whateverOtherFields
                }
              }
           }
         }
      }
   }
}
```

And you can keep going to get however many levels you want for whatever you're building.

CodeProKid:
Ah, got it. So by that logic, would you have to know how many levels the menu is to query all of the menu items?

src/Type/Menu/MenuType.php
```diff
@@ -0,0 +1,106 @@
...
+					],
+					'slug'            => [
+						'type'        => Types::string(),
+						'description' => esc_html__( 'The url friendly name of the menu. Equivalent to WP_Term->slug', 'wp-graphql' ),
+					],
+					'count' => [
+						'type' => Types::int(),
+						'description' => __( 'The number of items in the menu', 'wp-graphql' ),
+					],
+					'group'           => [
```
jasonbahl: I'd be ok with leaving the `group` out, and if someone needed it they can add it back via plugin, and _if_ someone presented enough evidence of why it's useful we could add it back then.

jasonbahl: actually, since it's a core feature, whether it's used a lot or not, I think I feel better keeping it. 

I will change it to be named `groupId` and change the type to `Types::int()` though. 

We do support a termGroupId field on other terms, so I think for folks that do use groups, it's still valuable. (I know I've never really used term groups, but doesn't mean some folks out there aren't using them)

src/Type/MenuItem/Connection/MenuItemConnectionResolver.php
```diff
@@ -0,0 +1,143 @@
...
+		 */
+		if ( ! empty( $args['where']['parentMenuItemId'] ) ) {
+			$id_parts = Relay::fromGlobalId( $args['where']['parentMenuItemId'] );
+			if ( is_array( $id_parts ) && ! empty( $id_parts['id'] ) ) {
+				$parent_id = $id_parts['id'];
+			}
+		}
+
+		$query_args['post_type'] = 'nav_menu_item';
+		$query_args['posts_per_page'] = 100;
```
jasonbahl: So this will affect each level, not the menu in it's entirety. 

I have a hard time thinking any individual level of a menu would have more than 100 items, but ya, there should definitely be a filter. 

I actually implemented one in the `get_query_amount` method which is part of the ConnectionResolver class that we're extending here, so I'll adjust to make use of that so that the filter can be applied for menus in the same way it is able to be filtered for other post_types with a filter like so: 

```
add_filter( 'graphql_connection_max_query_amount', function( $amount, $source, $args, \WPGraphQL\AppContext $context, \GraphQL\Type\Definition\ResolveInfo $info ) {

	if ( 'menuItems' === $info->fieldName ) {
		$amount = 1;
	}

	return $amount;

}, 10, 5 );
```

jasonbahl:
> Ah, got it. So by that logic, would you have to know how many levels the menu is to query all of the menu items?

You'd have to know how many levels your consuming application wants to support. Like I know it's pretty common for themes to support styles for 3 levels, but then things get pretty wonky after 3 levels. . .so if you had a UI that only supported 3 levels, you would write a query for 3 levels. 

Conceptually similar to the GraphQL introspection query that's used to query the Schema. It queries the info it needs, I believe up to 7 levels deep. So if your schema had deeply nested schema definitions deeper than ~7 levels, tools like GraphiQL might not work properly. 

See: https://github.com/graphql/graphql-js/blob/master/src/utilities/introspectionQuery.js#L14-L107

jasonbahl:
Closing this as it needs some more work. Will re-open a new PR with the changes in place.

ashyashoks:
@jasonbahl  I installed the latest release of the plugin in my wordpress instance but it gives the error 
"SyntaxError: Unexpected end of JSON input" . The early version was working fine. Did you change anything in the new release, is there any specific setting to do? Please help me.

jasonbahl:
@ashyashoks are you getting that error in general or related to menu queries? If it's not related to menu queries, can you create a new issue and include as much detail as you can? Where do you see the error? does it occur for only certain queries? Does the error occur if WPGraphQL is the only plugin installed? Etc.

ashyashoks:
@jasonbahl  getting that error in general not to menu queries.  I got this error in the chrome extension "chromeiQL" and also I didn't get the end point in my wordpress instance. The only other plugin installed in the instance is "types". Please help me.

jasonbahl:
@ashyashocks please open a new issue since what your describing has nothing to do with the issue being discussed here (menu support)

PR #168, opened 2017-07-12
#92-test-access-functions

jasonbahl:
This addresses #92


jasonbahl:
moves access-functions.php to 100% test coverage

Issue #129, opened 2017-05-23
Make sure boolean types return a boolean instead of null.


BE-Webdesign:
On the PostTypeType boolean values are checked against a ! empty -> value : null sort of thing.  This means that when the value is false `null` is returned instead.  This should probably be changed to just return the value.  This is probably happening elsewhere in the code base.

jasonbahl:
addressed in #179

PR #179, opened 2017-07-26
#129-make-sure-booleans-return-booleans

jasonbahl:
This adjusts the Type::boolean fields that were returning null, to make sure they properly return true or false instead.


jasonbahl:
addresses #129

jasonbahl:
addresses #129

PR #188, opened 2017-08-04
Feature/#17 attachment mutations

hughdevore:


hughdevore:
@jasonbahl, here's my initial run at this. I'm noticing a lot of composer files, should they have been git ignored?

jasonbahl:
@hughdevore can you run the following then commit and push again: 

```
rm -rf composer.lock
rm -rf vendor
composer install --no-dev
```

PR #191, opened 2017-08-10
User Mutations

CodeProKid:
This PR adds user mutations. I still have to write tests for this, but I think it's at a good enough point where the code can start getting reviewed.


src/Type/User/Mutation/UserUpdate.php
```diff
@@ -0,0 +1,132 @@
...
+						$error_message = $user_id->get_error_message();
+						if ( ! empty( $error_message ) ) {
+							throw new \Exception( esc_html( $error_message ) );
+						} else {
+							throw new \Exception( __( 'The user failed to update but no error was provided', 'wp-graphql' ) );
+						}
+					}
+
+					/**
+					 * If the $post_id is empty, we should throw an exception
```
jasonbahl: $user_id 😉

src/Type/User/Mutation/UserCreate.php
```diff
@@ -0,0 +1,118 @@
...
+	 * @return array
+	 */
+	private static function input_fields() {
+
+		$user_fields = UserMutation::input_fields();
+
+		/**
+		 * Create mutations require a login to be passed
+		 */
+		$user_fields['login']['type'] = Types::non_null( Types::string() );
```
jasonbahl: since the `UserMutation::input_fields()` is filterable we should probably explicitly make sure the login field exists, similar to how you're doing here: https://github.com/wp-graphql/wp-graphql/pull/191/files#diff-4fdbc70dee8938bb749e25fefa47a18bR114

Let's say someone did a heavy-handed filter like: 

```
add_filter( 'graphql_user_mutation_input_fields', function() {
  return $fields['the_only_mutation_field'] = [
     'type' => Types::string(),
     'description' => __( 'I just filtered out all mutation inputs except this field, bro', 'some-silly-plugin' ),
  ];
}, 10, 1 );
```

This would cause: `$user_fields['login']['type'] = Types::non_null( Types::string() );` to be a PHP warning as the array_key `login` wouldn't exist.

src/Type/User/Mutation/UserDelete.php
```diff
@@ -0,0 +1,120 @@
...
+			self::$mutation = Relay::mutationWithClientMutationId( [
+				'name' => 'deleteUser',
+				'description' => __( 'Delete a user object', 'wp-graphql' ),
+				'inputFields' => [
+					'id' => [
+						'type' => Types::non_null( Types::id() ),
+						'description' => __( 'The ID of the user you want to delete', 'wp-graphql' ),
+					],
+					'reassignId' => [
+						'type' => Types::int(),
```
jasonbahl: `Types::id()`

src/Type/User/Mutation/UserMutation.php
```diff
@@ -0,0 +1,209 @@
...
+
+		if ( ! empty( $input['nicename'] ) ) {
+			$insert_user_args['user_nicename'] = $input['nicename'];
+		}
+
+		if ( ! empty( $input['websiteUrl'] ) ) {
+			$insert_user_args['user_url'] = esc_url( $input['websiteUrl'] );
+		}
+
+		if ( ! empty( $input['email'] ) ) {
```
jasonbahl: turns out that if you enter a plain string instead of an email address for email, the email just gets set to null. . .

So, I'm thinking we need some validation here to ensure the input is an acceptable email address (probably need to dig and see what core is doing to validate the email when the update is made) before we allow it to be used as the `user_email` field for insertion.

I'm thinking, if the email isn't "valid" we just unset it and don't pass it to the mutation at all. . .and/or we throw an exception so the user knows the email was invalid.

See attached GIF. 

string-emails-not-working

CodeProKid:
@jasonbahl took a first pass at tests here, still needs some cleanup, and maybe a few more added, but feel free to check it out. Also still need to address some of the previous feedback.

src/Type/User/Mutation/UserCreate.php
```diff
@@ -0,0 +1,123 @@
...
+						}
+					]
+				],
+				'mutateAndGetPayload' => function( $input ) {
+
+					if ( empty( $input ) || ! is_array( $input ) ) {
+						throw new \Exception( __( 'Mutation not processed. There was no input for the mutation.', 'wp-graphql' ) );
+					}
+
+					if ( ! current_user_can( 'create_users' ) ) {
```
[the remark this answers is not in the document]
CodeProKid: I'll add a followup issue for this, since it's a separate mutation, and will have separate tests.

src/Type/User/Mutation/UserMutation.php
CodeProKid: I think the exception route would be best for now. Until someone has an objection to doing this, I say we roll with that.

jasonbahl: Ya, let's do some validation on the input to make sure it's an email and throw an exception if it's not proper

CodeProKid:
@jasonbahl addressed all feedback, and wrote a few tests! I feel pretty good about the coverage now, not sure how to rerun all the tests here though, they are all passing locally for me though.

jasonbahl:
looks like tests are barfing for Multisite. . .I'm gonna investigate a bit and see if I can figure out what's up. My guess is user creation, etc is handled a bit different for multisite?

jasonbahl:
Found a few things to adjust for multisite. 

Fwiw, to test in multisite locally you can run `export WP_MULTISITE=1` to set your environment variable so PHPUnit will run as a multisite. To switch back to non-multisite testing `export WP_MULTISITE=0`

You can see the travis.yml file as well to see what tests are run, etc. (need to get a good README together RE: unit tests. . .I'm still learning a LOT too).

src/Type/User/Mutation/UserMutation.php
```diff
@@ -0,0 +1,291 @@
...
+				],
+				'registered'  => [
+					'type'        => Types::string(),
+					'description' => __( 'The date the user registered. Format is Y-m-d H:i:s.', 'wp-graphql' ),
+				],
+				'roles'       => [
+					'type'        => Types::list_of( Types::string() ),
+					'description' => __( 'An array of roles to be assigned to the user.', 'wp-graphql' ),
+				],
+				'jabber'      => [
```
jasonbahl: AIM, YIM and JABBER were removed from core in 3.6 (https://developer.wordpress.org/reference/functions/wp_insert_user/#changelog) so not going to include these in the core user mutations.

Issue #160, opened 2017-07-01
Return plain text content related fields


duffn:
I'd like wp-graphql to return content related post fields, i.e. `content`, `excerpt`, that are plain text only without HTML tags. I'm specifically requesting fields in addition to the already available fields - `contentRaw`, `excerptRaw` for instance.

These would be helpful for developers using wp-graphql that are building mobile apps or spoken language apps that are only looking for the plain text related fields of a post.

jasonbahl:
My thought, at the moment, would be to either add an argument to the fields that would simply prepare the results of the field differently. 

For example: 

```
post(id:"..."){
  id
  content
  rawContent:content(outputFilters:[RAW])
}
```
Having an argument like `outputFilters` which would accept a list of filters (possibly an enum of available filters) to apply would allow for the consumer to have the response prepared for their needs without having to introduce new "fields" to the schema. 

This could be pretty powerful. 

We could have filters like "raw" that strip all markup, shortcodes, etc. 

We could also have filters that just strip markup, filters that just strip shortcodes, etc. 

So you could potentially do: 

```
post(id:"..."){
  id
  content
  contentWithoutShortcodes:content(outputFilters:[STRIP_SHORTCODES])
  contentWithoutHtmlTags:content(outputFilters:[STRIP_TAGS])
  contentWithoutHtmlTagsOrShortCodes:content(outputFilters:[STRIP_TAGS, STRIP_SHORTCODES])
}
```
Need to explore this more, but this could be a very valuable feature indeed.

jasonbahl:
@jaredcobb, also curious to get your thoughts on this as you've had to do your fair share of transformation of `the_content` and `excerpt`, etc for making it useful in various contexts (a.k.a. syndication, REST, etc)

keg:
@jasonbahl I think what you're suggesting makes sense with strip tag and strip shortcodes filters. Would `STRIP_TAGS` strip media too? The only other transformation that I've regularly hit is encoding, smart quotes etc. Also FWIW I'd make the default return type as close as possible to what is stored in the database.

jasonbahl:
@keg so the default output should be just the results of `$post->post_content` which would include raw shortcodes, etc? 

I think in _most_ places in WordPress, the expectation for "content" is the result of `$post->post_content` with `the_content` filter applied, which converts shortcodes to their rendered output, and includes any other "things" that have been added by other plugins and theme code.

[...]

PR #195, opened 2017-08-17
Provide global $post in resolvers that apply filters.

chriszarate:
Also refines descriptions.

### DIFF
--- a/src/Type/PostObject/PostObjectType.php
+++ b/src/Type/PostObject/PostObjectType.php
@@ -181,8 +181,12 @@ class PostObjectType extends WPObjectType {
 					],
 					'content'           => [
 						'type'        => Types::string(),
-						'description' => __( 'The content of the post. This is currently just the raw content. An amendment to support rendered content needs to be made.', 'wp-graphql' ),
-						'resolve'     => function( \WP_Post $post, $args, AppContext $context, ResolveInfo $info ) {
+						'description' => __( 'The content of the post (currently with the_content filters applied). An amendment to support filterability needs to be made.', 'wp-graphql' ),
+						'resolve'     => function( \WP_Post $wp_post, $args, AppContext $context, ResolveInfo $info ) {
+							global $post;
+
+							$post = $wp_post;
+
 							return apply_filters( 'the_content', $post->post_content );
 						},
 					],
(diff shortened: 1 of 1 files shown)


jasonbahl:
@chriszarate thanks for the PR! 

Looks like a recent change I made to the composer files is causing Travis CI not to run the unit tests correctly. Fixing that now, then will run the tests on this again. . . then, will provide feedback. 

Thanks!

jasonbahl:
@chriszarate would you be interested in writing some unit tests for this change? Would be good to make sure anything we change here works in the future as well as we make changes. 

Also, can you elaborate with an example of what the output is now when you query for the content on a post vs. what happens with this change in place (a unit test would serve as this example if you have time/interest in writing a test for this)

Thanks!

Jason

chriszarate:
Sure I'll work up a test. This is all subject to change given the comments in #160, so I'll try to write something that can be built upon.

jasonbahl:
@chris I think whatever we do for #160 would likely still respect the default output if there are no args passed. So if you just ask for the content you get a default output like you do now. If you do pass args to apply certain filters, then the output would be customized to match the applied filters.

jasonbahl:
@chriszarate it might be good to investigate further whether other fields rely on the $GLOBALS['post'] in the same way as the content and excerpt do. We _might_ want to consider setting that global when the $post object is prepared to be sent down the resolve tree, so it can happen once. Or should every field be responsible for setting the global if they need it, like you have now? 🤔 

Also, I'm thinking we probably need to capture the current state of the global before setting it, so that we can set it back to it's original state after the resolving is complete. While WPGraphQL is likely going to be used in a decoupled environment, it _can_ be used within the standard WordPress environment as well, so if someone wanted to use a GraphQL query to populate a shortcode in a page, or something to that tune, not resetting the global to what it was before could cause interesting behavior. 

ex: 

```
add_shortcode( 'graphql_basic_post_list', function( $atts ) {
	$query = '
	query basicPostList($first:Int){
		posts(first:$first){
			edges{
				node{
					id
					title
					date
				}
			}
		}
	}
	';

	$variables = [
		'first' => ! empty( $atts['first'] ) ? absint( $atts['first'] ) : 5,
	];

	$data = do_graphql_request( $query, 'basicPostList', $variables );
	$edges = ! empty( $data['data']['posts']['edges'] ) ? $data['data']['posts']['edges'] : [];

	if ( ! empty( $edges ) && is_array( $edges ) ) {
		$output = '<ul>';
		foreach ( $edges as $edge ) {
			$node = ! empty( $edge['node'] ) ? $edge['node'] : '';
			if ( ! empty( $node ) && is_array( $node ) ) {
				$output .= '<li id="' . $node['id'] . '">' . $node['title'] . ' ' . $node['date'] . '</li>';
			}
		}
		$output .= '</ul>';
	}

	return ! empty( $output ) ? $output : '';
});
```
If we don't reset the `$global`, anything on the page below the shortcode, like related posts, comments, etc would get funky. So whenever we override the $GLOBALS['post'] we need to make sure we reset it. 

FWIW, this shortcode works, and an example screenshot is attached: 

wpgraphql-basic-shortcode-example

jasonbahl:
@chriszarate any thoughts on my feedback?

chriszarate:
Closing in favor of #230.

PR #199, opened 2017-08-19
Activation and deactivation hooks

### LINKED ISSUE #196: Feature: flush rewrite rules on activation and deactivation
This will remove the requirement that the user manually flush the rewrite rules on plugin instal and activation.

AramZS:
Handle rewrite rule flushes automatically. This will take that responsibility off the user. Resolves #196.

Also, sets version into option table so future versions can execute actions on upgrades.

### DIFF
--- a/wp-graphql.php
+++ b/wp-graphql.php
@@ -196,6 +196,8 @@ if ( ! class_exists( 'WPGraphQL' ) ) :
 		 */
 		private function actions() {
 			// @placeholder where actions can be added throughout. This will be useful for mutations
+			register_deactivation_hook( __FILE__, array($this, 'activate' ) );
+			register_activation_hook( __FILE__, array( $this, 'deactivate' ) );
 		}
 
 		/**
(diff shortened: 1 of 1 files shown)


wp-graphql.php
```diff
@@ -196,6 +196,8 @@ private function includes() {
 		 */
 		private function actions() {
 			// @placeholder where actions can be added throughout. This will be useful for mutations
+			register_deactivation_hook( __FILE__, array( $this, 'activate' ) );
+			register_activation_hook( __FILE__, array( $this, 'deactivate' ) );
```
CodeProKid: you can use `[]` for arrays

wp-graphql.php
```diff
@@ -212,6 +214,22 @@ private function filters() {
 
 		}
 
+		public function activate() {
+			flush_rewrite_rules();
+			// Save the version of the plugin as an option in order to force actions
+			// on upgrade.
+			if ( false === get_option( 'wp_graphql_version', false ) ) {
+				add_option( 'wp_graphql_version', WPGRAPHQL_VERSION, null, 'no' );
```
CodeProKid: I don't think you have to actually check if the option exists already. You can just use `update_option` since it falls back to `add_option` if the option doesn't exist.

AramZS:
PR ready for consideration.

CodeProKid:
@AramZS looks good on first glance, just a few small things. I'll also let @jasonbahl do a final review on this and merge.

wp-graphql.php
AramZS: TIL: you can set autoload within the `update_option` function

AramZS:
@CodeProKid Changes made.

wp-graphql.php
```diff
@@ -196,6 +196,8 @@ private function includes() {
 		 */
 		private function actions() {
 			// @placeholder where actions can be added throughout. This will be useful for mutations
```
jasonbahl: we can get rid of this placeholder comment now that this method is in use and not actually a placeholder

AramZS:
@jasonbahl Removed excess comment.

jasonbahl:
@AramZS Thanks for this. Looks good. The unit test failure is unrelated, working on a fix, but merging this.

Issue #200, opened 2017-08-19
An empty request to the endpoint results in a malformed error response.


AramZS:

jasonbahl:
addressed by #202

PR #202, opened 2017-08-19
Return a user friendly error on an empty query

AramZS:
This assures that if the user hits the namespace with no query they are
presented with a friendly well-formatted error. Noted in ticket #200

### DIFF
--- a/tests/test-json-responses.php
+++ b/tests/test-json-responses.php
@@ -4,7 +4,7 @@
  *
  * @group ajax
  */
-class WPGraphQL_Test_Router extends WP_Ajax_UnitTestCase {
+class WPGraphQL_JSON_Responses extends WP_Ajax_UnitTestCase {
 
 	public function setUp() {
 		parent::setUp();
@@ -31,6 +31,8 @@ class WPGraphQL_Test_Router extends WP_Ajax_UnitTestCase {
 		set_query_var( 'graphql', true );
 		$GLOBALS['wp']->query_vars['graphql'] = true;
 
+		define( 'DOING_AJAX', true );
+
 		/**
 		 * Instantiate the router
 		 */
@@ -43,7 +45,6 @@ class WPGraphQL_Test_Router extends WP_Ajax_UnitTestCase {
 		try {
 			$this->_handleAjax( 'graphql_resolve_http_request' );
 		} catch ( WPAjaxDieContinueException $e ) {}
-		// $router::resolve_http_request();
 		/**
 		 * Make sure the constant gets defined when it's a GraphQL Request
 		 */
--- a/tests/test-json-responses.php
+++ b/tests/test-json-responses.php
@@ -13,7 +13,12 @@ class WPGraphQL_JSON_Responses extends WP_Ajax_UnitTestCase {
 	public function tearDown() {
 		parent::tearDown();
 	}
-
+	/**
+	 * [testResolveHttpRequestWithEmptyQuery description]
+	 *
+	 * @runInSeparateProcess
+	 * @return [type] [description]
+	 */
 	public function testResolveHttpRequestWithEmptyQuery() {
 		/**
 		 * Filter the request data
--- a/tests/test-json-responses.php
+++ b/tests/test-json-responses.php
@@ -1,0 +1,70 @@
+<?php
+/**
+ * Test case for Basic Router responses.
+ *
+ * @group ajax
+ */
+class WPGraphQL_Test_Router extends WP_Ajax_UnitTestCase {
+
+	public function setUp() {
+		parent::setUp();
+	}
+
+	public function tearDown() {
+		parent::tearDown();
+	}
+
+	public function testResolveHttpRequestWithEmptyQuery() {
+		/**
+		 * Filter the request data
+		 */
+		add_filter( 'graphql_request_data', function( $data ) {
+			$data['query'] = null;
+			$data['variables'] = null;
+			$data['operationName'] = null;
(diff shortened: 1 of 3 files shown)


AramZS:
@jasonbahl Finished writing unit tests, this is ready for code review.

AramZS:
hmm, after some checking I don't think that the errors here are a result of my changes?

jasonbahl:
@AramZS ya, looking into it. Something with the router tests is acting up.

tests/test-json-responses.php
```diff
@@ -0,0 +1,70 @@
+<?php
+/**
+ * Test case for Basic Router responses.
+ *
+ * @group ajax
+ */
+class WPGraphQL_Test_Router extends WP_Ajax_UnitTestCase {
```
jasonbahl: @AramZS looks like the issue is that you used the same class name here as we have here: https://github.com/wp-graphql/wp-graphql/blob/develop/tests/test-router.php#L3

Try changing this class name to something like `WP_GraphQL_Test_JSON_Responses`

tests/test-json-responses.php
```diff
@@ -0,0 +1,70 @@
...
+		/**
+		 * Set the query var to "graphql" so we can mock like we're visiting the endpoint via
+		 */
+		set_query_var( 'graphql', true );
+		$GLOBALS['wp']->query_vars['graphql'] = true;
+
+		/**
+		 * Instantiate the router
+		 */
+		$router = new \WPGraphQL\Router();
```
jasonbahl: Looks like this is never used, as the use of it is commented out?

tests/test-json-responses.php
```diff
@@ -0,0 +1,70 @@
...
+
+		add_action( 'wp_ajax_graphql_resolve_http_request', [ '\WPGraphQL\Router', 'resolve_http_request' ] );
+		/**
+		 * Process the request using our filtered data
+		 */
+		try {
+			$this->_handleAjax( 'graphql_resolve_http_request' );
+		} catch ( WPAjaxDieContinueException $e ) {}
+		// $router::resolve_http_request();
```
jasonbahl: Can remove this if it's not needed

jasonbahl:
@AramZS it also looks like the tests in `test-router.php` that test the `resolve_http_request` cause the tests to stop running altogether. 

test-router-abrupt-end

AramZS:
Weird. Ok, I'll check in to it!

tests/test-json-responses.php
AramZS: This gets used by the hook and I think it needs to be instantiated earlier.

AramZS:
@jasonbahl I fixed up the tests and figured out a way to unobtrusively make the router capable of responding on AJAX but not other non-header ready requests. 

The end point of that though was that I realized that the router function doesn't actually complete and push out a response on any of the tests in the router file. I saw other things are being checked, but I wasn't sure if that was the intent. That said, I fixed some changes I made previously to respect the current method. Those were the issues that were breaking the router test.

AramZS:
I still am getting `xdebug_get_headers()` related errors, but those are the ones that you said to ignore right?

jasonbahl:
@AramZS ya, we can ignore those for now. . .as those are environment specific (I just need to make sure the Travis instances have xdebug). . .not sure why those tests sometimes pass and sometimes fail though

jasonbahl:
> The end point of that though was that I realized that the router function doesn't actually complete and push out a response on any of the tests in the router file. I saw other things are being checked, but I wasn't sure if that was the intent. That said, I fixed some changes I made previously to respect the current method. Those were the issues that were breaking the router test.

Right now the intent was to test individual functions rather than full integration tests. . .where _most_ of our testing is more "integration" testing, where we're not testing individual functions on their own, but testing queries which map to several functions. 

Probably would be good to separate "integration" and "unit" tests at some point...

jasonbahl:
@AramZS I'm still getting funky output from the test, but it's passing now. . .not sure it's an issue, just haven't seen that before 🤔 

see: https://travis-ci.org/wp-graphql/wp-graphql/jobs/268428149

AramZS:
@jasonbahl I know where it is coming from, now when DOING_AJAX is true and `headers_sent` is true it executes the `wp_die` inside `wp_send_json` whereas before if the `headers_sent` was true it would never send a response (that's what I was talking about above). I believe that the headers are marked as set in the router test class, which is why it never outputted into the buffer. I noticed that from the CLI I never got this error in the tests I ran, but I see it is getting it when I run the full suite. 

Is there anywhere else besides the tests I created (which have output buffering built in, which is why they don't show that output) where DOING_AJAX is defined as true or the headers are set as part of the test? That must be where the echoed buffer is coming from.

jasonbahl:
> Is there anywhere else besides the tests I created (which have output buffering built in, which is why they don't show that output) where DOING_AJAX is defined as true or the headers are set as part of the test? That must be where the echoed buffer is coming from.

No. . .but I *think* we can tag the test with `@runInSeparateProcess` and that _might_ solve it?

AramZS:
@jasonbahl The thing is I don't think it is the `WPGraphQL_JSON_Responses` test. That one run on its own doesn't echo the buffer, so it is some other test causing the rogue output.

AramZS:
Yeah, I see it, the issue is that the DOING_AJAX const effects some other place where it is calling the response down the line. Adding the `runInSeparateProcess` directive.

jasonbahl:
@AramZS thanks for working on this! looks great.

PR #203, opened 2017-08-19
changed to get_the_title rather than direct post title field

mateomtb:


src/Type/PostObject/PostObjectType.php
```diff
@@ -190,7 +190,7 @@ private static function fields( $post_type_object ) {
 						'type'        => Types::string(),
 						'description' => __( 'The title of the post. This is currently just the raw title. An amendment to support rendered title needs to be made.', 'wp-graphql' ),
 						'resolve'     => function( \WP_Post $post, $args, AppContext $context, ResolveInfo $info ) {
-							return ! empty( $post->post_title ) ? $post->post_title : null;
+							return get_the_title( $post );
```
jasonbahl: @mateomtb since this field is a nullable field, we actually want to return null if the title were empty. 

So this should look something like: 
```
$title = get_the_title( $post );
return ! empty( $title ) ? $title : null; 
```

Issue #207, opened 2017-08-24
Get Post Objects by more than just the ID


jasonbahl:
We should be able to get a single Post by more than just the ID. 

I'm thinking the postObjects should have a `by` input that allows you to do something like these: 

```
post( by: { slug: "the-post-slug" } ) {
   id
   title
}
```

```
post( by: { id: "opaqueId" } ) {
   id
   title
}
```

```
post( by: { postId: 123 } ) {
   id
   title
}
```

CodeProKid:
Hmmmmm, I'm not sure I agree with this. I feel like if you want to get a post by anything other than the ID, it should be done with a `posts` query, since it could potentially return multiple posts in all instances, which makes the singular `post()` misleading.

jasonbahl:
@CodeProKid so, the use case is that in a React (or other JS app), you will might be at a route, say: `site.com/post/post-slug` and you need to hydrate that route with that post. 

Much like WordPress itself `/site.com/post-slug`

So, you need to be able to fetch that resource, and the only identifier you might have for it is the slug. 

Since title's are _not_ unique, I can take that one out, but I think the postId (ID in the database), Opaque ID (base64 relay ID), and the slug, should be fields to get a post by.

jasonbahl:
@CodeProKid I updated to remove the option to get by title.

CodeProKid:
@jasonbahl you will still have the issue of possible name collisions though, since WordPress has a unique permalink structure. So, there could be a post in a custom post type called "My Post" that would have a slug of `my-post`, and the full permalink would be something like: `custom-post-type/my-post`. Then you could have another post in the Post post type called "My Post" that would have a slug of `my-post` and a full permalink like `blog/my-post` or whatever. So if you *only* passed the slug as an argument, you would get two posts back since they both have the same value in the `post_name` column. I think the only way we could do this is if we also required the Type as well.

jasonbahl:
@CodeProKid ya, the Type is part of it anyway, as each post_type can only be queried via it's Type in the schema. 

like: 

```
post( by: { slug: "the-post-slug" } ) {
   id
   title
}
```

```
page( by: { slug: "the-post-slug" } ) {
   id
   title
}
```

CodeProKid:
[...]

PR #208, opened 2017-08-24
#207 - Get Post Objects by Global ID, Database ID, or URI.

jasonbahl:
This exposes a new “uri” field on the postObjectType, which resolves to `get_page_uri()`

Then, this adds a new `postBy, pageBy, mediaItemBy` field that allows single postObjects to be queried `by` one of 3 fields, ID, GlobalID or URI.

To resolve based on the URI, a cached version of `get_page_by_path` was added to the DataSource (borrowed and modified from VIP’s mu plugins)

This includes unit tests to ensure these queries work and that the nodes returned match each time.

### DIFF
--- a/src/Data/DataSource.php
+++ b/src/Data/DataSource.php
@@ -532,7 +532,7 @@ class DataSource {
 	 *
 	 * This is a modified version of the cached function from WordPress.com VIP MU Plugins here.
 	 *
-	 * @param string $path
+	 * @param string $uri
 	 * @param string $output Optional. Output type; OBJECT*, ARRAY_N, or ARRAY_A.
 	 * @param string $post_type Optional. Post type; default is 'post'.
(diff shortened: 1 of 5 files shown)


src/Data/DataSource.php
```diff
@@ -527,4 +527,40 @@ function( $node ) {
 		return self::$node_definition;
 	}
 
+	/**
+	 * Cached version of get_page_by_path so that we're not making unnecessary SQL all the time
+	 *
+	 * This is a modified version of the cached function from WordPress.com VIP MU Plugins here.
+	 *
+	 * @param string $path
```
Soean: `@param string $uri Object URI.`

jasonbahl: @soean good eye! Thanks for the review!

PR #230, opened 2017-09-21
Add post object field format arg

chriszarate:
_Edited to reflect discussion and changes:_

This provides a method to request `RAW` or `RENDERED` format of post object fields:

```
query {
  post(id: "...") {
    content(format: RAW)
  }
}
```

The `format` arg is available on `content`, `title`, and `excerpt`. If the `format` arg is omitted, then the default behavior will be `RENDERED` (which is simply to apply these filters: `the_content` on `content`, `the_title` on `title`, and `get_the_excerpt` and `the_excerpt` on `excerpt`).

It introduces a new enum type, `PostObjectFieldFormatEnumType`.

Perhaps more significantly, it sets up and resets post data during post object field resolution, so that any custom resolution undertaken by the user will have that done for them.

### DIFF
--- a/src/Type/Enum/PostObjectFieldFilterEnumType.php
+++ b/src/Type/Enum/PostObjectFieldFilterEnumType.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace WPGraphQL\Type\Enum;
+
+use WPGraphQL\Type\WPEnumType;
+
+/**
+ * Class PostObjectFieldFilterEnumType
+ *
+ * This defines an EnumType with allowed WordPress filters that can be applied
+ * to post field data.
+ *
+ * @package WPGraphQL\Type\Enum
+ * @since   0.0.18
+ */
+class PostObjectFieldFilterEnumType extends WPEnumType {
+
+	/**
+	 * This holds the enum values array.
+	 *
+	 * @var array $values
+	 */
+	private static $values;
+
+	public function __construct() {
+		$config = [
+			'name'        => 'postObjectFieldFilter',
+			'description' => __( 'The allowed WordPress filters on post field data.', 'wp-graphql' ),
+			'values'      => self::values(),
+		];
+		parent::__construct( $config );
+	}
+
+	/**
+	 * Creates a list of filters that can be used to filter post field data.
+	 *
+	 * @return array
+	 */
+	private static function values() {
+
+		if ( null === self::$values ) {
+
+			/**
+			 * Provide some default filters that are in WP Core.
+			 *
+			 * @since 0.0.18
+			 */
+			self::$values = [
+				'the_content' => [
+					'name'  => 'THE_CONTENT',
+					'description' => __( 'WordPress Core the_content filter', 'wp-graphql' ),
+					'value' => 'the_content',
+				],
+				'get_the_excerpt' => [
+					'name'  => 'GET_THE_EXCERPT',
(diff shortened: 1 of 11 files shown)


src/Type/Enum/PostObjectFieldFilterEnumType.php
```diff
@@ -0,0 +1,67 @@
+<?php
```
jasonbahl: This file name needs to change to: PostObjectFieldFormatEnumType

jasonbahl: Not sure how unit tests didn't bomb on this 🤔

chriszarate: !!! ... is composer saving the day here?

jasonbahl:
@chriszarate I added a few more tests, and all looks good. Merging! 

Thanks for your work on this!

bq14:
@jasonbahl @chriszarate 

Testing this with
```
curl 7.76.1
PHP 8.0.7
WP GraphQL v1.4.3
WP v5.7.2
```

for case: format = RAW, queries return 'null'

I'm not certain I didn't misread and maybe it's my usage :-/


for default query,
```
$CURL \
  --data '{ "query": "{ pages(where: {title: \"Test1\"}) { edges { node { content } } } }" }' \
https://wptest.loc/cms/graphql/ | json_pp


	{
	   "data" : {
	      "pages" : {
	         "edges" : [
	            {
	               "node" : {
	                  "content" : "\n<p>test 123 content </p>\n\n\n\n<p>test 456 content</p>\n"
	               }
	            }
	         ]
	      }
	   },
	   "extensions" : {
	      "debug" : []
	   }
	}
```

format = RENDERED, same return as default,
```
$CURL \
  --data '{ "query": "{ pages(where: {title: \"Test1\"}) { edges { node { content(format: RENDERED) } } } }" }' \
https://wptest.loc/cms/graphql/ | json_pp

	{
	   "data" : {
	      "pages" : {
	         "edges" : [
	            {
	               "node" : {
	                  "content" : "\n<p>test 123 content </p>\n\n\n\n<p>test 456 content</p>\n"
	               }
	            }
	         ]
	      }
	   },
	   "extensions" : {
	      "debug" : []
	   }
	}
```

format = RAW, returns 'null'
```
$CURL \
  --data '{ "query": "{ pages(where: {title: \"Test1\"}) { edges { node { content(format: RAW) } } } }" }' \
https://wptest.loc/cms/graphql/ | json_pp

	{
	   "data" : {
	      "pages" : {
	         "edges" : [
	            {
	               "node" : {
	                  "content" : null
	               }
	            }
	         ]
	      }
	   },
	   "extensions" : {
	      "debug" : []
	   }
	}
```

chriszarate:
> for case: format = RAW, queries return 'null'

@bq14 WPGraphQL's model layer prevents unauthorized access to certain data, following the protections of WordPress core. Raw post content is only exposed to users with `edit_posts` capability. Your example request is unauthenticated, therefore access to raw post content is not allowed:

https://github.com/wp-graphql/wp-graphql/blob/develop/src/Model/Post.php#L495

This PR predates the model layer, so apologies that there was no reference to this implementation detail.

If you want to expose raw post content, you'll need to authenticate your request or override the resolver for the `content` field to bypass this protection (be sure to evaluate the risks of doing so).

Issue #232, opened 2017-09-25
Typo in $wpdb->prepare


jasonbahl:
There's a typo (missing placeholder) here: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Config.php#L88

jasonbahl:
addressed by #233

PR #233, opened 2017-09-25
#232 - fix placeholder in wpdb->prepare

jasonbahl:
addresses #232

### DIFF
--- a/src/Data/Config.php
+++ b/src/Data/Config.php
@@ -85,7 +85,7 @@ class Config {
 							$order_compare = ( 'ASC' === $order ) ? '>' : '<';
 							$value = $cursor_post->{$by};
 							if ( ! empty( $by ) && ! empty( $value ) ) {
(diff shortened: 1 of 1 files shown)


src/Data/Config.php
```diff
@@ -85,7 +85,7 @@ public function graphql_wp_query_cursor_pagination_support( $where, \WP_Query $q
 							$order_compare = ( 'ASC' === $order ) ? '>' : '<';
 							$value = $cursor_post->{$by};
 							if ( ! empty( $by ) && ! empty( $value ) ) {
-								$where .= $wpdb->prepare( " AND {$wpdb->posts}.{$by} {$order_compare} '{$value}'", $value );
+								$where .= $wpdb->prepare( " AND {$wpdb->posts}.{$by} {$order_compare} '%s'", $value );
```
CodeProKid: I'm not sure you actually have to add the placeholder within the quotes here? Does it not work without them?

jasonbahl: @CodeProKid let me check. I think you're probably right

PR #240, opened 2017-10-05
CI-4209 Change author field for comments to support to non-users.

pjpak:


jasonbahl:
@pjpak Unit tests are failing on this, and code coverage decreased as the added code isn't tested.

We need to test both cases, when a comment author is a User and when a comment author is NOT a user. 

So we'll need to add a new unit test that does creates a comment that isn't associated to a user and make some assertions against it.

Similar to here (https://github.com/pjpak/wp-graphql/blob/01ba1e23c09fa6aef2a8ba919af95d3d910f2dae/tests/test-comment-object-queries.php#L83) but we need to set the `user_id` to 0, and populate the other fields in the comment that we want to make assertions against ($comment->comment_author, $comment->comment_author_email, $comment->comment_author_url, $comment->comment_author_IP)

Also, statically defining data for `User 109` is part of the failure, as that's not a constant and will change as we write more tests, etc. 

So instead, the assertion on those fields should be changed from : 

```
'email'   => 'user_109@example.org',
'name'    => 'User 109'
```

to something to this tune: 

```
'email' => get_userdata( $this->admin )->user_email,
'name' => get_userdata( $this->admin )->display_name,
```

But that's only adding fields to test a comment that's being queried when the author is a User, so we still need to test when the author is NOT a user, as outlined above.

src/Type/Comment/CommentType.php
```diff
@@ -81,7 +81,18 @@ private static function fields() {
...
+								$author = new \WP_User( $comment->user_id );
+							} else {
+								$author = new \WP_User();
+
+								if ( ! empty( $comment->comment_author ) ) {
+									$author->__set( 'display_name', $comment->comment_author );
+								}
+								if ( ! empty( $comment->comment_author_email ) ) {
+									$author->__set( 'user_email', $comment->comment_author_email );
+								}
```
jasonbahl: We need to populate the new WP_User's `user_url ` field, using `$comment->comment_author_url` as that's a field that can be queried against the author and the data to populate it exists here.

Also, we need to populate the ID field to be unique, as the ID is used to normalize data in a client side cache, so without a unique ID, client side cache will be funky. 

Without setting a unique ID, anytime a comment author is queried for, it will have an ID of 0, which means every single author that's queried will bust the cache of the previous author because they have the same ID.

I'm thinking we could use the email address and name as the unique identifier. . . something like: 

```
$comment_author_id = [ 'type' => 'comment_author', 'email' => $comment->comment_author_email, 'name' => $comment->comment_author ];
$author->_set( 'ID', $comment_author_id );
```

This will translate to a base64 encoded ID that has enough info to query for that author again, but will also ensure that it's unique enough that client side caches don't accidentally override other comment authors in the cache. 

Then of course this will need to be tested as well.

pjpak:
thanks @jasonbahl i checked in some changes, can you review when you get a chance.

src/Type/Comment/CommentType.php
jasonbahl: @pjpak did you see this comment too?

pjpak: I added some changes for comment_author_id at https://github.com/wp-graphql/wp-graphql/pull/240/files#diff-ee902c2db37a04884a3f7a15426e2b9eR94
I wasnt quite sure how to test that.

src/Type/Comment/CommentType.php
```diff
@@ -81,7 +81,23 @@ private static function fields() {
...
+								$author = new \WP_User( $comment->user_id );
+							} else {
+								$author = new \WP_User();
+
+								if ( ! empty( $comment->comment_author ) ) {
+									$author->__set( 'display_name', $comment->comment_author );
+								}
+								if ( ! empty( $comment->comment_author_email ) ) {
+									$author->__set( 'user_email', $comment->comment_author_email );
+									$comment_author_id = [ 'type' => 'comment_author', 'email' => $comment->comment_author_email, 'name' => $comment->comment_author ];
```
jasonbahl: So it looks like you can't set the ID like this. . .but this works: 

```
$author->ID = wp_json_encode([
    'type' => 'comment_author',
    'email' => ! empty( $comment->comment_author_email ) ? $comment->comment_author_email : null,
    'name' => ! empty( $comment->comment_author ) ? $comment->comment_author : null,
]);
```

Then we also need to include a test for the `id` field to make sure it's outputting what we expect, which should look something like: 

`dXNlcjp7InR5cGUiOiJjb21tZW50X2F1dGhvciIsImVtYWlsIjoiamFzb25iYWhsQG1hYy5jb20iLCJuYW1lIjoiSmFzb24gQmFobCJ9`

That's a base64encoded string that contains the following data when decoded:

`user:{"type":"comment_author","email":"jasonbahl@mac.com","name":"Jason Bahl"}`

That ID now contains enough data for the server to be able to identify how to re-fetch the data if needed, and it's unique enough for client side caches to invalidate when that comment author sees a mutation in the client.

jasonbahl:
@pjpak I'm going to close this. . .after thinking more about this I think we need to approach it with a union. I will update the original issue with more details on that approach.

PR #250, opened 2017-10-10
replaced is_a func with instanceof

luchianenco:


src/Type/PostObject/PostObjectQuery.php
```diff
@@ -106,7 +106,7 @@ public static function post_object_by( \WP_Post_Type $post_type_object ) {
 						throw new \Exception( __( 'No resource could by found', 'wp-graphql' ) );
 					}
 
-					if ( ! is_a( $post_object, 'WP_Post' ) ) {
+					if ( !$post_object instanceof \WP_Post ) {
```
CodeProKid: Needs a space between `!` and `$post_object`

CodeProKid:
Thanks for the contribution @luchianenco !

PR #254, opened 2017-10-10
Features/#156 options queries

hughdevore:


src/Data/DataSource.php
```diff
@@ -4,10 +4,12 @@
 use GraphQL\Type\Definition\ResolveInfo;
 use GraphQLRelay\Relay;
 
+use PHPUnit\Runner\Exception;
```
jasonbahl: @hughdevore I don't think this is supposed to be here, eh?

hughdevore: Hmmm, dunno how that got added, will remove!

jasonbahl:
@hughdevore I'm going to close this per our Slack conversation

Issue #256, opened 2017-10-11
Preflight OPTIONS requests are returning 500 status


jasonbahl:
HTTP clients such as Apollo and Axios send "pre-flight" requests, and we're not respecting those, but instead are returning a 500. 

We need to respect the pre-flight requests.

see: https://wp-graphql.slack.com/archives/C3NM1M291/p1507650342000560
see: https://apollographql.slack.com/archives/C10HTKHPC/p1507649812000123
see: https://apollographql.slack.com/archives/C10HTKHPC/p1507650789000675

CodeProKid:
This was handled in #257

PR #257, opened 2017-10-11
#256 - Preflight Request Support

jasonbahl:
This adds support for the preflight requests that HTTP clients such as Axios and Apollo use.

This also adds a filter to the status header so folks can override it as needed

Addresses #256


CodeProKid:
I know you merged this already, but LGTM.

Issue #272, opened 2017-10-17
Update GraphQL-PHP to v0.11.2


jasonbahl:
We're currently on v0.9.14 and there has been some updates to GraphQL PHP that we could take advantage of. 

There are some **breaking changes**, specifically around error reporting, so we will have to adjust some things to adjust our errors. 

See: https://github.com/webonyx/graphql-php/blob/master/UPGRADE.md#upgrade-v08x-v09x--v010x

jasonbahl:
addressed by #275

Issue #216, opened 2017-09-02
Return warnings in some cases


CodeProKid:
I believe some of the exceptions we are throwing are a bit heavy handed. I think in some instances we need to return some _warning_ data, but still run the query. The convo in the slack channel actually got me thinking about this. 

For instance; if someone queries for 200 posts, and the limit is still set to 100, ideally I think we should return the 100 posts, but also return a warning that says what was requested was over the hard limit.

I think the response data could look something like the following:
```
{
  "warnings": [
    {
        "message": "Too many posts queried, we've limited the results to the maximum number allowed...",
        "locations": [
        {
          "line": 10,
          "column":10,
        }
        ]
  ],
  "data": {
    "posts": {
      "edges": [
        {
          "node": {
            "id": "cG9zdDo0MTg="
          }
        },
        {
          "node": {
            "id": "cG9zdDo0MTc="
          }
        }
      ]
    }
  }
}
```
I think adding varying degrees of exceptions as a whole will be good for the codebase, and would allow us to give good debugging info without bombing out an app. Could also be good data for an optics like product that tracks all of the GraphQL queries.

jasonbahl:
Part of the first/last after/before and pageInfo partly solves this issue (the max query amount)  

When you ask for data you're never guaranteed to get back as much as you asked for, but the pageInfo will return wether there is more pages or not. 

So a consuming client will query first 1,000, they will receive 100 items (or whatever the sys admin deems an acceptable amount to allow without causing performance issues), and will be able to see in the pageInfo wether there hasNextPage is true and if so they can use the endCursor to query for more items. 

That said, I do think varying levels of warnings/errors should be explored more. I believe the GraphQL spec actually has more kinds of errors than we're currently making use of. Haven't explored a ton yet though. 

I do know I want to reorganize some of our mutations for sure though. We block full mutations to throw an error when we probably should capture the potential errors, do the partial mutation with what data we can, and throw an exception/warming for the parts we couldn't fulfill

Like user update for example. If I update first name, last name and email, but my email had a typo, we could update the first and last name, and throw a warning/exception about the email being invalid. 🤔

Def think we should explore the GraphQL spec re: warnings more in general though. 

At the moment I'm not sold that the max number of pages example should be a warning...Not completely against it but not completely for it either. I'll let it stew in my head for a bit.

jasonbahl:
[...]

PR #275, opened 2017-10-17
#272 - Update to use GraphQL-PHP v0.11.2

jasonbahl:
This updates WPGraphQL to use v0.11.2 of GraphQL-PHP.


CodeProKid:
@jasonbahl - this is a bit of a beast. Do you have a changelog you can link to, for what was actually changed in this library?

CodeProKid:
Seems like travis is barfing on some composer stuff?

jasonbahl:
@CodeProKid fixed the composer stuff. . .it's running again and I think it _should_ be good now. 

The Upgrade instructions for GraphQL-PHP are here: https://github.com/webonyx/graphql-php/blob/master/UPGRADE.md

The Changelog is here: https://github.com/webonyx/graphql-php/blob/master/CHANGELOG.md

The primary reason I wanted to update now is because there were some adjustments to work better with DataLoader, and some other core improvements to help with some other performance boosts, such as: 

- Significantly improved parser performance
- Apollo-style query batching support via server

---- 

The biggest breaking change was how Exceptions are handled. Default exceptions are no longer returned to the client as-is, they are masked as "Internal server error", so to return an exception with the request, we needed to convert Exceptions to `UserError()` to keep our implementation from breaking backward compatibility.

So a LOT of the changes in this PR are just one-liners like: https://github.com/wp-graphql/wp-graphql/pull/275/files#diff-9961a10648c95beb5c33cfbf42ec1532L155 then corresponding updates to Unit Tests to include the addition of the "category" that errors return with them now.

We now will have a lot more granular control over how we handle errors though, which is sweet (we have an open issue #216 about how to handle errors/warnings in various ways, so this update should help with that)

jasonbahl:
@CodeProKid merging this as I don't want it lingering for long. . .got some upcoming features that rely on some of these changes.

PR #276, opened 2017-10-17
CI-4215 - Adding tests for sanitize_input_fields and get_query_args in PostObject Connection

davidvexel:
Testing every single parameter of sanitize_input_fields, also testing get_query_args.

### DIFF
--- a/tests/test-post-connection-queries.php
+++ b/tests/test-post-connection-queries.php
@@ -332,4 +332,198 @@ class WP_GraphQL_Test_Post_Connection_Queries extends WP_UnitTestCase {
 
 	}
 
+	public function testSanitizeInputFieldsAuthorArgs() {
+		$mock_args = [
+			'authorName' => 'testAuthorName',
+			'authorIn' => [1, 2, 3],
+			'authorNotIn' => [4, 5, 6],
+		];
+
+		$app_context = new \WPGraphQL\AppContext();
+		
+		$app_info = new \GraphQL\Type\Definition\ResolveInfo( array() );
+
+		$actual = \WPGraphQL\Type\PostObject\Connection\PostObjectConnectionResolver::sanitize_input_fields( $mock_args, null, [], $app_context, $app_info );
+
+		/**
+		 * Make sure the returned values are equal to mock args
+		 */
+		$this->assertEquals( 'testAuthorName', $actual['author_name'] );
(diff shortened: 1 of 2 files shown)


tests/test-post-connection-queries.php
```diff
@@ -332,4 +332,198 @@ public function testPageWithChildren() {
...
+		$mock_args = [
+			'parent' => 2,
+			'parentIn' => [3, 4, 5],
+			'parentNotIn' => [6, 7, 8],
+			'in' => [9, 10, 11],
+			'notIn' => [12, 13, 14],
+			'nameIn' => ['testPost1', 'testPost2', 'testPost3'],
+		];
+
+		$app_context = new \WPGraphQL\AppContext();
```
jasonbahl: Not super essential, but to keep things a bit more DRY, instead of declaring `$app_context` and `$app_info` inside every single test, you can declare them once as a class var and just re-use them throughout the tests

For example, we setup some other re-usable values here: https://github.com/davidvexel/wp-graphql/blob/1fe1c837c83f7cfff157c8a15bc2390ccf220610/tests/test-post-connection-queries.php#L12-L34

davidvexel:
@jasonbahl getting close on this one, but I'm not sure how to test these lines:

https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/PostObject/Connection/PostObjectConnectionResolver.php#L147-L148

🤔

jasonbahl:
@davidvexel I think we can just get rid of those 2 lines 😄

davidvexel:
@jasonbahl good to know! I'll double check my current tests and add some comments on them, just wanted to know I'm on the right path, thanks!

tests/test-post-connection-queries.php
```diff
@@ -332,4 +336,322 @@ public function testPageWithChildren() {
...
+
+		$mock_args = array();
+
+		$actual = \WPGraphQL\Type\PostObject\Connection\PostObjectConnectionResolver::get_query_args( $source, $mock_args, $this->app_context, $this->app_info );
+
+		$this->assertEquals( $user_id, $actual['author'] );
+
+	}
+
+	// public function testGetQueryArgsNullObject() {
```
jasonbahl: don't need this commented out, let's just remove it?

jasonbahl:
@davidvexel this looks great. Thanks for working on it!

PR #277, opened 2017-10-17
Features/#156 - Adding Settings queries

hughdevore:
This PR addresses the need to be able to query any setting that gets registered using the WP register_setting method. It allows developers to filter which groups of settings, or settings in those groups, appear in the graphQL schema. Here is an example of the query getting all default site settings for a vanilla WP install: 

http://g.recordit.co/DnHXSHbf9D.gif


jasonbahl:
@hughdevore I'll teach you how to save an upload GIFs some day soon 😉 

graphql-options

jasonbahl:
@hughdevore I made some updates to this to get it to the finish line. Thanks for your hard work on it!

PR #282, opened 2017-10-18
CI-4209a

pjpak:


src/Type/Union/CommentAuthorUnionType.php
```diff
@@ -0,0 +1,86 @@
...
+	public function getPossibleTypes() {
+
+		if ( null === self::$possible_types ) {
+			self::$possible_types = [
+					'user' => Types::user(),
+					'commentAuthor' => Types::comment_author(),
+			];
+		}
+
+		$allowed_taxonomies = \WPGraphQL::$allowed_taxonomies;
```
jasonbahl: can get rid of lines 63-71

src/Type/Union/CommentAuthorUnionType.php
```diff
@@ -0,0 +1,86 @@
...
+
+		/**
+		 * Filter the possible_types as it's possible some systems might set things like "parent_id" to a different
+		 * object than a post_type, and might want to be able to hook in and add a non postObject type to the possible
+		 * resolveTypes.
+		 *
+		 * @param array $possible_types  An array of possible types that can be resolved for the union
+		 * @since 0.0.6
+		 */
+		self::$possible_types = apply_filters( 'graphql_term_object_union_possible_types', self::$possible_types );
```
jasonbahl: update this filter name to `graphql_comment_author_union_possible_types`

src/Type/Union/CommentAuthorUnionType.php
```diff
@@ -0,0 +1,86 @@
...
+	 */
+	public function __construct() {
+
+		$config = [
+			'name' => 'commentAuthorUnion',
+			'types' => function() {
+				return self::getPossibleTypes();
+			},
+			'resolveType' => function( $value ) {
+			if ( ! empty( $value->user_id ) ) {
```
jasonbahl: indentation looks funky here

src/Type/CommentAuthor/CommentAuthorQuery.php
```diff
@@ -0,0 +1,33 @@
+<?php
```
jasonbahl: Is this being used anywhere? I suppose we could add it to the `rootQuery` 🤔

src/Type/Comment/CommentType.php
```diff
@@ -77,15 +77,27 @@ private static function fields() {
...
 						'resolve' => function( \WP_Comment $comment, $args, AppContext $context, ResolveInfo $info ) {
 							$author = new \WP_User( $comment->user_id );
+							return ! empty( $author ) ? $author : null;
+						},
+					],
+					*/
+					// TODO
 
+					'author' => [
+						'type' => Types::comment_author(),
```
jasonbahl: `type` should be `Types::comment_author_union(),`
`description` could be changed to `The author of the comment`

Then the `resolve` function can be updated to something like: 

```
'resolve' => function( \WP_Comment $comment, $args, AppContext $context, ResolveInfo $info ) {
    
    // Here we determine if the Author is a User or not so we know what to pass through to the Union for resolving
    if ( ! empty( $comment->user_ID ) ) {
       $author = new \WP_User( $comment->user_ID )
    } else if ( ! empty( $comment ) ) {
       $author = $comment;
    } else { 
       $author = null;
    }
    return $author;
}
```

src/Type/Union/CommentAuthorUnionType.php
jasonbahl: Also, I screwed up on how this should look. . .the `$value` is actually what is passed from the field resolver that declares the union as the type. That is where we want to determine whether to return a WP_User or just pass through the `$comment` itself for the author Data to be plucked from.

So this can be changed to: 

```
resolveType' => function( $source ) {
    if ( $source instanceof \WP_User ) {
        $type = Types::user();
    } else if ( $source instanceof \WP_Comment ) {
        $type = Types::comment_author();
    } else {
        $type = null;
    }
    return $type;
}
```

pjpak:
@jasonbahl  Made recommended changes but still running into
GraphQL\Type\Definition\FieldDefinition::createMap(NULL, 'commentAuthor')
Also trying this change: https://github.com/pjpak/wp-graphql/blob/CI-4209a/src/Types.php#L300

Issue #55, opened 2017-02-23
Implement DataLoader


jasonbahl:
We should implement DataLoader to take advantage of the batching/caching functionality as it should boost performance significantly.

https://github.com/overblog/dataloader-php

jasonbahl:
Could be helpful for Post commections Coming from a term object. http://scribu.net/wordpress/sortable-taxonomy-columns.html

CodeProKid:
fwiw, we talked about this yesterday, and I think there may be more performance gains from doing something like hashing query arguments, and using that as a "key", and then storing the results of the query as a "value". Could either cache it just in the session, or in the object cache with `wp_cache_set()`.

jasonbahl:
Initial implementation of Deferred resolvers was implemented in #291 

We should probably open up new issues to address DataLoading for each of the Types / Connections.

PR #293, opened 2017-11-03
Initial Implementation of DataLoader (deferred resolvers) 

jasonbahl:
Addresses #55 (at least initially, more to come, should probably become their own issues...)

This adds initial support for Deferred resolvers (DataLoader concept). 

This initially adds a User loader to prove concept and pattern. 

But the benefits are immediately clear. For example, this exact same query before and after this PR goes from `28` SQL queries to `8` with a `28%` increase in overall execution time!

screen shot 2017-11-03 at 9 35 25 am
screen shot 2017-11-03 at 9 35 54 am

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
   "description": "GraphQL API for WordPress",
   "type": "wordpress-plugin",
   "license": "GPLv3",
-  "version": "0.0.20",
+  "version": "0.0.21",
   "authors": [
     {
       "name": "Jason Bahl",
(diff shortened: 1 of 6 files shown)


cr101:
@jasonbahl Is queryLog part of WPGraphQL Insights?

jasonbahl:
@cr101 The query log _is_ part of WPGraphQL insights. I believe you _might_ need to `define( 'SAVEQUERIES', true );` in your wp-config, or have something active that defines that for you, like Query Monitor.

Issue #296, opened 2017-11-06
Conflict with WooCommerce 3.2.3


jasonbahl:
There is some sort of conflict between WPGraphQL v0.0.21 and WooCommerce 3.2.3. 

Initially reported here: #295 

I'm not sure what the issue is yet but I've confirmed that when those are the only 2 plugins active, the response from a GraphQL query is: 

```
{
  "errors": [
    {
      "message": "Internal server error",
      "category": "internal"
    }
  ]
}
```

jasonbahl:
@cr101 I discovered what the conflict is. I have a fix coming shortly.

jasonbahl:
@cr101 I opened this PR to address this #297 

Planning to get it out today. Thanks!

jasonbahl:
Addressed by #297

PR #297, opened 2017-11-06
#296 - WooCommerce Conflict

jasonbahl:
This fixes the conflict with WooCommerce (and other plugins/theme code that have any post types or taxonomies with "show_in_rest" set to true. 

In the latest release we attempted to default any post_types and taxonomies that are set to "show_in_rest" to also "show_in_graphql" but because they don't have a proper "graphql_single_name" and "graphql_plural_name" set, this was causing issues. 

So, we're back to explicitly defining if a type should be included in GraphQL or not. 

Might be good to have some sort of GraphQL Admin page to set these things?


jasonbahl:
Addresses #296

PR #302, opened 2017-11-10
Add way to allow unauthorized requests

timneutkens:
We have a use case where the graphql api needs to be exposed to the public :+1:

### DIFF
--- a/src/Router.php
+++ b/src/Router.php
@@ -388,7 +388,7 @@ class Router {
 			 * set the status code to 200.
 			 */
 			$user = wp_get_current_user();
-			if ( $user && 0 !== $user->ID ) {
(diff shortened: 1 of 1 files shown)


jasonbahl:
@timneutkens the API is already publicly exposed. The 403 response is used to indicate that the request is not authenticated, but it doesn't prevent execution of the request. Public data will still be executed and returned.

If you do need to override the 403 status, there is already a filter in place here:
https://github.com/wp-graphql/wp-graphql/blob/develop/src/Router.php#L435 

Have you tried using that filter?

timneutkens:
Oh nice, I missed it 👌

jasonbahl:
@timneutkens Apollo expects a 200 for the pre-flight OPTIONS request, but can be a 403 for follow-up requests. That was addressed a few releases ago, when this filter was added. 

Are you on a current version of Apollo and WPGraphQL?

PR #306, opened 2017-11-12
Ability to sort by menu_order

indeox:
A few WP plugins (ex: [SCPOrder](https://en-gb.wordpress.org/plugins/simple-custom-post-order/)) allow users to re-arrange the sort order of the posts, and then assigns a value to the `menu_order` field in the database.

This PR would allow the GraphQL endpoint to return the posts in that specified order

```graphql
# Example query

posts(where: { 
  categoryName: "projects"
  orderby: { field: MENU_ORDER, order: ASC }
}) {
  menuOrder
  title
  link
} 
```


jasonbahl:
@indeox Hey! Thanks for the PR. I'll check it out and get you feedback (if needed) later today. 

Thanks!

indeox:
@jasonbahl Ah no worries. I was just playing around with a Proof of Concept for a side project and needed that patch for something to work. Thought I'd do a little contribution if you think it's useful.

Feel free to ignore it though. Otherwise, fantastic work on the plugin!

Issue #190, opened 2017-08-07
Hook up Travis CI with Schema Linting 


jasonbahl:
see: https://github.com/cjoudrey/graphql-schema-linter

CodeProKid:
Put this in the 1.1 release for now. Doesn't seem like this needs to be in there for 1.0. Feel free to move it into 1.0 if you think it needs to be though.

jasonbahl:
I was thinking that 1.0 made sense because that’s when we’re essentially saying it’s ready to go, and a tool like this will help us validate that it is ready to go and will help us keep it stable long term. . .so I think we should hook this up, and probably even make some breaking changes to adhere to the best practices it identifies, like capitalized type names (ours are all lowercase now)

jasonbahl:
addressed by #310

PR #310, opened 2017-11-14
#190 - Schema Linting

jasonbahl:
Setting up Codeception / Travis Environment


jasonbahl:
Finally! Got this working. 

Travis now uses `wp graphql generate-static-schema` to generate a static `schema.graphql` file, then we use `graphql-schema-linter` (npm script) to lint that schema. 

We can see output in Travis. 

Right now, I have this set up to run _after_ success, as I don't want it to cause failures yet. Once we have the linter passing, we'll move it to a script that will cause Travis to fail if linting doesn't pass.

hughdevore:
@jasonbahl, this breaks things when I merge and run tests locally 🙈 any ideas?

PR #312, opened 2017-11-15
Features/#156 options mutations

hughdevore:


jasonbahl:
@hughdevore thanks for working on this! Checking it out. Will have feedback in a bit!

hughdevore:
@jasonbahl, just cleaned up some of the comments/docs/naming.

PR #319, opened 2017-11-20
Remove Jason's local yml files and update README.

hughdevore:
This PR updates the README and removes Jason's local yml config files.


davidvexel:
Code looks good to me 👍

PR #339, opened 2017-12-14
Remove get_the_excerpt filter for image caption.

chriszarate:
The get_the_excerpt filter calls wp_trim_excerpt which, if the caption / excerpt is empty, uses global $post->post_content (a.k.a., the parent post) as the seed for crafting the excerpt.

This means that if you leave the caption field blank for a featured image, WPGraphQL currently populates the caption field with the excerpt of the parent post (or possibly even a different post).

I think it's safe to assume this isn't going against user's expectations, since you would need to setup post data *for the attachment* to use get_the_excerpt for image captions.


jasonbahl:
@chriszarate Looks good! Would you be interested in writing tests for this?

chriszarate:
Ok, I added a test that confirms empty caption is preserved. Also added a test for image description pulling from `post_content` since there wasn't one. Let me know if there's broader strategy here we should be going for. (Apologies for all the commits, I don't have codeception set up on this machine so I was flying blind.)

PR #357, opened 2018-01-10
fix User nicename type

### LINKED ISSUE #356: User nicename should be a string, not an int
From Slack:

> I’m trying to query Users by nicename, but I notice that nicename is set to `Types::int()` here. Should it be `string`? https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/User/Connection/UserConnectionArgs.php#L97

PR incoming shortly :)

ginatrapani:
Closes #356


ginatrapani:
@jasonbahl woohoo! first contribution and not the last! thanks again!

PR #362, opened 2018-01-11
#361 - This addresses some funkiness with GET requests not working pr…

### LINKED ISSUE #361: GET Requests with field arguments don't work
It appears that GET requests that contain Field arguments don't work properly. 

For example, this request works as a GET request: 

```
{
  users {
    edges {
      node {
        nicename
      }
    }
  }
}
```

But this doesn't:

```
{
  users(first: 1, where: {login: "jasonbahl"}) {
    edges {
      node {
        nicename
      }
    }
  }
}
```

We should investigate what is causing the issue (likely something here: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Router.php#L332-L370) and see if we can adjust to support GET requests a bit better

jasonbahl:
Closes #361


src/Router.php
```diff
@@ -362,11 +362,11 @@ public static function process_http_request() {
 					 * use in the executor.
 					 */
 				} else {
-					$decoded_variables = json_decode( $data['variables'] );
+					$decoded_variables = (array) json_decode( wp_kses_stripslashes( $data['variables'] ) );
```
CodeProKid: @jasonbahl you can actually pass `true` as the second param of `json_decode()` and it will return an associative array.

Issue #333, opened 2017-12-05
Post Object Mutations should allow setting/updating terms


jasonbahl:
Right now Terms can't be updated from a PostObjectMutation. 

We should be able to set terms in some fashion when creating/updating postObjects.

Need to discuss what the inputs should look like, but it needs to exist.

PR #369, opened 2018-01-17
#333 - Set term connections as nested input in PostObjectMutations

jasonbahl:
This adds support to set term connections via nested input in PostObjectMutations.

For example: 

```
mutation CreatePostWithNewCategoryAndTag {
  createPost(
    input: {
      title: "New Post", 
      status: PUBLISH, 
      tags: {
        append: true, 
        nodes: [
          {
            name: "My New Tag", 
            description: "New Tag Description"
          }
        ]
      }, 
      categories: {
        append: true, 
        nodes: [
          {
            name: "My New Category", 
            description: "New Category Description"
          }
        ]
      }, 
      clientMutationId: "uniqueId"
    }) {
    post {
      id
      title
      tags {
        edges {
          node {
            id
            name
            slug
            description
          }
        }
      }
      categories {
        edges {
          node {
            id
            name
            slug
            description
          }
        }
      }
    }
  }
}
```

Outputs: 

```
{
  "data": {
    "createPost": {
      "post": {
        "id": "cG9zdDo2NTE3",
        "title": "New Post",
        "tags": {
          "edges": [
            {
              "node": {
                "id": "cG9zdF90YWc6Mzc3",
                "name": "My New Tag",
                "slug": "my-new-tag",
                "description": "New Tag Description"
              }
            }
          ]
        },
        "categories": {
          "edges": [
            {
              "node": {
                "id": "Y2F0ZWdvcnk6Mzc2",
                "name": "My New Category",
                "slug": "my-new-category",
                "description": "New Category Description"
              }
            },
            {
              "node": {
                "id": "Y2F0ZWdvcnk6MQ==",
                "name": "Uncategorized",
                "slug": "uncategorized",
                "description": null
              }
            }
          ]
        }
      }
    }
  }
}
```

When creating or updating a post (or custom post type) each connected taxonomy will have a root input, which accepts `append` as a bool value, and `nodes` which expects an array of term nodes. 

The terms will be connected to existing terms based on the node's ID field, if present, then the slug field if present. If neither fields exist, the Name field will be used to either create a new term (if allowed) or connect to any existing term with that name. The description input field will be used only if a new term is created.

### DIFF
--- a/src/Type/PostObject/Mutation/PostObjectMutation.php
+++ b/src/Type/PostObject/Mutation/PostObjectMutation.php
@@ -447,7 +447,7 @@ class PostObjectMutation {
 								 * during a post object mutation, create the term to connect based on the
 								 * input
 								 */
-								if ( false === $term_exists && true === $allow_term_creation ) {
+								if ( ! $term_exists && true === $allow_term_creation ) {
 
 									/**
 									 * If the current user cannot edit terms, don't create terms to connect
@@ -527,10 +527,14 @@ class PostObjectMutation {
 			$created_term = wp_insert_term( $term_to_create['name'], $taxonomy, $term_args );
 		}
 
+		if ( is_wp_error( $created_term ) && isset( $created_term->error_data['term_exists'] ) ) {
+			return $created_term->error_data['term_exists'];
+		}
+
 		/**
 		 * Return the created term, or 0
 		 */
-		return ! is_wp_error( $created_term ) && ! empty( $created_term['term_id'] ) ? $created_term['term_id'] : 0;
+		return ! empty( $created_term['term_id'] ) ? $created_term['term_id'] : 0;
 
 	}
 
--- a/src/Type/PostObject/Mutation/PostObjectMutation.php
+++ b/src/Type/PostObject/Mutation/PostObjectMutation.php
@@ -120,11 +120,11 @@ class PostObjectMutation {
 									'description' => sprintf( __( 'The slug of the %1$s. If no ID is present, this field will be used to make a connection. If no existing term exists with this slug, this field will be used as a fallback to the Name field when creating a new term to connect to, if term creation is enabled as a nested mutation.', 'wp-graphql' ), $tax_object->graphql_single_name ),
 								],
 								'description'   => [
-									'type'        => Types::id(),
+									'type'        => Types::string(),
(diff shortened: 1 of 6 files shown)


src/Type/PostObject/Mutation/PostObjectMutation.php
```diff
@@ -95,6 +97,59 @@ public static function input_fields( $post_type_object ) {
...
+								'id'   => [
+									'type'        => Types::id(),
+									'description' => sprintf( __( 'The ID of the %1$s. If present, this will be used to connect to the %2$s. If no existing %1$s exists with this ID, no connection will be made.', 'wp-graphql' ), $tax_object->graphql_single_name, $post_type_object->graphql_single_name ),
+								],
+								'slug' => [
+									'type'        => Types::string(),
+									'description' => sprintf( __( 'The slug of the %1$s. If no ID is present, this field will be used to make a connection. If no existing term exists with this slug, this field will be used as a fallback to the Name field when creating a new term to connect to, if term creation is enabled as a nested mutation.', 'wp-graphql' ), $tax_object->graphql_single_name ),
+								],
+								'description'   => [
+									'type'        => Types::id(),
```
hughdevore: Shouldn't this be the `string` type? Also, maybe add this field to your tests.

src/Type/PostObject/Mutation/PostObjectMutation.php
```diff
@@ -95,6 +97,59 @@ public static function input_fields( $post_type_object ) {
...
+									'description' => sprintf( __( 'The description of the %1$s. This field is used to set a description of the %1$s if a new one is created during the mutation.', 'wp-graphql' ), $tax_object->graphql_single_name ),
+								],
+								'name'   => [
+									'type'        => Types::id(),
```
hughdevore: Same type issue here, maybe add this field to your tests.

src/Type/PostObject/Mutation/PostObjectMutation.php
hughdevore: Never-mind about the tests, it looks like your tests cover those lines and they're passing. 🤔

jasonbahl: yes, it should be a string. 

The ID type is flexible and can be a string or integer, but this _should_ be a string

src/Type/PostObject/Mutation/PostObjectMutation.php
jasonbahl: same, should be a string, so I'll update

jasonbahl:
@hughdevore updated the Types for name and description. Good catch!

PR #372, opened 2018-01-19
WP-CLI Improvements

renatonascalves:
WP-CLI has no internationalization support. See https://github.com/wp-cli/handbook/issues/193
Added also minor updates to the cli command.
`wp graphql generate` alias added.


jasonbahl:
@renatonascalves thanks! I didn't realize this. Looks like part of the Travis build failed, but from what I can tell it's not related to your change. Checking it out to make sure, then I'll merge.

PR #376, opened 2018-01-26
fix: hasPassword type should be boolean

### LINKED ISSUE #374: Query for posts where hasPassword: false should not return password-protected posts
screen shot 2018-01-25 at 4 59 13 pm

ginatrapani:
Closes #374

### DIFF
--- a/tests/wpunit/PostObjectConnectionQueriesTest.php
+++ b/tests/wpunit/PostObjectConnectionQueriesTest.php
@@ -273,7 +273,6 @@ class PostObjectConnectionQueriesTest extends \Codeception\TestCase\WPTestCase {
 			'post_title'    => 'Password protected',
 			'post_type'     => 'post',
 			'post_status'   => 'publish',
-			'has_password'  => true,
 			'post_password' => 'password',
 		] );
 
@@ -288,15 +287,33 @@ class PostObjectConnectionQueriesTest extends \Codeception\TestCase\WPTestCase {
 		 * GraphQL query posts that have a password
 		 */
 		$variables = [
-			'hasPassword' => true,
+			'where' => [
+				'hasPassword' => true,
+			],
 		];
-		$graphql_query_posts_with_password = $this->postsQuery( $variables );
 
-		$this->assertNotEmpty( $graphql_query_posts_with_password );
-		$this->assertEquals(
-			$wp_query_posts_with_password->posts[0]->ID,
-			$graphql_query_posts_with_password['data']['posts']['edges'][0]['node']['postId']
-		);
+		$request = $this->postsQuery( $variables );
+
+		$this->assertNotEmpty( $request );
+		$this->assertArrayNotHasKey( 'errors', $request );
+
+		$edges = $request['data']['posts']['edges'];
+		$this->assertNotEmpty( $edges );
+
+		/**
+		 * Loop through all the returned posts
+		 */
+		foreach ( $edges as $edge ) {
+
+			/**
+			 * Assert that all posts returned have a password, since we queried for
+			 * posts using "has_password => true"
+			 */
+			$password = get_post( $edge['node']['postId'] )->post_password;
+			$this->assertNotEmpty( $password );
+
+		}
+
 	}
 
 	public function testPageWithChildren() {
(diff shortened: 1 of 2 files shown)


jasonbahl:
@ginatrapani any chance you'd be interesting in adding a unit test that does the following:

* create post with password
* use a GraphQL request to query for posts that have a password, using this argument
* assert that that any posts returned do indeed have passwords

If you're not familiar with or comfortable writing unit tests, no worry, just thought I'd check

jasonbahl:
also, @ginatrapani thanks for spending time digging and working on this. very much appreciated!!!

ginatrapani:
@jasonbahl sure thing! I added the test in 1d9453ca97b7b0c96043108271f8d9e76642d453 -- but it is failing, and I'm not familiar enough with debugging in Codeception to be able to see what's wrong, but I suspect it's something really obvious to someone who has more experience. Take a look and let me know if you have any hints. (Also, getting the tests set up changed several files, that changeset is in bc9c7951810ac2c612f77802466240966f4e7460 )

jasonbahl:
@ginatrapani I'll take a look and see what's up. On first glance the test looks good. 

As far as the codeception stuff goes, I need to document (like most things) it better. . .but when you're done testing, if you run `rm -rf composer.lock vender` then `composer install --no-dev` it should remove the test files. 

---- 
Aside: 

Ultimately, we shouldn't be versioning the vendor files (kind of an anti-pattern), but I opted to version them so that folks not familiar with composer could just download the plugin from Github, activate and start using. . .but we should probably setup some build script that installs the vendor dependencies and removes dev dependencies for folks to be able to download. . . 🤔 

That way, devs can check this repo out and have all the dev dependencies they need, and end users can download a zip prepared for install & activation.

jasonbahl:
@ginatrapani I checked Travis CI, and it looks like the composer changes is what's causing the fail. . .so if you run those commands, then the PR _should_ just have the 2 files changed, `src/Type/PostObject/Connection/PostObjectConnectionArgs.php` and the test file. . .then things should pass again.

Thanks so much for your work on this!

ginatrapani:
Sure thing @jasonbahl! I force-pushed the branch minus the test changes so the PR is just 2 changed files. Looks like I'm still getting that failure though. 😢 I can take a deeper look again tomorrow! thanks for your help and have a good night!

tests/wpunit/PostObjectConnectionQueriesTest.php
```diff
@@ -265,6 +267,38 @@ public function testMaxQueryAmount() {
...
+			'post_type'     => 'post',
+			'post_status'   => 'publish',
+			'has_password'  => true,
+			'post_password' => 'password',
+		] );
+
+		/**
+		 * WP_Query posts with a password
+		 */
+		$wp_query_posts_with_password = new WP_Query( [
```
jasonbahl: try setting `"posts_per_page" => 1` here

I think there's some other test data that's causing the queries to return more than 1 item, and due to how WPGraphQL cursor pagination works, the default order of a GraphQL query isn't a perfect match with default WP_Query, so I think multiple items are being returned, just not in the correct order?

The other option, would be to not run a WP_Query at all here, but instead do something like: 

```
$graphql_query_posts_with_password = $this->postsQuery( $variables );
$edges = $graphql_query_posts_with_password['data']['posts']['edges'];

foreach ( $edges as $post_edge ) {
   $post = get_post( $post_edge['node']['postId'] );
   $this->assertTrue( ! empty( $post->post_password) );
}
```

This way, even if there is other test data, we still are verifying that only posts with a password set are being returned when using the `hasPassword` argument in a GraphQL request.

ginatrapani: So, I tried this out this morning and it didn't work. The test is failing b/c posts without a password are in fact getting returned. I'm not sure why, b/c when I query for posts with a password against my own WordPress install on Graphiql, the fix works. Maybe there's something about how I'm setting up the test data that isn't correct?

jasonbahl: Hmm. Maybe. Let me take another look

jasonbahl:
@ginatrapani I made some adjustments to the test. 

The primary issue was the shape of the arguments: https://github.com/wp-graphql/wp-graphql/pull/376/commits/391a428a7c18c8456d942ffd241beb00ec82077f#diff-4232742370ce60dbe32afbf8e87e3476L291

I also adjusted the assertions we're making a bit just to make sure that we're not asserting a specific post is returned, but rather that all posts returned indeed do have a password.

jasonbahl:
merging. Thanks for working on this @ginatrapani!

ginatrapani:
Ahhh rookie mistakes! Thanks for fixing and merging @jasonbahl.

PR #392, opened 2018-02-09
Add support for category__and in post queries

jrabramson:
Just some simple support for category__and

### DIFF
--- a/src/Type/PostObject/Connection/PostObjectConnectionArgs.php
+++ b/src/Type/PostObject/Connection/PostObjectConnectionArgs.php
@@ -119,6 +119,11 @@ class PostObjectConnectionArgs extends WPInputObjectType {
 					'description' => __( 'Array of category IDs, used to display objects from one 
 										category OR another', 'wp-graphql' ),
 				],
+				'categoryAnd'   => [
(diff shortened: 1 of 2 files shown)


jasonbahl:
@jrabramson so, at the moment, we're trying to leave out some arguments like this that are known to be problematic. 

See: https://vip.wordpress.com/documentation/code-review-what-we-look-for/#functions-that-use-joins-taxonomy-relation-queries-cat-tax-queries-subselects-or-api-calls

For systems that feel they _need_ this and it won't be harmful to their system, they can filter in the support for their system, but we've opted to stay on the cautious side of only exposing queries we feel relatively good about performance wise. 

I'm open to more discussion though, if you feel there's good reason to include this even with the potential performance issues in mind. . .

jasonbahl:
@jrabramson also, I should've started first by saying Thanks for the PR and interest in using/contributing to the plugin!!! Very much appreciated!

jrabramson:
No problem, and thanks for bringing that to my attention!

jrabramson:
I'll implement this on our app and we'll monitor its impact, thanks again for showing me that list, very helpful.

PR #395, opened 2018-02-13
Set post status for Post Object Create mutations late, to allow for side-effects before final status is set

### LINKED ISSUE #393: Add context and info params to all "xxx_mutation_update_additional_data" hooks
### When you open an _issue_ for a feature request, please add as much detail as possible:
- [x] A descriptive title
- [x] A description of the problem you're trying to solve, including *why* you think this is a problem
- [x] An overview of the suggested solution
- [ ] If the feature changes current behavior, reasons why your solution is better

We are currently unable to utilize the Relay app context and app info in mutation hooks. Being able to do this would allow us to vet incoming mutation requests more easily.

As a developer, I would like to be able to utilize the app `context` and `info` parameters returned by the `Relay::mutationWithClientMutationId()` method by passing them along to the following hooks:

`graphql_media_item_mutation_update_additional_data`
`graphql_post_object_mutation_update_additional_data`
`graphql_user_object_mutation_update_additional_data`

### LINKED ISSUE #394: Set status of PostObjects at the end of a create mutation
### When you open an _issue_ for a feature request, please add as much detail as possible:
- [x] A descriptive title
- [x] A description of the problem you're trying to solve, including *why* you think this is a problem
- [x] An overview of the suggested solution
- [x] If the feature changes current behavior, reasons why your solution is better

## Description
When posts are created via GraphQL, the status of the post should be set late in the process, allowing for side effects (setting term objects, updating various meta, sideloading images, etc) to happen before a post is set to publish. 

The way this should work is that the status from the input (if any) should be cached for later use. The post should be inserted with the site's default post_status, then side effects should happen, then the status, (if any) should be set, if enabled (filterable to true by default). 

This will allow for certain side effects that might take a while (sideloading images, for example), to prevent a post from being published until the side effect is completed.

jasonbahl:

During PostObjectMutations, any post_status input will be cached, and the post will be created with the default status (draft, but filterable to accompany plugins like Edit Flow, etc).

Then the side-effects will be triggered via `PostObjectMutation::update_additional_post_object_data()`. 

Then, if `$should_set_intended_status` is true (filterable), the post_status that was input with the mutation (if different than the default) will be set as the status of the post.

This allows for side effects that may take a long time (sideloading images, etc) to happen in a deferred process (like a WPCron or a WP Queue Task - https://github.com/dfmedia/wp-queue-tasks) and those side effects can come back later and set the status when they complete.

closes #394 
closes #393 

Where has this been tested?
---------------------------
**Operating System:**
Mac OSX 10.12.6

**WordPress Version:**
4.9.4

### DIFF
--- a/src/Type/PostObject/Mutation/PostObjectCreate.php
+++ b/src/Type/PostObject/Mutation/PostObjectCreate.php
@@ -91,6 +91,31 @@ class PostObjectCreate {
 					$post_args = PostObjectMutation::prepare_post_object( $input, $post_type_object, $mutation_name );
 
 					/**
+					 * Filter the default post status to use when the post is initially created. Pass through a filter to
+					 * allow other plugins to override the default (for example, Edit Flow, which provides control over
+					 * customizing stati or various E-commerce plugins that make heavy use of custom stati)
+					 *
+					 * @param string $default_status The default status to be used when the post is initially inserted
+					 * @param \WP_Post_Type $post_type_object The Post Type that is being inserted
+					 * @param string $mutation_name The name of the mutation currently in progress
+					 */
+					$default_post_status = apply_filters( 'graphql_post_object_create_default_post_status', 'draft', $post_type_object, $mutation_name );
+
+					/**
+					 * We want to cache the "post_status" and set the status later. We will set the initial status
+					 * of the inserted post as the default status for the site, allow side effects to process with the
+					 * inserted post (set term object connections, set meta input, sideload images if necessary, etc)
+					 * Then will follow up with setting the status as what it was declared to be later
+					 */
+					$intended_post_status = ! empty( $post_args['post_status'] ) ? $post_args['post_status'] : $default_post_status;
+
+					/**
+					 * Set the post_status as the default for the initial insert. The intended $post_status will be set after
+					 * side effects are complete.
+					 */
+					$post_args['post_status'] = $default_post_status;
+
+					/**
 					 * Insert the post and retrieve the ID
 					 */
 					$post_id = wp_insert_post( wp_slash( (array) $post_args ), true );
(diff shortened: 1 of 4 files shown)


src/Type/PostObject/Mutation/PostObjectCreate.php
```diff
@@ -122,6 +147,43 @@ public static function mutate( \WP_Post_Type $post_type_object ) {
...
+
+						$update_args = [
+							'ID' => $post_id,
+							'post_status' => $intended_post_status,
+							// Prevent the post_date from being reset if the date was included in the create post $args
+							// see: https://core.trac.wordpress.org/browser/tags/4.9/src/wp-includes/post.php#L3637
+							'edit_date' => ! empty( $post_args['post_date'] ) ? $post_args['post_date'] : false,
+						];
+
+						$update_args = array_merge( $update_args );
```
CodeProKid: I think you are missing the second array you are trying to merge in here, eh?

src/Type/PostObject/Mutation/PostObjectCreate.php
```diff
@@ -122,6 +147,43 @@ public static function mutate( \WP_Post_Type $post_type_object ) {
...
+						wp_update_post( $update_args );
```
CodeProKid: Shouldn't this get moved out of this if statement so it fires even if the filter is set to false?

src/Type/PostObject/Mutation/PostObjectCreate.php
jasonbahl: good catch, I did have a merge, but changed it.

src/Type/PostObject/Mutation/PostObjectCreate.php
jasonbahl: No. The intent is that a plugin, like DFM GraphQL Extensions, that needs to process sideloaded images, can filter this to false, meaning the created post will remain a draft. Then the process that sideloads the images can come back and set the post as `Publish` or whatever the intended status was. 

Otherwise, if the status was set to publish and the deferred (cron / queued tasks, etc) side effects were not complete, then you'd have a published post with incomplete data.

jasonbahl:
@CodeProKid pushed an update, cleaning up the ArrayMerge, and including some additional context in some of the hooks/filters (requested by @hughdevore for some of the work he's doing to sideload galleries)

jasonbahl:
@CodeProKid updated to fix the broken tests too.

Issue #406, opened 2018-02-19
WP_Query pagination uses post_date and post ID in comparison, which can exclude posts


chriszarate:
The `where` clause returned by `Config.php#graphql_wp_query_cursor_pagination_support` compares against both the `post_date` and the post `ID` of the cursor post. For example, take the following query (using default orderby field of `DATE` and default `order` of `DESC`):

```
query {
  posts(
    after: "cG9zdDoxMA==" // post:10
 ) { ... }
}
```

The pagination `where` clause produced for this GraphQL query looks for posts with a `post_date` less than that of post 10 _and_ a post `ID` greater than `10`. This is helpful in making sure that posts that might have the same `post_date` do not get duplicated across boundaries.

However, a flaw exists around the assumption that posts with a lower post `ID` will also have an earlier `post_date`. It is sometimes the case (often for us) that posts are published out of order (e.g., post 5 lives as a draft for a long time and is published after posts 6 and 7). The current pagination logic will miss posts when the out-of-order post is the cursor post.

This issue is present in all versions of WPGraphQL.

jasonbahl:
closed by #409

PR #407, opened 2018-02-19
Adjust WP_Query cursor-based pagination to use only post_date (and not post ID)

chriszarate:
This is an attempt to address #406 (see for details).

This adjusts the where clause generated by WP_Query pagination logic to use only `post_date` and not post `ID` to accommodate out-of-order post publishing.

I've added some tests that will produce failures in the current codebase but pass with these changes. They simulate publishing posts out-of-order, such that post ID `12` is not guaranteed to be published after post ID `11` (as an example).

One side-effect of this change is that posts could get duplicated across pagination boundaries _if_ the cursor post shares the exact same publish date (to the second) with another post. In that instance, there is no way to determine which side of the boundary the post belongs to (without, say, storing query results in a state store). However, this seems to be a commonly accepted side-effect of cursor-based pagination.

### DIFF
--- a/src/Data/Config.php
+++ b/src/Data/Config.php
@@ -89,7 +89,7 @@ class Config {
 							}
 						}
 					} else {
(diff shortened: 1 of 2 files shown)


jasonbahl:
@chriszarate thanks for working on this! 

I'll check it out in detail later tonight or tomorrow. Really appreciate the details and the included tests!

jasonbahl:
@chriszarate github/travis are being weird with this after I merged the Docker stuff. . .going to close this PR and re-open it.

PR #409, opened 2018-02-21
Fix adjustments to WP_Query pagination to use only post_date.

jasonbahl:
This is an attempt to address #406 (see for details).

This adjusts the where clause generated by WP_Query pagination logic to use only `post_date` and not post `ID` to accommodate out-of-order post publishing.

I've added some tests that will produce failures in the current codebase but pass with these changes. They simulate publishing posts out-of-order, such that post ID `12` is not guaranteed to be published after post ID `11` (as an example).

One side-effect of this change is that posts could get duplicated across pagination boundaries _if_ the cursor post shares the exact same publish date (to the second) with another post. In that instance, there is no way to determine which side of the boundary the post belongs to (without, say, storing query results in a state store). However, this seems to be a commonly accepted side-effect of cursor-based pagination.

### DIFF
--- a/tests/wpunit/DataConfigTest.php
+++ b/tests/wpunit/DataConfigTest.php
@@ -1,0 +1,87 @@
+<?php
+
+use WPGraphQL\Data\Config;
+
+class DataConfigTest extends \Codeception\TestCase\WPTestCase {
+	public static function setUpBeforeClass() {
+		parent::setUpBeforeClass();
+
+		if ( ! defined( 'GRAPHQL_REQUEST' ) ) {
+			define( 'GRAPHQL_REQUEST', true );
+		}
+	}
+
+	/**
+	 * Create n posts, making sure that publish dates are offset by at least one
+	 * second. The posts can either simulate being published in the same order as
+	 * the post IDs (e.g., 1, 2, 3, ... 10) or out of order (e.g., 10, 9, ... 1).
+	 */
+	private function create_posts( $count, $operator, $offset_multiplier = 1 ) {
+		$iterable = array_keys( array_fill( 0, $count, null ) );
+		$timestamp = time();
+
+		$posts = array_map( function ( $offset ) use ( $timestamp, $offset_multiplier ) {
+			return $this->factory->post->create_and_get(
+				array(
+					'post_date' => date( 'Y-m-d H:i:s', $timestamp + ( $offset * $offset_multiplier ) ),
+				)
+			);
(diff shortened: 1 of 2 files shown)


jasonbahl:
@chriszarate here's the PR re-opened. Taking a look again. Seeing if I have any thoughts/concerns about the pagination changes.

jasonbahl:
@chriszarate it looks like some of the `DataConfigTest` tests are intermittently passing and failing, so I think we have some edge case we need to investigate. 

I haven't looked in detail quite yet. It _could_ potentially be something we can change in the assertions, or it might be an actual bug we need to fix with the pagination change you made. 

I'll take a look later today or tomorrow if you don't get to it first. 

You can check the past few travis builds and you can see that they sometimes pass and sometimes fail. 🤔

jasonbahl:
examples: 
https://travis-ci.org/wp-graphql/wp-graphql/builds/344456372
https://travis-ci.org/wp-graphql/wp-graphql/builds/343646701
https://travis-ci.org/wp-graphql/wp-graphql/builds/343572914

chriszarate:
Hmmm, that's a bummer. I'll take a look tomorrow.

tests/wpunit/DataConfigTest.php
```diff
@@ -18,7 +18,10 @@ public static function setUpBeforeClass() {
 	 */
 	private function create_posts( $count, $operator, $offset_multiplier = 1 ) {
 		$iterable = array_keys( array_fill( 0, $count, null ) );
-		$timestamp = time();
+
+		// Make sure starting timestamp is sufficiently in the past so that posts
+		// do not receive a post_status of "future".
+		$timestamp = time() - 1000;
```
jasonbahl: nice, I had a feeling it was something like this. thanks for fixing!

chriszarate:
@jasonbahl Found the reason for the intermittent failing tests. I was creating posts with a time offset so that I could reorder them, but in doing so I was creating posts with publish dates in the future. Depending on how long the tests took to run, this _sometimes_ resulted in posts with a `post_status` of `future`, which made them invisible to queries looking for a `post_status` of `publish`.

Simple fix: put the starting timestamp in the past to avoid the possibility of a `post_status` of `future`. Previously the tests would fail about 1 out of 6 times. I've run the tests about 50 times now with now failures. Feel free to run the tests on Travis a bunch.

jasonbahl:
@chriszarate thanks for fixing this!

PR #427, opened 2018-04-07
#405 - PostObject Types should only have fields they have support for

### LINKED ISSUE #405: PostObject fields should only be included if post_type supports them
There are several fields that are "optional" when registering a Post Type: https://codex.wordpress.org/Function_Reference/register_post_type#supports

These fields should only be added to a PostTypeObject in the Schema if the Post Type has registered support for them, and fallback to just the defaults (title/content) if nothing is explicitly defined in the "supports" array of the post_type definition.

jasonbahl:
# BREAKING CHANGE

This is a breaking change as it removes fields from the Schema. (both query fields and mutation input field)

This adjusts the schema for PostObjects to only expose fields if the post_type registry registers support for them. 

The fields affected are those that `post_type_supports` allows: https://codex.wordpress.org/Function_Reference/post_type_supports

Some fields (comments, thumbnail) were already added conditionally based on post_type support, but this carries that logic to the other post_types.  

closes #405

### DIFF
--- a/src/Type/PostObject/Mutation/PostObjectMutation.php
+++ b/src/Type/PostObject/Mutation/PostObjectMutation.php
@@ -33,29 +33,9 @@ class PostObjectMutation {
 		if ( ! empty( $post_type_object->graphql_single_name ) && empty( self::$input_fields[ $post_type_object->graphql_single_name ] ) ) {
 
 			$input_fields = [
-				'authorId'      => [
-					'type'        => Types::id(),
-					'description' => __( 'The userId to assign as the author of the post', 'wp-graphql' ),
-				],
-				'commentCount'  => [
-					'type'        => Types::int(),
-					'description' => __( 'The number of comments. Even though WPGraphQL denotes this field as an integer, in WordPress this field should be saved as a numeric string for compatability.', 'wp-graphql' ),
-				],
-				'commentStatus' => [
-					'type'        => Types::string(),
-					'description' => __( 'The comment status for the object', 'wp-graphql' ),
-				],
-				'content'       => [
-					'type'        => Types::string(),
-					'description' => __( 'The content of the object', 'wp-graphql' ),
-				],
 				'date'          => [
 					'type'        => Types::string(),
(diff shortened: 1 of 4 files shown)


src/Type/PostObject/Mutation/PostObjectMutation.php
```diff
@@ -89,16 +69,53 @@ public static function input_fields( $post_type_object ) {
...
-					'type'        => Types::string(),
-					'description' => __( 'The title of the post', 'wp-graphql' ),
-				],
 				'toPing'        => [
 					'type'        => Types::list_of( Types::string() ),
 					'description' => __( 'URLs queued to be pinged.', 'wp-graphql' ),
 				],
 			];
 
+			if ( post_type_supports( $post_type_object->name, 'author' ) ) {
```
chriszarate: i might suggest to keep the definitions in the array above and `unset` them here if they aren't supported. this keeps all the definitions together and in alpha order for readability. just a mild preference though

jasonbahl: @chriszarate I like that idea

src/Type/PostObject/PostObjectType.php
```diff
@@ -363,7 +291,14 @@ private static function fields( $post_type_object ) {
 							return ! empty( $uri ) ? $uri : null;
 						},
 					],
-					'terms'             => [
+				];
+
+				/**
+				 * Only if the post_type has taxonomies associated with it should it get these fields
+				 */
+				if ( ! empty( $object_taxonomies = get_object_taxonomies( $post_type_object->name ) ) ) {
```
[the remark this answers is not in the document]
jasonbahl: @chriszarate I'll adjust. I think that makes things much cleaner

jasonbahl:
@chriszarate So I have a refactored version of this now, but looking at it, my concern is that anyone else looking at the codebase will now have a harder time understanding why these fields exist sometimes and why they don't, because the logic of adding the field if the post_type supports it is now decoupled more from the field.

Guess we just gotta decide what trade-off is better. . .having the array of fields be defined alphabetically making it easier to scan the list, or having the logic that sets the field if the post_type supports it be closer to the field definition. 🤔

jasonbahl:
this needs to be completely re-implemented to fit the v0.1.0 re-architecture of the plugin. Closing this PR, but we can reference it again next time we give a go at #405

PR #431, opened 2018-04-20
Prevent CORS-error in Chrome when using Apollo

wvoelcker:
We saw an error in Chrome complaining that the error "Access-Control-Allow-Credentials" was an empty string when it needed to be "true"; this corrects for that.

We saw an error in Chrome, when using graphql cross-domain, complaining that the error "Access-Control-Allow-Credentials" was an empty string when it needed to be "true"; this corrects for that.

No

No

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …
Mac OS, Chrome

**WordPress Version:** …
4.9.4

### DIFF
--- a/src/Router.php
+++ b/src/Router.php
@@ -203,6 +203,7 @@ class Router {
 		$headers = [
 			'Access-Control-Allow-Origin'  => '*',
 			'Access-Control-Allow-Headers' => implode( ', ', $access_control_allow_headers ),
+			'Access-Control-Allow-Credentials' => 'true',
 			'Content-Type'                 => 'application/json ; charset=' . get_option( 'blog_charset' ),
 			'X-Robots-Tag'                 => 'noindex',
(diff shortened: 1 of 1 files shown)


jasonbahl:
@wvoelcker first off, thanks for the PR! Much appreciated. 

Can you provide insight into what your requests from Apollo look like. I’ve not ran into this issue with Apollo. 

Looks like something specific to authentication. In that case I think whatever Auth mechanism you’re using should probably be responsible for filtering this header in but leave it out by default. 

For example, if you were using wp-graphql-jwt-authentication plugin, I think that plugin should handle filtering in this header. 

Open to discussion if you feel like it should be in the core plugin and not something to be filtered in.

wvoelcker:
@jasonbahl thank you for this; I'll have to ask my colleagues who saw the error for more information and get back to you.

Issue #436, opened 2018-05-02
Fix Orderby mapping for post__in, post_name__in and post_parent__in


jasonbahl:
The orderby mapping for `post__in`, `post_name__in`, and `post_parent__in` should _not_ include the order (ASC/DESC) value as it's not respected. 

### Proposed Solution
In PostObjectConnectionResolver.php where the `$args['where']['orderby']` inputs are mapped to the WP_Query args, these fields should be mapped to just a `orderby => field` instead of `orderby => [ field => order ]`

jasonbahl:
closed by #437

PR #437, opened 2018-05-02
#436 - fix orderby mapping of post__in, post_name__in and post_parent__in

jasonbahl:
This fixes the orderby mapping of post__in, post_name__in and post_parent__in so that the order value is not included as that is not respected by WP_Query. See: https://codex.wordpress.org/Class_Reference/WP_Query#Order_.26_Orderby_Parameters

addresses #436

Here's screenshots of before/after showing the order not properly working, then working with the fix in place.

screen shot 2018-05-02 at 10 01 51 am

screen shot 2018-05-02 at 10 02 12 am


CodeProKid:
Was this causing those three fields not to work?

jasonbahl:
@CodeProKid yes

Probably should be good to get some specific tests added 🤔

PR #438, opened 2018-05-02
Release/v0.0.28

jasonbahl:
- Updates to the main README to make it easier to find some resources about plugin usage
- Introduce `graphql_access_control_allow_headers` filter
- Adjust some sanitization re: GET request handling
- Fixes a bug with undefined variables for GET requests
- Adds `stati` arg to PostObjectConnectionArgs
- Fixes bug re: ordering by `in`, `notIn` and `parentIn`


tests/functional/PostListWithDifferentStatusCept.php
```diff
@@ -0,0 +1,58 @@
...
+	$I->havePostInDatabase( [
+		'post_type'    => 'post',
+		'post_status'  => $post_status,
+		'post_title'   => "test {$post_status} post",
+		'post_content' => "test {$post_status} content"
+	] );
+}
+
+/**
+ * Set the content-type so we get only posts with status published and draft.
```
chriszarate: comment makes it sound like content-type is at play

chriszarate: https://github.com/wp-graphql/wp-graphql/pull/439

src/Type/PostObject/Connection/PostObjectConnectionArgs.php
```diff
@@ -243,6 +243,13 @@ private static function fields( $connection ) {
 					'type' => Types::post_status_enum(),
 				],
 
+				/**
+				 * List of post status parameters
+				 */
+				'stati' => [
```
chriszarate: I understand the logic here to mirror WP_Query functionality. I missed the initial PR; my only question is whether we considered consolidating on a single arg that always takes an array of statuses, even if you are only querying one, e.g.:

```
status: ['draft']
```

```
status: ['draft', 'publish']
```

jasonbahl: @chriszarate right now the `status` arg accepts an individual status and this introduces a plural `stati` which accepts a list, and does so in a non-breaking fashion.

tests/functional/PostListWithDifferentStatusCept.php
jasonbahl: Thanks for fixing

src/Type/PostObject/Connection/PostObjectConnectionArgs.php
chriszarate: Makes sense, just raising opportunities for consolidation. Glad to have this functionality!

PR #446, opened 2018-05-24
Nav Menu support

chriszarate:
Implements navigation menu support with two new types: `Menu` and `MenuItem`. Each have singular and plural root queries. `MenuItems` are connected to both `Menus` and to themselves—this allows for the implementation of `childItems` (hierarchical nav menu queries):

```
{
  menus( location: MY_MENU ) {
    edges {
      node {
        id
        menuItems {
          edges {
            node {
              id
              childItems {
                edges {
                  node {
                    id
                    connectedObject {
                      ... on Post {
                        id
                        title
                        content
                      }
                    }
                  }
                }
              }
              connectedObject {
                ... on Post {
                  id
                  title
                  content
                }
              }
            }
          }
        }
      }
    }
  }
}
```

MenuItems can also be queried directly with an `id` or `location`:

```
{
  menuItems( location: MY_MENU ) {
    edges {
      node {
        id
        connectedObject {
          ... on Post {
            id
            title
            content
          }
        }
      }
    }
  }
}
```

Menu locations are implemented as an Enum. This allows the schema to communicate which menu locations have been registered for the current active theme.

The `connectedObject` field returns a Union of types that can be added the nav menus: post types, taxonomy terms, and custom links (itself a `MenuItem`). This provides consistent communication of the information about a nav menu object (which is usually the thing the user is trying to query).

MenuItems are tightly tied to the hierarchy of their Menu (parent-child relationships are stored in post meta), so we cannot resolve MenuItems if we cannot resolve the Menu that they belong to. In other words, this query returns an empty array:

```
{
  menuItems {
    edges {
      node {
        id
      }
    }
  }
}
```

(We probably could implement this, but it would require alot of expensive meta querying.)

Addresses #126.

### DIFF
--- a/src/Type/MenuItem/Connection/MenuItemConnectionDefinition.php
+++ b/src/Type/MenuItem/Connection/MenuItemConnectionDefinition.php
@@ -1,0 +1,115 @@
+<?php
+
+namespace WPGraphQL\Type\MenuItem\Connection;
+
+use GraphQLRelay\Relay;
+use WPGraphQL\Types;
+use WPGraphQL\Type\WPInputObjectType;
+
+/**
+ * Class MenuItemConnectionDefinition
+ *
+ * @package WPGraphQL\Type\MenuItem\Connection
+ * @since 0.0.29
+ */
+class MenuItemConnectionDefinition {
+
+	/**
+	 * Stores some date for the Relay connection for term objects
+	 *
+	 * @var array $connection
+	 * @access private
+	 */
+	private static $connection;
+
+	/**
+	 * Stores the where_args Input Object Type
+	 *
+	 * @var \WPGraphQL\Type\WPInputObjectType $where_args
+	 */
+	private static $where_args;
+
+	/**
+	 * Stores the fields for the $where_args
+	 *
+	 * @var array $where_fields
+	 */
+	private static $where_fields;
+
+	/**
+	 * Method that sets up the relay connection for term objects
+	 *
+	 * @return mixed
+	 */
+	public static function connection() {
+
+		if ( null === self::$connection ) {
+
+			$connection = Relay::connectionDefinitions( [
+				'nodeType' => Types::menu_item(),
+				'name'     => 'MenuItems',
+			] );
+
+			$args = [
+				'where' => [
+					'name' => 'where',
+					'type' => self::where_args(),
+				],
+			];
+
+			self::$connection = [
+				'type'        => $connection['connectionType'],
+				'description' => __( 'A collection of menu item objects', 'wp-graphql' ),
+				'args'        => array_merge( Relay::connectionArgs(), $args ),
+				'resolve'     => [ __NAMESPACE__ . '\\MenuItemConnectionResolver', 'resolve' ],
+			];
+		}
+
+		return ! empty( self::$connection ) ? self::$connection : null;
+	}
+
+	/**
+	 * Defines the "where" args that can be used to query menuItems
+	 *
+	 * @return WPInputObjectType
+	 */
+	private static function where_args() {
+
+		if ( null === self::$where_args ) {
+			
+			self::$where_args = new WPInputObjectType( [
+				'name'   => 'MenuItemQueryArgs',
+				'fields' => function() {
+					return self::where_fields();
+				},
+			] );
+		}
+
+		return ! empty( self::$where_args ) ? self::$where_args : null;
+
+	}
+
+	/**
+	 * This defines the fields to be used in the $where_args input type
+	 *
+	 * @return array|mixed
+	 */
+	private static function where_fields() {
+		if ( null === self::$where_fields ) {
+			$fields = [
+				'id' => [
+					'type'        => Types::int(),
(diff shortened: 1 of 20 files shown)


src/Type/Union/MenuItemObjectUnionType.php
```diff
@@ -0,0 +1,97 @@
...
+	 * @since 0.0.5
+	 */
+	public static function get_possible_types() {
+		if ( is_array( self::$possible_types ) ) {
+			return self::$possible_types;
+		}
+
+		// We could restrict this further using `show_in_nav_menus`, but it's of
+		// questionable utility since we'd only be creating more work for those
+		// that want to implement custom resolution of menu items.
```
jasonbahl: I think I’m ok with just restricting to `show_in_graphql` As you have it. Thanks for calling this out in the comment though. 😀

src/Type/Union/MenuItemObjectUnionType.php
```diff
@@ -0,0 +1,97 @@
...
+use GraphQL\Type\Definition\UnionType;
+use WPGraphQL\Types;
+
+/**
+ * Class MenuItemObjectUnionType
+ *
+ * Navigation menus comprise menu items that reference an object, which can be
+ * a post object, a taxonomy term object, or a custom link.
+ */
+class MenuItemObjectUnionType extends UnionType {
```
jasonbahl: We should add a WPUnionType to the core plugin to help centralize filters, etc. 🤔

src/Type/MenuItem/MenuItemType.php
```diff
@@ -0,0 +1,189 @@
...
+									$resolved_object = get_term( $object_id );
+									break;
+							}
+
+							/**
+							 * Allow users to override how nav menu items are resolved.
+							 * This is useful since we often add taxonomy terms to menus
+							 * but would prefer to represent the menu item in other ways,
+							 * e.g., a linked post object (or vice-versa).
+							 */
```
jasonbahl: This resolver should probably include the other args that come along (`$args`, `$context`, `$info`) then pass those to this filter so anyone hooking to this filter has that extra resolve info to use if needed. 

Might be good to pass the $object_id and $object_type to the filter as well

chriszarate:
@jasonbahl Made some adjustments. Also restricted the menu location Enum to locations registered in the active theme, which was my intent.

However, it seems that you cannot have an "empty" Enum type (an error is thrown), so if there are no registered locations, I add a dummy location called `EMPTY`. Another option would be to deregister the `Menu` and `MenuItem` types when there are no registered locations, but that seems extreme and could lead to confusion.

jasonbahl:
@chriszarate thanks for the updates! I'll try and review today or tomorrow, but realistically it will probably be next week when I get to it.

jasonbahl:
@chriszarate re: 

> However, it seems that you cannot have an "empty" Enum type (an error is thrown), so if there are no registered locations, I add a dummy location called EMPTY. Another option would be to deregister the Menu and MenuItem types when there are no registered locations, but that seems extreme and could lead to confusion.

So, it looks like the WP-Admin doesn't show the Menu interface if there are not menus registered via `register_nav_menu`, which requires a `menu_location`

So, if there are no registered menus (and therefore no registered menu locations), then perhaps menus shouldn't be added to the Schema. Match behavior of wp-admin?

jasonbahl:
@chriszarate I checked this locally and I'm not properly getting menuItem labels. 

Haven't investigated too far yet, but here's a screenshot of my menu and a screenshot of the query with null values: 

screen shot 2018-06-04 at 2 46 41 pm

screen shot 2018-06-04 at 2 54 55 pm

chriszarate:
Will take a look at these

jasonbahl:
@chriszarate I made some adjustments to this PR. It resolves the label issue I was having and introduces the ability to query for nodes without going through edges. 

One thing to note is that it wasn't properly getting `wp_setup_nav_menu_item` applied. Changing one line in the PostObjectConnectionResolver fixed that. 

I see one of the tests is now failing so I'm looking into fixing that.

src/Type/MenuItem/Connection/MenuItemConnectionDefinition.php
```diff
@@ -49,17 +49,17 @@ public static function connection() {
...
-                        'nodes'        => [
-                            'type'        => Types::list_of( Types::menu_item() ),
-                            'description' => __( 'The nodes of the connection, without the edges', 'wp-graphql' ),
-                            'resolve'     => function( $source, $args, $context, $info ) {
-                                return ! empty( $source['nodes'] ) ? $source['nodes'] : [];
-                            },
-                        ],
-                    ];
-                },
+				'connectionFields' => function() {
```
jasonbahl: @chriszarate good catch here. . .apparently my PHP Storm settings on my new machine aren't updated. 🤔

jasonbahl:
@chriszarate I think I'm pretty happy with the state of this PR. 

@CodeProKid @hughdevore @0aveRyan any thoughts before I merge?

PR #448, opened 2018-05-31
Add docker-compose file for local / user testing.

chriszarate:
Provides a separate, isolated local environment for user testing in the browser, accessible by default at http://wpgraphql.example. Environment can be started with:

```
docker-compose -f docker-compose.dev.yml up -d
```

Where has this been tested?
---------------------------
Docker on OS X

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,0 +1,57 @@
+# This file provides a local environment that is *separate* and *isolated* from
+# the one provided by docker-compose.yml. This allows you to have one
+# environment for automated tests and another for user testing / browsing.
+#
+# Edit your /etc/hosts file to add an entry for wpgraphql.example, e.g.:
+#
+# 127.0.0.1 localhost wpgraphql.example
+#
+# Next, start the environment. The `-f` flag is necessary to restrict your
+# commands to this docker-compose file:
+#
+# docker-compose -f docker-compose.dev.yml up -d
+#
+# After about 30 seconds, you will be able to access http://wpgraphql.example
+# in your browser.
+#
+# Stop the environment with:
+#
+# docker-compose -f docker-compose.dev.yml stop
+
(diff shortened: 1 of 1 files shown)


jasonbahl:
@chriszarate pardon my ignorance re: Docker, but when I checkout this PR and run the command, I get the following response: 

```
Pulling mysql (mariadb:10.2)...
ERROR: Get https://registry-1.docker.io/v2/library/mariadb/manifests/10.2: unauthorized: incorrect username or password
```

Do I need to have something configured locally already?

jasonbahl:
@chriszarate looks like I just had to logout and login via `docker logout` and `docker login`

I'll update the description at the top of the `docker-compose.dev.yml` file to specify this step for others that are unaware that you need to be logged in to docker via CLI for this to work.

jasonbahl:
@chriszarate I updated the description at the top of the file to call out that users need to make sure they're logged into Docker via `docker login`. 

Everything is working 👌 for me. Thanks for putting this together!

chriszarate:
@jasonbahl Sorry I missed your comments above. You don't need to `docker login` to use public images. What likely happened is that you had logged in at some point but your credentials expired, and Docker was asking you to resolve it before downloading any image (similar to how GitHub might react if you were cloning a public repo using the SSH url and there was an issue with your SSH key).

jasonbahl:
Ah. Ok. Let me try again with being logged out and see if that works for me. If so, I’ll update my comment to clarify better in case anyone else finds themselves in the same position. I don’t think I had logged in before though as this is a new machine I just got last Monday 🤔

jasonbahl:
@chriszarate ya, works fine when I am logged out. So I suppose I must've been logged in with bad creds like you mentioned. ¯\_(ツ)_/¯

I'll update the note at the top of the file to clarify that.

PR #450, opened 2018-06-08
Check for duplicate declaration of graphql_init

jmlallier:
This PR prevents a fatal error due to duplicate declaration of the `graphql_init` method in `wp-graphql.php`. This error is a possibility in the case that multiple themes or plugins include `wp-graphql` as a composer repository or the code is otherwise used in multiple locations in a single WP instance. 

Where has this been tested?
---------------------------
**Operating System:** MacOS 10.13.4

**WordPress Version:** 4.9.6

### DIFF
--- a/wp-graphql.php
+++ b/wp-graphql.php
@@ -731,17 +731,19 @@ if ( ! class_exists( 'WPGraphQL' ) ) :
 	}
 endif;
 
-/**
- * Function that instantiates the plugins main class
- *
- * @since 0.0.1
- */
-function graphql_init() {
-
+if ( ! function_exists( 'graphql_init' ) ) {
 	/**
-	 * Return an instance of the action
+	 * Function that instantiates the plugins main class
+	 *
+	 * @since 0.0.1
 	 */
-	return \WPGraphQL::instance();
+	function graphql_init() {
+
+		/**
+		 * Return an instance of the action
+		 */
+		return \WPGraphQL::instance();
+	}
 }
 graphql_init();
 


jasonbahl:
@jmlallier one thought re your comment, if you have your composer.json set the install path for the plugin to your plugins directory I believe you _shouldn't_ run into any issues re: having the codebase loading multiple times.

jmlallier:
@jasonbahl Hmm. I ran into the issue because we're experimenting with embedding the plugin in our plugin's codebase to provide an internal API for our custom tables. If another plugin or theme were to require it as well, there would be a conflict (at least that's what my happened when I installed WPGraphQL as a plugin separate from ours). We can't really require users to install WPGraphQL in order for our plugin to work, so we're going the dependency route. What are your thoughts on this?

jasonbahl:
Ya. . .WordPress needs a dependency management system 🤔 

In my head, was thinking initially that you could configure your plugin's composer.json to have a custom install path for WPGraphQL (like maybe `wp-content/vendor/` instead of `wp-content/plugins` so that it doesn't show in the plugin UI in the admin) then for all other plugins that require WPGraphQL as a dependency, you could use the same install path that way if you had one or many plugins that require WPGraphQL as a dependency it would _always_ be installed in the same place. 

_But_, I think that would require making some assumptions about the setup of the site so that you could statically define an install path in the composer.json files. . .and you simply wouldn't have enough info to know where your plugin is being loaded from to be able to define an install path within composer. 🤔 

For example, if you used something like: https://github.com/mnsami/composer-custom-directory-installer I was thinking you could just set the install path as `../vendor/wp-graphql` or something to that tune. . .but that gets funky if one of your plugins is installed as a regular plugin, one as an mu-plugin, one as a required plugin from within a theme, etc. . .

So, I think the way you're approaching it is probably the best way to go until WordPress (not happening anytime soon) figures out a better way of handling dependencies 🤔 

I'll go ahead and merge this.

jasonbahl:
@jmlallier one thing to note though, is that you could potentially have different versions of WPGraphQL active at once. . .if you have 2 plugins that each require WPGraphQL, and `a-plugin` requires `WPGraphQL v0.0.28` and `b-plugin` requires `WPGraphQL v0.0.29`, your site would be running `v0.0.28` as `a-plugin` would instantiate `v0.0.29` first.

Issue #435, opened 2018-05-01
Add featuredImage field to Post mutation


hughdevore:
We are currently able to query for a post's featured image:

```graphql
{
  posts {
    edges {
      node {
        postId
        title
        featuredImage {
          id
          title
          link
        }
      }
    }
  }
}
```

However, we are not able mutate this field when we create a post. I propose we add this field so that users are able to add/update/delete the featured image of a post. This could look something like:

```graphql
mutation {
  createPost(input: {
    title: "New Post with Fancy Image",
    featuredImage: {
      url: "images.com/sweet-image.jpg"
      title: "Sweet Featured Image",
    }
  }) {
    clientMutationId
    post {
      id
      title
      featuredImage {
        id
        title
        link
      }
    }
  }
}
```

The fields for the featuredImage mutation should match the fields of a mediaItem mutation.

jasonbahl:
@hughdevore do you plan to implement this via sideloading, or via a file upload and multipart form?

hughdevore:
@jasonbahl, planning on adding this via sideloading next sprint. I may have time to do the multipart as well but I think we should make it a separate issue.

jasonbahl:
@hughdevore so, I think for the core plugin we could support a nested `featuredImage` input on post mutations that accepts the ID of the featuredImage and sets the relationship during post creation/upates. . .but I don't think we should support uploading an image as a nested input. 

I believe the most common scenario for most applications would be to have separate mutations for creating images and creating posts, and just set the relationship to each other during their respective mutations. 

So, I think we can add a nested `featuredImage` input field, but it should expect an ID and simply set the relationship (if the ID is a valid media item).

hughdevore:
Agreed, I think that's a solid solution, keeping the mutations separate will help with maintenance in the long run. I'll work on a PR for adding this.

DSchau:
Just to chime in on this -- it's currently impossible to set a featured image on a post with WPGraphQL entirely, correct? Is there a workaround in lieu of this feature continuing to be worked upon?

jasonbahl:
@DSchau you can try using the REST API for the time being?

https://gist.github.com/ahmadawais/0ccb8a32ea795ffac4adfae84797c19a

jasonbahl:
I think the GraphQL ecosystem has settled on some “formal” ways of handling uploads, so I plan to revisit this soon.

DSchau:
[...]

PR #453, opened 2018-06-12
Feature/#435 - featured image field on post mutation

hughdevore:
# Add the `featuredImage` field to the post object if it supports featured images
This PR resolves issue #435.

### Featured Image creation on the createPost mutation
```graphql
mutation createPostWithFeaturedImage {
  createPost(input: {
  "input": {
    "title": "A post with a Featured Image",
    "content": "This is a post that was created with a featured image attached to it.",
    "featuredImage": {
      "sourceUrl": "https://media.giphy.com/media/Z6f7vzq3iP6Mw/giphy.gif",
      "title": "Awesome"
    },
    "clientMutationId": "createThatPostandImage"
  }
}) {
    post {
      id
      postId
      title
      link
      featuredImage {
        id
        mediaItemId
        title
        sourceUrl
      }
    }
  }
}
```

### Featured Image creation on the updatePost mutation
```graphql
mutation updatePostWithFeaturedImage($input: UpdatePostInput!) {
  updatePost(input: {
    "id": "cG9zdDoyNA==",
    "featuredImage": {
      "sourceUrl": "https://media.giphy.com/media/Z6f7vzq3iP6Mw/giphy.gif",
      "title": "Awesome"
    },
    "clientMutationId": "updateThatPostandCreateAnImage"
  }) {
    post {
      id
      postId
      title
      link
      featuredImage {
        id
        mediaItemId
        title
        sourceUrl
      }
    }
  }
}
```

Where has this been tested?
---------------------------
**Operating System:** Mac OSX (Sierra)

**WordPress Version:** 4.9.6

### DIFF
--- a/tests/wpunit/PostObjectMutationsTest.php
+++ b/tests/wpunit/PostObjectMutationsTest.php
@@ -1142,7 +1142,7 @@ class PostObjectMutationsTest extends \Codeception\TestCase\WPTestCase {
 					'post' => [
 						'featuredImage' => [
 							'title'     => $new_image_variables['featuredImage']['title'],
-							'sourceUrl' => $actual['data']['updatePost']['post']['updatePost']['sourceUrl'],
+							'sourceUrl' => $actual['data']['updatePost']['post']['featuredImage']['sourceUrl'],
 						],
 					],
 				],
--- a/tests/wpunit/PostObjectMutationsTest.php
+++ b/tests/wpunit/PostObjectMutationsTest.php
@@ -1142,7 +1142,7 @@ class PostObjectMutationsTest extends \Codeception\TestCase\WPTestCase {
 					'post' => [
 						'featuredImage' => [
 							'title'     => $new_image_variables['featuredImage']['title'],
-							'sourceUrl' => 'http://wpgraphql.test/wp-content/uploads/'. date("Y") . '/' . date('m') . '/giphy.gif',
+							'sourceUrl' => $actual['data']['updatePost']['post']['updatePost']['sourceUrl'],
 						],
 					],
 				],
@@ -1150,6 +1150,8 @@ class PostObjectMutationsTest extends \Codeception\TestCase\WPTestCase {
 		];
 
 		$this->assertEquals( $expected, $actual );
+		$this->assertArrayHasKey( 'sourceUrl', $actual['data']['updatePost']['post']['featuredImage'] );
+		$this->assertInternalType( 'string', $actual['data']['updatePost']['post']['featuredImage']['sourceUrl'] );
 
 		/**
 		 * Create a media item to add to the post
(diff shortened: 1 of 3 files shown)


hughdevore:
@jasonbahl, all but one test is passing 🤔 

https://travis-ci.org/wp-graphql/wp-graphql/jobs/391432484

tests/wpunit/PostObjectMutationsTest.php
```diff
@@ -8,8 +8,34 @@ class PostObjectMutationsTest extends \Codeception\TestCase\WPTestCase {
...
+	public $status;
+	public $pingStatus;
+	public $parentId;
 
 	public function setUp() {
+
+		/**
+		 * Clear the uploads folder
+		 */
+		$_wp_content_upload_dir = WP_CONTENT_DIR . '/uploads';
```
chriszarate: this seems like it could catch some people off-guard if they aren't running tests in isolation. call out in docs?

tests/wpunit/_bootstrap.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+/**
+ * Clear the uploads folder
+ */
+$_wp_content_upload_dir = WP_CONTENT_DIR . '/uploads';
```
chriszarate: maybe `define( 'UPLOADS', 'wp-content/testuploads' );` so that we don't disturb any existing uploads

hughdevore:
@chriszarate, good call. I've changed my tests to use different assertions and leave the uploads folder alone.

jasonbahl:
@hughdevore so, one thing I'm still struggling with is if this actually belongs in the core WPGraphQL plugin or if it should be something that's filtered in for edge cases like ours at DFM where we want to create an image in the same request as the article. 

I think 99% of use cases would treat this as 2 mutations: 
- `createImage`
- `createPost( input: { featuredImage: "idOfCreatedImage" } )`

Like any UI I can think of you would create the image separate from the Post, and when you create the post you would just set the reference to the already create image, not handle the uploading/creating of the image at the same time. 

We had a unique case for our content syndication, but I think we're more the edge than the norm and I'm not sure the maintenance burden in the core plugin is worth it. 

I also think creating images from an external URL is a bit of an edge case. . .like WordPress core doesn't (that I know of) have a single UI built-in where you can ingest an image from an external URL into your media library. There are plugins that allow this, but not WordPress core.

@chriszarate @0aveRyan @CodeProKid thoughts on this? Does a feature like this belong in the core WPGraphQL plugin (being able to create media items using nested input values on a createPost/updatePost mutation) or should that be left to plugin territory?

hughdevore:
@jasonbahl, going to go ahead and close this since we've decided that this should be a custom extension instead of a core feature.

PR #482, opened 2018-08-03
Feature/#462 fix code coverage

jasonbahl:


chriszarate:
@jasonbahl I did some more poking around here and I think I zeroed in on the issue:

When we run `wpunit` tests, we are treating this plugin/repo as the entire application and bootstrapping WordPress in prior to running tests. That's how our Codeception config is written: as if WPGraphQL is the entire application and WordPress is just a dependency that gets jammed in before tests run.

When we run `functional` and `acceptance` tests, though, we are running WPGraphQL as a plugin—WordPress is the application. But our Codeception config is still the same and it thinks WPGraphQL is the entire application. Further complicating things is the fact that, under `functional` and `acceptance` tests, the web path to WPGraphQL is `/wp-content/plugins/wp-graphql/`, but it can't be loaded directly. But that's what is expected in order to leverage `c3.php`. Loading directly isn't possible, of course, because WordPress won't be loaded.

The only solution I can think of would be to try to run acceptance and functional tests with WordPress bootstrapped in the same way as it is for wpunit tests, so that it remains the application root. That way `c3` will be at the application root. But this more or less defeats the purpose of these tests, since WPGraphQL is no longer loaded in a real WP environment. As far as I can tell, `c3` isn't designed to run inside of an application submodule—it expects to be the entire application.

chriszarate:
FYI, going through these steps is what helped things click for me. If you do the same—and look at the source of c3 to see how it is loading Codeception—you'll quickly realize that we aren't providing the setup that c3 is expecting.

https://github.com/Codeception/c3#debug

jasonbahl:
@chriszarate I'm still at a point where I'm getting told there's no Code Coverage Driver available. 🤔 

```
  [SebastianBergmann\CodeCoverage\RuntimeException]  
  No code coverage driver available 
```

jasonbahl:
Here's an example of trying to run the functional tests with coverage:

```
Functional (docker) Tests (3) ------------------------------
✖ BasicPostListCept: Get public data without passing authentication headers (0.26s)
✖ MenuWithMenuItemsCept: Get nav menu with menu items (0.15s)
✖ PostListWithDifferentStatusCept: Get posts with different post_status (0.10s)
------------------------------------------------------------


Time: 5.95 seconds, Memory: 12.00MB

There were 3 failures:

---------
1) BasicPostListCept: Get public data without passing authentication headers
 Test  tests/functional/BasicPostListCept.php
 Step  See response is json 
 Fail  Invalid json: <br />
<b>Fatal error</b>:  Uncaught SebastianBergmann\CodeCoverage\RuntimeException: No code coverage driver available in /var/www/html/wp-content/plugins/wp-graphql/vendor/phpunit/php-code-coverage/src/CodeCoverage.php:1026
Stack trace:
#0 /var/www/html/wp-content/plugins/wp-graphql/vendor/phpunit/php-code-coverage/src/CodeCoverage.php(141): SebastianBergmann\CodeCoverage\CodeCoverage-&gt;selectDriver()
#1 /var/www/html/wp-content/plugins/wp-graphql/c3.php(232): SebastianBergmann\CodeCoverage\CodeCoverage-&gt;__construct()
#2 /var/www/html/wp-content/plugins/wp-graphql/c3.php(333): __c3_factory(NULL)
#3 /var/www/html/wp-content/plugins/wp-graphql/wp-graphql.php(34): require_once('/var/www/html/w...')
#4 /var/www/html/wp-settings.php(305): include_once('/var/www/html/w...')
#5 /var/www/html/wp-config.php(95): require_once('/var/www/html/w...')
#6 /var/www/html/wp-load.php(37): require_once('/var/www/html/w...')
#7 /var/www/html/wp-blog-header.php(13): require_once('/var/www/html/w...')
#8 /var/www/html/index.php(17): require('/var/www/html/w in <b>/var/www/html/wp-content/plugins/wp-graphql/vendor/phpunit/php-code-coverage/src/CodeCoverage.php</b> on line <b>1026</b><br />
. System message: Syntax error.
Failed asserting that 4 matches expected 0.

```

jasonbahl:
@chriszarate this issue looks like it might be related: 
https://github.com/Codeception/Codeception/issues/2220

chriszarate:
@jasonbahl Hmm I think I got a little further on my branch but didn't realize it. Check out:

https://github.com/Quartz/wp-graphql/tree/feature/%23462-fix-code-coverage

Your issue with `No code coverage driver available` is because PHPUnit is a little dumb: even though you are asking Codeception for remote code coverage via `c3`, PHPUnit will see the coverage flag and refuse to run unless it sees a local code coverage provider. You either need to provide Xdebug in the container or run with the `phpdbg -qrr` wrapper (I've tried both).

That will get you past that message, but your next stumbling block is getting 0% code coverage, which I think is explained in my comment above. Not sure that Codeception issue is related, but could be missing something.

jasonbahl:
Ok, I'm back on the same page. Got rid of the `no code coverage driver` messages. 

Still confused by things, but I feel like I'm at least in a better spot to try and wrap my head around it a bit more.

jasonbahl:
closing in favor of #519

Issue #493, opened 2018-08-16
Add user password reset mutation


kellenmace:
I think it would be useful to add a mutation for reseting a user's password. I envision it working like this:

1. User submits a form to request a password reset and receives the email that contains a link to reset their password (see #491)
2. User clicks the password reset link that contains the key and username as `$_GET` params and is sent back to the headless WP frontend app (this can be done using the `retrieve_password_message` filter to change the URL in the email the user gets from WP's default `wp-login.php` page to some endpoint on the frontend app instead)
3. The frontend app grabs the `key` and `login` `$_GET` params in the URL and stores them in the app state.
4. User submits a password reset form, entering in their new `password`.
5. The app fires off this WPGraphQL mutation, passing along the `key`, `login` and `password` to reset the password.
6. Response is sent back to the GraphQL client indicating whether the password was reset successfully. A payload of data for the user is provided if it was, or an error if it wasn't.

Related issue: #491

kellenmace:
I submitted a PR for this 🕺🏼.

Here is a gif showing this new mutation in action:
![resetuserpassword](https://user-images.githubusercontent.com/5306336/44343219-5b90de00-a45b-11e8-9e69-4e8e2e8bf7d4.gif)

Issue #491, opened 2018-08-15
Add mutation to send password reset email


kellenmace:
The intent of this mutation is to mimic WP's `retrieve_password()` function that runs when the "Lost Password" form is submitted: https://cloudup.com/cDFg7wCo1ig

I envision it working like this:

1. The user is presented with a form field in which they can enter their username or password.
2. When the form is submitted, the `sendPasswordResetEmail` mutation is used and accepts that username/password as an argument.
3. `sendPasswordResetEmail` gets the user, calls WP's `get_password_reset_key()`, which generates a password reset hash, stores it as the `user_activation_key` for that user in the database, then returns the key.
4. The user is sent an email with a password reset link that contains the reset key and their login, such as `domain.com/wp-login.php?action=rp&key=12345&login=myusername`. Both the email subject and body are filterable, though, so the password reset link can easily be filtered to point to the frontend URL for a headless WP site instead, for instance.

kellenmace:
I submitted a PR for this feature.

Here is a gif showing the new feature in action –
![sendpasswordresetemail](https://user-images.githubusercontent.com/5306336/44218490-cc788300-a147-11e8-8ef7-908de758e214.gif)

PR #494, opened 2018-08-16
Feature/#491 add send password reset email mutation

kellenmace:
Adds a mutation for sending password reset emails to users and provides tests for it.

Yes, #491

Here is a gif demonstrating this mutation's functionality:
sendpasswordresetemail

Where has this been tested?
---------------------------
**Operating System:** macOS 10.13.6

**WordPress Version:** 4.9.8

Any other comments?
-------------------
I'd love some feedback on this PR.

**The Name**
It was tough to find a good name for this mutation. I was toying with the idea of `userLostPassword`, but decided against it, since that doesn't describe what the mutation does.  WP's function for doing this is named `retrieve_password()`, but I think that's misleading, since it doesn't actually retrieve the user's password – it just emails them a link that can be used to reset it. So I ultimately went with `sendPasswordResetEmail` since I think that most accurately reflects what the mutation does.

I plan to also submit a PR for a `resetUserPassword` mutation that will accept the key and username and perform the reset. See #493.

**The Functionality compared to `retrieve_password()`**
I would have loved to leverage WP's `retrieve_password()` function for handling the bulk of the functionality for this feature, but there are a few things that made it unsuitable:

1. It assumes that the username/password is always in `$_POST['user_login']`. That's inconsistent with the rest of the WPGraphQL user mutations that use `username` as the key, though. My code uses `username` to stay consistent with all other user mutations.
2. Toward the end of the function, if the password reset email doesn't send, `wp_die()` is called. When used in a GraphQL context, that results in a 500 Internal Server Error. In my code, I'm throwing an 'The email could not be sent.' UserError instead in that case, so the GraphQL client always gets a response with usable data rather than a 500 error.

So I based my code on WP's `retrieve_password()`, but put the changes mentioned above in place, and broke up the logic into many smaller methods rather than one very long one.

**The args**
`retrieve_password()` works whether `$_POST['user_login']` is set to a user's username or their email address. I considered making `sendPasswordResetEmail` accept either a `username` arg which would always be the username or an `email` arg which would always be the email. I decided against it though, since doing it that way would require you to know in the GraphQL client whether what the user had entered is their username or their email ahead of time and pass it through as the appropriate arg. So I kept things simple and stuck with a single "username" field. It works like this:

1. The user is presented with a password reset form with a single field in which they can enter their username or password
2. When it's submitted, the value of that field can be sent to the `sendPasswordResetEmail` mutation as the `username` arg without the need to know whether it's the user's username or email – they'll get the password reset email regardless and be on their way.

Thanks!

### DIFF
--- a/src/Type/RootMutationType.php
+++ b/src/Type/RootMutationType.php
@@ -19,6 +19,7 @@ use WPGraphQL\Type\TermObject\Mutation\TermObjectUpdate;
 use WPGraphQL\Type\User\Mutation\UserCreate;
 use WPGraphQL\Type\User\Mutation\UserDelete;
 use WPGraphQL\Type\User\Mutation\UserUpdate;
+use WPGraphQL\Type\User\Mutation\SendPasswordResetEmail;
 
 /**
  * Class RootMutationType
@@ -138,9 +139,10 @@ class RootMutationType extends WPObjectType {
 			/**
 			 * User Mutations
 			 */
-			$fields[ 'createUser' ] = UserCreate::mutate();
-			$fields[ 'updateUser' ] = UserUpdate::mutate();
-			$fields[ 'deleteUser' ] = UserDelete::mutate();
+			$fields[ 'createUser' ]             = UserCreate::mutate();
+			$fields[ 'updateUser' ]             = UserUpdate::mutate();
+			$fields[ 'deleteUser' ]             = UserDelete::mutate();
+			$fields[ 'sendPasswordResetEmail' ] = SendPasswordResetEmail::mutate();
 
 			self::$fields = $fields;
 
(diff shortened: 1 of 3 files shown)


jasonbahl:
@kellenmace this is great! Thanks for taking time to write up your thought process as well. Very helpful! 

I'll try and review this more thoroughly shortly. 

Thanks!

jasonbahl:
@kellenmace I'm prepping for the v0.1.0 release which introduces the new `register_graphql_*` APIs and the concept of a central TypeRegistry. 

I think this feature is pretty close to what we want, but want to wait to merge until I get that release fleshed out.

jasonbahl:
@kellenmace just checking in again. the v0.1.0 release is _pretty_ close to being released, so I think we should follow up on this shortly after it's released to get it up to par with how things have changed.

jasonbahl:
@kellenmace closing this as I refactored it to fit with the new post-v0.1.0 architecture in #586

PR #497, opened 2018-08-17
#495 & #496

### LINKED ISSUE #495: Batch Requests don't pass data to do_graphql_request properly
When Batch Requests come in, they are an array of queries/operations/variables. 

The `do_graphql_request` action expects to be fired with the query, operation and variables for each query being resolved.

### LINKED ISSUE #496: TermObjectConnectionResolver tries to make use of Post ID even if the source is not a Post
When trying to output terms in a flat list, the TermObjectConnectionResolver attempts to `wp_get_object_terms( $source->ID )` even if the `$source` is not a Post. 

This should only happen if `$source instanceof \WP_Post` to make sure we don't cause errors when the `$source` is another entity, such as a parent Term that doesn't have an `->ID` property.

jasonbahl:
closes #495 - Batch Requests not passing context to `do_graphql_request` properly.

Since batch requests come in as an array, we need to iterate over the array and send out the do_graphql_request action for each

closes #496 - PHP Warnings are being thrown because we sometimes try to use $source->ID even if the $source is a Term and has no ->ID property.


CodeProKid:
@jasonbahl looks like some requests timed out causing this to fail, do you want to re-run them? Otherwise, changes look good to me 👍

PR #501, opened 2018-08-25
Fix error message for non-existent post

chriszarate:
Fixes a grammatical error in an error message that has been bugging me for a long time. 😂

It's worth noting that this is actually a breaking change for us! This error corresponds to a 404 response for us on the client. But because this is a error and not an empty response, we have to catch all errors and inspect them in order to determine whether we should 404. The only way to do this is to regex the string, making it a bit fragile.

Have we considered implementing error codes (either on the error object or leading the message)?

Any other comments?
-------------------
Added test, which would at least catch future changes to the message.


jasonbahl:
@chriszarate Thanks!

Issue #503, opened 2018-08-28
Excerpt is always of the first post


seimsel:
I tried to get the excerpt of all my posts using this query:

```
{
  posts {
    edges {
      node {
        id
        title
        excerpt
      }
    }
  }
}
```

The result was:

```
{
  "data": {
    "posts": {
      "edges": [
        {
          "node": {
            "id": "cG9zdDo1",
            "title": "This is a Test",
            "excerpt": "<p>Welcome to WordPress. This is your first post. Edit or delete it, then start writing!</p>\n"
          }
        },
        {
          "node": {
            "id": "cG9zdDox",
            "title": "Hello world!",
            "excerpt": "<p>Welcome to WordPress. This is your first post. Edit or delete it, then start writing!</p>\n"
          }
        }
      ]
    }
  }
}
```

When I look at it with the normal Wordpress frontend the two post do **not** have the same content.

izzygld:
Try adding this to your function.php file:

`  add_action( 'do_graphql_request', function () {
    add_filter( 'get_the_excerpt', function ( $excerpt, $post ) {

      return !empty( $excerpt ) ? $excerpt : wp_trim_words( strip_shortcodes($post->post_content) );
    }, 10, 2 );
  } );`

jasonbahl:
Sounds like the global post isnt getting set properly as the posts are being resolved. 

I’ll see what I can find out.

jasonbahl:
I beleive this was fixed by #504 but need to confirm

CodeProKid:
@jasonbahl I wonder if we should maybe use this ticket to write a test that asserts this does not happen? Would be pretty easy to do.

jasonbahl:
Ya, good call.

CodeProKid:
This was fixed, we just need to write a test to make sure it doesn't happen in the future. That will happen in #576

PR #506, opened 2018-08-30
#504 - Ensure post objects are resolved using DataSource::resolve_pos…

### LINKED ISSUE #504: postBy behaves in an inconsistent way
Hey!

I came across a scenario in which postBy returns inconsistent responses. In short, if there's any filters for the post that modify the post's content, then `postBy(id:"cG9zdDox")` works correctly while `postBy(slug:"hello-world")`  does not.

We use WooCommerce Memberships plugin, which basically can be used to restrict post / page content to certain users or user groups. That plugin basically modifies returned `content` for Post and displays full `content` only when logged in with proper user. Out-of-the-box that works good with REST API and mostly with WP GraphQL as well (which is great!). If I try e.g. `posts` or `postBy(id:"xyz")` those work correctly and `content` is properly modified. However, if I do `postBy(slug:"post-slug")` then I receive the post content without any modifications, so that part doesn't seem to be working correctly.

This far I was able to add a work-around for this, but still didn't get to the root cause. Unfortunately I don't have full repo to reproduce this :/ 
Here's my commit for the work-around if it helps at all: https://github.com/Hellenic/wp-graphql/commit/232016da3cd9fca6b5cb494e70416bc046a1e416
With that commit in place, `postBy(slug..)` also properly modifies the content, even though that most likely is not the right way to address this problem.

Versions:
* WordPress 4.9.8
* WP GraphQL - develop branch

jasonbahl:
This ensures post objects are resolved using DataSource::resolve_post_object() to make sure the global $post is properly set and setup_postdata is properly run. This ensures filters that rely on the Global post object, etc to run properly.

closes #504

### DIFF
--- a/src/Data/DataSource.php
+++ b/src/Data/DataSource.php
@@ -179,6 +179,14 @@ class DataSource {
 		}
 
 		/**
+		 * Mimic core functionality for templates, as seen here:
+		 * https://github.com/WordPress/WordPress/blob/6fd8080e7ee7599b36d4528f72a8ced612130b8c/wp-includes/template-loader.php#L56
+		 */
+		if ( 'attachment' === $post_type ) {
+			remove_filter( 'the_content', 'prepend_attachment' );
+		}
+
+		/**
 		 * Set the resolving post to the global $post. That way any filters that
 		 * might be applied when resolving fields can rely on global post and
 		 * post data being set up.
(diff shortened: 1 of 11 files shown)


jasonbahl:
I believe this fixes #503. . .need to confirm

PR #515, opened 2018-09-15
Fix $post_type in graphql_map_input_fields_to_wp_query

epeli:
Not sure if this is bug but I noticed that the implementation does not match with the doc block of the `graphql_map_input_fields_to_wp_query`.

It states that the post type is passed as the third argument

https://github.com/wp-graphql/wp-graphql/blob/aa69a640b6877292a84a710c7270ac7403680b95/src/Type/PostObject/Connection/PostObjectConnectionResolver.php#L360

but the actual third argument is the `$source`.

### DIFF
--- a/src/Type/PostObject/Connection/PostObjectConnectionResolver.php
+++ b/src/Type/PostObject/Connection/PostObjectConnectionResolver.php
@@ -357,16 +357,16 @@ class PostObjectConnectionResolver extends ConnectionResolver {
 		 *
 		 * @param array       $query_args The mapped query arguments
 		 * @param array       $args       Query "where" args
-		 * @param string      $post_type  The post type for the query
 		 * @param mixed       $source     The query results for a query calling this
 		 * @param array       $all_args   All of the arguments for the query (not just the "where" args)
 		 * @param AppContext  $context    The AppContext object
 		 * @param ResolveInfo $info       The ResolveInfo object
+		 * @param string      $post_type  The post type for the query
 		 *
 		 * @since 0.0.5
 		 * @return array
 		 */
-		$query_args = apply_filters( 'graphql_map_input_fields_to_wp_query', $query_args, $args, $source, $all_args, $context, $info );
+		$query_args = apply_filters( 'graphql_map_input_fields_to_wp_query', $query_args, $args, $source, $all_args, $context, $info, self::$post_type );
 
 		/**
 		 * Return the Query Args


chriszarate:
This is a borderline breaking change, but probably worth it. There's no other way to suss out the type of the query without this. You'd have to use `graphql_post_object_connection_query_args` instead, which is a little bit more of a dangerous filter.

On Slack, @epeli also made a suggestion of dynamic map_input_fields queries:

> Another alternative would be to add dynamic filters for the mapping like
> 
> ```
> $query_args = apply_filters( 'graphql_map_' . self::$post_type . '_input_fields_to_wp_query', $query_args, $args, $source, $all_args, $context, $info );
> ```

CodeProKid:
A non-breaking way to do this would be to just throw the variable at the end of the params, and adjust the docblock, but that might be annoying if you have to them list out all the other params in your callback just to get to the last one that's arguably the most useful in this instance.

I personally prefer using a param rather than a dynamic filter/hook name in most instances since it seems a little more flexible, but open to whatever people think would work best here.

epeli:
I actually would prefer the param too. Static filter names are easier to debug: It's very annoying debug why some dynamic filter is not being called. You have dig into the framework code to findout the actualy filter name...

I really hope that we can go with this and not add it to the end of the params list. It's very important part of the api and putting it as the second param it signals its importance to the user.

If anyone else is looking for a workaround for now the `$info` object can be used:

```
    function ($query_args, $args, $source, $all_args, $context, $info) {
        // only manipulate my input fields
        if ($info->fieldName !== 'myfields') {
            return $query_args;
        }
```

The `$info->fieldName` seem to contain the plural version of the field name.

chriszarate:
> The $info->fieldName seem to contain the plural version of the field name

It contains the name of the root query you are accessing. So it's plural only if you are hitting the plural root query, e.g., `posts`.

src/Type/PostObject/Connection/PostObjectConnectionResolver.php
```diff
@@ -366,7 +366,7 @@ public static function sanitize_input_fields( array $args, $source, array $all_a
 		 * @since 0.0.5
 		 * @return array
 		 */
-		$query_args = apply_filters( 'graphql_map_input_fields_to_wp_query', $query_args, $args, $source, $all_args, $context, $info );
+		$query_args = apply_filters( 'graphql_map_input_fields_to_wp_query', $query_args, $args, self::$post_type, $source, $all_args, $context, $info );
```
jasonbahl: If we add `self::$post_type` to the end, after `$info` this wouldn't be a breaking change. 

As in, anyone with existing code using this filter wouldn't break. I think this is a good addition to the filter, let's just add it as the 7th param instead of the 3rd.

jasonbahl:
```suggestion
		$query_args = apply_filters( 'graphql_map_input_fields_to_wp_query', $query_args, $args, $source, $all_args, $context, $info, self::$post_type );
```

jasonbahl: also would need to update the DocBloc

epeli:
Rebased this PR to match the suggested change.

PR #521, opened 2018-09-21
Make the post object field format enums filterable

danvisintainer:
This allows the post object format enum values to be filterable, if you'd like to have more post content formats available to query (AMP, in our case).

Where has this been tested?
---------------------------
**Operating System:**  CentOS 7

**WordPress Version:**  4.9.8


jasonbahl:
@danvisintainer hey! Thanks for the PR. 

You can already filter values of an Enum via this filter here:

https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/WPEnumType.php#L49

Let me know if that doesn't work for you and we can re-open this.

danvisintainer:
@jasonbahl Works great! Thanks for the quick response.

Issue #505, opened 2018-08-29
Menus not returning ChildItems


hblackett:
I've just run into an issue where childItems of a menuItems object were not returning.

<img width="1117" alt="screen shot 2018-08-29 at 14 57 37" src="https://user-images.githubusercontent.com/4136219/44792475-e8dedb80-ab9b-11e8-9deb-49ec7ed3d27d.png">

I changed the menu name in WP (from 'Header Menu' to 'ALW Menu') and suddenly the childItems work as expected:

<img width="983" alt="screen shot 2018-08-29 at 14 59 18" src="https://user-images.githubusercontent.com/4136219/44792644-470bbe80-ab9c-11e8-89cb-2123cb22a21f.png">

<img width="1119" alt="screen shot 2018-08-29 at 14 49 13" src="https://user-images.githubusercontent.com/4136219/44792311-84bc1780-ab9b-11e8-970f-6c052ab7df28.png">

If I change the menu name back to 'Header Menu' it breaks the childItems.

It's a localhost install running WP GraphQL 0.0.34 on WP 4.9.8 with all other plugins disabled. I've tried recreating the issue on a separate install but everything is working as expected – can't seem to find the magical menu name that will break things!

izzygld:
i just came across the same issue.

i had to delete the `Header_menu` and remake the menu.

(i was using [see here](https://gist.github.com/izzygld/287aa48a90d170dabd9ba479196ea695)

But we should definitely look into it.

jasonbahl:
Wait, so you’re saying on a fresh install “Header Menu” does work as expected?

hblackett:
Ha, yes! Things are working as expected on a fresh install.

But on the localhost install where I found the issue, I tried deleting and rebuilding the menu and `childItems` would only work with some menu names ('Header Menu', 'Main Nav' and others didn't work; possibly conflicting with previous menus that I had deleted?).

izzygld:
i found this to be true too.

When renaming the menu and quering to the new menu name things went back to being good 👍

jasonbahl:
Strange. I'll see what I can find out. Curious if this is specifically an issue with WPGraphQL or an underlying issue. Thanks for reporting and thanks for the added details.

ElisaMassafra:
In my experience it works only if it's the first menu in the list. I see that it always returns the items for the first menu (MenuItemConnectionResolver -> probably line 46, where it always returns the childItems for $menus[0]->slug from $menus?

[...]

PR #522, opened 2018-09-21
Fix for bug: Menus not returning ChildItems #505

ElisaMassafra:
The menu childItems were working only on the first menu instead of selecting the current menu


jasonbahl:
closing this in favor of #524

PR #539, opened 2018-10-15
Release/v0.1.0

jasonbahl:
# Release Notes

## New Registration API
Similar to WordPress core `register_*` APIs, such as `register_post_type`, `register_taxonomy`, `register_meta`, etc, WPGraphQL now has it's own `register_graphql_*` API. 

The following `regiser_graphql_*` functions are introduced with this release:

`register_graphql_type`
`register_graphql_object_type`
`register_graphql_input_type`
`register_graphql_enum_type`
`register_graphql_union_type`
`register_graphql_field`
`register_graphql_fields`
`register_graphql_schema`
`register_graphql_connection`
`register_graphql_mutation`
`deregister_graphql_field`

Docs to come, but for the time being, check the source code to see how things work. Specifically TypeRegistry.php. All Types have been refactored to be registered with these new functions. 

### Deprecated hooks & filters

Some filters and hooks have been removed, intentionally, because they were redundant. 

- *graphql_root_queries*: this filter has been removed. The RootQuery type is an ObjectType and already inherits the filters established in WPObjectType, so to add fields to the RootQuery, one could use the `graphql_RootQuery_fields` filter, or could (recommended) use the new `register_graphql_field` or `register_graphql_fields` functions to add fields to the root. 
- *graphql_post_object_mutation_input_fields*: this filter has been removed. The input fields are simply fields of an InputObjectType, which inherit the filters established in WPObjectType, so to add fields, one can use the filters `graphql_UpdatePostInput_fields` or `graphql_CreatePostInput_fields`, or (recommended) use the new `register_graphql_field` or `register_graphql_fields` functions to add fields to the inputs. 
- *graphql_comment_mutation_input_fields*: See previous note
- *graphql_term_object_mutation_input_fields* See above
- *graphql_setting_mutation_input_fields*: See above

### Changes to various Type names
- various type names have changed for consistency. 
- All input types now include `Input` in their name. 
- All Enum types now include `Enum` in their name
- All Union types now include `Union` in their name
- All connections are registered via the `register_graphql_connection()` function, and therefore all connection types and fields have consistent naming, where before the names were a bit random
  - Connection input args have consistent names now. 
     - For example: `RootCommentsCommentArgs` is now `RootQueryToCommentConnectionWhereArgs` (see: https://github.com/wp-graphql/wp-graphql/compare/release/v0.1.0?expand=1#diff-00fe868538645d14c1c8af8d3a19bde6) 
     - `RootCategoriesTermArgs` is now `RootQueryToCategoryConnectionWhereArgs` https://github.com/wp-graphql/wp-graphql/compare/release/v0.1.0?expand=1#diff-9d7db44df670461f784357a0a7e3c860L83

## Moved files / namespaces
- Connection resolvers have all been moved under `WPGraphQL\Data`. If you extend any of the existing resolver classes, you should be able to just update your namespace, such as changing `\WPGraphQL\Type\PostObject\Connection\PostObjectConnectionResolver` to `\WPGraphQL\Data\PostObjectConnectionResolver` as seen here: https://github.com/wp-graphql/wp-graphql/compare/release/v0.1.0?expand=1#diff-4232742370ce60dbe32afbf8e87e3476L382


src/Mutation/MediaItemDelete.php
```diff
@@ -1,74 +1,49 @@
...
 				'id'          => [
-					'type'        => Types::non_null( Types::id() ),
+					'type'        => [
+						'non_null' => 'ID',
+					],
 					'description' => __( 'The ID of the mediaItem to delete', 'wp-graphql' ),
 				],
 				'forceDelete' => [
-					'type'        => Types::boolean(),
+					'type'        => 'Boolean',
```
chriszarate: really like this change in particular

src/Type/Union/MenuItemObjectUnion.php
```diff
@@ -0,0 +1,46 @@
...
+
+// Add taxonomies that are allowed in WPGraphQL.
+foreach ( get_taxonomies( $args ) as $type ) {
+	$possible_types[ $type ] = Types::term_object( $type );
+}
+
+// Add the custom link type (which is just a menu item).
+$possible_types['MenuItem'] = Types::menu_item();
+
+register_graphql_union_type( 'MenuItemObjectUnion', [
```
chriszarate: Is it possible to alias union types? One of my regrets in my Menus PR was naming this `MenuItemObjectUnion`. That's not actually what it is and it's more broadly useful than that since it's effectively a union of all supported post types, users, and terms.

If we can't alias it later it might be nice to get it changed before `0.1.0`.

jasonbahl: @chriszarate In my head it makes sense to still have a `MenuItemObjectUnion` as we _should_ (if we aren't already) probably only respect types that are registered with `show_in_nav_menus => true`, eh?

I do think there is room for a Union of _all_ possible PostTypeObjects and TermObjects, but might make sense to be _another_ union, not this one?

jasonbahl: I do see the code comment here about this topic: https://github.com/wp-graphql/wp-graphql/blob/61786c27f1101aff61fd80167707dd056de94189/src/Type/Union/MenuItemObjectUnionType.php#L66-L68 🤔

jasonbahl: So, since the Schema itself is queryable, I do think there is value in restricting this to types that are registered with `show_in_nav_menus => true`, because GraphQL can serve as a debug tool just as much as it can serve as an API for clients. 

For example, as an admin on a WordPress site (or even a UI developer), I may want to quickly find out what Types are registered to show in nav menus, and a quick query to do that would be: 

```
{
  __type(name: "MenuItemObjectUnion") {
    possibleTypes {
      name
      description
      kind
    }
  }
}
```

jasonbahl: Here's an example of that Query run with and without `'show_in_nav_menus' => true`

You can see that "attachments / mediaItems)" are not registered to `show_in_nav_menus`

I think I'd argue that this Type _should_ stay, and we should add `show_in_nav_menus => true` to it, and introduce another Type which supports an intersection of all PostTypeObjects + TermObjects

screen shot 2018-10-22 at 10 34 20 am
screen shot 2018-10-22 at 10 34 33 am

chriszarate: Makes sense.

Issue #517, opened 2018-09-18
ConnectionResolver doesn't work non-WP objects or arrays


kidunot89:
While trying to add define a custom connection by extending `WPGraphQL\Data\ConnectionResolver`. I hit a snag that made my jump through a few unnecessary hoops to get it to work. 

## Issues
1. [`get_array_slice`](https://github.com/wp-graphql/wp-graphql/blob/aa69a640b6877292a84a710c7270ac7403680b95/src/Data/ConnectionResolver.php#L96) - returns an empty array if the `$query` isn't and `object` and  [`default`](https://github.com/wp-graphql/wp-graphql/blob/aa69a640b6877292a84a710c7270ac7403680b95/src/Data/ConnectionResolver.php#L119) just sets `$array_slice` to `$items` repeatedly. This will be issue if the `$query` is every a mixed list of WP and non-WP objects, and arrays get completely ignored.
2. [`get_query_info`](https://github.com/wp-graphql/wp-graphql/blob/aa69a640b6877292a84a710c7270ac7403680b95/src/Data/ConnectionResolver.php#L271) - same as the non-WP objects and array get ignored here but there a filter here that can be used to remedy that, but I fill like I should have to use a filter if I'm extending the class for its intended use.

Both issues can be get worked around but the workarounds aren't pretty 

## Patches
1. `get_array_slice` - unlike `get_query_info` and its called using the `self::` keyword from `resolve()` so it can't be overloaded. `get_connection` can be though and it receives the argument from `get_array_slice`. So my solution was to simple overloaded that and do the `get_array_slice` functionality there. 
``` 
public static function get_connection( $query, $null, $source, $args, $context, $info ) {
	$info = self::get_query_info( $query );
	$items = $info['items'];
	$array = [];
	if ( ! empty( $items ) && is_array( $items ) ) {
		foreach ( $items as $item ) {
			if( is_array( $item ) ) {
				$array[] = $item;
			}
		}
	}
	$meta = self::get_array_meta( $query, $args );
	$connection = ArrayConnection::connectionFromArraySlice( $array, $args, $meta );
	$connection['nodes'] = $array;
	return $connection;
}
```
[source](https://github.com/kidunot89/wp-graphql-theme/blob/5082c80ab178f7c193a02237a423092774be13ae/client/public/inc/wp-graphql/Types/Widget/Connection/WidgetConnectionResolver.php#L175)

2. `get_query_info` - use the `graphql_connection_query_info` filter add modified output
[...]

PR #540, opened 2018-10-18
bug fixed for issue #517

kidunot89:
https://github.com/wp-graphql/wp-graphql/issues/517

`get_array_slice()` and `get_query_info()` can handle `array` query objects

#517 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Data/ConnectionResolver.php
+++ b/src/Data/ConnectionResolver.php
@@ -116,6 +116,7 @@ abstract class ConnectionResolver implements ConnectionResolverInterface {
 						default:
 							$array_slice[] = $item;
 					}
+				// For non-stdClass nodes
 				} elseif ( is_array( $item ) ) {
 					$array_slice[] = $item;
 				}
--- a/src/Data/ConnectionResolver.php
+++ b/src/Data/ConnectionResolver.php
@@ -116,6 +116,8 @@ abstract class ConnectionResolver implements ConnectionResolverInterface {
 						default:
 							$array_slice[] = $item;
 					}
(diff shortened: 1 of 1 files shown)


src/Data/ConnectionResolver.php
```diff
@@ -298,6 +299,9 @@ public static function get_query_info( $query ) {
 					$query_info['total_items'] = ! empty( $query->get_total() ) ? $query->get_total() : count( $query->get_results() );
 					$query_info['items']       = $query->get_results();
 					break;
+				case is_array( $query ):
```
jasonbahl: I don't believe this `case is_array( $query ):` will ever evaluate to true as we have a check above: `if ( true === is_object( $query ) ) {`

I think you'd have to approach the same way you did in your other change and have an else statement for arrays.

jasonbahl: Also, can you elaborate on the use case where this isn't working so I have some context? like maybe share some code or something I can look at?

kidunot89: Say this is `get_query` returns an array of arrays
```
protected static function get_query() {
   $query = [
     [ ...itemOne],
     [ ...itemTwo],
     [ ...itemThree],
   ];
   return $query;
}

the current `array_slice` return an empty array. And it called with the `self` keyword in `resolve` so it can't be effectively overloaded. as it is now the only solution is to overload `get_connection` ignore the `$array` and doing reslicing the query results there.

Issue #557, opened 2018-10-30
WPGraphQL Tax Query and Meta Query plugins not compat with v0.1.0


jasonbahl:
Some changes to v0.1.0 of WPGraphQL are breaking for the following extensions:

- [ ] WPGraphQL Meta Query extension
- [ ] WPGraphQL Tax Query extension

These plugins should undergo a refactor to work with v0.1.0

jasonbahl:
Closed by #591

Issue #578, opened 2018-11-08
WPInputObjectType::construct doesn't call prepare_fields


galaxiecruzin:
I'm not sure if there is a new way to register input fields with the type registry implementation, so I'm testing registering an input type "lang" to filter posts by language. The following snippet worked prior to 0.1.0. To reproduce add this snippet as a plugin

```
add_filter( 'graphql_input_fields', function( $fields ) {
	$fields['lang'] = [
		'type' => \WPGraphQL\Types::string(),
		'description' => 'Select posts by Polylang language',
	];
	$fields['allLanguages'] = [
		'type' => \WPGraphQL\Types::boolean(),
		'description' => 'Show posts regardless of the Polylang language',
	];
	return $fields;
} );
```

Then run a query like
```
query Posts($lang: String) {
  posts(where: {lang: $lang}) {
    edges {
      node {
        title
      }
    }
  }
}
---
{
  "lang": "en"
}
```
See https://gist.github.com/epeli/87909f92b19a216dc7f5b2b6c204ad31

While digging into this I found in ./src/Type/WpInputObjectType.php

$config['fields'] was an array prior to 0.1.0 so this if condition prevents prepare_fields() from being called.

```
	public function __construct( $config = [] ) {
		if ( ! empty( $config['fields'] ) && is_array( $config['fields'] ) ) {
			$config['fields'] = self::prepare_fields( $config['fields'], $config['name'], $config );
		}
		parent::__construct( $config );
	}
...
```

A print_r of $config['fields'] here gives me

```
Closure Object
(
    [static] => Array
        (
            [config] => Array
                (
                    [description] => Input for the updateSettings mutation
                    [fields] => Array
...
...
```

I haven't yet been able to figure out what this _should_ be accepting... but figured I'd post this to start a thread... If someone can point me in the direction of what this should be expecting... I can take a crack at trying to update.

jasonbahl:
@galaxiecruzin hey, thanks for reporting!

ya, there's a new API for registering Types and Fields. It's fairly backward compatible, but there definitely are some breaking changes.

Here's the recommended way to add fields now (sorry, [the docs](https://www.wpgraphql.com/docs/api-reference/) are still incomplete and have been slow to get together):

```
// hook in here to make sure it's a GraphQL request and the GraphQL registry is available
add_action( 'graphql_register_types', function() {

   // first $arg is the Type you want to add the fields to, second arg is an array of fields
[...]

PR #591, opened 2018-11-28
Restore graphql_input_fields filter

chriszarate:
- [X] Make sure you are making a pull request against the **develop branch** (left side). Also you should start *your branch* off *our develop*.
- [X] Make sure you are requesting to pull request from a **topic/feature/bugfix branch** (right side). Don't pull request from your master!

I'm not sure how I missed this in reviewing the Type registry changes or if this snuck in after I reviewed them, but the `graphql_input_fields` filter (what we use to implement custom `where` args) is now fairly broken.

The issue is that the `fields` definition for every object type is now wrapped in a closure (see [TypeRegistry#L366](https://github.com/wp-graphql/wp-graphql/blob/develop/src/TypeRegistry.php#L366)), which delegates to `WPObjectType::prepare_fields` to apply various filters. Then `WPInputObjectType` attempts to apply its own filters by calling `self::prepare_fields` in its constructor, but `fields` is now a closure (not an array) and the method call is skipped. As a result, the `graphql_input_fields` filter is never applied.

This PR adds logic to `WPObjectType::prepare_fields` to call `WPInputObjectType::prepare_fields` if the object `kind` is `input`. This is a little inelegant but preserves the API of `WPInputObjectType` which could be used by implementors (it is by us).

It may make more sense to move `graphql_input_fields` into `WPObjectType::prepare_fields`, but this would make `WPInputObjectType` an empty shell (the `format_enum_name` method was completely unused).

We could also eliminate the `graphql_input_fields` filter and force users to the individual type filters, but this central filter is rather useful so that `where` args can be easily added across multiple types (e.g., all custom post types).

### DIFF
--- a/src/Type/WPObjectType.php
+++ b/src/Type/WPObjectType.php
@@ -3,7 +3,6 @@ namespace WPGraphQL\Type;
 
 use GraphQL\Type\Definition\ObjectType;
 use WPGraphQL\Data\DataSource;
-use WPGraphQL\Type\WPInputObjectType;
 
 /**
  * Class WPObjectType
--- a/src/Type/WPObjectType.php
+++ b/src/Type/WPObjectType.php
@@ -92,12 +92,11 @@ class WPObjectType extends ObjectType {
 	 *
 	 * @param array  $fields
 	 * @param string $type_name
-	 * @param array $config
 	 *
 	 * @return mixed
 	 * @since 0.0.5
 	 */
-	public static function prepare_fields( $fields, $type_name, $config = [] ) {
+	public static function prepare_fields( $fields, $type_name ) {
 
 		if ( null === self::$prepared_fields ) {
 			self::$prepared_fields = [];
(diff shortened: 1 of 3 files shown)


chriszarate:
Noting that this seems like it might fix https://github.com/wp-graphql/wp-graphql/issues/557, but I haven't verified that.

chriszarate:
Also related to https://github.com/wp-graphql/wp-graphql/issues/578.

@jasonbahl's response there indicates he might be in favor of deprecating `graphql_input_fields`?

jasonbahl:
@chriszarate I'm not 100% opposed to keeping the `graphql_input_fields` filter. It did seem to be redundant when I was going through the refactor.

src/Type/WPObjectType.php
```diff
@@ -140,6 +142,16 @@ public static function prepare_fields( $fields, $type_name ) {
...
 
+			/**
+			 * If the object defines input fields, apply a centralized filter for
+			 * input fields. This was previously handled by WPInputObjectType, but
+			 * the fields definition is now trapped in a closure so all filters need
+			 * to happen in one place.
+			 */
+			if ( isset( $config['kind'] ) && 'input' === $config['kind'] ) {
+				$fields = WPInputObjectType::prepare_fields( $fields, $type_name, $config );
+			}
```
jasonbahl: @chriszarate in what situation does `kind === 'input'` and evoke a `WPObjectType`? 

if you `register_graphql_type()` and specify `'kind' => 'input'` or `'register_graphql_input_type()'` where `'kind' => 'input'` is set for you, then `WPObjectType` shouldn't even be getting instantiated, but `WPInputObjectType` should be, as seen here: 

https://github.com/wp-graphql/wp-graphql/blob/develop/src/TypeRegistry.php#L378-L380

I could easily be missing something though. . .can you clarify how you get to this point where `kind === input` in a WPObjectType?

chriszarate: Updated to centralize `prepare_fields` calls per our conversation.

jasonbahl:
@chriszarate thanks for updating this.

Issue #326, opened 2017-12-04
Add Revision support


jasonbahl:
The core WP REST API supports revisions: https://developer.wordpress.org/rest-api/reference/post-revisions/

They require the user to be authenticated, so we should respect similar auth patterns the REST API has established, but revisions should be able to be accessed from the core WPGraphQL Schema. 

I think it should be a field of the PostObjectType like so, something to this tune:

```
{
    post( id:"..." ) {
       id
       title
       revisions( first: 5 ) {
          edges {
            node {
              id
              title
           }
         }
      }
   }
}
```

I think there should also be a root query to get a single revision, like so, if you know the ID of the revision: 
```
{
   revision( id: "..." ) {
      id
      title
  }
}
```

Or you could get a revision from other parameters, such as the revisedContent type and ID.
```
{
   revision( where: { revisedContent: { type: POST id: "..." } } ) {
       id
       title
   }
}
```

justinwhall:
This would be awesome :) 

It would facilitate building a Gatsby client route to preview posts. I'm assuming something like the following would retrieve the most recent revision:

```
{
    post( id:"..." ) {
       id
       title
       revisions( first: 1 ) {
          edges {
            node {
              id
              title
           }
         }
      }
   }
}
```

jasonbahl:
@justinwhall have you used Gatsby 2.0 with WPGraphQL?

I haven't had a chance to work with Gatsby 2.0, but it looks like you can use any GraphQL source now, eh?

jasonbahl:
closed by #695

PR #593, opened 2018-12-01
#326 - Revision Support

jasonbahl:

- This adds revision support, including full respect for capability checks.
- TypeRegistry.php - adds a new `connectionFieldConfig` field to allow additional config to be passed to the root connection field when registering a connection
- Unset status and stati $args for Revision connections
- Remove the filter that was previously added to Config.php to attempt to restrict revisions
- Update DataSource::resolve_post_object(); to call the new PostObject model, which returns a PostObject in the correct shape, and handles all the cap checks in a central place
- Update PostObjectConnectionResolver to ensure both nodes and edges resolve the PostObject properly before being returned.
- Introduce a new PostObject model which handles permission checking before returning a PostObject
- Move resolvers out of the Type/Object/PostObject definition and into the Model/PostObject
- run composer install --no-dev && composer dump-autoload

#326 (follow-up to #588) 

**Authenticated Admin user querying for Revisions**

revisions-auth-admin

**Authenticated Author querying for Revisions**

revisions-authenticated-author

**Unauthenticated user Querying for revisions**

revisions-no-auth

Any other comments?
-------------------

This lays some foundation for handling better auth checks in a more centralized way. 

I think there's more work that can be done still, but this lays some initial foundation so that we can handle permissions for objects in a central manner. 

A PostObject can be referenced from many Types in the Schema, so instead of defining cap checks over and over and over within resolvers, we can now define them in the Model layer, and let the Types resolve to the Model layer. 

Over time I see more and more Models being created for core entities. . . Comments, MediaItems, User, etc. 

This will allow for the Schema to better mirror UI needs and have less of a 1-to-1 mirror of the data storage as well. 😄


jasonbahl:
closing in favor of the work @CodeProKid is doing with the Model Layers.

PR #628, opened 2018-12-19
#619 - Throw Better Exceptions in TypeRegistry

### LINKED ISSUE #619: Throw exceptions in register_graphql_* methods when they're missing critical fields
The new `register_graphql_*` methods introduced in v0.1.0 don't have strong validation on the $args that are input. 

We should add better validation around the registration of Types and Fields and throw appropriate exceptions when a field is missing fields. 

One common one is the `name` field. If a field or Type is registered without a name, I believe it leads to an internal_server_error, instead of a specific error message. 

It would benefit developers if they had better introspection into _why_ they're getting errors, and improper schema validation is a common mistake.

jasonbahl:
- Router.php add the $debug flag to FormattedError::createFromException() so errors are returned with a valuable message other than just "internal server error" when GRAPHQL_DEBUG is true (can't believe how long this has been broken. . .)
- Add some more helpful InvariantViolation exceptions in TypeRegistry.php when fields/types aren't registered properly
- Fix a bug with how the fields were being registered to MediaItems numerous times
- Add some tests to make sure the Exceptions are being thrown as expected when fields are registered poorly

-----

Closes #619

### DIFF
--- a/src/TypeRegistry.php
+++ b/src/TypeRegistry.php
@@ -287,6 +287,10 @@ class TypeRegistry {
 		add_filter( 'graphql_' . $type_name . '_fields', function ( $fields ) use ( $type_name, $field_name, $config ) {
 
 			if ( isset ( $fields[ $field_name ] ) ) {
+				if ( true === GRAPHQL_DEBUG ) {
+					throw new InvariantViolation( sprintf( __( 'You cannot register duplicate fields on the same Type. The field \'%1$s\' already exists on the type \'%2$s\'. Make sure to give the field a unique name.' ), $field_name, $type_name ) );
+				}
+
 				return $fields;
 			}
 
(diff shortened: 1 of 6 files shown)


src/TypeRegistry.php
```diff
@@ -287,6 +287,10 @@ public static function register_field( $type_name, $field_name, $config ) {
 		add_filter( 'graphql_' . $type_name . '_fields', function ( $fields ) use ( $type_name, $field_name, $config ) {
 
 			if ( isset ( $fields[ $field_name ] ) ) {
+				if ( true === GRAPHQL_DEBUG ) {
+					throw new InvariantViolation( sprintf( __( 'You cannot register duplicate fields on the same Type. The field \'%1$s\' already exists on the type \'%2$s\'. Make sure to give the field a unique name.' ), $field_name, $type_name ) );
```
hughdevore: 👏

jasonbahl:
@hughdevore I updated this PR. Removed one of the exceptions that was actually going to cause us a lot of problems, and updated the tests to work properly. . .also caught a few edge cases I addressed as well with some places where there may be empty arrays we were trying to loop through

PR #635, opened 2019-01-04
CI-6012 | Refactoring Docker around Makefile; Using WP 5.0.2; Adding XDebug support

mngi-arogers:

- [x ] Make sure you are making a pull request against the **develop branch** (left side). Also you should start *your branch* off *our develop*.
- [ x] Make sure you are requesting to pull request from a **topic/feature/bugfix branch** (right side). Don't pull request from your master!

- Refactoring Docker configuration around Makefile to improve readability and maintenance
- Using WP 5.0.2 Docker image
- Adding XDebug support

No

N/A

Any other comments?
-------------------
N/A

Where has this been tested?
---------------------------
**Operating System:** Ubuntu Linux 18

**WordPress Version:** 5.0.2 with PHP 5.6, 7.1, and 7.2


chriszarate:
I sincerely appreciate the hard work on improving the Docker environment! A couple questions:

1. What is the value to the project of switching to `Makefile`? While it's certainly a (long-)accepted dev tool, I worry that it is more opaque to developers (at least in the WP space) than plain old shell scripts.

2. Is there value in _simplifying_ the Docker approach? (I think there is.) The current approach (while wonderful in its ability to solve for code coverage issues) is perhaps too complex. For users just starting with Docker, it is probably mystifying; I have a decent amount of experience and it took me a long time to fully understand what is happening between the containers when running tests. (Rsyncing files between container volumes, in particular, seems like an unnecessary complication.) I worry that it will quickly become a part of the project that Only One Person Knows.

mngi-arogers:
@chriszarate Hey thanks for taking a look at this!

1. Regarding the `Makefile`: In this PR I've changed the Docker directory structure to try to make file dependencies more obvious (e.g. certain shell scripts work with certain `docker-compose.yml` files but not others, etc). As a part of this change, some scripts that were at the root of the project were moved into subdirectories. So something was needed to allow a developer to run those Docker scripts with the appropriate environment variables from the root of the project. I used the `Makefile` to satisfy those needs since this is one use-case it solves and because it's widely available. However, you have a point about it being difficult to understand. I'll look at an approach using a shell script instead of a `Makefile`.

1. That `rsync` stuff (for the `test-environment` task) is a bit hairy because it needs to satisfy both of the following items:
   * account files changed by developer via their IDE
   * account for files downloaded by PHP composer without altering the developer's host filesystem (e.g. typically `vendor` files)

I couldn't figure out another way to allow a developer to frequently run a subset of tests in a fully isolated environment in an efficient way. But in any case, if the ROI of the `test-environment` task is low, it can be removed.

As for the general Docker approach I also don't want to have a situation where only one person knows how to do something. That said, I've worked under the assumption that most people on the project wouldn't want to run `docker run` or `docker-compose` directly since there's a number of "Dockerisms" to workaround. I also assumed plugin developers wanted to get up and running as quickly as possible (e.g. run WP with my plugin changes; run all of my tests, etc) regardless of whether Docker was being used. If that's not the case and people prefer to use those Docker commands directly to gain Docker experience, that approach can be taken as well.

mngi-arogers:
Closing this PR without merge in favor of the simpler approach.

Issue #638, opened 2019-01-05
Add filter to allow users to customise typeBy query args


hsimah:
### When you open an _issue_ for a feature request, please add as much detail as possible:
- [ ] A descriptive title
- [ ] A description of the problem you're trying to solve, including *why* you think this is a problem
- [ ] An overview of the suggested solution
- [ ] If the feature changes current behavior, reasons why your solution is better

I have a need to query a type by custom fields (not just ID, object ID, slug, uri). I want to extend the `typeBy` query instead of rolling my own.

I will need to add fields to `$post_by_args` and a custom resolver to `RootQuery`.

I suggest adding two filters, one for args and one for custom resolution.

hsimah:
To use:

```
    add_filter( 'graphql_userTutorialBy_args', function ( $post_by_args, $post_type_object ) {
        $post_by_args['tutorialId'] = [
            'type'        => 'Int',
            'description' => __( 'Gets the UserTutorial by linked Tutorial ID', 'wp-graphql' ),
        ];

        return $post_by_args;
    }, 10, 2 );
```

```
    add_filter( 'graphql_userTutorialBy_resolve', function( $post_type_object, $source, array $args, $context, $info ) {
        if ( ! empty( $args['tutorialId'] ) ) {
            global $wpdb;
            $user_id = get_current_user_id();

            $user_tutorial = null; // do your magic here - I use $wpdb to read a custom table

            if ( $user_tutorial && $user_tutorial->ID ) {
                return DataSource::resolve_post_object( $user_tutorial->ID, 'UserTutorial' );
            }
        }

        return null;
    }, 10, 5 );
```

jasonbahl:
@hsimah I'm closing this. Referencing my comment on the PR for context: https://github.com/wp-graphql/wp-graphql/pull/639#issuecomment-451993929

PR #639, opened 2019-01-05
#638 add two new post type specific filters

hsimah:

Allows users to customise the `typeBy` query input and resolution.

#638 

Any other comments?
-------------------
Hello again

Where has this been tested?
---------------------------
**Operating System:** Windows 10, Red Hat

**WordPress Version:** 5.0.2

### DIFF
--- a/src/Type/Object/RootQuery.php
+++ b/src/Type/Object/RootQuery.php
@@ -192,6 +192,9 @@ if ( ! empty( $allowed_post_types ) && is_array( $allowed_post_types ) ) {
 			];
 		}
 
+        /**
+         * Add any user customized input args for this type
+         */
 		$post_by_args = apply_filters( 'graphql_' . lcfirst( $post_type_object->graphql_single_name ) . 'By_args', $post_by_args, $post_type_object );
 
 		register_graphql_field( 'RootQuery', $post_type_object->graphql_single_name . 'By', [
@@ -218,6 +221,9 @@ if ( ! empty( $allowed_post_types ) && is_array( $allowed_post_types ) ) {
 					$slug        = esc_html( $args['slug'] );
 					$post_object = DataSource::get_post_object_by_uri( $slug, 'OBJECT', $post_type_object->name );
 				} else {
+                    /**
+                     * Invoke custom resolver if no default args have been supplied
+                     */
 					$post_object = apply_filters( 'graphql_' . lcfirst( $post_type_object->graphql_single_name ) . 'By_resolve', $post_type_object, $source, $args, $context, $info );
 				}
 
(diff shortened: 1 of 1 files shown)


jasonbahl:
Currently, filters technically exist already for this, although not as elegant as they could be. 

For example, you can currently filter in postBy args like so:

add_filter( 'graphql_RootQuery_fields', function( $fields ) {

  if ( isset( $fields['postBy'] ) {
    $args = ! empty( $fields['postBy']['args'] ? $fields['postBy']['args'] : [];
    $args['myNewArg'] = [
       'type' => 'String',
       'description' => __( 'Description of myNewArg', 'your-textdomain' ),
    ];
  }

} );

Then you can filter the args for the resolver like so: 

```
add_action( 'graphql_resovle_field', function( $result, $source, $args, $context, $info, $type_name, $field_key, $field, $field_resolver ) {

  // If the arg is set, do something with it and return the custom results
  if ( 'postBy' === $field_key && ! empty( $args['myNewArg'] ) {
    $result = do_something_with_your_new_arg( $source, $args['myNewArg'] );
  }

  return $result;

}, 10, 9 );
```

---------

I do think the filters you're adding make it more explicit that you can filter that particular entry point, but I think I'd prefer to solve this more centrally. . .possibly via introducing something like a `register_graphql_arg` function. 

I'm thinking we could probably have something like: `register_graphql_connection_arg`, `register_graphql_mutation_arg`, register_graphql_input_arg` (maybe others)

And then you could configure both the fields that should apply to the arg _AND_ what should be done with the arg's value during execution.

Something like: 

```
register_graphql_connection_arg( [
  'connectionName' => 'RootQueryToPostConnection',
  'field' => [
    'type' => 'string',
    'description' => 'description of the arg',
  ],
  'handleInput' => function( $arg ) {
     do_something_with_arg( $arg );
  }
] );
```

Totally pseudo code, but I'm thinking some entry point like this could do the heavy lifting of filtering the Schema to add the field to the Schema, but also filter the execution to add the `handleInput` (or whatever we should call it) do run during execution if the arg is present in the query/mutation. 

Need to think about it a bit more, but something along that line would be my ideal approach.

hsimah:
@jasonbahl I didn't know about the `graphql_resolve_field` filter - that was the main reason for doing this enhancement. Given the functionality exists, I will change my filters and close this PR.

I do like your idea of having a central way to extend queries. I would offer to help but I am swamped at the moment.

jasonbahl:
@hsimah one thing to be aware of though, is that the current `graphql_resolve_field` still executes the original resolve function. 

We probably should introduce a filter here that allows you to override the resolver without the default resolver executing. 

See: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Utils/InstrumentSchema.php#L126-L149

hsimah:
I'm looking at that now.

If the RootQuery resolver doesn't resolve an object it throws a UserError. I'll admit I don't know a lot about PHP exceptions, but wouldn't that then not return to InstrumentSchema to fire off the filter?

Issue #632, opened 2019-01-03
RootQuery 'By' uses invalid type for database ID


hsimah:
Looking at `RootQuery.php` and the definition of `$post_by_args` it looks like the types are incorrect:

- a GraphQL ID should be type `ID` (ie base64 encoded string), not type `String`
- a WP database ID should be type `Int`, not type `ID`

jasonbahl:
closed by #633

Issue #109, opened 2017-03-21
Inputs that expect an ID, should all be of the ID Type and should accept the Integer or Relay ID


jasonbahl:
In various Input args, we're currently setup to accept the WordPress ID as the input, but we should be accepting Global Relay IDs. 

For example, the `where` args for the `posts` query has an input of `authorIn` which accepts a list of Ints. . .this should accept a list of `ID`s and should parse out the WP ID from the Global ID.

Same goes for basically all inputs where the expected input is an ID

More examples, but not necessarily a comprehensive list

```
authorIn
authorNotIn
categoryAnd
categoryIn
in
notIn
parentIn
parentNotIn
tagAnd
tagIn
tagNotIn
```

CodeProKid:
I will try to come up with a comprehensive list in #221

jasonbahl:
related to #532

hsimah:
I don't think we should necessarily stop accepting the database IDs as arguments. There are a lot of plugins out there which rely on linking to posts by their ID. For example I integrate with MB Relationships and FacetWP. I personally don't like the idea of using `atob` in my client to modify an ID returned from the server to request an item.

I do think that there is scope for both input types. Perhaps we could come up with a naming standard where `id` is the relay ID and `databaseId` (or similar) is the int value. It would mean duplicated fields, but would ensure flexibility for the end user.

jasonbahl:
@hsimah ya, I _believe_ that if the Input type is ID that it can accept either an integer _or_ a String, so we can have the same input field accept either the database ID _or_ the Relay ID

Then let the execution handle what to do. 

Like, during execution, as we handle the input we can check something like the following (pseudo code): 

```
if ( is_int( $arg['inputId'] ) {
  $id = absint( $arg['inputId'] );
} else if ( is_relay_id( $args['inputId'] ) {
  $id = Relay::fromGlobalId()['id'];
} else {
  throw new UserError( 'The ID is invalid. . .' );
}
```

jasonbahl:
related: #532 and #263

jasonbahl:
Closing in favor of #997 and #998

PR #647, opened 2019-01-09
update post object connection arg types for ids

hsimah:
Related to #632 

`in`, `notIn`, `parentIn` and `parentNotIn` all should be `[Int]` and not `[ID]`. 

No

### DIFF
--- a/src/Connection/PostObjects.php
+++ b/src/Connection/PostObjects.php
@@ -246,25 +246,25 @@ class PostObjects {
 			],
 			'parentIn'     => [
 				'type'        => [
-					'list_of' => 'ID',
+					'list_of' => 'Int',
 				],
 				'description' => __( 'Specify objects whose parent is in an array', 'wp-graphql' ),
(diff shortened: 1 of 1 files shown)


jasonbahl:
@hsimah open this against develop. 

I think this one needs more discussion. We actually have issues open discussing the opposite. . .transitioning these inputs _from_ `INT` to `ID`, as that would allow flexibility of the input to be the WordPress database ID _or_ the Relay-style Global ID, where an `Int` wouldn't allow for the Relay ID to be passed.

jasonbahl:
see: https://github.com/wp-graphql/wp-graphql/issues/109

hsimah:
After sleeping on it - currently, while incorrectly documented, this works if you pass `"1056"` to `in`. WP resolves by the WP ID and returns the correct payload.

If we merge this it will be a breaking change to anyone who is using that as a work around.

I'll leave further comments on #109 .

PR #648, opened 2019-01-09
Release/0.2.0

jasonbahl:
# 🚀 RELEASE: v0.2.0

## Breaking Changes

- the filter `graphql_request_results` can return data in a different shape than previously returned. If making use of this filter, verify it's working as expected or make appropriate changes for your use case.
   - **WPGraphQL Insights (https://github.com/wp-graphql/wp-graphql-insights) uses this filter, so if you have this plugin active, be sure to update it to v0.3.0 needs to work properly with WPGraphQL v0.2.0+**
- the `do_graphql_request` action now executes _after_ the Schema has been built. In early days this hook was used sometimes to hook in and extend the Schema. If you have anything hooked here that modifies the Schema in some way, try hooking into `graphql_get_schema` instead.  

## New Features
- A new Request class has been introduced, simplifying and consolidating the entry points for execution of GraphQL requests. Thanks @chriszarate!
- A new `graphql()` function has been added as an entry point to the `Request` class. Ex: `$resposne = graphql([ .'query' => '{posts{edges{node{id,title}}}}' ]);`
- update menu items order to `post__in` instead of `menu_order`
- Update Type definitions for `post_by_args`. Thanks @hsimah!
   - ID was incorrectly expecting String instead of ID
   - postId was incorrectly expecting ID instead of Int
- Pass `currentConnection` and `connectionArgs` down through context to nested fields of a connection. See notes on usage here: https://github.com/wp-graphql/wp-graphql/pull/637


CodeProKid:
Skimmed through this. Looks good to me. Some great work in here by a few contributors!

Issue #649, opened 2019-01-09
Make more composer install friendly


mikelking:
Currently when installing the plugin via composer it includes a lot of other files that are probably not required for normal plugin to operation. Essentially everything under the repo path: https://github.com/wp-graphql/wp-graphql/ is included as part of the plugin installation. 

Adding wp-graphql to my composer was relatively straight forward:

1. added the repo tp the VCS section:
```
    {
      "type": "vcs",
      "url": "https://github.com/wp-graphql/wp-graphql"
    },
```
2. Added the item and version tag to the require section like I would any other plugin:
```
    "wp-graphql/wp-graphql" : "0.1.4",
```

I was able to quickly rebuild and deploy the site in minutes. However, there are a lot of files included like the Docker manifest and the associated shell scripts that probably are not required for normal plugin operation. Creating this issue as a reminder to clean things for production release.

mikelking:
I have created [PR650](https://github.com/wp-graphql/wp-graphql/pull/650) for this.

PR #650, opened 2019-01-10
added the .gitattributes file to test composer packaging and distribution

mikelking:
This is directly related to the root repo's issue [649](https://github.com/wp-graphql/wp-graphql/issues/649)

As per the recommendation I am testing this in accordance with https://madewithlove.be/gitattributes/


mikelking:
As far as I can tell the intended use of gitattributes does not state anything along the lines mentioned in the previous site. According to the git manual it is really a way to describe text conversions and conventions necessary for DIFF. So far it have not added the exclusion desired.

mikelking:
Ok I have conducted considerable testing with my personal project I have conducted https://github.com/mikelking/bacon and not sure exactly how but successfully used .gitattribute to change the composer artifacts. This construct is a bit of a pain in that it needs to be merged into an install referenceable version for testing with composer.

.gitattributes
```diff
@@ -0,0 +1,18 @@
+/tests export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/phpcs.ruleset.xml  export-ignore
+/phpunit.xml.dist export-ignore
+/ISSUE_TEMPLATE.md export-ignore
+/PULL_REQUEST_TEMPLATE.md export-ignore
+#/changelog.md export-ignore
+#/readme.md export-ignore
+/vendor export-ignore
```
jasonbahl: so, at the moment, if you were to download the Zip file from the Github releases, the `vendor` directory _is_ included, as we want folks to be able to download the Zip, install in WordPress and be up and running. 

So we still need to include the `vendor` directory, at least for the time being

mikelking:
@jasonbahl ,

I made the change by commenting our the vendor path from the exclusion. I also left the readme & change log as I believe those are beneficial to end users. Many examples on the web leave those out.

jasonbahl:
@CodeProKid any thoughts you want to add to this?

CodeProKid:
I think there are still some additional things that can get ignored.

- bin/
- img/
- CODE_OF_CONDUCT.md
- CONTRIBUTING.md

PR #651, opened 2019-01-12
CI-6012 | Supporting WP 5.0.2; Adding Docker XDebug for Linux; Streamlining Docker logic

mngi-arogers:

- [x ] Make sure you are making a pull request against the **develop branch** (left side). Also you should start *your branch* off *our develop*.
- [x ] Make sure you are requesting to pull request from a **topic/feature/bugfix branch** (right side). Don't pull request from your master!

- Flattening Docker directory structure and removing Docker wrapper script to make Docker logic easier to understand
- Using simpler implementation for running subsets of tests within Docker
- Adding XDebug+Docker support for Linux. MacOS will come later.
- Adding support for WP 5.0.2
- Updating the Docker Compose version used by Travis.

No

N/A

Any other comments?
-------------------
N/A

Where has this been tested?
---------------------------
**Operating System:** Linux

**WordPress Version:** WP 5.0.2


mngi-arogers:
Arg. The Docker build is running the `composer install` stuff twice resulting in longer PR test times. Let me fix that.

mngi-arogers:
Alright, this PR is ready for review.

.travis.yml
```diff
@@ -21,20 +21,19 @@ cache:
 env:
   matrix:
     # These apply to the latest WP version that is available as a Docker image: https://hub.docker.com/_/wordpress/
-    - PHP_VERSION=5.6
```
jasonbahl: 🤔 

So since we support PHP 5.6+, I want to make sure we know if we have tests breaking for 5.6. 

I'm not 💯 opposed to bumping our minimum version, but with WordPress core moving to PHP 5.6 min version in April, it seems to make sense to continue to support 5.6+ unless we have a serious need to change.

mngi-arogers: Ok I'll use WP 5.0.2. That version still has PHP 5.6-based Docker images.

mngi-arogers:
@chriszarate @jasonbahl This PR is ready for review.  By the way @jasonbahl , I removed that `WP_BROWSER_VERSION` environment variable. I didn't see it being used anywhere. Let me know if I need to leave it in.

jasonbahl:
Ah. Good call. I ended up removing that line but forgot to cleanup and remove the arg. Thanks!

mngi-arogers:
@jasonbahl Arg. Found another issue (from another Docker project) which applies to this PR. 

When a developer changes some files via an editor outside of the container and those specific files (as opposed to directories) are mounted into a Docker container, the container will not see the change. This is because some editors save to a temp file before renaming the file. This changes the host OS's filesystem records in a way where the container cannot see the change. 

This is probably why file mounts (containers, VMs, etc) are generally at the directory level and not at the file level.

More info:
https://github.com/moby/moby/issues/15793#issuecomment-453435094

jasonbahl:
@mngi-arogers how is what we’re doing different than mapping volumes? I don’t know enough about Docker to know how to accomplish anything, but I know that in Flywheel I just click a button to map my a volume and all my changes show up just fine. Not sure the technical differences between what they’re doing and what we’re doing though 🤷‍♂️

mngi-arogers:
@jasonbahl For the use-case where you simply need to run WP with your plugin code, you are correct--it is a matter of mapping the host OS's plugin directory into the right container location. No other dependencies need to be installed. That's probably what Flywheel is doing. That's also being done in this PR as the "local-app" use case.

However with the test-environment use-case (running subsets of tests within a container), special care has to be taken to make sure that tests can be run without the host filesystem becoming "contaminated" with random files from the container: For example, Codeception tests create/modify these files:
* `vendor/` directory which contains Codeception files that were installed via PHP Composer 
* `tests/_support/_generated/` directory which contains PHP files generated by Codeception
* `tests/_output/` directory which contains Codeception test output files

The way I tried to address it was to add mappings based on specific files. Here's an example:
   `../wp-graphql.php:/tmp/wordpress/wp-content/plugins/wp-graphql/wp-graphql.php:ro`
This led to the problem I mentioned previously. 

Now (with a couple of workarounds), I've updated the PR to map on a directory basis and not a file basis. Here's an example: `../:/tmp/wordpress/wp-content/plugins/wp-graphql/`this will avoid the problem I mentioned.

This PR is ready for review.

jasonbahl:
@mngi-arogers thanks for the added details!

jasonbahl:
@mngi-arogers reviewing this right now.

PR #662, opened 2019-01-23
🐛 FIX: Fixes PHP warnings in the Request class

### LINKED ISSUE #657: PHP Warnings for missing 2nd arg for `after_execute_actions`
in Request.php, the `after_execute_actions` should define the second argument to be optional to avoid php warnings: 

`private function after_execute_actions( $response, $key = null ) {`

jasonbahl:
This fixes a bug with `after_execute_actions` throwing PHP warnings that was causing problems when WP_DEBUG_DISPLAY is turned on.

Closes #657

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -160,7 +160,7 @@ class Request {
 	 *
 	 * @return array
 	 */
-	private function after_execute_actions( $response, $key ) {
+	private function after_execute_actions( $response, $key = null ) {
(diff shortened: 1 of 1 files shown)


jasonbahl:
Self merging 🤦‍♂️ . . .but this is already live on the DFM servers and I want to get this next release out since VIP has been asking about it...

PR #676, opened 2019-01-30
✨ Add TimezoneEnum Type

### LINKED ISSUE #675: Add "TimezoneEnum" Type
In the General Settings, there's an option for users with the `manage_options` capability to set the Timezone for the site. 

The Timezone setting is allowed to be one of a pre-defined set of values. This should be exposed to WPGraphQL as an Enum so the input can be validated against the static values.

See: https://github.com/WordPress/WordPress/blob/c204ac4bc7972c9ca1e6b354ec8fb0851e255bc5/wp-includes/functions.php#L5191

jasonbahl:
This adds the Type definition, but doesn't define any field to use it yet. . .this is to support the Settings Patch and get the settings all in "proper" order

Closes #675

### DIFF
--- a/src/Type/Enum/TimezoneEnum.php
+++ b/src/Type/Enum/TimezoneEnum.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace WPGraphQL\Type;
 
 /**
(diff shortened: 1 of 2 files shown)


jasonbahl:
@CodeProKid moved this to 0.2.3 🤷‍♂️

PR #661, opened 2019-01-23
Wordpress Core Settings Patch

kidunot89:
### Your checklist for this pull request
Thanks for sending a pull request! Please make sure you click the link above to view the contribution guidelines, then fill out the blanks below.

🚨Please review the [guidelines for contributing](../CONTRIBUTING.md) to this repository.

- [ ] Make sure you are making a pull request against the **develop branch** (left side). Also you should start *your branch* off *our develop*.
- [ ] Make sure you are requesting to pull request from a **topic/feature/bugfix branch** (right side). Don't pull request from your master!

What does this implement/fix? Explain your changes.
---------------------------------------------------
Registers the following options as settings using the WordPress Settings API, so they can be loaded to the $wp_registered_settings object and read by `DataSource::get_allowed_settings_by_group` and `DataSource::get_allowed_settings` functions.

**Discussion Settings**
- default_pingback_flag
- comments_notify
- moderation_notify
- comment_moderation
- require_name_email
- comment_whitelist
- comment_max_links
- moderation_keys
- blacklist_keys
- show_avatars
- avatar_rating
- avatar_default
- close_comments_for_old_posts
- close_comments_days_old
- thread_comments
- thread_comments_depth
- page_comments
- comments_per_page
- default_comments_page
- comment_order
- comment_registration
- show_comments_cookies_opt_in

**Media Settings**
- thumbnail_size_w
- thumbnail_size_h
- thumbnail_crop
- medium_size_w
- medium_size_h
- large_size_w
- large_size_h

**Permalink Settings**
- permalink_structure
- category_base
- tag_base

**Privacy Settings**
- wp_page_for_privacy_policy

**Reading Settings**
- posts_per_rss
- rss_use_excerpt
- show_on_front
- page_on_front
- page_for_posts
- blog_public
	
**Writing Settings**
- default_email_category
- default_link_category

Does this close any currently open issues?
------------------------------------------
…

Any relevant logs, error output, GraphiQL screenshots, etc?
-------------------------------------
(If it’s long, please paste to https://ghostbin.com/ and insert the link here.)

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 18.04

**WordPress Version:** 5.0.3

PR #677, opened 2019-01-30
✨ Add AvailableLocalesEnum Type

### LINKED ISSUE #674: Add "AvailableLocalesEnum" Type
On the General Settings page in the WP-Admin, there is a `Site Language` select that allows users with the `manage_options` capability select the language.

The option is hydrated based on a static list of Locales that are available.

In order to provide a similar experience in WPGraphQL, we need to expose the available options as an Enum and have the `language` field in the settings be of the type `AvailableLocalesEnum`

Something to this tune should work for the Type definition. . .then the Settings API can make use of that Type

```
<?php
namespace WPGraphQL\Type;

if ( ! function_exists( 'wp_get_available_translations' ) ) {
	require_once ABSPATH . '/wp-admin/includes/translation-install.php';
}

$values       = [];
$translations = wp_get_available_translations();

$values[ WPEnumType::get_safe_name( 'en_US' ) ] = [
	'value'       => 'en_US',
	'description' => _x( 'English (United States)', 'wp-graphql' ),
];

if ( ! empty( $translations ) && is_array( $translations ) ) {
	foreach ( $translations as $key => $value ) {
		$values[ WPEnumType::get_safe_name( $key ) ] = [
			'value'       => $key,
			'description' => $value['english_name']
		];
	}
}

register_graphql_enum_type( 'AvailableLocalesEnum', [
	'description' => __( 'Available translations for the site', 'wp-graphql' ),
	'values'      => $values
] );
```

jasonbahl:
This adds the Type definition, but doesn't define any field to use it yet. . .this is to support the Settings Patch and get the settings all in "proper" order

Closes #674 

Related to #661

### DIFF
--- a/src/Type/Enum/AvailableLocalesEnum.php
+++ b/src/Type/Enum/AvailableLocalesEnum.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace WPGraphQL\Type;
+
+if ( ! function_exists( 'wp_get_available_translations' ) ) {
+	require_once ABSPATH . '/wp-admin/includes/translation-install.php';
+}
+
+$values       = [];
+$translations = wp_get_available_translations();
+
+$values[ WPEnumType::get_safe_name( 'en_US' ) ] = [
+	'value'       => 'en_US',
+	'description' => _x( 'English (United States)', 'wp-graphql' ),
+];
+
+if ( ! empty( $translations ) && is_array( $translations ) ) {
+	foreach ( $translations as $key => $value ) {
+		$values[ WPEnumType::get_safe_name( $key ) ] = [
+			'value'       => $key,
+			'description' => $value['english_name']
+		];
+	}
+}
+
+register_graphql_enum_type( 'AvailableLocalesEnum', [
+	'description' => __( 'Available translations for the site', 'wp-graphql' ),
+	'values'      => $values
+] );
--- a/src/TypeRegistry.php
+++ b/src/TypeRegistry.php
@@ -82,6 +82,7 @@ class TypeRegistry {
 		 * Register core WPGRaphQL Types
 		 */
 		require_once( WPGRAPHQL_PLUGIN_DIR . 'src/Type/Object/Avatar.php' );
+		require_once( WPGRAPHQL_PLUGIN_DIR . 'src/Type/Enum/AvailableLocalesEnum.php' );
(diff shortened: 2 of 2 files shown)


src/Type/Enum/AvailableLocalesEnum.php
```diff
@@ -0,0 +1,29 @@
...
+$values[ WPEnumType::get_safe_name( 'en_US' ) ] = [
+	'value'       => 'en_US',
+	'description' => _x( 'English (United States)', 'wp-graphql' ),
+];
+
+if ( ! empty( $translations ) && is_array( $translations ) ) {
+	foreach ( $translations as $key => $value ) {
+		$values[ WPEnumType::get_safe_name( $key ) ] = [
+			'value'       => $key,
+			'description' => $value['english_name']
```
CodeProKid: Should probably use a ternary here and provide a default value if `english_name` doesn't exist.

src/Type/Enum/AvailableLocalesEnum.php
```diff
@@ -0,0 +1,29 @@
...
+	require_once ABSPATH . '/wp-admin/includes/translation-install.php';
+}
+
+$values       = [];
+$translations = wp_get_available_translations();
+
+$values[ WPEnumType::get_safe_name( 'en_US' ) ] = [
+	'value'       => 'en_US',
+	'description' => _x( 'English (United States)', 'wp-graphql' ),
+];
```
CodeProKid: I'm curious why this had to be defined independently? Does this not come through with `wp_get_available_translations`? Or is this meant to be a fallback?

jasonbahl:
@CodeProKid the more I think about this, the less I think this _should_ be an Enum. I think the option _should_ be validated against this before saving, but I feel like exposing the available locales in the Schema _could_ be a leak of implementation details that a site owner wouldn't necessarily want exposed. Seems like we should instead offer a `availableLocales` field that outputs a list of locales, and has proper cap-checks to make sure the user asking for the list has access, instead of exposing raw in the Schema. 

🤔

jasonbahl:
Closing this. I feel like this is exposing _potentially_ sensitive data that a site owner _may_ not want public. 

I think instead we should add a `getAvailableLocales` field to the RootQuery that returns the locales if the requesting user has permission to access (manage_options)

PR #679, opened 2019-01-31
Support more WP_Query options

craigmcnamara:

Support WP_Query `category__not_in` and `tag__not_in` filters.

I'll go look, if not I was going to open an issue.

Nope

Any other comments?
-------------------
This is a great project, going from zero to a pull request with tests was painless. Thanks!

Where has this been tested?
---------------------------
**Operating System:** OSX

**WordPress Version:** 5.0.3


CodeProKid:
I think we excluded these args on purpose because they aren't very performant. A better approach is to loop through returned posts and just exclude them if they are attached to a certain term.

We also support the full `tax_query` arg in this extension plugin -> https://github.com/wp-graphql/wp-graphql-tax-query

This does bring up a good point though, where we _can_ support args like this, and just approach it differently on the server side, rather than translating the args straight to WP_Query, we could use the args to loop through query results and exclude posts from terms _after_ the query has returned. This would get a little funky with pagination / returning the correct # of posts per page though.

@jasonbahl what are your thoughts?

renatonascalves:
I think adding a "hack" for the sake of it, increasing maintenance and possibly future problems is not a good idea either.

The performance implication is true even in the PHP API, so it is presumed the person is aware of that and is doing something to mitigate that, like adding cache.

craigmcnamara:
I've got this change running in production, behind a cache. I think it's probably better to fully support WP_Query options and expect people to use their judgement than to leave things out, because it makes the developer experience inconsistent.

I'm actually philosophically opposed to filtering things in code that can/should be filtered by the DB query. Codeland DB filtering adds all sorts of inconsistent pagination bugs and other unexpected behavior.

chriszarate:
I wouldn't characterize this as a hack, and I think @craigmcnamara has laid out a compelling case for its usefulness and how it could be used performantly.

However, I strongly believe that _out-of-the-box_ performance and security should be a critical factor for the feature set we pursue. This is the only way we keep the installation bar low across a wide spectrum of users and hosts—from hobbyists to WordPress VIP. Merging this PR would instantly raise flags for anyone that reviews code for performance.

Further, this PR could very easily be transformed into a plugin (e.g., `wp-graphql-term-not-in`), which is how I'd recommend it be implemented.

jasonbahl:
I think this should be plugin territory as well, same reason we've left out meta queries from the core plugin. Folks can still use Meta Queries via filtering in or using the WPGraphQL Meta Query plugin, but it's up to the owner of the site to take action to enable something that's commonly known to be a performance problem. 

However, if you do use Cursor Pagination with WPGraphQL, the "hack" @CodeProKid presented shouldn't cause issues with pagination. If we strip out "nodes" on the server before giving a response, we still have a cursor for each item returned and you can paginate appropriately using the cursors for the items returned. 

VIP actually recommends this route in our code as well. The homepages of our news sites show many categories of posts, and for de-duplication, we over-query, then strip out results and it's been quite a bit more performant than when we were using `not__in` queries. 

FWIW, The Model Layer we're [working on](https://github.com/wp-graphql/wp-graphql/search?q=model+layer&type=Issues) now will operate in a similar fashion, stripping out nodes that you don't have access to. 

For example, should you query for a private post (by ID perhaps) as an unauthenticated user, the Model Layer would determine that you don't have access to the post and would strip it from the response as if it never existed. Much like what happens in the Admin (or front-end) if you're a user that doesn't have access to it. 

----

> I think it's probably better to fully support WP_Query options and expect people to use their judgement and their profiling tools than to leave things out because of a performance concern.

If we were assuming the clients consuming the API were also the same folks exposing it, I'd be on board with this, but since the API is a publicly consumable API the moment you activate the plugin, we have a responsibility to make sure interacting with the API is safe for the site owner.

If the site owner feels it's safe for them to expose more options, I definitely think they should be able to via hooks/filters or extension plugins that do it for them (like @chriszarate proposed doing for this feature). 

The core REST API has made many similar performance decisions to make sure what's exposed by default is "responsible", such as limiting responses to a max of 100 items (even though WP_Query will let you make unbound queries, etc).

cr101:
I'm also using WordPress as a CMS and I need these args in my web app and it would be very useful to fully support them in the core plugin.
As @craigmcnamara said let the site owner who needs to use these args in their project decide how they solve any potential performance concerns.

jasonbahl:
@cr101 WP_Query isn't fully exposed to clients to interact with where a GraphQL API is. WP_Query is only interacted with by folks with server access writing PHP, so it's much more flexible in what's exposed to the developer interacting with it, but since GraphQL is exposed as a public API by default, enabling the full capabilities of WP_Query is both dangerous and irresponsible. If site-owners want to filter in capabilities that aren't supported out of the box, we'll always support their ability to do so, but exposing capabilities that could allow malicious clients to take down your site is not something I'm interested in supporting.

> let the site owner who needs to use these args in their project decide how they solve any potential performance concerns

That's what filters are there for. If you feel it's safe to expose to the consumers interacting with the site, filter it in. 

---

Is there a WordPress core UI that makes use of this feature even? Can you point me to a WP-Admin screen or one of the twenty-* core themes that uses these query args?

If so, then I'll consider adding support in some fashion, like maybe those args will be passed to the WP_Query by authenticated users but not public users or something to that tune 🤔

renatonascalves:
I was against it considering what I thought as a "hack" but the arguments outlined so far convinced me. We don't need to copy PHP API design into the GraphQL API. And if one would like to do that, one can filter it. So it is an easy "fix".

craigmcnamara:
Could someone explain to me why `tagIn` is ok but `tagNotIn` is a performance concern? It strikes me as odd that the negated version of something would be significantly different in real world performance.

jasonbahl:
@craigmcnamara after more discussion, I'm actually in favor of this now.

The `notIn` that I think we should do at the PHP level by overfetching and stripping is object IDs. 

So if you say `{ posts( first: 10 where: { notIn: [1, 2, 3] }) { ...` we could fetch `13` items, and if any are `1`, `2`, or `3`, we can strip them, then return the first 10 remaining results. 

But that doesn't work for relationships like tags, because we can't overfetch and have any gaurantee that the overfetch will have matching results to strip and return the user valid results. 

So for related `notIn` queries, it's cheaper at the SQL level than the PHP level, but for non-relationships, it (can be) cheaper at the PHP level to overfetch then strip out.

jasonbahl:
Merging this and we'll support for now. I do want to re-do the connection args at some point, but in the mean time I think this is useful to include.

PR #681, opened 2019-02-03
 Apply wp_unslash to query (GET) variables to undo wp_magic_quotes

chriszarate:

Adding variables to a `GET` request results in an error from `graphql-php`, e.g.:

```
GraphQL Request parameter "variables" must be object or JSON string parse
d to object, but got "{\\"foo\\":\\"bar \\"}"
```

This is because WordPress adds slashes to superglobals for safety reasons (`wp_magic_quotes`):

https://github.com/WordPress/WordPress/blob/92c3c46abe9f3b9e27764cbd483204530e8c1c2a/wp-includes/load.php#L761-L763

I had accounted for this and was stripping slashes on the `extensions` parameter, but didn't realize the same issue existed for `variables` and all other `GET` parameters. This PR applies `wp_unslash` to all `GET` query variables.

Even though WordPress also escapes `$_POST`, the same problem does not exist there because `graphql-php` reads the request body directly from `php://input`:

https://github.com/webonyx/graphql-php/blob/master/src/Server/Helper.php#L463

I looked around but didn't find one. This is critical for persisted query support, which is how I turned it up.

Screenshot of existing issue
---------------------
screenshot from 2019-02-02 20-53-11

Any other comments?
-------------------
Added tests. They fail without this fix.

Where has this been tested?
---------------------------
**WordPress Version:** 5.0.3

### DIFF
--- a/src/Server/WPHelper.php
+++ b/src/Server/WPHelper.php
@@ -23,8 +23,11 @@ class WPHelper extends Helper {
 	 * @throws RequestError
 	 */
 	public function parseRequestParams( $method, array $bodyParams, array $queryParams ) {
+		// Apply wp_unslash to query (GET) variables to undo wp_magic_quotes. We
+		// don't need to do this for POST variables because graphql-php reads the
+		// HTTP body directly.
 		$parsed_body_params = $this->parse_params( $bodyParams );
-		$parsed_query_params = $this->parse_extensions( $queryParams );
+		$parsed_query_params = $this->parse_extensions( wp_unslash( $queryParams ) );
 
 		/**
 		 * Allow the request data to be filtered. Previously this filter was only
@@ -67,7 +70,7 @@ class WPHelper extends Helper {
 	 */
 	private function parse_extensions( $params ) {
 		if ( isset( $params['extensions'] ) && is_string( $params['extensions'] ) ) {
-			$tmp = json_decode( stripslashes( $params['extensions'] ), true );
+			$tmp = json_decode( $params['extensions'], true );
 			if ( ! json_last_error() ) {
 				$params['extensions'] = $tmp;
 			}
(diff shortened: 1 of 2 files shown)


src/Server/WPHelper.php
```diff
@@ -23,8 +23,11 @@ class WPHelper extends Helper {
 	 * @throws RequestError
 	 */
 	public function parseRequestParams( $method, array $bodyParams, array $queryParams ) {
+		// Apply wp_unslash to query (GET) variables to undo wp_magic_quotes. We
+		// don't need to do this for POST variables because graphql-php reads the
+		// HTTP body directly.
 		$parsed_body_params = $this->parse_params( $bodyParams );
-		$parsed_query_params = $this->parse_extensions( $queryParams );
+		$parsed_query_params = $this->parse_extensions( wp_unslash( $queryParams ) );
```
jasonbahl: @chriszarate at one point we addressed this via this PR: https://github.com/wp-graphql/wp-graphql/pull/362/files

Probably didn't get fully translated during the change to the formal Request class? 

Curious what the difference between `wp_unslash` and `wp_kses_stripslashes` are and which one we should use 🤔

chriszarate: @jasonbahl Yeah looks like a regression on my part. 😭 

`wp_kses_stripslashes` is not recursive and brings it's own implementation. Seems like `stripslashes` is a better choice since it speaks directly to the source of the issue (`magic_quotes`).

```
function wp_unslash( $value ) {
    return stripslashes_deep( $value );
}
```

```
function wp_kses_stripslashes($string) {
    return preg_replace('%\\\\"%', '"', $string);
}
```

jasonbahl: nice! LGTM

Issue #618, opened 2018-12-11
Protect WP_User->user_pass from leaking


chriszarate:
Since type resolvers return the primitive WordPress object (e.g., `WP_User`), it opens up the possibility that a user's `user_pass` field could be indirectly exposed by WPGraphQL.

To be clear, that's not possible with WPGraphQL out-of-the-box, but a malicious plugin or theme could use `register_graphql_field` to add a field resolver to the `User` type and expose the user's hashed password.

Ideally we would delete or hide private fields like `user_pass` before passing them to field resolvers.

CodeProKid:
This should be handled in #604 adding to the milestone so we make sure it's handled. 👍

jasonbahl:
closed by #683

Issue #604, opened 2018-12-04
Model Layer: Users


CodeProKid:
Create a model layer for User objects

jasonbahl:
closed by #683

Issue #375, opened 2018-01-25
Return a post's visibility in the response


ginatrapani:
Related to #347 

Right now there's no way to see a post's "visibility" (as it's labeled in the WP admin): whether a post is private, password-protected, or public. 

We chatted about this a bit in the Slack: https://wp-graphql.slack.com/archives/C3NM1M291/p1516917534000156

Jason says:

We will likely return a `null` value for the node object if you don't have permission to view it, but we should provide an `edge` field that indicates the visibility. 

Something like: 

```{
  posts {
     edges {
        isProtected
        node {
           id
           title
        }
     }
  }
}
```

or something to that tune  would give a response like:

```{
  "data":{
     "posts": {
         edges: [
               {
                 isProtected: true,
                 node: null,
               },
               {
                 isProtected:false,
                node: {
                   id: "asdfasdfa",
                   title: "Hello World"
                }
            }
        ]
     }
  }
}```

CodeProKid:
This makes a ton of sense. Would be very valuable to know this on the render side of things. This should also be handled with the Model layer initiative. #280

CodeProKid:
One thing that still needs to be fleshed out here, is the verbiage to use for Private, Protected, And Public data. @jasonbahl and I have talked about this a lot, and we think those are essentially the 3 states a piece of data can have (with varying degrees of protected-ness in the protected status).

We just need to firm up what word to use for these statuses. Also, where to output this info in the schema. To me, it makes sense to be on the Node, but I could see how an argument could be made for this to be edge data.

Issue #280, opened 2017-10-18
Introduce Model Layer


jasonbahl:
It's common practice in GraphQL to separate the Schema/Type definitions and the resolvers. 

I think we should introduce a Model layer would help separate the Schema and the actual population of Data. 

This would allow the Schema to focus on _defining_ the contract between the Client and the Server and the Model to focus on _fulfilling_ that contract.

This would also centralize any logic we have about how things are resolved. 

This was the original reasoning behind the `DataSource` that currently exists. . .but we've just been inconsistent in utilizing it to its potential

CodeProKid:
I'm curious if this is something you still really want to do @jasonbahl

CodeProKid:
Reopening this as it now has a clearer vision. Creating smaller issues for each object type that needs a model.

jasonbahl:
Closed by #683 (and like 100 other stories)

PR #683, opened 2019-02-05
Feature/604 - Model layer abstraction and User implementation

CodeProKid:

This PR implements the new Model abstract class, as well as the User class which is implemented on top of it.

There's a couple of moving parts here, so I'll try to break things out, and outline my thoughts here.

## Model Class ##
I created a `Model` abstract for all of the other models to extend. This class attempts to centralize all of the functionality Models will share so that creating a new model will be simple and straight forward.

One focus of the model layer is to centralize hooking & filtering into data resolvers so there's a clear and consistent way to override/modify the data being resolved by any field. I added two new filters to achieve this: `graphql_pre_return_field_from_model` and `graphql_return_field_from_model`. The former can be used to return your own data for the field, and prevent the registered callback from firing at all. The latter is to modify the results of the registered callback after they are returned. These filters will be applied to all fields by `wrap_fields` method, basically we loop through all the fields in the model, and wrap the callbacks with our own callback in the model. This will help keep all of the individual models nice and clean.

Another concept we introduced in the `Model` is the idea of varying data visibility levels. I introduced three new fields to let a consumer know what sort of visibility into the data they have. The three fields are named `isPublic`, `isRestricted`, and `isPrivate`. The data will get added to the model automatically, so classes extending the model don't have to worry about it. Eventually I will probably move this functionality out to another class so we can register the fields to all types using the Model layer from one spot. One weird caveat is that the `isPrivate` field is kind of useless currently, because if we determine the object is private, we simply return `null` for the entire object, and no fields are returned.

## User Model ##
This PR also introduces the first actual data model which is for the `User` object. It's important to note that there are breaking changes with this PR from this. Access to all available fields for users is now restricted to users with the `list_users` capability, so if someone has an application that is openly querying Users without authentication, this will break their application. As we've talked about before this is probably a good breaking change, since we shouldn't be exposing all of that stuff by default anyways.

## Still in progress ##
I'm still cleaning up a few small things up, and I still need to unify filter/hook naming in the `Model` class as well as decide what to pass for context to those hooks & filters, and make sure the context is passed in a consistent order for the best possible developer experience.

## Moving Forward ##
After this is merged I will retrofit the Posts model that @jasonbahl did to work on top of this layer, and then continue to other objects outlined in the Model Layer milestone.

There's some weird composer things going on here (I think related to the version of composer I'm using), I'll try to clean that up as well before this goes in.

I'd love feedback on this, so if you have thoughts just comment here!

#280, #618, #604, #375 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** OSX

**WordPress Version:** 5.0.3


src/Model/Model.php
```diff
@@ -0,0 +1,299 @@
...
+		if ( ! is_array( $fields ) || empty( $fields ) ) {
+			return $fields;
+		}
+
+		$clean_array = [];
+		foreach ( $fields as $key => $callback ) {
+
+			$clean_array[ $key ] = function() use ( $key, $callback ) {
+				/**
+				 * @TODO: Revisit all of the filter/hook context here.
```
jasonbahl: what in particular do you want to revisit here?

vendor/composer/autoload_commands_classmap.php
```diff
@@ -6,4 +6,96 @@
 $baseDir = dirname($vendorDir);
 
 return array(
+    'CLI_Command' => $vendorDir . '/wp-cli/wp-cli/php/commands/src/CLI_Command.php',
```
jasonbahl: @CodeProKid looks like you need to run `composer install --no-dev` and commit again

Issue #684, opened 2019-02-07
Model Layer: Posts


CodeProKid:
Need to move the work @jasonbahl did on modeling Posts over to the new Model abstract implemented in #683

jasonbahl:
closed by #689

Issue #610, opened 2018-12-04
Model Layer: Menu Items


CodeProKid:
Create a model layer for menu items

jasonbahl:
closed by #686

Issue #486, opened 2018-08-09
Auth / accessing private posts


almorgan:
Hey guys,

I was just exploring the JWT plugin and was just testing a few things out / seeing how it works.

By default though (without the JWT plugin) I seem to be able to access private posts through the graphql queries whilst the default wp-json returns a "rest_forbidden" -> "Sorry, you are not allowed to do that."

Even if I install a plugin such as "Force Login" which redirects every page to the login and blocks all rest endpoints I still seem to be able to access all the sites data through the graphql queries (not logged in nor have I saved a JWT token).

Not sure if I'm missing something or potentially that bit hasn't been done yet?

Loving the plugin so far though!

Many thanks

jasonbahl:
@almorgan ya, we have some open issues to address this still. 

#375 and #245 are 2 that come to mind.

jasonbahl:
@almorgan happy to chat with you more if you're interested in working on helping with the solution. 😄

almorgan:
Hey @jasonbahl 

Thanks for the quick response.

Sure thing! quite new to graphql so can't promise anything but more than happy to have a look if you can point me in the right direction.

jasonbahl:
@almorgan are you in the wp-graphql Slack?

PR #686, opened 2019-02-09
Implement Post object model on new Model layer

CodeProKid:

This PR converts the PostObject model that @jasonbahl did to a new Post model that extends the new Model class. There is also a new model for Menu items as well. I have also added some tests for the private & restricted scenarios. Somer other underlying code also had to change in here for things to work properly to exclude private objects from edges/nodes.

#486, #684, #610 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** OSX

**WordPress Version:** 5.0.3


src/Model/Post.php
```diff
@@ -0,0 +1,390 @@
...
+					return ! empty( $this->post->post_date ) && '0000-00-00 00:00:00' !== $this->post->post_date ? $this->post->post_date : null;
+				},
+				'dateGmt'       => function () {
+					return ! empty( $this->post->post_date_gmt ) ? Types::prepare_date_response( $this->post->post_date_gmt ) : null;
+				},
+				'contentRendered' => function() {
+					$content = ! empty( $this->post->post_content ) ? $this->post->post_content : null;
+					return ! empty( $content ) ? apply_filters( 'the_content', $content ) : null;
+				},
+				'contentRaw' => function() {
```
jasonbahl: So, I think there's still some gray area to restricted. 

Like the RAW content should only be accessible to authenticated users. So even if the post is considered Public, the raw content is still restricted if the request isn't authenticated. 

The REST API handles this via an `edit` context. 

I think we can handle it via a simple `if ( empty( get_current_user_id() ) ) { return null; }` check, or something to that tune

I think some fields will need more precise treatment than the 3 common `public`, `restricted` and `private` states of the object at large.

And these cap checks should happen at the Model, not the GraphQL resolvers.

src/Model/Post.php
```diff
@@ -0,0 +1,390 @@
...
+					return ! empty( $this->post->post_content ) ? $this->post->post_content : null;
+				},
+				'titleRendered' => function() {
+					$id    = ! empty( $this->post->ID ) ? $this->post->ID : null;
+					$title = ! empty( $this->post->post_title ) ? $this->post->post_title : null;
+					return apply_filters( 'the_title', $title, $id );
+				},
+				'titleRaw' => function() {
+					return ! empty( $this->post->post_title ) ? $this->post->post_title : null;
```
jasonbahl: Same comment as above. Should only be allowed for authenticated requests, even if it's otherwise considered a public post.

src/Model/Post.php
```diff
@@ -0,0 +1,390 @@
...
+				},
+				'excerptRendered' => function() {
+					$excerpt = ! empty( $this->post->post_excerpt ) ? $this->post->post_excerpt : null;
+					$excerpt = apply_filters( 'get_the_excerpt', $excerpt, $this->post );
+					return apply_filters( 'the_excerpt', $excerpt );
+				},
+				'excerptRaw' => function() {
+					return ! empty( $this->post->post_excerpt ) ? $this->post->post_excerpt : null;
```
jasonbahl: same as above.

src/Model/Post.php
```diff
@@ -0,0 +1,390 @@
...
+					$uri = get_page_uri( $this->post->ID );
+					return ! empty( $uri ) ? $uri : null;
+				},
+				'commentCount'  => function () {
+					return ! empty( $this->post->comment_count ) ? absint( $this->post->comment_count ) : null;
+				},
+				'featuredImage' => function () {
+					$thumbnail_id = get_post_thumbnail_id( $this->post->ID );
+					return ! empty( $thumbnail_id ) ? DataSource::resolve_post_object( $thumbnail_id, 'attachment' ) : null;
+				},
```
jasonbahl: Doesn't look like we're exposing the `post_password` at the Model level. Probably something that wasn't exposed before, but we need to expose it to users with proper caps to see it.

If we had a UI in Nucleus for example where users could set a password on a Post, they'd need to be able to see the current password, if they had proper caps to see it.

src/Model/Post.php
```diff
@@ -0,0 +1,390 @@
...
+			];
+
+			if ( 'attachment' === $this->post->post_type ) {
+				$attachment_fields = [
+					'caption' => function() {
+						$caption = apply_filters( 'the_excerpt', apply_filters( 'get_the_excerpt', $this->post->post_excerpt, $this->post ) );
```
jasonbahl: So the REST API outputs the caption as `RAW` and `RENDERED`. . .much like we do for the excerpt. We should follow suit

And along with that only expose `RAW` to authenticated users.

src/Model/Post.php
```diff
@@ -0,0 +1,390 @@
...
+						return ! empty( $caption ) ? $caption : null;
+					},
+					'altText' => function() {
+						return get_post_meta( $this->post->ID, '_wp_attachment_image_alt', true );
+					},
+					'description' => function() {
+						return ! empty( $this->post->post_content ) ? apply_filters( 'the_content', $this->post->post_content ) : null;
```
jasonbahl: So the REST API outputs the caption as `RAW` and `RENDERED`. . .much like we do for the content of a Post. We should follow suit

And along with that only expose `RAW` to authenticated users.

src/Model/Post.php
CodeProKid: Yeah, it wasn't exposed before. I can add it in.

src/Model/Post.php
CodeProKid: Agreed

jasonbahl: Now, the question is do we use the condition, (is_logged_in()) to apply here in the callback, or apply to the allowed_restricted_fields?

Feel like the latter is more centrally managed and plugins, etc can more easily declare whether the fields should/should not be exposed by filtering the `allowed_restricted_fields` array vs having to filter individual resolvers 🤔

src/Model/MenuItem.php
```diff
@@ -0,0 +1,131 @@
...
+	 *                                  to build an object with those keys and their respective
+	 *                                  values.
+	 *
+	 * @access public
+	 * @return void
+	 * @throws \Exception
+	 */
+	public function __construct( \WP_Post $post, $filter = null ) {
+
+		if ( empty( $post ) ) {
```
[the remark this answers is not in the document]
CodeProKid: In PHP 7+ if the given value is of the incorrect type, PHP will throw a TypeError exception. See -> http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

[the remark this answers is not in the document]
jasonbahl: ah nice. and php 5.6 will just move on?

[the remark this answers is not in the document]
jasonbahl: ah, php 5: 

> in PHP 5, this will be a recoverable fatal error,

CodeProKid:
@jasonbahl made changes here. Ready for a follow up review. See latest commits.

PR #708, opened 2019-02-26
Do not execute graphql_request_data twice for GET requests

### LINKED ISSUE #707: graphql_request_data filter is executed twice for GET requests
Using the latest develop (947e07058befdeec2977062eeaddfc1e17d2b6b3) branch when graphql query is sent as a GET request the `graphql_request_data` is executed twice because `$parsed_body_params` is empty and `$parsed_query_params` is not

https://github.com/wp-graphql/wp-graphql/blob/947e07058befdeec2977062eeaddfc1e17d2b6b3/src/Server/WPHelper.php#L43-L46

This causes the proposed [lock mode implementation](https://github.com/Quartz/wp-graphql-persisted-queries/pull/1) in wp-graphql-persisted-queries to crash on GET requests because `graphql_request_data` is executed first with an empty array.

epeli:
Fixes #707

Avoids executing `graphql_request_data` filter twice on GET requests.

graphql-php uses only the query params on GET requets so executing the filter only for the query params on GET requests should be ok:

https://github.com/webonyx/graphql-php/blob/616fc10837c73f0a333c38f8aefa285cf70c5717/src/Server/Helper.php#L123

Added a second param to the filter so the users can access the query params on POST requests too if needed.

https://github.com/wp-graphql/wp-graphql/issues/707

Where has this been tested?
---------------------------
**Operating System:** …

Ubuntu Bionci, the Docker test env

**WordPress Version:** …

5.x

### DIFF
--- a/src/Server/WPHelper.php
+++ b/src/Server/WPHelper.php
@@ -29,6 +29,12 @@ class WPHelper extends Helper {
 		$parsed_body_params = $this->parse_params( $bodyParams );
 		$parsed_query_params = $this->parse_extensions( wp_unslash( $queryParams ) );
 
+		$request_context = [
+			'method' => $method,
+			'query_params' => ! empty( $parsed_query_params ) ? $parsed_query_params : null,
+			'body_params' => ! empty( $parsed_body_params ) ? $parsed_body_params : null
+		];
+
 		/**
 		 * Allow the request data to be filtered. Previously this filter was only
 		 * applied to non-HTTP requests. Since 0.2.0, we will apply it to all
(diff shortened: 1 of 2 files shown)


src/Server/WPHelper.php
```diff
@@ -39,10 +45,12 @@ public function parseRequestParams( $method, array $bodyParams, array $queryPara
...
 		 * @param array $data An array containing the pieces of the data of the GraphQL request
+		 * @param array $request_context An array containing the both body and query params
 		 */
-		$parsed_body_params = apply_filters( 'graphql_request_data', $parsed_body_params );
-		if ( ! empty( $parsed_query_params ) ) {
-			$parsed_query_params = apply_filters( 'graphql_request_data', $parsed_query_params );
+		if ( 'GET' === $method ) {
+			$parsed_query_params = apply_filters( 'graphql_request_data', $parsed_query_params, $request_context );
+		} else {
+			$parsed_body_params = apply_filters( 'graphql_request_data', $parsed_body_params, $request_context );
```
chriszarate: nice improvement! could we be a little more explicit about how GET and POST are treated differently? i feel like it would help others to be clear about what we expect from `graphql-php` and could help others interested in using this filter. e.g.:

```
if ( 'GET' === $method ) {
	$request_params = apply_filters( 'graphql_request_data', $parsed_query_params, $request_context );
	return parent::parseRequestParams( $method, [], $request_params );
}

$request_params = apply_filters( 'graphql_request_data', $parsed_body_params, $request_context );
return parent::parseRequestParams( $method, $request_params, [] );
```

sending query params on a POST request is totally valid and could be useful to separate core concerns from biz-logic. reading this, i know that for POST request i need to use 'graphql_request_data' to merge in any query params

epeli: Absolutely.

epeli: PR rebased.

PR #714, opened 2019-03-06
Setup postdata on field levels

epeli:
We noticed that the "Excerpt is always of the first post" (#503) issue has regressed and it was not caught by tests because the test added in #625 does not properly test the usage of the global post.

It does not use it because the `get_the_excerpt` filter uses the `wp_trim_excerpt` function by default and if a pre-existing excerpt is passed it just returns it:

https://github.com/WordPress/WordPress/blob/cddee8b72e59c9968b1c786c01992dae6f71ee13/wp-includes/formatting.php#L3685-L3686

Setting the `post_excerpt` string to an empty string causes `wp_trim_excerpt` to call `get_the_content( '' );` and thus the test fails.

I don't know how this should be really fixed but adding this to the `excerpt` resolver makes it pass again:

https://github.com/epeli/wp-graphql/commit/d174a6c41f82dbdbee4fcaaa816fa71e0fbc8139

but I'm pretty sure it's not the right solution so I didn't include it in this PR.

**UPDATE: I have now added a proper fix to this PR**

Where has this been tested?
---------------------------

The latest dev branch.

### DIFF
--- a/tests/wpunit/PostObjectConnectionQueriesTest.php
+++ b/tests/wpunit/PostObjectConnectionQueriesTest.php
@@ -732,12 +732,12 @@ class PostObjectConnectionQueriesTest extends \Codeception\TestCase\WPTestCase {
 
 		$post_1_args = [
 			'post_content' => 'Post content 1',
-			'post_excerpt' => 'Post excerpt 1',
+			'post_excerpt' => '',
 		];
 
 		$post_2_args = [
 			'post_content' => 'Post content 2',
-			'post_excerpt' => 'Post excerpt 2',
+			'post_excerpt' => '',
 		];
 
 		$post_1_id = $this->createPostObject( $post_1_args );
--- a/src/Data/DataSource.php
+++ b/src/Data/DataSource.php
@@ -179,14 +179,6 @@ class DataSource {
 			remove_filter( 'the_content', 'prepend_attachment' );
 		}
 
-		/**
-		 * Set the resolving post to the global $post. That way any filters that
-		 * might be applied when resolving fields can rely on global post and
-		 * post data being set up.
-		 */
-		$GLOBALS['post'] = $post_object;
-		setup_postdata( $post_object );
-
 		return $post_object;
 
 	}
--- a/src/Request.php
+++ b/src/Request.php
@@ -139,9 +139,14 @@ class Request {
 		 * This allows for a GraphQL query to be used in the middle of post content, such as in a Shortcode
 		 * without disrupting the flow of the post as the global POST before and after GraphQL execution will be
 		 * the same.
+		 *
+		 * We cannot use wp_reset_postdata here because it just resets the post from the global query which can
+		 * be anything the because the resolvers themself can set it to whatever. So we just manually reset the
+		 * post with setup_postdata we cached before this request.
 		 */
 		if ( ! empty( $this->global_post ) ) {
 			$GLOBALS['post'] = $this->global_post;
+			setup_postdata( $this->global_post );
 		}
 
 		/**
--- a/src/Utils/InstrumentSchema.php
+++ b/src/Utils/InstrumentSchema.php
@@ -18,6 +18,12 @@ use WPGraphQL\Type\WPObjectType;
 class InstrumentSchema {
 
 	/**
+	* Cache post for the resolvers so we can call the setup_postdata only when the actual
+	* source post changes
+	*/
+	private static $cached_post = null;
+
+	/**
 	 * @param \WPGraphQL\WPSchema $schema
 	 *
 	 * @return \WPGraphQL\WPSchema
@@ -115,7 +121,8 @@ class InstrumentSchema {
 						 * This ensures that functions like get_the_content() work correctly
 						 * so graphql queries can be used in the loop without issues.
 						 */
-						if ( is_a( $source, 'WP_Post' ) ) {
+						if ( is_a( $source, 'WP_Post' ) && self::$cached_post !== $source ) {
+							self::$cached_post = $source;
 							$GLOBALS['post'] = $source;
 							setup_postdata( $source );
 						}
--- a/src/Utils/InstrumentSchema.php
+++ b/src/Utils/InstrumentSchema.php
@@ -111,6 +111,16 @@ class InstrumentSchema {
 					$field->resolveFn = function( $source, array $args, AppContext $context, ResolveInfo $info ) use ( $field_resolver, $type_name, $field_key, $field ) {
 
 						/**
+						 * Setup the global post to the current post (if a post)
+						 * This ensures that functions like get_the_content() work correctly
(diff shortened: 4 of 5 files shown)


epeli:
I have now added a fixing commit to this PR too. It ensures that the global post is set for all field resolvers.

epeli:
Did some more digging and found `setup_postdata` call from the `resolve_post_object`

https://github.com/wp-graphql/wp-graphql/blob/c9ca28a1d1bdcf34a773220e28412b401dc4e907/src/Data/DataSource.php#L182-L188

I'm pretty sure this is too early to call it because during the [testPostExcerptsAreDifferent](https://github.com/wp-graphql/wp-graphql/blob/8726e8412147f066c40c5b657c6b75e748aa53f5/tests/wpunit/PostObjectConnectionQueriesTest.php#L731-L757) test it is called immediately for both test posts before the excerpt resolvers are called so it makes sense it returns wrong values because the last one "gets stuck" there for all resolver calls.

In the InstrumentSchema's `resolveFn` it is always setup up right before each resolver is called. There might be room for some optimization here but this seems ensure the right behaviour.

So I've added an commit that removes `setup_postdata` call from the `resolve_post_object`.

epeli:
So yeah... I basically rewrote this PR. We found more bugs when `do_graphql_request` is called within the loop because it does not reset the global query properly. I improved the existing tests to cover that case and removed one extraneous test that didn't make much sense (explained it in the commit message).

So this PR now rewrites how WPGraphQL calls setup_postdata and resets it. Previously it was done on the request level but because resolvers can resolve data for different posts the setup_postdata must called for each resolver and thus must reset right after the resolver is called.

tests/wpunit/PostObjectConnectionQueriesTest.php
```diff
@@ -732,12 +732,12 @@ public function testPostExcerptsAreDifferent() {
...
 		$post_1_args = [
 			'post_content' => 'Post content 1',
-			'post_excerpt' => 'Post excerpt 1',
+			'post_excerpt' => '',
 		];
 
 		$post_2_args = [
 			'post_content' => 'Post content 2',
-			'post_excerpt' => 'Post excerpt 2',
+			'post_excerpt' => '',
```
CodeProKid: I feel like we should leave the explicitly set `post_excerpt` arg here so we don't get some false positives on this working when it's actually not. The best way to test the global $post data is being setup correctly is by testing the content & excerpt data is actually different. With this change the excerpts are the same for both of the posts we're querying.

tests/wpunit/RequestTest.php
```diff
@@ -88,18 +88,6 @@ public function testRequestError() {
...
-	 * The request should not clobber the global post since it can be called in the Loop.
-	 */
-	public function testRequestPreservesGlobalPost() {
-		$GLOBALS['post'] = 'testing';
-
-		$this->factory->post->create();
-		$this->create_example_request()->execute();
-
-		$this->assertEquals( 'testing', $GLOBALS['post'] );
-	}
```
CodeProKid: Seems like it would be good to keep this test? Why delete it?

tests/wpunit/PostObjectConnectionQueriesTest.php
epeli: What I found was exactly the opposite. This test was passing when it should have failed. When excerpt has a value it does not need to generate the excerpt as it can return the value directly. In that case the `get_the_excerpt` filter here:

https://github.com/wp-graphql/wp-graphql/blob/c9ca28a1d1bdcf34a773220e28412b401dc4e907/src/Type/Object/PostObject.php#L265

does not use the global post for anything because the default filter in WP uses the `wp_trim_excerpt` function:

https://github.com/WordPress/WordPress/blob/d9a1f99d9ce352a0e151169c3612b00c4c146ddf/wp-includes/default-filters.php#L185


And the `wp_trim_excerpt` function checks for empty string here and bails out if there is something already:

https://github.com/WordPress/WordPress/blob/cddee8b72e59c9968b1c786c01992dae6f71ee13/wp-includes/formatting.php#L3685-L3712

tests/wpunit/RequestTest.php
epeli: I originally removed it because it wasn't working with my implementation because it does exercise a weird use case: Setting global post to a string but not calling `setup_postdata` which doesn't make any sense? 

The current implementation in this PR passes it now but due to its weirdness and because it seems that testPostQueryPostDataReset already tests this situation well enough I decided to leave it removed. 

https://github.com/wp-graphql/wp-graphql/blob/c9ca28a1d1bdcf34a773220e28412b401dc4e907/tests/wpunit/PostObjectQueriesTest.php#L1321

But if you think there's some merit to it I'll restore it.

tests/wpunit/PostObjectConnectionQueriesTest.php
epeli: But we could probably write a completely new and better test for this situation.

Fox example we could just assert that when querying multiple posts the global post is set to it inside the `graphql_resolve_field` filter.

CodeProKid: Ahhhhh, I see. So in this test, the post_excerpt should be the same as the post_content for each, shouldn't it?

epeli: Yeah, if the excerpt field is empty it will be generated from the content field on demand so it should be always different in this test.

tests/wpunit/RequestTest.php
jasonbahl: @epeli Ya, the use for this test was to make sure _whatever_ is set as the Global post would be the same _after_ a GraphQL request was made. 

Since you can use a GraphQL Query within the context of a post, such as within a Shortcode or whatever, we want to make sure that whatever the Global post is _before_ GraphQL execution is the same as it is _after_ GraphQL execution. In this case, we set the global to `testing` and assert that it's the same _after_ GraphQL executes. 

I believe at one point we weren't properly resetting, so if you used GraphQL within a shortcode, anything on the page after that (such as related posts after the conent, etc) would think the Global Post was the last item resolved by GraphQL. 

This test _should_ prevent that regression.

epeli: Fair enough. The test will indeed ensure wp-graphql stays resilient even if the user does something wonky with the global post. I'll restore it now.

tests/wpunit/PostObjectConnectionQueriesTest.php
epeli: I'll rebase my commits a bit. Will ping you when I think this is ready to merge.

tests/wpunit/RequestTest.php
jasonbahl: ya, we don't really care _what_ the global post was set to before a GraphQL request, we just want to ensure it's the same before and after. 😄 Thanks for fixing!

PR #715, opened 2019-03-06
Add 'link' root query

esamattis:

It adds an ability to query all post types by their permalinks.

No.

Any other comments?
-------------------

This makes it easier to implement clients for headless WordPress installations. Without this the client implementor must be aware of the rewrite setup WordPress is using.

This was discussed in the Slack earlier https://wp-graphql.slack.com/archives/C3NM1M291/p1551470085101500

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

5.0.5, Ubuntu Bionic.

### DIFF
--- a/src/Data/DataSource.php
+++ b/src/Data/DataSource.php
@@ -759,11 +759,30 @@ class DataSource {
 
 	}
 
-		/**
+	/**
      * Returns array of nav menu location names
      */
     public static function get_registered_nav_menu_locations() {
 			global $_wp_registered_nav_menus;
 			return array_keys( $_wp_registered_nav_menus );
+	}
+
+	/**
+	 * Get post object by its link.
+	 *
+	 * @param string $link from get_permalink() for example
+	 * @param string $post_type Optional. Post type; default is 'post'.
+	 *
+	 * @return \WP_Post|null WP_Post on success or null on failure
+	 * @see  https://codex.wordpress.org/Function_Reference/url_to_postid
+	 */
+	public static function get_post_object_by_link( $link, $post_type = 'post' ) {
+		$post_id = url_to_postid( $link );
+
+		if ( $post_id ) {
+			return self::resolve_post_object( $post_id, $post_type );
 		}
+
+		return null;
+	}
 }
--- a/src/Type/Object/RootQuery.php
+++ b/src/Type/Object/RootQuery.php
@@ -182,6 +182,10 @@ if ( ! empty( $allowed_post_types ) && is_array( $allowed_post_types ) ) {
 			'uri'                                         => [
 				'type'        => 'String',
 				'description' => sprintf( __( 'Get the %s by its uri', 'wp-graphql' ), $post_type_object->graphql_single_name ),
+			],
+			'link'                                         => [
+				'type'        => 'String',
+				'description' => sprintf( __( 'Get the %s by its link. Uses url_to_postid() to resolve the link.', 'wp-graphql' ), $post_type_object->graphql_single_name ),
 			]
 		];
 
@@ -215,6 +219,9 @@ if ( ! empty( $allowed_post_types ) && is_array( $allowed_post_types ) ) {
 				} elseif ( ! empty( $args['slug'] ) ) {
 					$slug        = esc_html( $args['slug'] );
 					$post_object = DataSource::get_post_object_by_uri( $slug, 'OBJECT', $post_type_object->name );
+				} elseif ( ! empty( $args['link'] ) ) {
+					$link   = esc_html( $args['link'] );
+					$post_object = DataSource::get_post_object_by_link( $link, $post_type_object->name );
 				}
 
 				if ( empty( $post_object ) || is_wp_error( $post_object ) ) {
--- a/tests/wpunit/PostObjectQueriesTest.php
+++ b/tests/wpunit/PostObjectQueriesTest.php
@@ -1538,4 +1538,74 @@ class PostObjectQueriesTest extends \Codeception\TestCase\WPTestCase {
 
 	}
 
+	/**
+	 * Test post can be queried its permalink
+	 */
+	public function testPostByPermalinkQuery() {
+		$post_id = $this->createPostObject( [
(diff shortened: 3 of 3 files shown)


tests/wpunit/PostObjectQueriesTest.php
```diff
@@ -1532,4 +1532,74 @@ public function testGraphQLQueryShortcodeInContent() {
...
+	/**
+	 * Test post can be queried its permalink
+	 */
+	public function testPostByPermalinkQuery() {
+		$post_id = $this->createPostObject( [
+			'post_type' => 'post',
+			'post_title' => 'Post title',
+		] );
+
+		$permalink = get_permalink($post_id);
```
jasonbahl: @epeli I believe you said this _should_ also work with the relative path? If so can we get a test for that as well to make sure it continues to work in the future?

Also, naming is hard. Is `permalink` a good name or is that a "WordPress" term? 🤔

jasonbahl:
@epeli looks like some merge conflicts though

tests/wpunit/PostObjectQueriesTest.php
epeli: I was thinking that test like that just tests the get_permalink() implementation.

Yeah, `permalink` is bit WordPressish. I chose it because the `url_to_postid` documentation says 

> $url  (string) (Required) Permalink to check.

And the current graphql api exposes a permalink field so I think it nicely analogous with it.

jasonbahl: @epeli on the PostObjects we expose it as `link`

We are fairly inconsistent here though. Some places we do `url`, some we do `sourceUrl` and here we do `link`

🤔

---

Did I mention that naming is hard? 😂 

---

Maybe for consistency with how it's exposed on the PostObjects, have it supported here as `link` instead of `permalink`?

epeli: Damn I'm pretty sure I saw permalink in somewhere in the graphql apis...

But if so `link` is better indeed.

jasonbahl: ya, it's `link` here: https://github.com/wp-graphql/wp-graphql/blob/79d2c4d0f2093553d94718962398863b5c7c3db8/src/Type/Object/PostObject.php#L386

epeli: It's now a `link`.

jasonbahl:
I really like the idea for this, but I want to stew on it a bit more. 

I think we likely want a root entry that is a Union of ANY Type that has a permalink. . .essentially a `nodeBy( url: "/path/to/thing")`

and then you could do:

```
nodeBy( url: "/path/to/thing") {
   ...on Page {
     ...pageFields
   }
   ...on Post {
      ...postFields
   }
   ...on Tag {
      ...tagFields
   }
   ...on User {
      ...userFields
   }
}
```

or something like that

Might be good to allow it for each Type as well, if you know for sure it's a Post ahead of time, but seems like if you have a URL you wouldn't necessarily know what Type it is, unless something in the URL indicates it, like `.com/post/path-to-post` or whatever 🤔

jasonbahl:
This is making me think we REALLY need to come up with some better methods of utilizing GraphQL interfaces and unions too 🤔

epeli:
We have use cases for both and `nodeBy` with unions is actually something I've been planning to propose too.

jasonbahl:
@epeli can you expand on a use case where you would have a URL that you know is a post, and could submit as input for `postBy( link: "url" )` but you couldn't use `uri`?

I think we discussed this before, but refresh my memory a bit 🤔

jasonbahl:
I guess in my head, something in the URL needs to indicate that it's a Post for you to even know that you should ask for `postBy( link: "url" )` vs asking for a page, or term or other post type, etc 🤔 

So just some context around the use case will help me think this through

epeli:
It's purely for developer convenience. We could use `uri` but it's awkward because with it one needs to know how the rewrite setup is configured for a given custom post type and modify the url to be compatible with `uri` semantics. Usage of the proposed `link` is dead simple. Just pass in the url in whatever format it happens to be in and it just works.

I will be teaching this stuff to junior devs and I don't want to explain how to figure out where rewrite setup is, how it relates to `uri` etc.

jasonbahl:
Ok. 🤔 

I think I want to think about it a bit more before making a final decision on merging.

src/Data/DataSource.php
```diff
@@ -759,11 +759,30 @@ public static function get_post_object_by_uri( $uri, $output = OBJECT, $post_typ
...
+	}
+
+	/**
+	 * Get post object by its link.
+	 *
+	 * @param string $link from get_permalink() for example
+	 * @param string $post_type Optional. Post type; default is 'post'.
+	 *
+	 * @return \WP_Post|null WP_Post on success or null on failure
+	 * @see  https://codex.wordpress.org/Function_Reference/url_to_postid
```
renatonascalves: This is not a rule but the `@see` tag usually comes before the `@return`: https://docs.phpdoc.org/references/phpdoc/tags/see.html

src/Type/Object/RootQuery.php
```diff
@@ -215,6 +219,9 @@
 				} elseif ( ! empty( $args['slug'] ) ) {
 					$slug        = esc_html( $args['slug'] );
 					$post_object = DataSource::get_post_object_by_uri( $slug, 'OBJECT', $post_type_object->name );
+				} elseif ( ! empty( $args['link'] ) ) {
+					$link   = esc_html( $args['link'] );
```
renatonascalves: If this is a url, `esc_url` should be used.

jasonbahl:
@esamattis I'm going to close this in favor of the upcoming v0.6.0 release where I add many new entry points to get nodes by URI, etc. 

Check out the PR for that, and if you feel like it's missing features that this would have accomplished, let me know and we can work through it. 

Thanks!

https://github.com/wp-graphql/wp-graphql/pull/1086

PR #721, opened 2019-03-13
Implement meta query ordering for the cursor based pagination

### LINKED ISSUE #287: orderby meta value?
I have a query that looks like this:

```
		 $recent = array(
					   'post_type' => 'post',
					   'paged' => ( get_query_var('paged') ? get_query_var('paged') : 1),
					   'posts_per_page' => '12',
						'orderby' => 'meta_value_num',
						'meta_key' => 'ratings',
						'order' => 'DESC',
						'cat' => $end
						
					   );
		$the_query = new WP_Query( $recent );
```

Is it possible to orderby meta key/value ? currently I am only seeing orderby options is limited to : Date, title, author, parent, slug, in.

Is it possible to extend the orderby options with custom functions of some sort?

Thanks

### LINKED ISSUE #729: Post order returned from cursors is not stable
When paginating data using Apollo I randomly get same items from different cursor pages

image

I cannot isolate this to a test because this happens randomly. I first noticed this while implementing meta value based ordering in #721 but today I noticed it happening without meta values too so I'm opening this new issue.

I'm not using custom ordering here but I'm using wp-graphql-tax-query based filtering.

Luckily my refactoring in #721 fixes this issue too. It changes the generated SQL query from

```sql
SELECT   wp_posts.ID FROM wp_posts  LEFT JOIN wp_term_relationships ON (wp_posts.ID = wp_term_relationships.object_id) WHERE 1=1  AND wp_posts.post_parent = 0  AND ( 
  ( 
    wp_term_relationships.term_taxonomy_id IN (23)
  )
) AND wp_posts.post_type = 'apartment' AND ((wp_posts.post_status = 'publish')) AND wp_posts.post_date <= '2019-03-18 19:04:34' AND wp_posts.ID != 481 GROUP BY wp_posts.ID ORDER BY wp_posts.post_date DESC LIMIT 0, 6
```

to

```sql
SELECT   wp_posts.ID FROM wp_posts  LEFT JOIN wp_term_relationships ON (wp_posts.ID = wp_term_relationships.object_id) WHERE 1=1  AND wp_posts.post_parent = 0  AND ( 
  ( 
    wp_term_relationships.term_taxonomy_id IN (23)
  )
) AND wp_posts.post_type = 'apartment' AND ((wp_posts.post_status = 'publish')) AND wp_posts.ID < 461 GROUP BY wp_posts.ID ORDER BY wp_posts.post_date DESC, wp_posts.ID DESC  LIMIT 0, 6
```

The post_date AND  `AND wp_posts.post_date <= '2019-03-18 19:04:30'` is gone and `, wp_posts.ID DESC` is added to the `GROUP BY` clause.

100% sure this is the best way to do but seems to fix my app. I can now get all values using cursors without duplicates.

epeli:
Closes #730
Fixes #729
Fixes #287

This PR does not add any public api to use meta queries but allows users to add meta query ordering in the `graphql_map_input_fields_to_wp_query` filter.

Example:

```php
add_filter('graphql_map_input_fields_to_wp_query', function( $query_args, $args ) {
  if (isset($args['custom_order']) {
    $query_args['orderby'] = [ 'meta_value' => 'ASC', ];
	$query_args['meta_key'] = 'test_meta';
  }

  return $query_args;
}, 10, 2);
```

Before this pull request the above would work for the first page but when the second page is requested using the after-cursor it would return the same page again. This pull request adds partial meta query ordering support into the `graphql_wp_query_cursor_pagination_support` method.

## Implementation

It is partial because meta queries can be written in various ways and we cannot reuse the WP core implementation because WPGraphQL passes the post id in its cursors and not the database limit&offsets (for performance reasons).

Here's the relevant documentation for meta query ordering:

https://codex.wordpress.org/Class_Reference/WP_Query#Order_.26_Orderby_Parameters

And the improvements made in WP 4.2:

https://make.wordpress.org/core/2015/03/30/query-improvements-in-wp-4-2-orderby-and-meta_query/

It can be a quite challenge to have comlete meta query support but I hope we can land even a partial support which covers most scenarios.

## Testing strategy

This is going to need multiple tests so I've added an `assertMetaQuery()` helper which can be used to test various meta query combinations with minimal code. It executes given meta query using `do_graphql_request()` and a plain `WP_Query` and compares the results.

Unfortunately the `testPostOrderingByNumberMetaKey` test I've added seems to be bit fragile now. Run it enough times and will randomly fail. I could not find any differences from the PHP code side so it might be related to the generated SQL query. It seems that the post offset in pages shift by one on some runs.

The error is always the same:

```
- Expected | + Actual
@@ @@
-'19,6,7,8,9'
+'6,19,7,8,9'
```

---------

This PR is not very pretty especially commitwise so I'm going to squash the commits once this PR is ready. So for now this is a Draft pull request.

Closes #729


jasonbahl:
@epeli you move too fast ⚡️ 🏃 🚄!!!

jasonbahl:
I'll try to review this shortly. I'm prepping a release here shortly as well. Get out some of these smaller features before @CodeProKid and I get our massive "Model Layer + DataLoader" release together

tests/wpunit/PostObjectConnectionQueriesTest.php
```diff
@@ -755,5 +755,89 @@ public function testPostExcerptsAreDifferent() {
...
+
+		add_filter( 'graphql_map_input_fields_to_wp_query', function( $query_args ) use ( $meta_fields ) {
+			return array_merge( $query_args, $meta_fields );
+		}, 10, 1 );
+
+		// Must use dummy where args here to force
+		// graphql_map_input_fields_to_wp_query to be executes
+		$query = "
+		query getPosts(\$cursor: String) {
+			posts(after: \$cursor, first: 5, where: {author: {$this->admin}}) {
```
jasonbahl: @epeli I don't see where the `order` is being set here in the query? What am I missing? 🤔

epeli: Its added by the `graphql_map_input_fields_to_wp_query` filter

jasonbahl: ah, I think I get it. You're testing that the capability works even if the query itself doesn't specify order, because the server may want to set a default order? eh?

jasonbahl: I guess I was assuming the test would have order set in the query like:

```
posts(after: \$cursor, first: 5, where: {author: {$this->admin}, orderBy: { "someMetaKey" } }) {
```

or something to that tune, but I think I'm on the same page now

epeli: Yeah, in real world code would do it like that. Just taking a shortcut in the tests.

epeli:
Added couple more tests and the random failing seem to happen more often now.

Executing wpunit with `shuffle: true` seems make it happen more often

    run wpunit --env docker tests/wpunit/PostObjectConnectionQueriesTest.php:testPostOrderingBy -o "settings: shuffle: true"

It's only one test that fails.

epeli:
Yay. Random failing fixed.

epeli:
Well. Actually an interesting point here: I never saw the plain WP_Query to return results in random order with the same meta values...

epeli:
Worked bit more on this.

## Misc

Moved meta query tests to a separate file

## Meta clauses

Implemented basic meta clause based ordering eg. this now works:

```php
			'orderby' => [ 'test_clause' => 'ASC', ],
			'meta_query' => [
				'test_clause' => [
					'key' => 'test_meta',
					'compare' => 'EXISTS',
				]
			]

```

This syntax allows ordering by multiple meta values but I was not able to make it work. My attempt to implement it is in a separate branch here:

https://github.com/epeli/wp-graphql/commit/3be85340672c70adcde630a31b249a0a443b090e

WP uses join aliases for the different different meta key&value pairs. My implementation now generates this SQL query

https://gist.github.com/epeli/630d5f1dc54a1cb884e651577403f537

For this meta query:

```php
			'orderby' => [
				'test_clause1' => 'ASC',
				'test_clause2' => 'ASC',
			],
			'meta_query' => [
				'test_clause1' => [
					'key' => 'test_meta1',
					'compare' => 'EXISTS',
				],
				'test_clause2' => [
					'key' => 'test_meta2',
					'compare' => 'EXISTS',
				]
			]
```


But it returns no results for the second page. Not sure why it does not work.

## Random result order

I was able to stabilize it by forcing a secondary ordering by the post IDs. This might be something which is required for normal field (`post_date` etc.) ordering too.

See https://github.com/wp-graphql/wp-graphql/pull/721/commits/490f0e60a6284ef62847d3be281c20a0dc76f314

epeli:
I implemented few more meta query features and the Config.php started to get out of hand so I extracted the cursor login to its own PostCursor class (basically rewrote it).

This is starting to look pretty good to me. Support for multiple meta query order fields is still missing but we can handle that in a separate PR. I think this is a pretty good base to build on now.

This now PR unblocks us from using wpgraphql in some of our projects now. So I'm marking this PR ready for review now.

epeli:
Found a bug related to ordering by post_title. It's now fixed and has a test. Renamed the test file to PostObjectCursorTest because it has tools for all kinds of ordering testing - not just meta field based.

jasonbahl:
@epeli cool if we close this in favor of #744?

epeli:
Yup

PR #722, opened 2019-03-14
⚡️ DataLoader

jasonbahl:
# Add DataLoader support

DataLoader is a concept of batching and caching within requests to reduce the number of underlying database queries needed to fetch the required data. 

In GraphQL, it's common that resolvers lead to a **n+1** problem, where one query is executed, then "n" more queries are executed to resolve the connected objects.

WPGraphQL was largely guilty of this problem. 

Take the following query as an example:

```
{
  posts( first: 10 ) {
    nodes {
       id
       title
       author {
         name
       }
     }
  }
}
```

This would execute 1 query to get the 10 posts, then 10 queries to get the author for Post 1, Post 2, etc

With this PR, this would execute 2 queries. 

The first query would execute to get the posts, then the `post_author` for each post is cached in a buffer, then a single deferred resolver is executed to fetch all 10 (or fewer if there are duplicates) authors, thus saving us up to 9 queries!

Prior to this PR, we had an initial implementation for Deferred resolvers for post authors, but that was the _only_ place we were making use of this concept of Deferred resolvers for batch queries. 

Now, it's implemented _pretty much_ everywhere. There are other improvements as well, so even the DataLoader setup for authors is more efficient than it was previously.

## Examples
Here are some before/after examples showing the improvements:

### Example 1:
- **before**:  20 SQL Queries
- **after**: 7 SQL queries
```
{
  posts {
    nodes {
      id
      title
      date
      author {
        name
      }
    }
  }
}
```

### Example 2
- **before**:  114 SQL Queries
- **after**: 67 SQL queries
```
{
  categories {
    nodes {
      id
      name
      posts {
        nodes {
          id
          title
          author {
            name
            posts {
              nodes {
                id
              }
            }
          }
        }
      }
    }
  }
}
```


renatonascalves:
That's fantastic!

src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -0,0 +1,495 @@
...
+		$max_query_amount = apply_filters( 'graphql_connection_max_query_amount', 100, $this->source, $this->args, $this->context, $this->info );
+
+		return min( $max_query_amount, absint( $this->get_amount_requested() ) );
+
+	}
+
+	/**
+	 * get_amount_requested
+	 *
+	 * This checks the $args to determine the amount requested, and if
```
hughdevore: ... go on...

src/Data/Connection/TermObjectConnectionResolver.php
```diff
@@ -166,10 +164,12 @@ public static function get_query_args( $source, array $args, AppContext $context
 		 * If the query is a search, the source isn't another Term, and the parent $arg is not explicitly set in the query,
 		 * unset the $query_args['parent'] so the search can search all posts, not just top level posts.
 		 */
-		if ( ! $source instanceof Term && isset( $query_args['search'] ) && ! isset( $input_fields['parent'] ) ) {
+		if ( ! $this->source instanceof Term && isset( $query_args['search'] ) && ! isset( $input_fields['parent'] ) ) {
 			unset( $query_args['parent'] );
 		}
 
+		$query_args['fields'] = 'ids';
```
hughdevore: Add a comment explaining that we want an array of IDs back from the term query

src/Data/Connection/TermObjectConnectionResolver.php
```diff
@@ -182,134 +182,45 @@ public static function get_query_args( $source, array $args, AppContext $context
...
 		 */
-		$query_args = apply_filters( 'graphql_term_object_connection_query_args', $query_args, $source, $args, $context, $info );
+		$query_args = apply_filters( 'graphql_term_object_connection_query_args', $query_args, $this->source, $this->args, $this->context, $this->info );
 
 		return $query_args;
-
 	}
 
 	/**
-	 * This runs the query and returns the response
```
hughdevore: Any reason you removed this?

src/Data/Connection/TermObjectConnectionResolver.php
```diff
@@ -182,134 +182,45 @@ public static function get_query_args( $source, array $args, AppContext $context
...
-				'hasNextPage'     => $has_next_page,
-				'startCursor'     => ! empty( $first_edge['cursor'] ) ? $first_edge['cursor'] : null,
-				'endCursor'       => ! empty( $last_edge['cursor'] ) ? $last_edge['cursor'] : null,
-			],
-			'nodes'    => $items,
-		];
-
-		return $connection;
-
+	public function get_items() {
```
hughdevore: Doc block

src/Request.php
```diff
@@ -139,9 +139,14 @@ private function after_execute( $response ) {
 		 * This allows for a GraphQL query to be used in the middle of post content, such as in a Shortcode
 		 * without disrupting the flow of the post as the global POST before and after GraphQL execution will be
 		 * the same.
+		 *
+		 * We cannot use wp_reset_postdata here because it just resets the post from the global query which can
+		 * be anything the because the resolvers themself can set it to whatever. So we just manually reset the
```
hughdevore: extra `the`

src/Type/Object/Comment.php
```diff
@@ -22,10 +25,36 @@
 		'commentedOn' => [
 			'type'        => 'PostObjectUnion',
 			'description' => __( 'The object the comment was added to', 'wp-graphql' ),
+			'resolve'     => function ( Comment $comment, $args, AppContext $context, ResolveInfo $info ) {
```
hughdevore: extra space between function and (

src/Type/Object/Comment.php
```diff
@@ -22,10 +25,36 @@
...
+					return null;
+				}
+				$id = absint( $comment->comment_post_ID );
+				return DataSource::resolve_post_object( $id, $context );
+			}
 		],
 		'author'      => [
 			'type'        => 'CommentAuthorUnion',
 			'description' => __( 'The author of the comment', 'wp-graphql' ),
+			'resolve'     => function ( Comment $comment, $args, AppContext $context, ResolveInfo $info ) {
```
hughdevore: extra space between function and (

src/Type/Object/Comment.php
```diff
@@ -48,7 +77,7 @@
 					'description' => __( 'Format of the field output', 'wp-graphql' ),
 				]
 			],
-			'resolve'     => function( $comment, $args ) {
+			'resolve'     => function ( Comment $comment, $args ) {
```
hughdevore: extra space between function and (

src/Type/Object/Comment.php
```diff
@@ -75,6 +104,17 @@
 		'parent'      => [
 			'type'        => 'Comment',
 			'description' => __( 'Parent comment of current comment. This field is equivalent to the WP_Comment instance matching the WP_Comment->comment_parent ID.', 'wp-graphql' ),
+			'resolve'     => function ( Comment $comment, $args, AppContext $context, ResolveInfo $info ) {
```
hughdevore: extra space between function and (

src/Type/Object/RootQuery.php
```diff
@@ -61,10 +63,10 @@
 					],
 				],
 			],
-			'resolve'     => function ( $source, array $args, $context, $info ) {
+			'resolve'     => function ( $source, array $args, AppContext $context, ResolveInfo $info ) {
 				$id_components = Relay::fromGlobalId( $args['id'] );
-
-				return DataSource::resolve_post_object( $id_components['id'], 'nav_menu_item' );
+				$id       = absint( $id_components['id'] );
```
hughdevore: Remove extra space before `=`

src/Type/Object/RootQuery.php
```diff
@@ -162,10 +168,15 @@
...
-			'resolve'     => function ( $source, array $args, $context, $info ) use ( $post_type_object ) {
+			'resolve'     => function ( $source, array $args, AppContext $context, ResolveInfo $info ) use ( $post_type_object ) {
 				$id_components = Relay::fromGlobalId( $args['id'] );
 
-				return DataSource::resolve_post_object( $id_components['id'], $post_type_object->name );
+				if ( ! isset( $id_components['id'] ) || ! absint( $id_components['id'] ) ) {
+					throw new UserError( __( 'The ID input is invalid', 'wp-graphql' ) );
+				}
+
+				$post_id       = absint( $id_components['id'] );
```
hughdevore: Remove extra space before `=`

src/Type/Object/TermObject.php
```diff
@@ -62,13 +66,28 @@ function register_taxonomy_object_type( $taxonomy_object ) {
...
+					foreach ( $ancestor_ids as $ancestor_id ) {
+						$ancestors[] = DataSource::resolve_term_object( $ancestor_id, $context );
+					}
+				}
+
+				return ! empty( $ancestors ) ? $ancestors : null;
+			}
 		] );
 
 		// @todo
```
hughdevore: Do we have an issue for this? If so, we should reference it in the comment

src/TypeRegistry.php
```diff
@@ -661,7 +680,7 @@ public static function register_connection( $config ) {
 				],
 			], $where_args ),
 			'description' => sprintf( __( 'Connection between the %1$s type and the %2s type', 'wp-graphql' ), $from_type, $to_type ),
-			'resolve'     => function( $root, $args, $context, $info ) use ( $resolve_connection, $connection_name ) {
+			'resolve'     => function ( $root, $args, $context, $info ) use ( $resolve_connection, $connection_name ) {
```
hughdevore: Remove extra space before (

tests/wpunit/AvatarObjectQueriesTest.php
```diff
@@ -205,6 +205,8 @@ public function testAvatarQueryNotFound() {
 		 */
 		$actual = do_graphql_request( $query );
 
+		codecept_debug( $actual );
```
hughdevore: Remove debugging code

tests/wpunit/CommentConnectionQueriesTest.php
```diff
@@ -109,6 +109,9 @@ public function testFirstComment() {
 		];
 
 		$results        = $this->commentsQuery( $variables );
+
+		codecept_debug( $results );
```
hughdevore: Remove debug code

tests/wpunit/MediaItemMutationsTest.php
```diff
@@ -482,6 +482,8 @@ public function testCreateMediaItemAttachToParent() {
 			],
 		];
 
+		codecept_debug( $actual );
```
hughdevore: Remove debug code

tests/wpunit/CommentMutationsTest.php
```diff
@@ -343,6 +343,8 @@ public function testUntrashComment()
 
 		$actual = do_graphql_request( $mutation, 'restoreCommentTest', $variables );
 
+		codecept_debug( $actual );
```
hughdevore: Remove debug code

tests/wpunit/MenuConnectionQueriesTest.php
```diff
@@ -29,6 +29,8 @@ public function testMenusQueryById() {
 
 		$actual = do_graphql_request( $query );
 
+		codecept_debug( $actual );
```
hughdevore: Remove debug code

tests/wpunit/MenuItemConnectionQueriesTest.php
```diff
@@ -103,6 +103,8 @@ public function testMenuItemsQueryWithNoArgs() {
 
 		$actual = do_graphql_request( $query );
 
+		codecept_debug( $actual );
```
hughdevore: Remove debug code

tests/wpunit/MenuItemConnectionQueriesTest.php
```diff
@@ -133,6 +135,8 @@ public function testMenuItemsQueryById() {
...
```
hughdevore: Remove debug code

tests/wpunit/MenuItemConnectionQueriesTest.php
```diff
@@ -161,6 +165,8 @@ public function testMenuItemsQueryByLocation() {
...
```
hughdevore: Remove debug code

tests/wpunit/MenuItemConnectionQueriesTest.php
```diff
@@ -218,10 +224,14 @@ public function testMenuItemsQueryWithChildItems() {
...
```
hughdevore: Remove debug code

tests/wpunit/MenuItemConnectionQueriesTest.php
```diff
@@ -218,10 +224,14 @@ public function testMenuItemsQueryWithChildItems() {
...
+
 		// Perform some common assertions.
 		$this->compareResults( $created['menu_item_ids'], $created['post_ids'], $actual );
 
 		// The fourth menu item has the expected number of child items.
+		codecept_debug( $child_count );
```
hughdevore: Remove debug code

tests/wpunit/MenuItemQueriesTest.php
```diff
@@ -8,6 +8,8 @@ public function testMenuItemQuery() {
 		$menu_id = wp_create_nav_menu( 'my-test-menu' );
 		$post_id = $this->factory()->post->create();
 
+		codecept_debug( $post_id );
```
hughdevore: Remove debug code

tests/wpunit/MenuItemQueriesTest.php
```diff
@@ -20,8 +22,12 @@ public function testMenuItemQuery() {
 			]
 		);
 
+		codecept_debug( $menu_item_id );
```
hughdevore: Remove debug code

tests/wpunit/MenuItemQueriesTest.php
```diff
@@ -20,8 +22,12 @@ public function testMenuItemQuery() {
...
+
 		$menu_item_relay_id = Relay::toGlobalId( 'nav_menu_item', $menu_item_id );
 
+		codecept_debug( $menu_item_relay_id );
```
hughdevore: Remove debug code

tests/wpunit/MenuItemQueriesTest.php
```diff
@@ -39,6 +45,8 @@ public function testMenuItemQuery() {
 
 		$actual = do_graphql_request( $query );
 
+		codecept_debug( $actual );
```
hughdevore: Remove debug code

tests/wpunit/NodesTest.php
```diff
@@ -64,6 +66,8 @@ public function testNodeQueryWithVariables() {
 		 */
 		$actual = do_graphql_request( $query, 'getPageByNode', $variables );
 
+		codecept_debug( $actual );
```
hughdevore: Remove debug code

tests/wpunit/NodesTest.php
```diff
@@ -317,6 +321,8 @@ public function testUserNodeQuery() {
 		";
 		$actual    = do_graphql_request( $query );
 
+		codecept_debug( $actual );
```
hughdevore: Remove debug code

tests/wpunit/NodesTest.php
```diff
@@ -491,8 +497,10 @@ public function testUnsuccessfulCommentResolver() {
 
 		$actual = do_graphql_request( $query );
 
+		codecept_debug( $actual );
```
hughdevore: Remove debug code

tests/wpunit/PostObjectQueriesTest.php
```diff
@@ -477,6 +458,8 @@ public function testPostQueryWithComments() {
 		 */
 		$actual = do_graphql_request( $query );
 
+		codecept_debug( $actual );
```
hughdevore: Remove debug code

tests/wpunit/PostObjectQueriesTest.php
```diff
@@ -1188,6 +1171,8 @@ public function testPostByQueryWithIDForADifferentType() {
...
```
hughdevore: Remove debug code

tests/wpunit/TermObjectConnectionQueriesTest.php
```diff
@@ -117,7 +117,7 @@ public function testfirstCategory() {
 		];
 		$results   = $this->categoriesQuery( $variables );
 
-		var_dump( $results );
+		codecept_debug( $results );
```
hughdevore: Remove debug code

tests/wpunit/TermObjectConnectionQueriesTest.php
```diff
@@ -132,6 +132,8 @@ public function testfirstCategory() {
 		] );
 		$terms = $query->get_terms();
 
+		codecept_debug( $terms );
```
hughdevore: Remove debug code

tests/wpunit/TermObjectQueriesTest.php
```diff
@@ -196,6 +196,8 @@ public function testTermQuery() {
 		 */
 		$actual = do_graphql_request( $query );
 
+
```
hughdevore: Remove extra lines

tests/wpunit/TermObjectQueriesTest.php
```diff
@@ -411,6 +413,8 @@ public function testTermQueryWithParentTerm() {
 			],
 		];
 
+
```
hughdevore: Remove extra lines

tests/wpunit/TermObjectQueriesTest.php
```diff
@@ -445,6 +449,8 @@ public function testTermQueryWhereTermDoesNotExist() {
 		 */
 		$actual = do_graphql_request( $query );
 
+		codecept_debug( $actual );
```
hughdevore: Remove debug code

tests/wpunit/UserObjectQueriesTest.php
```diff
@@ -349,6 +349,8 @@ public function testUserQueryWithPages() {
 		 */
 		$actual = do_graphql_request( $query );
 
+		codecept_debug( $actual );
```
hughdevore: Remove debug code

tests/wpunit/UserObjectQueriesTest.php
```diff
@@ -464,28 +466,15 @@ public function testUserQueryWhereUserDoesNotExist() {
...
```
hughdevore: Remove debug code

src/Data/Connection/PostObjectConnectionResolver.php
```diff
@@ -0,0 +1,380 @@
...
+	 *
+	 * @return array|null
+	 */
+	public function sanitize_post_stati( $stati ) {
+		if ( empty( $stati ) ) {
+			$stati = [ 'publish' ];
+		}
+		$statuses = wp_parse_slug_list( $stati );
+		$post_type_obj = get_post_type_object( $this->post_type );
+		$allowed_statuses = array_filter( array_map(function( $status ) use ( $post_type_obj ) {
```
CodeProKid: missing space between `array_map(` and `function()`

BTW, this is the most javascript PHP code I've ever seen.

src/Data/Connection/PostObjectConnectionResolver.php
```diff
@@ -0,0 +1,380 @@
...
+			if ( $status === 'publish' ) {
```
CodeProKid: yoda conditionals dawg

src/Data/Connection/PostObjectConnectionResolver.php
```diff
@@ -0,0 +1,380 @@
...
+				return $status;
+			}
+			if ( current_user_can( $post_type_obj->cap->edit_posts ) || 'private' === $status && current_user_can( $post_type_obj->cap->read_private_posts ) ) {
```
CodeProKid: Is there any logic that needs to be handled for `future` posts?

src/Data/Connection/UserConnectionResolver.php
```diff
@@ -148,10 +149,9 @@ public static function sanitize_input_fields( array $args, $source, array $all_a
 		 * @since 0.0.5
 		 * @return array
 		 */
-		$query_args = apply_filters( 'graphql_map_input_fields_to_wp_comment_query', $query_args, $args, $source, $all_args, $context, $info );
+		$query_args = apply_filters( 'graphql_map_input_fields_to_wp_comment_query', $query_args, $args, $this->source, $this->args, $this->context, $this->info );
```
CodeProKid: Any way we could centralize this in the abstract?

src/Data/Connection/UserRoleConnectionResolver.php
```diff
@@ -0,0 +1,80 @@
...
+	 * @return bool
+	 */
+	public function should_execute() {
+		if ( ! current_user_can( 'list_users' ) ) {
+			return false;
+		}
+		return true;
+	}
+
+//	/**
```
CodeProKid: Why is this commented out? Can we remove this?

src/Data/DataSource.php
```diff
@@ -443,10 +466,15 @@ public static function resolve_user_role( $name ) {
 	 * @param AppContext  $context The AppContext passed down to the query
 	 * @param ResolveInfo $info    The ResloveInfo object
 	 *
+	 * @throws \Exception
 	 * @return array
 	 */
 	public static function resolve_user_role_connection( $source, array $args, AppContext $context, ResolveInfo $info ) {
-		return UserRoleConnectionResolver::resolve( $source, $args, $context, $info );
+//		return UserRoleConnectionResolver::resolve( $source, $args, $context, $info );
```
CodeProKid: Delete this

src/Data/Loader/AbstractDataLoader.php
```diff
@@ -0,0 +1,275 @@
...
+
+		return $asArray ? iterator_to_array( $generator ) : $generator;
+	}
+
+	/**
+	 * @param $keys
+	 * @param $result
+	 *
+	 * @return \Generator
+	 */
```
CodeProKid: Docblock needs some cleanup

src/Data/Loader/AbstractDataLoader.php
```diff
@@ -0,0 +1,275 @@
...
+		foreach ( $keys as $key ) {
+			$key = $this->keyToScalar( $key );
+			yield isset( $result[ $key ] ) ? $this->normalizeEntry( $result[ $key ], $key ) : null;
+		}
+	}
+
+	/**
+	 * @return array
+	 * @throws \Exception
+	 */
```
CodeProKid: Docblock needs cleanup

src/Model/Comment.php
```diff
@@ -172,16 +160,9 @@ protected function init() {
...
-					if ( ! empty( $this->comment->comment_parent ) ) {
-						$parent_obj = get_comment( $this->comment->comment_parent );
-						if ( is_a( $parent_obj, 'WP_Comment' ) ) {
-							$parent = new Comment( $parent_obj );
-						}
-					}
-					return $parent;
-				},
+				'userId' => function() {
+					return ! empty( $this->comment->user_id ) ? $this->comment->user_id : null;
```
CodeProKid: Should this default to `0` instead of `null`? We are pretty inconsistent on wether the default is `null`, or the empty value within the scalar type `0`, `''`, etc...

src/Model/MenuItem.php
```diff
@@ -41,7 +39,7 @@ class MenuItem extends Model {
 	 * @throws \Exception
 	 */
 	public function __construct( \WP_Post $post ) {
-		$this->post = $post;
+		$this->post = wp_setup_nav_menu_item( $post );
```
CodeProKid: Might be better to have this happen in `DataSource` 🤔

src/Model/Post.php
```diff
@@ -88,6 +87,22 @@ public function __construct( \WP_Post $post ) {
...
+		$GLOBALS['post'] = $this->post;
+		setup_postdata( $this->post );
+
+		/**
+		 * Mimic core functionality for templates, as seen here:
+		 * https://github.com/WordPress/WordPress/blob/6fd8080e7ee7599b36d4528f72a8ced612130b8c/wp-includes/template-loader.php#L56
+		 */
+		if ( 'attachment' === $this->post->post_type ) {
+			remove_filter( 'the_content', 'prepend_attachment' );
+		}
```
CodeProKid: Should all of this stuff not happen in `DataSource`?

src/Model/Post.php
```diff
@@ -235,6 +249,7 @@ public function init() {
 					return ! empty( $this->post->post_date_gmt ) ? Types::prepare_date_response( $this->post->post_date_gmt ) : null;
 				},
 				'contentRendered' => function() {
+					setup_postdata( $this->post );
```
CodeProKid: This seems redundant?

src/Model/Post.php
```diff
@@ -256,6 +272,7 @@ public function init() {
 					'capability' => $this->post_type_object->cap->edit_posts,
 				],
 				'excerptRendered' => function() {
+					setup_postdata( $this->post );
```
CodeProKid: Why are you setting up the post data in the individual fields when it happens in the __construct already? Seems like this would hurt performance if querying multiple fields that set up the post data.

src/Request.php
```diff
@@ -139,9 +139,14 @@ private function after_execute( $response ) {
...
+		 * post with setup_postdata we cached before this request.
 		 */
 		if ( ! empty( $this->global_post ) ) {
 			$GLOBALS['post'] = $this->global_post;
+			setup_postdata( $this->global_post );
```
CodeProKid: Lol, seems like this is happening all over the place...

src/Data/Connection/MenuItemConnectionResolver.php
```diff
@@ -70,30 +84,44 @@ private static function get_menu_items( $source, array $args ) {
...
+		 * @param mixed       $source     The source that's passed down the GraphQL queries
+		 * @param AppContext  $context    The AppContext passed down the GraphQL tree
+		 * @param ResolveInfo $info       The ResolveInfo passed down the GraphQL tree
+		 */
+		$args = apply_filters( 'graphql_menu_item_connection_args', $this->args, $this->source, $this->context, $this->info );
 
 		// Prevent the query from matching anything by default.
 		$query_args = [
 			'post_type' => 'nav_menu_item',
-			'post__in'  => array( 0 ),
```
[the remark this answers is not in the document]
jasonbahl: apparently at one point this was needed to prevent some edge case. The `post__in` gets defined later, but only in certain cases, and I guess for certain cases this was needed as the default. 

I don't recall specifics now, but it seemed to be causing me issues after refactoring other internal.

src/Data/Connection/UserRoleConnectionResolver.php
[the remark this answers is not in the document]
jasonbahl: Ah, that was from the old class. I had it here as I refactored just for quick reference. Didn't mean to leave here. Moving to fast, per usual. 🤦‍♂️

src/Data/DataSource.php
```diff
@@ -7,10 +7,15 @@
 use GraphQL\Type\Definition\ResolveInfo;
 use GraphQLRelay\Relay;
 
+use Illuminate\Support\Facades\App;
```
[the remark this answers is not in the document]
jasonbahl: absolutely nothing. I must've been Typing `AppContext` and hit enter early and PHPStorm added it. 🤦‍♂️

src/Data/Loader/AbstractDataLoader.php
```diff
@@ -0,0 +1,275 @@
...
+		}
+	}
+
+	/**
+	 * @param $key
+	 *
+	 * @return mixed
+	 */
+	protected function keyToScalar( $key ) {
+		return $key;
```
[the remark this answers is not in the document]
jasonbahl: Ya, the docblock summary would help. It is intentional though.

src/Data/Loader/AbstractDataLoader.php
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	/**
+	 * @param $entry
```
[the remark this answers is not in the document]
jasonbahl: ya, it's meant to allow other loaders extending this to transform the data from the DB before being cached.

like if the raw DB query returned `{ id: 123 }` and we wanted the cache to be `{ ID: 123 }` the specific loader could override this method here and provide that transformation in this method

We're doing a lot of our transformations currently in the Model layer though

src/Type/Object/Comment.php
jasonbahl: Looks like I had a PHPStorm rule set weird for this one. Fixed the rule.

src/Type/Object/TermObject.php
jasonbahl: ha! That was resolved a while (few months) ago, just never cleaned up the todo.

src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -0,0 +1,495 @@
...
+	 *   return new WP_Query( $this->get_query_args() );
+	 *   return new WP_Comment_Query( $this->get_query_args() );
+	 *   return new WP_Term_Query( $this->get_query_args(); );
+	 *
+	 * Whatever it is will be passed through filters so that fields throughout
+	 * have context from what was queried and can make adjustments as needed, such
+	 * as exposing `totalCount` in pageInfo, etc.
+	 *
+	 * @var mixed
+	 */
```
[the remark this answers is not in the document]
jasonbahl: 🥇

src/Data/Connection/CommentConnectionResolver.php
```diff
@@ -0,0 +1,237 @@
...
+		/**
+		 * Pass the graphql $this->args to the WP_Query
+		 */
+		$query_args['graphql_args'] = $this->args;
+
+		/**
+		 * Handle setting dynamic $query_args based on the source (higher level query)
+		 */
+		if ( true === is_object( $this->source ) ) {
+			switch ( true ) {
```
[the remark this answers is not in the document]
jasonbahl: sadly, this cannot be done. 😢

src/Data/Connection/PostObjectConnectionResolver.php
jasonbahl: I get so bummed when my es6 doesn't execute in PHP!! 😬 

Functional programming ftw!

src/Data/Connection/PostObjectConnectionResolver.php
jasonbahl: ha! I think I copied this from the core REST API 😂 

The force must not be strong with it. . .

src/Data/Connection/PostObjectConnectionResolver.php
jasonbahl: Yezzir. . .

jasonbahl: Ok, so this is saying that if the user is querying for `publish` posts, allow it. 

If the status is anything else, and the user can edit posts of that post_type, allow that status (including future). 

If the status is private, they have to have "read_private_posts" and if they do, allow that status as well

Otherwise, return null

so, if I queried:

```
{
  posts( stati: [PUBLISH, FUTURE, PRIVATE]) {
    nodes {
      title
    }
  }
}
```

And I was a public user, I would end up executing 

```
WP_Query([ 'post_status' => ['publish'] ]);
```

But if I had `edit_posts` cap, it would be:

```
WP_Query([ 'post_status' => ['publish', 'future'] ]);
```

And if I had "read_private_posts" cap, it would be: 

```
WP_Query([ 'post_status' => ['publish', 'future', 'private'
] ]);
```

src/Data/Connection/UserConnectionResolver.php
jasonbahl: ya, I think that's a good idea. 🤔

jasonbahl: Ok, so I've added a filter to the AbstractConnectionResolver for this. Just trying to think if we fully deprecate that filter, or include some deprecated filter functionality that calls that old filter from the new one. 

For example, I added `apply_filters( 'graphql_connection_query_args', $this->get_query_args(), $this )`

So we could have a `Deprecated.php` that does something like: 

```
add_filter( 'graphql_connection_query_args', function( $args, $connection ) {
    apply_filters( 'graphql_map_input_fields_to_wp_comment_query', $args, $connection->source, $connection->args, $connection->context, $connection->info );
}, 10, 2 );
```

jasonbahl: I think I'm in favor of removing and keeping clean, documenting the update path, then coming up with a better deprecation method after v1.0.0 when we're communicating that we're stable

src/Model/MenuItem.php
jasonbahl: 🤔 Let me think about that a bit.

jasonbahl: Ya, I don't think so. DataSource now is essentially concerned with the Deferred resolution of nodes. The model is more concerned with actually determining what data should be returned to represent the node. 

Like here's the entire `DataSource::resolve_menu_item()` currently:

```
public static function resolve_menu_item( $id, AppContext $context ) {
		if ( empty( $id ) || ! absint( $id ) ) {
			return null;
		}
		$menu_item_id = absint( $id );
		$context->MenuItemLoader->buffer( [ $menu_item_id ] );

		return new Deferred( function () use ( $menu_item_id, $context ) {
			return $context->MenuItemLoader->load( $menu_item_id );
		} );
	}
```

It basically just asks the Loader to load the item.

The Loader then gets the object and passes it through the Model Layer, so it's the model layers job now to prepare the data for response (or return null)

jasonbahl: If there's a strong argument the other way, I'll consider, but I think the DataSource methods have a bit more of a clearer role now

src/Model/Comment.php
jasonbahl: well, if the ID is _actually_ 0 we'd want to return 0, but if the ID is 0 or falsey because it's not a user, we should return null, I think...

🤷‍♂️

src/Model/Post.php
jasonbahl: No. DataSource is essentially an extension of DataLoader, and it's primary concern is asking for data, not preparing the data. 

The Model Layer's job, now, is to respond with the prepared data.

Also, because of Deferred resolution, we're not composing objects all at once and returning them all at once, we may return one field initially, then return other fields, then come back for even more fields. 

So the Model Layer pretty much _has_ to be where this happens now.

src/Model/Post.php
jasonbahl: It's not really though . . . because of Deferred resolution now, there's no guarantee that we will be returning all fields of the same post at the same time. 

We might return the `title` of post 1, 2, 3

Then return `content` of post 1, 2, 3

So each field needs to reset the postdata if it's needed.

Otherwise the globals are all funky.

However, `setup_postdata` is luckily smart enough to read from the cache, so DataLoader is now adding the necessary info to the cache up front, and `setup_postdata` just sets the Globals to already cached data before the fields resolve. 

So we're not costing additional queries like we once were.

src/Model/Post.php
jasonbahl: See the previous comment. It's call cached now, we just have to reset the globals to play nice with deferred resolution

src/Request.php
jasonbahl: ya, see previous comments.

PR #737, opened 2019-03-25
Feature/additional loaders

kidunot89:

- `graphql_data_loaders` - for adding a loader class to the AppContext. It can also be used to overwrite loaders assigned to specific post-types and taxonomies
```
/**
 * @param string $loader - name loader is stored under in $context
 */
public static function resolve_post_object_loader( $loader, $post_id, $context, $post_type ) {
    $wc_post_types = array(
        'shop_coupon',
	'product',
	'product_variation',
	'shop_order',
	'shop_order_refund',
    );

    foreach( $wc_post_types as $post_type ) {
	$loaders[ $post_type ] = new WC_Loader( $context );
    }

    return $loaders;
}
```

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 18.04

**WordPress Version:** 5.1

### DIFF
--- a/src/Data/DataSource.php
+++ b/src/Data/DataSource.php
@@ -66,11 +66,14 @@ class DataSource {
 		}
 
 		$comment_id = absint( $id );
-		$context->CommentLoader->buffer( [ $comment_id ] );
+		$loader     = $context->getLoader( 'comment' );
+		$loader->buffer( [ $comment_id ] );
 
-		return new Deferred( function () use ( $comment_id, $context ) {
-			return $context->CommentLoader->load( $comment_id );
-		} );
+		return new Deferred(
+			function () use ( $loader, $comment_id ) {
+				return $loader->load( $comment_id );
+			}
+		);
 
 	}
 
@@ -199,13 +202,12 @@ class DataSource {
 			return null;
 		}
 		$post_id = absint( $id );
-
-		$loader = apply_filters( 'resolve_post_object_loader', 'PostObjectLoader', $post_id, $context, $post_type );
-		$context->{$loader}->buffer( [ $post_id ] );
+		$loader  = $context->getLoader( $post_type );
+		$loader->buffer( [ $post_id ] );
 
 		return new Deferred(
-			function () use ( $loader, $post_id, $context ) {
-				return $context->{$loader}->load( $post_id );
+			function () use ( $loader, $post_id ) {
+				return $loader->load( $post_id );
 			}
 		);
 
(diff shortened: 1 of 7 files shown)


src/Data/DataSource.php
```diff
@@ -193,17 +193,21 @@ public static function resolve_plugins_connection( $source, array $args, AppCont
...
 		if ( empty( $id ) || ! absint( $id ) ) {
 			return null;
 		}
 		$post_id = absint( $id );
-		$context->PostObjectLoader->buffer( [ $post_id ] );
 
-		return new Deferred( function () use ( $post_id, $context ) {
-			return $context->PostObjectLoader->load( $post_id );
-		} );
+		$loader = apply_filters( 'resolve_post_object_loader', 'PostObjectLoader', $post_id, $context, $post_type );
```
jasonbahl: Hey, so I was thinking the filter should happen in the Context, not here, because the Loaders can be used outside of these DataSource methods

Any resolve function should be able to do something like `$context->loaders['post_object']->buffer([$post_id]);`

So, instead of having an action in context, we would have one filter, and `AppContext->loaders` would be an array of loaders, passed through the filter. So folks could add custom loaders, or even replace loaders using that filter. 

Then the loaders are centralized and accessible to all resolvers.

jasonbahl: @kidunot89 mentioning you to make sure you get notified of my comment 😄

src/Model/Post.php
```diff
@@ -73,6 +73,14 @@ class Post extends Model {
 	 */
 	protected $post_type_object;
 
+	/**
+	 * The ID used to identify the fields' origin.
+	 *
+	 * @var array $fields
+	 * @access public
+	 */
+	protected $fields_id_name = 'post';
```
jasonbahl: @kidunot89 what is this for? I don't see it being used 🤔

kidunot89: My bad. I forgot to remove it.

src/Model/Post.php
```diff
@@ -125,6 +132,7 @@ public function __construct( \WP_Post $post ) {
 
 		parent::__construct( 'PostObject', $post, $restricted_cap, $allowed_restricted_fields, $post->post_author );
 		$this->init();
+		parent::prepare_fields();
```
kidunot89: This is so more fields can be added in extended class before they are processed

jasonbahl: @kidunot89 why did this need to move?

jasonbahl: I think you might want to just create a NEW model in that case. . .maybe? 🤔

jasonbahl: Maybe worth discussing with @CodeProKid a bit too

CodeProKid: You can use the `graphql_return_modeled_data` filter for this -> https://github.com/wp-graphql/wp-graphql/blob/feature/model-layer/src/Model/Model.php#L401

jasonbahl: ^ what @CodeProKid said. 

That filter will allow third party plugins, etc to extend the models as well. So they can define fields in one place, define their access checks for those fields, etc. . .then keep that separate from the resolvers

kidunot89: Oh, I completely missed that filter. @CodeProKid @jasonbahl Thanks . I changed it back.

jasonbahl:
@kidunot89 I'm closing this and re-opening as a PR to the "model-layer" branch since the DataLoaer stuff was merged there.

PR #751, opened 2019-03-29
🐛 FIX ISSUE WITH ROLE BEING SET DURING REGISTER USER MUTATIONS

jasonbahl:
This removes roles from allowed input in the registerUser mutation, and also unsets role when the mutation calls wp_update_user, so even plugins that filter in a role input can't save a role during this particular mutation.

### DIFF
--- a/src/Mutation/UserRegister.php
+++ b/src/Mutation/UserRegister.php
@@ -24,7 +24,7 @@ class UserRegister {
      * @return array
      */
     public static function get_input_fields() {
-        return array_merge( UserCreate::get_input_fields(), [
+        $input_fields = array_merge( UserCreate::get_input_fields(), [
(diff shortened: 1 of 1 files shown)


acafourek:
Was just considering how I might implement `role` as a field on RegisterUser when I spotted this merge. I know this was from awhile ago, but any memory into why it was decided to explicitly disallow role on input?

jasonbahl:
@acafourek WordPress has a default user role that will be used. When users register, they will be granted that default role. The registering user doesn't have permission to declare what role they would like to be.

acafourek:
Ah of course; I was conflating "create" vs "register". Makes sense to rely on the WP setting for enforcement in registration - thanks!

PR #771, opened 2019-04-03
Feature/#375 - Expose `isRestricted`

### LINKED ISSUE #375: Return a post's visibility in the response

jasonbahl:
- This adds `isRestricted: Boolean` fields to the various ObjectTypes that have an associated model. This allows for queries to know whether a node isRestricted or not. This gives client some insight into whether there is actually no data for a particular field, or whether there's no data returned because it's restricted.

I think there's room to expand this further down the road so the client can have even more info on what fields are public/restricted, etc. Could help build UIs properly if the consumer knows that x field is null because it's listed as a restricted field, vs it's null because there is simply no underlying data.

- This also fixes a bug with the PostObjectConnectionResolver to prevent querying a connection if the user specifies ONLY post_stati that they don't have permission to view.

====

closes #375


jasonbahl:
meant to open to the Model Layer branch 🤦‍♂️

PR #789, opened 2019-04-08
Prevent autoloader if GraphQL class exists

homerjam:

This prevents us having to set an environment variable to prevent autoloading dependencies. Useful when installing as a composer package.

Not AFAIK

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** Docker centos

**WordPress Version:** latest

### DIFF
--- a/wp-graphql.php
+++ b/wp-graphql.php
@@ -216,7 +216,7 @@ if ( ! class_exists( 'WPGraphQL' ) ) :
 			 * The codeception tests are an example of an environment where adding the autoloader again causes issues
 			 * so this is set to false for tests.
 			 */
-			if ( defined( 'WPGRAPHQL_AUTOLOAD' ) && true === WPGRAPHQL_AUTOLOAD ) {
+			if ( ! class_exists('GraphQL\GraphQL') && defined( 'WPGRAPHQL_AUTOLOAD' ) && true === WPGRAPHQL_AUTOLOAD ) {
 				// Autoload Required Classes.
 				require_once WPGRAPHQL_PLUGIN_DIR . 'vendor/autoload.php';
 			}


wp-graphql.php
```diff
@@ -216,7 +216,7 @@ private function includes() {
 			 * The codeception tests are an example of an environment where adding the autoloader again causes issues
 			 * so this is set to false for tests.
 			 */
-			if ( defined( 'WPGRAPHQL_AUTOLOAD' ) && true === WPGRAPHQL_AUTOLOAD ) {
+			if ( ! class_exists('GraphQL\GraphQL') && defined( 'WPGRAPHQL_AUTOLOAD' ) && true === WPGRAPHQL_AUTOLOAD ) {
```
jasonbahl: 🤔 

So, you're installing GraphQL-PHP as a composer dependency separate from WPGraphQL?

WPGraphQL can only really guarantee that it will work with one specific version of GraphQL PHP, so it makes me a bit nervous to allow folks to install them separately as that feels like it will increase our support burden. 

Happy to learn more about your use case to see if it makes sense to support this situation. 

Right now, WPGraphQL is (somewhat) tightly coupled with a specific version of GraphQL-PHP, and as we move forward we can only really promise to work with the specific version we test against.

homerjam: I'm letting composer handle all the dependencies in my project. The reason is to avoid the same libs being loaded by different plugins and to keep the compiled code size to a minimum (I'm running WordPress in a serverless environment so this is important to keep cold boot times to a minimum).

As such this is really just a convenience. It could also prevent issues if wp-graphql is included alongside another plugin that uses the graphql dependency.

jasonbahl: I'm just nervous that I'm going to have to bear the weight of support tickets for folks bundling their own version of graphql-php and wondering why a feature in WPGraphQL isn't working. 

I'll need to think about this. Simple change, but could potentially open up support requests that I waste more time on than I should 🤔

homerjam: Ok, sounds sensible - with the `WPGRAPHQL_AUTOLOAD` flag I can make it work in my setup. It might be worth leaving the PR open in case others run into a similar issue inadvertently. Thanks for all your work!

jasonbahl:
@homerjam going to close this pr. I believe it should still be surfaced if folks are searching things like Autoloader, etc.

PR #862, opened 2019-05-23
#818 - Cursor Pagination Broken with Custom Orderby

### LINKED ISSUE #818: Cursor pagination not working with custom ordering
When paginating with custom orderby paramaters on posts queries, the results don't come back as expected. 

More details to come. . .but shortly, but wanted to get an issue filed.

jasonbahl:
This removes the `post_` prefix from the `compare_with` method to avoid issues with cursor pagination.

Closes #818 

----

There's still an edge case bug causing orderby title to skip a post if 2 adjacent posts have the same title. 

Very edge case-y, so will follow up with another bug to address.


jasonbahl:
closing in favor of #864

PR #863, opened 2019-05-23
#786 - Support sticky posts if enabled

### LINKED ISSUE #786: Sticky Posts are not being handled anymore in 0.3.0
At first I thought, that the addition of 

https://github.com/wp-graphql/wp-graphql/blob/86a7d9f0a63bac9ed8f42a1b178bcdbe5162a28f/src/Data/Connection/PostObjectConnectionResolver.php#L117

brought up this error, but it seems to be way deeper than this. It appears that since 0.3.0 fetching the posts related to a query occurs in two steps: first the IDs of the posts are gathered, then the posts for the given ids are loaded.

Because of this change, sticky posts are never handled, as the querying for sticky posts is performed at a later time in wp-query.

https://github.com/WordPress/WordPress/blob/94b592ac684d7cc9a82b5ba42161d320a4c329f4/wp-includes/class-wp-query.php#L2906-L2916

When `fields` is set to `ids` in the query vars, the query returns before fetching and adding the sticky posts which only happens after line 3077 in class-wp-query. 
https://github.com/WordPress/WordPress/blob/94b592ac684d7cc9a82b5ba42161d320a4c329f4/wp-includes/class-wp-query.php#L3077

Thus sticky posts will never be handled. 

To be honest I have no clue how to solve this, as this requires far deeper knowledge for the way wp-graphql loads the data, so at the moment I'm quite overwhelmed with how to solve this.

jasonbahl:
Closes #786 

This adds a new `graphql_include_sticky_posts` filter. Default is `false`, but if set to `true` WPGraphQL will include sticky posts at the top of Post Object queries (of the "post" post_type). 

For paginated queries, the sticky post will only be included in the first query of forward paginated queries.

### DIFF
--- a/src/Data/Connection/PostObjectConnectionResolver.php
+++ b/src/Data/Connection/PostObjectConnectionResolver.php
@@ -62,7 +62,21 @@ class PostObjectConnectionResolver extends AbstractConnectionResolver {
 	 * @return array
 	 */
 	public function get_items() {
-		return ! empty( $this->query->posts ) ? $this->query->posts : [];
+
+		$sticky = [];
+
+		/**
+		 * Filter whether GraphQL should include sticky posts at the top of the query.
+		 *
+		 * @param boolean                      $include Whether to include sticky posts. False by default.
+		 * @param PostObjectConnectionResolver $this    Instance of the class to provide context for filters to decide when to allow/disallow sticky post support
+		 */
+		$include_sticky_posts = apply_filters( 'graphql_include_sticky_posts', false, $this );
+		if ( true === $include_sticky_posts && 0 == $this->get_offset() && ! isset( $this->args['last'] ) ) {
+			$sticky = get_option( 'sticky_posts' );
+		}
+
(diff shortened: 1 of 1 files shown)


epeli:
Wouldn't this be more flexible as a query arg?

jasonbahl:
Is it already morning there? 😱

jasonbahl:
@epeli maybe both. 🤔

My thought was that as a maintainer of a server you may not want to support exposing sticky posts. 

But on the flip side, as a consumer you may want to be able to ask for them 🤔


I’m open to feedback for sure. 

My though still is that it feels a bit more like the domain of the server to decide rather than the consumer. 

🤔

epeli:
Consider this scenario:

I want to list all posts whether they are sticky or not by the publish date and create a second list of only sticky posts to a sidebar. If server makes this decision that would be impossible with a single graphql query.

nwoeddie:
@jasonbahl Where would I add the `graphql_include_sticky_posts` filter?

AlbertoAbruzzo:
Any news about that? I can't query for sticky posts...

PR #864, opened 2019-05-23
Fix orderby with post object cursors

### LINKED ISSUE #818: Cursor pagination not working with custom ordering

### LINKED ISSUE #861: Pagination not working for pageBy and taxQuery queries
So I have the following query:

```
query DivisionQuery($first: Int, $after: String) {
  pageBy(uri: "what-we-do/visual-effects/") {
    title
    childPages(first: $first, after: $after, where: {orderby: [{field: MENU_ORDER, order: ASC}]}) {
      pageInfo {
        hasNextPage
        endCursor
      }
      nodes {
        title
      }
    }
  }
}
// Variables
{
  "first": 1,
  "after": null
}
```

Which gives me this result:
```
{
  "data": {
    "pageBy": {
      "title": "Visual Effects",
      "childPages": {
        "pageInfo": {
          "hasNextPage": true,
          "endCursor": "YXJyYXljb25uZWN0aW9uOjU4OQ=="
        },
        "nodes": [
          {
            "title": "Adidas &#8211; Creators"
          }
        ]
      }
    }
  }
}
```

So I do the next query, with `after` set to the `endCursor`:
```
// Variables
{
  "first": 1,
  "after": "YXJyYXljb25uZWN0aW9uOjU4OQ=="
}
```

But the result is:
```
{
  "data": {
    "pageBy": {
      "title": "Visual Effects",
      "childPages": {
        "pageInfo": {
          "hasNextPage": false,
          "endCursor": null
        },
        "nodes": []
      }
    }
  }
}
```

It should actually give me the next page. If I run the first query with `first: 2` and `after: null` I get the right pages, with a result like this:
```
{
  "data": {
    "pageBy": {
      "title": "Visual Effects",
      "childPages": {
        "pageInfo": {
          "hasNextPage": true,
          "endCursor": "YXJyYXljb25uZWN0aW9uOjE0Mjg="
        },
        "nodes": [
          {
            "title": "Adidas &#8211; Creators"
          },
          {
            "title": "Superbowl 2018"
          }
        ]
      }
    }
  }
}
```

I haven't been able to get pagination to work with `taxQuery` plugin either, so I wonder if I'm doing this right at all?

epeli:
The docker test environment is broken for me so I'm sending this as a PR now to see how the tests execute.

I have following posts

image

and I execute this query for the second page

```graphql
{
  posts(first: 5, after: "YXJyYXljb25uZWN0aW9uOjI4", where: {orderby: {field: TITLE, order: ASC}}) {
    pageInfo {
      endCursor
    }
    edges {
      node {
        title
        content
      }
    }
  }
}
```

Before this PR it returns this

image

which is wrong. The generated SQL query is 

```sql
SELECT wp_posts.ID
FROM wp_posts
WHERE 1 = 1
	AND wp_posts.post_parent = 0
	AND wp_posts.post_type = 'post'
	AND ((wp_posts.post_status = 'publish'))
	AND CAST(wp_posts.post_date AS DATETIME) <= CAST('2019-05-23 06:04:47' AS DATETIME)
	AND (
		CAST(wp_posts.post_date AS DATETIME) < CAST('2019-05-23 06:04:47' AS DATETIME)
		OR (wp_posts.ID < 28)
		)
ORDER BY wp_posts.post_title ASC
	,wp_posts.ID DESC LIMIT 0
	,6
```

Which is clearly wrong too. No `post_title`!

After this PR is applied the second page matches what is seen in the wp-admin and the SQL query is

```sql
SELECT wp_posts.ID
FROM wp_posts
WHERE 1 = 1
	AND wp_posts.post_parent = 0
	AND wp_posts.post_type = 'post'
	AND ((wp_posts.post_status = 'publish'))
	AND wp_posts.post_title >= 'foo'
	AND (
		wp_posts.post_title > 'foo'
		OR (wp_posts.ID < 28)
		)
ORDER BY wp_posts.post_title ASC
	,wp_posts.ID DESC LIMIT 0
	,6
```

and the result

image

Closes #818
Closes #861
Closes #862

### DIFF
--- a/src/Data/Cursor/PostObjectCursor.php
+++ b/src/Data/Cursor/PostObjectCursor.php
@@ -142,6 +142,8 @@ class PostObjectCursor {
 			$this->compare_with_date();
 		}
 
+		$this->builder->add_field( "{$this->wpdb->posts}.ID", $this->cursor_offset, 'ID' );
+
 		return $this->to_sql();
 	}
 
@@ -150,7 +152,6 @@ class PostObjectCursor {
 	 */
 	private function compare_with_date() {
 		$this->builder->add_field( "{$this->wpdb->posts}.post_date", $this->get_cursor_post()->post_date, 'DATETIME' );
-		$this->builder->add_field( "{$this->wpdb->posts}.ID", $this->cursor_offset, 'ID' );
 	}
 
 	/**
@@ -163,14 +164,13 @@ class PostObjectCursor {
 	 */
 	private function compare_with( $by, $order ) {
 
-		$post_field = 'post_' . $by;
-		$value      = $this->get_cursor_post()->{$post_field};
+		$value      = $this->get_cursor_post()->{$by};
 
 		/**
 		 * Compare by the post field if the key matches an value
 		 */
 		if ( ! empty( $value ) ) {
-			$this->builder->add_field( "{$this->wpdb->posts}.post_{$by}", $value, null, $order );
+			$this->builder->add_field( "{$this->wpdb->posts}.{$by}", $value, null, $order );
 
 			return;
 		}
--- a/tests/wpunit/PostObjectCursorTest.php
+++ b/tests/wpunit/PostObjectCursorTest.php
@@ -226,6 +226,21 @@ class PostObjectCursorTest extends \Codeception\TestCase\WPTestCase {
 		] );
 	}
 
+	public function testPostOrderingByDuplicatePostTitles() {
+		foreach ($this->created_post_ids as $index => $post_id) {
+			wp_update_post( [
+				'ID' => $post_id,
+				'post_title' => 'duptitle',
+
+			] );
+		}
+
+		$this->assertQueryInCursor( [
+			'orderby' => 'post_title',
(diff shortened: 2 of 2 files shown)


epeli:
Ah. The test enviroment is broken on Travis too

image

jasonbahl:
@epeli we can address the edge case separate, but I believe if you have 2 items with the same title and you use the cursor of one of them and do an after query, it skips the other. 

Like do the same query, but use the cursor of one of the `foo` and do an after query with it, and I think it will skip some of the other `foo` posts

epeli:
And it's green.

We had some bad tests and because the orderby feature was written using TDD the implementatio also came out as bad. This PR now fixes the tests and the implementation.

Sorry for the inconvenience everybody.

jasonbahl:
> Sorry for the inconvenience everybody.

No need to apologize. This is a HUGE feature for this plugin. Super cool to have it working well! 

Thanks so much!!!!

PR #877, opened 2019-06-11
added a fix for error given when encounter permission to edit others …

joseffb-mla:
2nd try of PR to add a absint fix to the userid check.


src/Mutation/PostObjectCreate.php
```diff
@@ -212,10 +213,16 @@ public static function mutate_and_get_payload( $post_type_object, $mutation_name
              * If the post being created is being assigned to another user that's not the current user, make sure
              * the current user has permission to edit others posts for this post_type
              */
-            if ( ! empty( $input['authorId'] ) && get_current_user_id() !== $input['authorId'] && ! current_user_can( $post_type_object->cap->edit_others_posts ) ) {
-                // translators: the $post_type_object->graphql_plural_name placeholder is the name of the object being mutated
-                throw new UserError( sprintf( __( 'Sorry, you are not allowed to create %1$s as this user', 'wp-graphql' ), $post_type_object->graphql_plural_name ) );
-            }
+	        $authorID = absint($input['authorId']);
```
kidunot89: This breaks everything. If the `authorId` is provided as a *Relay ID* it will be a string.

src/Mutation/PostObjectCreate.php
```diff
@@ -212,10 +213,16 @@ public static function mutate_and_get_payload( $post_type_object, $mutation_name
...
+	        $id_components = Relay::fromGlobalId ( $input['authorId'] );
+	        if (is_array( $id_components ) && ! empty( $id_components['id'])) {
+		        $authorID = $id_components['id'];
+	        }
+
```
kidunot89: This retrieves the *WP ID* for the author using a *Relay ID*, and should only be executed if the `authorId` is a *Relay ID*

joseffb-mla:
Got it. This is my first time with Relay IDs, so Ill get that fixed up next
week when I’m back from vacation. Thanks!

jasonbahl:
@joseffb-mla try `composer install --no-dev`. It looks like this is adding composer dev dependencies

Issue #178, opened 2017-07-26
Filter Plugins by Status


jasonbahl:
It would be nice to be able to filter plugins by status, like the wp-admin plugins.php does. 

Stati that can be filtered: `active`, `recently active`, `inactive`, `must-use`, `drop-ins`

CodeProKid:
Moving this into 1.0, since the query already exists in the plugin. Would be good to add this functionality before the 1.0 release. I also think it would be a good first issue for an outside contributor to pickup. We can always punt it later if no one picks it up though.

jasonbahl:
closed by #2298

PR #895, opened 2019-07-05
filter plugins by status

let00:

Allows plugins to be filtered by their status (active, recently active, inactive, must-use, drop-in and upgrade)

#178 

N/A

Any other comments?
-------------------
I am a first time contributor at WPGraphQL. Would really appreciate any feedback.

Where has this been tested?
---------------------------
**Operating System:** Docker, MacOS Mojave

**WordPress Version:** 5.2.2

### DIFF
--- a/src/Connection/Plugins.php
+++ b/src/Connection/Plugins.php
@@ -19,13 +19,39 @@ class Plugins {
 	 * @access public
 	 */
 	public static function register_connections() {
+
+		$connection_args = self::get_connection_args();
+
 		register_graphql_connection( [
 			'fromType'      => 'RootQuery',
 			'toType'        => 'Plugin',
 			'fromFieldName' => 'plugins',
+			'connectionArgs'   => $connection_args,
 			'resolve'       => function ( $root, $args, $context, $info ) {
 				return DataSource::resolve_plugins_connection( $root, $args, $context, $info );
 			},
 		] );
+	}
+
+	/**
+	 * Given an optional array of args, this returns the args to be used in the connection
+	 *
+	 * @access public
+	 * @param array $args The args to modify the defaults
+	 *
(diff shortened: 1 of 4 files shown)


jasonbahl:
@let00 thanks for the PR! I'll review it shortly and provide whatever feedback necessary. 😄

src/Connection/Plugins.php
```diff
@@ -19,13 +19,39 @@ class Plugins {
...
+	 */
+	public static function get_connection_args( $args = [] ) {
+
+		return array_merge( [
+
+			/**
+			 * Status parameters
+			 */
+			'status'       => [
+				'type' => 'PluginStatusEnum',
```
jasonbahl: @let00 let's add a description as well for self documenting Schema 😄

jasonbahl:
@let00 I wonder if `status` should be `active` and `inactive` and we have another argument for `pluginKind` or similar, which is used to differentiate `dropin` and `must-use`?

🤔

let00:
@jasonbahl That seems like a good idea. So something like this?
```
status: [active, recently active, inactive]
pluginKind: [must-use, dropin]
upgrade: [true, false]
```
Does that sound good?

jasonbahl:
Hmm. This is interesting though, because Core WordPress actually does call all of these things "status" in the admin on the plugins page 🤔 

Need to think about this a bit more. 

Sometimes I think it's wise to depart from how core has done things in the past, but sometimes it's good to maintain consistency. 

🤔

jasonbahl:
might be good to prototype a UI for how this _could_ work if we had different arguments than _just_ `status` to see what makes sense 🤔

jasonbahl:
Closing in favor of #2298

Issue #897, opened 2019-07-08
Cursor pagination does not preserve order with the "search" where arg


let00:
## First query for posts 1-11

I am making a query with the search arg which returns 13 results in total. The query below fetches the first 11 results.

```
query MyQuery {
  wpgraphql {
    posts(first: 11, after: null, where: {search: "deserunt"}) {
      edges {
        cursor
      }
      nodes {
        title
      }
      pageInfo {
        startCursor
        endCursor
      }
    }
  }
}
```
Here is the result for this query.

```
{
  "data": {
    "wpgraphql": {
      "posts": {
        "edges": [
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjM0OA=="
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjM0Ng=="
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjM1NA=="
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjM1Mg=="
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjM0MA=="
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjM1MA=="
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjMzOA=="
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjM0Mg=="
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjM0NA=="
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjMzNA=="
          },
          {
            "cursor": "YXJyYXljb25uZWN0aW9uOjMzMA=="
          }
        ],
        "nodes": [
          {
            "title": "Fugiat deserunt ipsum voluptatem"
          },
          {
            "title": "Ut sit et eligendi rerum deserunt"
          },
          {
            "title": "Deserunt ut omnis repellat"
          },
          {
            "title": "Porro eligendi minima laudantium"
          },
          {
            "title": "Dolores dolor animi sunt"
          },
          {
            "title": "Eveniet qui esse iure magni consequatur voluptatum"
          },
          {
            "title": "Eos assumenda qui qui provident"
          },
          {
            "title": "Qui perferendis omnis ab sint"
          },
          {
            "title": "Occaecati quasi ut dolorem et veritatis doloremque"
          },
          {
            "title": "Quia totam et id sint odio"
[...]

PR #898, opened 2019-07-09
#897 - Cursor pagination does not preserve order with the "search" where arg

jasonbahl:
WP_Query applies default ordering by title when the "s" arg is passed to it, which was causing funky issues with WPGraphQL Cursor Pagination.

This sets the `search_orderby_title` to false, and defaults to orderby date. Custom orderby options can be passed to the connection as well.

#897 

To test, I created 4 posts with the word "floof" as the content. 

In the admin, you can see a search for "floof" shows posts: `Four, Three, Two, One"

Screen Shot 2019-07-09 at 2 13 50 PM

### Query the last 2

Here, I query for the `last: 2` items, and as expected, we get Posts `Two` and `One`.

#### Query
```graphql
query GET_POSTS($first: Int, $last: Int, $after: String, $before: String, $where: RootQueryToPostConnectionWhereArgs) {
  posts(last: $last, before: $before, first: $first, after: $after, where: $where) {
    pageInfo {
      hasPreviousPage
      hasNextPage
      startCursor
      endCursor
    }
    edges {
      cursor
      node {
        id
        date
        title
      }
    }
  }
}
```

#### Variables
```json
{
  "last": 2,
  "before": "YXJyYXljb25uZWN0aW9uOjEzOTc=",
  "where": {
    "search": "floof"
  }
}
```

#### Screenshot of results
Screen Shot 2019-07-09 at 2 15 29 PM

### Query with before cursor

Now, if we execute the same query, but use the `startCursor` from the previous query results as the value for the `before` input we get items `Four` and `Three`:

#### Variables
```json
{
  "last": 2,
  "before": "YXJyYXljb25uZWN0aW9uOjEzOTc=",
  "where": {
    "search": "floof"
  }
}
```

#### Screenshot of paginated query
Screen Shot 2019-07-09 at 2 17 54 PM

Any other comments
---------------------------
Tests are included for forward and backward pagination using the `{where: { search: ""} }` argument

Where has this been tested?
---------------------------
**Operating System:** Mac OSX 10.14.5
**WordPress Version:** WordPress v5.2.2

### DIFF
--- a/src/Data/Cursor/PostObjectCursor.php
+++ b/src/Data/Cursor/PostObjectCursor.php
@@ -121,20 +121,22 @@ class PostObjectCursor {
 		$orderby = $this->get_query_var( 'orderby' );
 		$order   = $this->get_query_var( 'order' );
 
-
 		if ( ! empty( $orderby ) && is_array( $orderby ) ) {
+
 			/**
 			 * Loop through all order keys if it is an array
 			 */
 			foreach ( $orderby as $by => $order ) {
 				$this->compare_with( $by, $order );
 			}
+
 		} else if ( ! empty( $orderby ) && is_string( $orderby ) ) {
 
 			/**
 			 * If $orderby is just a string just compare with it directly as DESC
 			 */
 			$this->compare_with( $orderby, $order );
+
 		}
 
 		/**
@@ -166,7 +168,7 @@ class PostObjectCursor {
 	 */
 	private function compare_with( $by, $order ) {
 
-		switch( $by ) {
+		switch ( $by ) {
 			case 'author':
 			case 'title':
 			case 'type':
@@ -178,7 +180,7 @@ class PostObjectCursor {
 				break;
 		}
 
-		$value      = $this->get_cursor_post()->{$by};
+		$value = $this->get_cursor_post()->{$by};
 
 		/**
 		 * Compare by the post field if the key matches an value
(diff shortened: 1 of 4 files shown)


src/Data/Cursor/PostObjectCursor.php
```diff
@@ -121,6 +121,7 @@ public function get_where() {
 		$orderby = $this->get_query_var( 'orderby' );
 		$order   = $this->get_query_var( 'order' );
 
+
```
hughdevore: 👀

src/Data/Cursor/PostObjectCursor.php
```diff
@@ -129,6 +130,7 @@ public function get_where() {
 				$this->compare_with( $by, $order );
 			}
 		} else if ( ! empty( $orderby ) && is_string( $orderby ) ) {
+
```
hughdevore: 👀

src/Data/Cursor/PostObjectCursor.php
hughdevore: 👀

tests/wpunit/PostObjectSearchTest.php
```diff
@@ -0,0 +1,221 @@
...
+		]);
+
+		$this->created_post_ids = $this->create_posts();
+
+		$this->app_context = new \WPGraphQL\AppContext();
+
+		$this->app_info = new \GraphQL\Type\Definition\ResolveInfo( array() );
+
+		$this->query = '
+		query GET_POSTS($first: Int, $last: Int, $after: String, $before: String $where:RootQueryToPostConnectionWhereArgs) {
```
let00: @jasonbahl  seems like a comma is missing before the where arg

jasonbahl: good call. Commas in variable declarations are optional, so not a big deal, but probably should adjust for consistency 😄

PR #909, opened 2019-07-16
Fix docker test env

epeli:
mysql-client has been renamed to mariadb-client

Fix the `run-docker-test-environment.sh` script error with

```
Package mysql-client is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
```


jasonbahl:
Thanks!!!

PR #910, opened 2019-07-16
Create comment perm fix 2ndtry pr

joseffb-mla:
Fixed the permissions checking to allow a post create if the user has capability to edit/update for others.


tests/acceptance.suite.dist.yml
```diff
@@ -15,15 +15,15 @@ modules:
         WPDb:
             dsn: 'mysql:host=127.0.0.1;dbname=wpgraphql_test'
             user: 'root'
-            password: ''
+            password: 'mysql'
```
jasonbahl: instead of modifying this file, you can copy it to `acceptance.suite.yml` and it will be `.gitignored` so you can use your local creds without versioning them

composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "08fdc196dedea64b0e5f907e24e52446",
+    "content-hash": "4105e60ad0aa71a672c8dacd01b6d30c",
```
jasonbahl: try running `composer install --no-dev` before committing. It will prevent the dev dependencies from being included.

src/Mutation/PostObjectCreate.php
```diff
@@ -5,345 +5,358 @@
 use GraphQL\Deferred;
 use GraphQL\Error\UserError;
 use GraphQL\Type\Definition\ResolveInfo;
+use GraphQLRelay\Relay;
 use WPGraphQL\AppContext;
 use WPGraphQL\Data\DataSource;
 use WPGraphQL\Data\PostObjectMutation;
 
 class PostObjectCreate {
-    /**
```
jasonbahl: This is hard to see what changed as it's showing the entire file was changed instead of just a few lines. My guess is because of your IDE settings formatting it differently?

tests/wpunit.suite.dist.yml
```diff
@@ -1,23 +0,0 @@
-# Codeception Test Suite Configuration
```
jasonbahl: this should not be getting deleted

tests/wpunit/PostObjectMutationsTest.php
```diff
@@ -501,6 +501,92 @@ public function testCreatePostObjectWithoutProperCapabilities() {
...
+		 * can test the mutation and make sure they can create a post
+		 * since they should have have proper 'other user' permissions now
+		 */
+		wp_set_current_user( $this->subscriber );
+
+		/**
+		 * Prepare mutation for GQL request
+		 */
+		$mutation = 'mutation UPDATE_POST( $input:UpdatePostInput! ) {
+          updatePost(input: $input) {
```
jasonbahl: formatting is funky here

src/Type/Object/RootQuery.php
```diff
@@ -226,6 +226,7 @@
 					$post_id     = isset( $post_object->ID ) ? absint( $post_object->ID ) : null;
 				}
 				$post = DataSource::resolve_post_object( $post_id, $context );
+
```
jasonbahl: probably don't need to commit this change?

PR #946, opened 2019-08-22
Allow filtering of User connection query args

### LINKED ISSUE #945: Allow filtering of query args for User connections
Currently, there is the ability to filter the query args for both Posts and Terms using the following filters:

`graphql_term_object_connection_query_args`
`graphql_post_object_connection_query_args`

These are very useful, especially when trying to pull the total number of items a query can return as per the suggestions in #723. 

I'm currently trying to pull the total number of users returned by a query, to do this I need to set `$query_args['count_total']` to `true`. Unfortunately, there's no way to do this as it stands as we can only filter the query args for Posts and Terms.

~~I suggest adding a filter to all connections so that users can adjust those queries as they need.~~

I've looked at the code and it seems that only the User type is missing a filter to do this!

I will open a PR for this shortly 🙂

darylldoyle:
This PR adds the ability to filter query args for the User connection. E.G. this is useful if someone wants to set `count_total` to `true` within their user queries.

This will close #945 

Where has this been tested?
---------------------------
**Operating System:** macOS

**WordPress Version:** 5.2.2

Have also ran the test suite locally which passed.

### DIFF
--- a/src/Data/Connection/UserConnectionResolver.php
+++ b/src/Data/Connection/UserConnectionResolver.php
@@ -77,6 +77,20 @@ class UserConnectionResolver extends AbstractConnectionResolver {
 			$query_args['has_published_posts'] = true;
 		}
 
+		/**
+		 * Filter the query_args that should be applied to the query. This filter is applied AFTER the input args from
+		 * the GraphQL Query have been applied and has the potential to override the GraphQL Query Input Args.
+		 *
+		 * @param array       $query_args array of query_args being passed to the
+		 * @param mixed       $source     source passed down from the resolve tree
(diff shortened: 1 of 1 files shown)


jasonbahl:
@darylldoyle hey! There's already a central filter for all connection resolvers here: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Connection/AbstractConnectionResolver.php#L152

Pretty sure this existing filter will provide what you need.

darylldoyle:
@jasonbahl you are of course, 100% correct! I spent far too long missing that filter 🤦‍♂ 

Thanks for your help and the good work 🙂 

I'll close this and the PR now.

PR #948, opened 2019-08-23
Feature/#846 allow schema to be cleared

jasonbahl:
This turned out to be a MASSIVE PITA, but I think it will greatly help with the future of testing the plugin and the plugins extensions.

I will elaborate more Monday when I'm back at the computer, but wanted to get this PR out for folks like @kidunot89 to check out.

I will document the upgrade path, etc on Monday.

For now there are a few tests broken re: Settings, which I will also address on Monday.


jasonbahl:
Closing as this has been merged into the v0.4.0 release branch

PR #950, opened 2019-08-28
sourceURL returns file attachment url if file is not an image

### LINKED ISSUE #947: `sourceUrl` is false for non-image file uploads
The `sourceUrl` for PDF uploads always returns `"false"` (mimeType and other info is still in tact however) 

Image mediaItems are fine. 

Is this the right way to fetch non-image upload paths? Thanks!

Screen Shot 2019-08-22 at 10 08 23 pm

jydmnd:

- [ X] Make sure you are making a pull request against the **develop branch** (left side). Also you should start *your branch* off *our develop*.
- [X ] Make sure you are requesting to pull request from a **topic/feature/bugfix branch** (right side). Don't pull request from your master!

--------------------------------------------------
The sourceURL property would return false if an uploaded file was not an image, since it was using wp_get_attachment_image_src() on non-image files.

fixes #947

Where has this been tested?
---------------------------
**Operating System:** Mac OS Mojave, CloudLinux

**WordPress Version:** 5.2.2


jasonbahl:
Oooh! Thanks for getting to this. 

Any chance you'd like to contribute a test for this as well? I'd be happy to pair program with you to write the test if you're unsure about how to contribute in that way. 

Let me know. 

Thanks!

jydmnd:
Sure! I'd love to, but I would definitely need your help.

jydmnd:
@jasonbahl I tried installing the test suite locally, but I'm getting this error when running wpunit : 
PHP Fatal error:  Declaration of AccessFunctionsTest::setUp() must be compatible with tad\WPBrowser\Compat\Codeception\Unit::setUp(): void in [...]/wp-graphql/tests/wpunit/AccessFunctionsTest.php on line 3
I'm running PHP 7.2 on Mac OS Mojave and ran the setup script with _latest_ as the wp version, if that helps. I haven't tried with Docker yet.

It works if I add a :void at the end of the appropriate function declarations, but I'm assuming I'm not supposed to do this for every test! Maybe something from the test suite got a breaking update? Composer installed WP Browser with version 2.2.20.

As for the test itself, the bug happens when a post has a post_type of 'attachment'. I don't think there's a actual test for this specifically?

jasonbahl:
@jydmnd if you’re available tomorrow afternoon (MST) we could do a pairing session and get the tests running and get some written for this. 

Or we can schedule something next week. 

What time zone are you in?

jydmnd:
@jasonbahl I'm available tomorrow from 1PM to 5PM EST.

jasonbahl:
@jydmnd are you in the WPGraphQL Slack? If not you can join here: https://wpgql-slack.herokuapp.com/

If you can DM me in there we can coordinate a Zoom call to pair program. 

Thanks!

src/Model/Post.php
```diff
@@ -408,10 +408,10 @@ protected function init() {
 					'mediaType'           => function() {
 						return wp_attachment_is_image( $this->data->ID ) ? 'image' : 'file';
 					},
+                    'mediaItemUrl' => function() {
```
jasonbahl: spacing looks off here

src/Type/Object/PostObject.php
```diff
@@ -164,13 +164,17 @@ function register_post_object_types( $post_type_object ) {
 					'type'        => 'String',
 					'description' => __( 'Type of resource', 'wp-graphql' ),
 				],
+                'mediaItemUrl'    => [
+                    'type'        => 'String',
+                    'description' => __( 'Url of the mediaItem', 'wp-graphql' ),
+                ],
```
jasonbahl: spacing looks off here

tests/wpunit/MediaItemQueriesTest.php
```diff
@@ -336,4 +336,37 @@ public function testMediaItemQuery( $image_meta = [] ) {
...
 	}
 
+    /**
+     * testPostQuery
+     *
+     * This tests creates a single attachment and retrieves said post URL via a GraphQL query
+     *
+     * @since 0.3.6
+     */
+    public function testMediaItemUrl() {
```
jasonbahl: This is good! 

We should include another test to test something of another mime type, like a video or mp3 or something. 

I think the common complaint has been that `sourceUrl` wasn't working for non-images

jasonbahl:
@jydmnd there was a few formatting issues, so I just pushed up some quick changes. I think this is looking pretty good. I want to verify a few things locally though before merging.

jasonbahl:
Ok, so I wonder if it makes sense to deprecate the `sourceUrl` field here. Let it continue resolving, but mark it deprecated so that it no longer shows in GraphiQL Docs. 🤔

jydmnd:
@jasonbahl I don't think it should be deprecated, since sourceUrl can still be used for pictures when providing a size. It returns a different URL depending on the size, which is fine.

jasonbahl:
@jydmnd ah ya. 🤔 

I think this feature will need to be revisited with a more critical eye again in the future, but for now I think the value of this is high enough for users to merge as-is. 

I think down the road we'll need to formally address a MediaItem interface Type and have separate types for each Type of Media Item that support different shapes. 

Will need to think about it a bit more and probably work on an RFC.

jasonbahl:
closes #947

Issue #22, opened 2016-12-20
Set up wpgraphql.com


jasonbahl:

jasonbahl:
Check out https://www.wpgraphql.com

PR #954, opened 2019-09-07
Adds support for custom CPT schema definitions.

kidunot89:

This implements a few changes related to opting-out of the Custom Post-Type schema design defined in `src/Type/Object/PostObject.php`  while still using some of the connected types like `MenuItemObjectUnion`.

#### Example Usage
This are the two filters being used in WooGraphQL to bypass the default post-type definition in `TypeRegistry`.

Register Product type.
```
/**
 * Registers WooCommerce post-types to be used in GraphQL schema
 *
 * @param array  $args      - allowed post-types.
 * @param string $post_type - name of taxonomy being checked.
 *
 * @return array
 */
public static function register_post_types( $args, $post_type ) {
		$args['show_in_graphql']                = true;
		$args['graphql_single_name']         = 'Product';
		$args['graphql_plural_name']          = 'Products';
		$args['skip_graphql_type_registry'] = true;
	}
}
add_filter( 'register_post_type_args', 'register_post_types' , 10, 2 );
```

Opt-out of `WPGraphQL::get_allowed_post_types()`
```
/**
 * Filters "allowed_post_types" list if schema hasn't been defined.
 *
 * @param array $post_types  Post types registered in GraphQL schema.
 *
 * @return array
 */
public static function skip_type_registry( $post_types ) {
	return array_diff(
		$post_types,
		get_post_types(
			[
				'show_in_graphql'            => true,
				'skip_graphql_type_registry' => true,
			]
		)
	);
}
add_filter( 'graphql_post_entities_allowed_post_types', 'skip_type_registry', 10 );
```

In the example above the Product type is registered to the GraphQL schema but it's not defined by the with the rest of the CPTs in the [TypeRegistry](https://github.com/wp-graphql/wp-graphql/blob/develop/src/TypeRegistry.php#L159). However, the Product post-type can still be retrieve by direct calls to `get_post_types()` like in [MenuItemObjectUnion](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/Union/MenuItemObjectUnion.php#L14).

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 18.04

**WordPress Version:** 5.2

### DIFF
--- a/src/Type/Union/MenuItemObjectUnion.php
+++ b/src/Type/Union/MenuItemObjectUnion.php
@@ -52,7 +52,7 @@ register_graphql_union_type(
 				return Types::term_object( $object->taxonomyName );
 			}
 
-			return null;
+			return apply_filters( 'resolve_menu_item_type', null, $object );
 		},
 	]
 );
--- a/src/Type/Union/MenuItemObjectUnion.php
+++ b/src/Type/Union/MenuItemObjectUnion.php
@@ -52,7 +52,7 @@ register_graphql_union_type(
 				return Types::term_object( $object->taxonomyName );
 			}
 
-			return apply_filters( 'resolve_menu_item_type', null, $object );
+			return $object;
 		},
 	]
 );
--- a/src/Type/Union/PostObjectUnion.php
+++ b/src/Type/Union/PostObjectUnion.php
@@ -5,7 +5,7 @@ namespace WPGraphQL\Type;
 use WPGraphQL\Types;
 
 $possible_types     = [];
-$allowed_post_types = \WPGraphQL::get_allowed_post_types();
+$allowed_post_types = get_post_types( [ 'show_in_graphql' => true ] );
 if ( ! empty( $allowed_post_types ) && is_array( $allowed_post_types ) ) {
 	foreach ( $allowed_post_types as $allowed_post_type ) {
 		if ( empty( $possible_types[ $allowed_post_type ] ) ) {
--- a/src/TypeRegistry.php
+++ b/src/TypeRegistry.php
@@ -198,15 +198,6 @@ class TypeRegistry {
 		}
 
 		/**
-		 * Register all Union Types
-		 * Unions need to be registered after other types as they reference other Types
-		 */
-		require_once WPGRAPHQL_PLUGIN_DIR . 'src/Type/Union/CommentAuthorUnion.php';
-		require_once WPGRAPHQL_PLUGIN_DIR . 'src/Type/Union/MenuItemObjectUnion.php';
-		require_once WPGRAPHQL_PLUGIN_DIR . 'src/Type/Union/PostObjectUnion.php';
-		require_once WPGRAPHQL_PLUGIN_DIR . 'src/Type/Union/TermObjectUnion.php';
-
-		/**
 		 * Register core connections
 		 */
 		Comments::register_connections();
@@ -243,6 +234,15 @@ class TypeRegistry {
 		if ( ! did_action( 'graphql_register_types' ) ) {
 			do_action( 'graphql_register_types' );
 		}
+
+		/**
+		 * Register all Union Types
+		 * Unions need to be registered after other types as they reference other Types
+		 */
+		require_once WPGRAPHQL_PLUGIN_DIR . 'src/Type/Union/CommentAuthorUnion.php';
+		require_once WPGRAPHQL_PLUGIN_DIR . 'src/Type/Union/MenuItemObjectUnion.php';
+		require_once WPGRAPHQL_PLUGIN_DIR . 'src/Type/Union/PostObjectUnion.php';
+		require_once WPGRAPHQL_PLUGIN_DIR . 'src/Type/Union/TermObjectUnion.php';
 
 	}
 
--- a/src/Type/WPUnionType.php
+++ b/src/Type/WPUnionType.php
@@ -18,9 +18,11 @@ class WPUnionType extends UnionType {
 	/**
 	 * WPUnionType constructor.
 	 *
+	 * @param array $config The Config to setup a Union Type
+	 *
 	 * @since 0.0.30
 	 */
-	public function __construct( $config ) {
+	public function __construct( $config = [] ) {
 		/**
 		 * Set the Types to start with capitals
 		 */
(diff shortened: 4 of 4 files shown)


src/Type/Union/MenuItemObjectUnion.php
```diff
@@ -52,7 +52,7 @@
 				return Types::term_object( $object->taxonomyName );
 			}
 
-			return null;
+			return apply_filters( 'resolve_menu_item_type', null, $object );
```
jasonbahl: 🤔 This seems like something we need to resolve centrally. This _should_ be a filter applied to _all_ Union Types. We should consider applying this filter centrally within WPUnionType class, instead of adding a filter for each Union. 

Can you also elaborate on what this change does and why it's necessary?

src/TypeRegistry.php
```diff
@@ -244,6 +235,15 @@ public static function init() {
...
 
+		/**
+		 * Register all Union Types
+		 * Unions need to be registered after other types as they reference other Types
+		 */
+		require_once WPGRAPHQL_PLUGIN_DIR . 'src/Type/Union/CommentAuthorUnion.php';
+		require_once WPGRAPHQL_PLUGIN_DIR . 'src/Type/Union/MenuItemObjectUnion.php';
+		require_once WPGRAPHQL_PLUGIN_DIR . 'src/Type/Union/PostObjectUnion.php';
+		require_once WPGRAPHQL_PLUGIN_DIR . 'src/Type/Union/TermObjectUnion.php';
+
```
jasonbahl: So, realistically, Types _should_ be able to be registered in any order. For example, what if you have a Union that returns a Connection Type? Or a Connection that returns a Union Type? 

Types should be able to be registered in any order. Their fields _should_ be built as a callback, once all Types have been registered, thus allowing for circular Types. 

I think we might have some places where we're not fully respecting the concept of circular Types, and that might be what needs to be addressed more than moving things lower in the Registry.

jasonbahl:
> This implements a few changes related to opting-out of the Custom Post-Type schema design defined in src/Type/Object/PostObject.php while still using some of the connected types like MenuItemObjectUnion.

@kidunot89 I think maybe you didn't commit all the files you thought you did based on this comment in the description?

src/TypeRegistry.php
kidunot89: I agree with that principle however the Introspection query does not. If the one of the types in the Union is defined during the `graphql_register_type` action. The introspection query resolve that type to `null`. This is the error I receive when the `Product` post-type found in the MenuItemObjectUnion using on line [#22](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/Union/MenuItemObjectUnion.php#L22)

```
Error: Introspection result missing possibleTypes: {"kind":"UNION","name":"MenuItemObjectUnion","description":null,"fields":null,"inputFields":null,"interfaces":null,"enumValues":null,"possibleTypes":null}
    at w (http://example.com/wordpress/wp-content/plugins/wp-graphiql/assets/app/build/static/js/main.b1825402.js?ver=5.2.3:1:586976)
    at b (http://example.com/wordpress/wp-content/plugins/wp-graphiql/assets/app/build/static/js/main.b1825402.js?ver=5.2.3:1:586263)
    at r (http://example.com/wordpress/wp-content/plugins/wp-graphiql/assets/app/build/static/js/main.b1825402.js?ver=5.2.3:1:585723)
    at http://example.com/wordpress/wp-content/plugins/wp-graphiql/assets/app/build/static/js/main.b1825402.js?ver=5.2.3:1:588833
    at Array.map (<anonymous>)
    at r (http://example.com/wordpress/wp-content/plugins/wp-graphiql/assets/app/build/static/js/main.b1825402.js?ver=5.2.3:1:588810)
    at http://example.com/wordpress/wp-content/plugins/wp-graphiql/assets/app/build/static/js/main.b1825402.js?ver=5.2.3:1:564667
```

kidunot89:
I added more detail to PR summary maybe that will shine more light on what I've done.

src/Type/Union/MenuItemObjectUnion.php
kidunot89: Naw, I would agree if all unions resolved the same way but they are each very different in how they operate, however `MenuItemObjectUnion` targets post-types using `get_post_types( [ 'show_in_graphql' => true, 'show_in_nav_menus' => true ] )`, and I think that's the right way to do it. In [PostObjectUnion](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/Union/PostObjectUnion.php#L8), I think is `get_post_types( [ 'show_in_graphql` => true ] )` should be used instead of `WPGraphQL::get_allowed_post_types()`, but that's the only change I would make.

jasonbahl: Well, where you placed this filter is within the resolver, which is decoupled from the use of `get_post_types()` in this file. 

I think if we add a filter to pass over all `resolveType` callbacks for `all` Unions, then any union could be hooked into, not _just_ this particular union.

src/TypeRegistry.php
jasonbahl: Ok, ya, we've got a different error to address here then. I think I may have addressed it in my upcoming PR that allows the Schema to be cleared.

src/Type/Union/MenuItemObjectUnion.php
jasonbahl: @kidunot89 ok, so I opened a PR to your fork here showing what I was thinking: https://github.com/kidunot89/wp-graphql/pull/2

This moves the filter centrally so _any_ Union in the entire Schema can benefit from this filter. 

Then, in your codebase, all you would need to do is change the name of the filter from `resolve_menu_item_type` to `graphql_union_resolve_type` and it should work the same, but with the added benefit that all Unions throughout the Schema are also filterable.

jasonbahl: @kidunot89 let me know what you think here.

src/Type/Union/PostObjectUnion.php
```diff
@@ -5,7 +5,7 @@
 use WPGraphQL\Types;
 
 $possible_types     = [];
-$allowed_post_types = \WPGraphQL::get_allowed_post_types();
+$allowed_post_types = get_post_types( [ 'show_in_graphql' => true ] );
```
jasonbahl: @kidunot89 wait, why is this changing? Don't we want your filter to be applied here too?

Like if a user were to do a query across many post_types, WordPress might return results of posts, pages and Products, and this union should know that the Product post_type should be included in the union, which your filter applied to WPGraphQL::get_allowed_post_types() would include. 

Can you elaborate on why this change is necessary?

I would argue that it's not

kidunot89: If `\WPGraphQL::get_allowed_post_types()` is the Product post-type won't be seen due the hook in the example in the description. If `get_post_types( [ 'show_in_graphql' => true ] )` is the hook is bypassed, and the `Product` post-type is included. Which is the desired behavior.

src/Type/Union/MenuItemObjectUnion.php
kidunot89: Merged

jasonbahl:
Closing. Including these changes manually in WPGraphQL v0.4.0 release

Issue #958, opened 2019-09-17
Provide public getters for properties of AbstractConnectionResolver class


jasonbahl:
The AbstractConnectionResolver provides central filters that apply to _all_ connections extending the class, making it possible for plugins / themes to filter various functionality of Connections in a central way. 

The instance of the Connection is passed through to the filters, but many of the properties of the class aren't publicly exposed, making it difficult for extending by checking the values of the properties.

Take this filter for example, where we want to check if the connection was passed specific arguments, and if it was, apply it the the query_args that will be used by WP_Query (or other query class):

```
add_filter( 'graphql_connection_query_args', function( $query_args, $connection ) {

  if ( isset( $connection->args['where']['someCustomInput'] ) ) {
    $query_args['something_custom'] = $connection->args['where']['someCustomInput'];
  }

  return $query_args;

} );
```

The `$connection->args` are not public, so this won't work. We will see a 

Exposing public getters for the properties that should be available will allow users to make use of these filters more appropriately. 

For example, this change (using ->getArgs()) will work properly if that's a public method: 

```
add_filter( 'graphql_connection_query_args', function( $query_args, $connection ) {

  if ( isset( $connection->getArgs()['where']['someCustomInput'] ) ) {
    $query_args['something_custom'] = $connection->getArgs()['where']['someCustomInput'];
  }

  return $query_args;

} );
```

jasonbahl:
closed by #959

jasonbahl:
Thanks @CodeProKid

PR #959, opened 2019-09-17
Feature/#958 add getters to abstract connection resolver

jasonbahl:
This adds public getters to the AbstractConnectionResolver class, allowing for external code that makes use of available filters to be able to access properties of the class when making decisions on how to filter underlying data.


jasonbahl:
closes #958

PR #962, opened 2019-09-20
#961 - Negative performance impact in the User model

### LINKED ISSUE #961: Negative performance impact in the User Model
When querying for Posts (or custom post types) or Users, we first check to see if the user is considered private or not. 

To check this, we check to see if the user has published posts. 

In WordPress, if a User has published a Post, they're considered a publicly accessible entity (not all fields of the User are considered public though).

The current approach for making this check is to use the core method `count_user_posts`

On small sites, this is a quick check. On large sites, with large datasets, this can be slow. 

To reduce the performance impact, we should provide an alternative method that queries for published posts by the post_author more efficiently. We can do this by first limiting the query to 1 item. We don't care how many posts the author has published, because a single published post makes them non-private. We can also cache that so that we don't have to query each time we check. WordPress.com VIP has a cached version of `count_user_posts` we can draw inspiration from: https://github.com/svn2github/wordpress-vip-plugins/blob/9c0a32757bdd78a0e2babf72a593c2c6a83287df/vip-do-not-include-on-wpcom/wpcom-caching.php#L403-L425

We don't want to rely on the VIP function itself, as more often than not, a site will not be running the VIP code. 

In addition to the more efficient, cached query, we should also provide a filter that allows for developers to bypass the check altogether and provide their own way of determining if a User should be considered public.

jasonbahl:
@chriszarate reported a performance issue on sites with larger data sets within the User model.

This provides a more efficient method of checking if a User has published posts, which makes them non-private entities. This PR provides a filter for plugins/themes to bypass the check altogether, and also makes the check much more efficient by caching the check and making the actual query more efficient than the currently used `count_user_posts` query.

----
closes #961

### DIFF
--- a/src/Model/User.php
+++ b/src/Model/User.php
@@ -78,21 +78,105 @@ class User extends Model {
 	 */
 	protected function is_private() {
 
+		/**
+		 * Filter whether the user is private.
+		 *
+		 * If true, the user will be considered private without further checks.
+		 *
+		 * @param mixed|null|boolean $is_private If set, return without executing the manual query
+		 *                                       to see if the user has published posts.
+		 * @param User               $this       Instance of the User Model
+		 */
+		$is_private = apply_filters( 'graphql_user_model_is_private', null, $this );
+
+		/**
+		 * If the filter returns true, we can return now as the user is considered private.
+		 *
+		 * Otherwise we need to continue and check if the User should be considered private or not
+		 */
+		if ( null !== $is_private ) {
+			return (bool) $is_private;
+		}
+
+		/**
+		 * If the requesting user does not have "list_users" capabilities and the current user is not
+		 * the user being requested, we need to check if the user being requested has published posts
+		 * which makes them a non-private entity.
+		 */
 		if ( ! current_user_can( 'list_users' ) && false === $this->owner_matches_current_user() ) {
 
 			/**
-			 * @todo: We should handle this check in a Deferred resolver. Right now it queries once per user
-			 *      but we _could_ query once for _all_ users.
-			 *
-			 *      For now, we only query if the current user doesn't have list_users, instead of querying
-			 *      for ALL users. Slightly more efficient for authenticated users at least.
+			 * Get allowed Post Types
 			 */
-			if ( ! count_user_posts( absint( $this->data->ID ), \WPGraphQL::get_allowed_post_types(), true ) ) {
-				return true;
+			$post_types = \WPGraphQL::get_allowed_post_types();
+			unset( $post_types['revision'] );
+			unset( $post_types['attachment'] );
+
+			/**
+			 * If the user has no published posts of any allowed post type,
+			 * they are considered to be private entities.
+			 */
+			if ( 0 === $this->count_user_posts( absint( $this->data->ID ), $post_types, true ) ) {
+				$is_private = true;
 			}
 		}
 
-		return false;
+		return $is_private;
+
+	}
+
+	/**
+	 * Cached version (and variation) of count_user_posts, which is uncached but doesn't always
+	 * need to hit the db
+	 *
+	 * count_user_posts is generally fast on smaller sites, but slows on large data sets.
+	 *
+	 * It can also be easy to end up with many redundant queries if it's called several times per
+	 * request. This allows bypassing the db queries in favor of the cache
+	 *
+	 * @param int @userid ID of the User
+	 * @param mixed|string|array         $post_type   The post type(s) to check for published posts in
+	 * @param boolean                    $public_only Whether to count only publicly published posts
+	 *
+	 * @return int|mixed|boolean
+	 */
+	protected function count_user_posts( $userid, $post_type = 'post', $public_only = false ) {
+
+		if ( ! is_numeric( $userid ) ) {
+			return 0;
+		}
+
+		$cache_key   = 'graphql_user_' . (int) $userid;
+		$cache_group = 'user_posts_count';
+
+		/**
+		 * If there's no cache, query to see if the author has 1 published post.
+		 *
+		 * This method of querying is more efficient on large datasets
+		 * then the core count_user_posts() method as this stops at the first match,
+		 * where the count_user_psts() method needs to find all matches and
+		 * it can take a while to calculate.
+		 *
+		 * We only care if there's at least 1 published post, not the total.
+		 */
+		if ( false === ( $count = wp_cache_get( $cache_key, $cache_group ) ) ) {
+			global $wpdb;
+
+			$where   = get_posts_by_author_sql( $post_type, true, $userid, $public_only );
+			$results = $wpdb->get_results( "SELECT id FROM $wpdb->posts $where LIMIT 1" );
+
+			$count = $results ? count( $results ) : 0;
+		}
+
+		/**
+		 * Filters the number of posts a user has written.
+		 *
+		 * @param int          $count       The user's post count.
+		 * @param int          $userid      User ID.
+		 * @param string|array $post_type   Single post type or array of post types to count the number of posts for.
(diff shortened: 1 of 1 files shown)


src/Model/User.php
```diff
@@ -78,21 +78,105 @@ public function __construct( \WP_User $user ) {
...
+		/**
+		 * Filter whether the user is private.
+		 *
+		 * If true, the user will be considered private without further checks.
+		 *
+		 * @param mixed|null|boolean $is_private If set, return without executing the manual query
+		 *                                       to see if the user has published posts.
+		 * @param User               $this       Instance of the User Model
+		 */
+		$is_private = apply_filters( 'graphql_user_model_is_private', null, $this );
```
CodeProKid: This is kind of already handled here -> https://github.com/wp-graphql/wp-graphql/blob/develop/src/Model/Model.php#L222 though the query will still run, so maybe we just need to rework that lower level filter, but I think we should keep all of the Model Layer hooks & filters at the abstract level so they are consistent across all types.

src/Model/User.php
```diff
@@ -78,21 +78,105 @@ public function __construct( \WP_User $user ) {
...
-			 * @todo: We should handle this check in a Deferred resolver. Right now it queries once per user
-			 *      but we _could_ query once for _all_ users.
-			 *
-			 *      For now, we only query if the current user doesn't have list_users, instead of querying
-			 *      for ALL users. Slightly more efficient for authenticated users at least.
+			 * Get allowed Post Types
+			 */
+			$post_types = \WPGraphQL::get_allowed_post_types();
+			unset( $post_types['revision'] );
+			unset( $post_types['attachment'] );
```
CodeProKid: I can understand revisions, but why unset attachments? I think it might be better to do an array intersect on all public post types maybe?

This gets pretty complicated actually, as we _technically_ shouldn't expose any authors that don't have published posts exposed in GraphQL. So if someone goes and locks down the `Pages` type, and a user only has posts in that post type, we technically shouldn't expose them. In a headless context, there would be no other way to enumerate that user. 🤔

src/Model/User.php
```diff
@@ -78,21 +78,105 @@ public function __construct( \WP_User $user ) {
...
+	 * need to hit the db
+	 *
+	 * count_user_posts is generally fast on smaller sites, but slows on large data sets.
+	 *
+	 * It can also be easy to end up with many redundant queries if it's called several times per
+	 * request. This allows bypassing the db queries in favor of the cache
+	 *
+	 * @param int @userid ID of the User
+	 * @param mixed|string|array         $post_type   The post type(s) to check for published posts in
+	 * @param boolean                    $public_only Whether to count only publicly published posts
```
CodeProKid: spacing is messed up here

src/Model/User.php
```diff
@@ -78,21 +78,105 @@ public function __construct( \WP_User $user ) {
...
+		 * If there's no cache, query to see if the author has 1 published post.
+		 *
+		 * This method of querying is more efficient on large datasets
+		 * then the core count_user_posts() method as this stops at the first match,
+		 * where the count_user_psts() method needs to find all matches and
+		 * it can take a while to calculate.
+		 *
+		 * We only care if there's at least 1 published post, not the total.
+		 */
+		if ( false === ( $count = wp_cache_get( $cache_key, $cache_group ) ) ) {
```
CodeProKid: This falls apart if someone un-publishes the post. We would have to have some code hooked into transition post status to bust this cache.

Also, where does this cache get set? I only see the get call.

src/Model/User.php
jasonbahl: @CodeProKid If we were to use the existing `graphql_data_is_private` filter, it _still_ executes this `is_private` method, which will execute the query to check for published posts, even if a filter were applied to override the results.

Perhaps we could do what I'm doing here in the Model.php file though, to centralize it better. 

Basically, I want to provide an "escape hatch" for folks to _completely_ override the automated checks we're doing on their behalf. 

For example, if you just have a hard rule that ALL USERS should be considered private, always, then you could avoid the lookup for published posts altogether and just return true with no overhead. 

If you were to do that with the current filter, and just `return__true` it still executes the query. 

I didn't think it would, but upon testing and actually monitoring, it indeed does. 🤷‍♂

src/Model/User.php
jasonbahl: Attachments don't get the `publish` status, just inherit. 

So, it's a waste of energy to check for `publish` attachments, as there won't be any.

src/Model/User.php
jasonbahl: 🤔 I used the `composer run fix-cs` command here, so it should be formatted according to the WP code standards...

src/Model/User.php
jasonbahl: ya. . .that's what I was thinking too. 😬 

Where's DFM Transients when you need them?

jasonbahl: > Also, where does this cache get set? I only see the get call.

uhhh. . .looks like I caught a case of the "moving too fasts" 🤢

src/Model/User.php
jasonbahl: Ahh! Spacing is because of `@userid` should have been `$userid`

src/Model/User.php
CodeProKid: yeah, that's what I'm proposing. Move the existing filter in Model.php to before we call `$this->is_private();` and only call `$this->is_private();` if the filter is empty, otherwise use the results of the filter.

src/Model/User.php
CodeProKid: ah, I see. Would be good to add an inline comment about this for context, since this seems super arbitrary.

src/Model/User.php
CodeProKid: You will also get stuck in a bad state if an author has 0 published posts, and then publishes a post, as it will still be cached as `0`. So you'll need to bust the cache in that instance as well.

We definitely want to cache `0` as a legitimate result since querying for published posts for a user that has none will do a full table scan to result in the `0`.

src/Model/User.php
chriszarate: :+1: we would use this filter and it would solve the entire perf problem for us

src/Model/User.php
chriszarate: i would consider calling to `wpcom_vip_count_user_posts` behind a `function_exists` check. that function basically implements what you've done here but i believe they handle the cache busting:

https://github.com/svn2github/wordpress-vip-plugins/blob/master/vip-do-not-include-on-wpcom/wpcom-caching.php#L410

i don't have a strong opinion on needing to bust the cache, i'd probably lean towards not busting it and just documenting the delay.

can i ask what the threat model is here? why exactly do users without published posts have an expectation of privacy? the whole reason that this method is unperformant is that it _always_ runs, even when the user connected to a post (in which case, they obviously have pubbed a post).

is there a way to make this run conditional depending on the `$source` object?

jasonbahl:
> can i ask what the threat model is here? why exactly do users without published posts have an expectation of privacy? the whole reason that this method is unperformant is that it always runs, even when the user connected to a post (in which case, they obviously have pubbed a post).

@chriszarate WordPress treats users as private entities by default until they publish posts. If you install WordPress and create a new user, that user is not discoverable in any way until they have published a post. We're trying to respect existing access control patterns that WordPress has set.

zacscott:
Hello,

I have raised another pull request to hopefully help resolve this issue. This was based on the PR you originally opened @jasonbahl, simply adding the `graphql_user_model_is_private` filter with no caching / query changes etc.  https://github.com/wp-graphql/wp-graphql/pull/1685

I see @chriszarate has also opened a pull request a short time ago. I think there is a place for both of our solutions.

Being able to completely bypass the is_private() is ideal for our situation as we'll either have all users available or have it based on something much simpler, e.g. the user role.

PR #970, opened 2019-10-11
Fix/961 negative performance impact user model

TylerBarnes:

This PR improves the performance of post queries for sites with a lot of posts. When posts are queried, an extra SQL query is running for each post which gets the post author and then counts that users posts in order to determine wether they’re a published author or not so that we can determine wether that users account is public data or not. For sites where a single user has for ex 100,000 posts, this was problematic as it was then potentially counting those posts repeatedly when the fields were resolving. This PR uses the User Loader to make a single SQL query for all users at once to check if they're published authors, instead of an extra query for each post.

#961 

Any other comments?
-------------------
This is my first PR to this repo. Nitpicky CR highly welcomed. Please let me know why this code isn't good, haha :)

Where has this been tested?
---------------------------
**Operating System:** macOs mohave

**WordPress Version:** 5.2.3

### DIFF
--- a/src/Data/Loader/UserLoader.php
+++ b/src/Data/Loader/UserLoader.php
@@ -71,16 +71,16 @@ class UserLoader extends AbstractDataLoader {
 		
 		$this->load_published_author_ids( $keys );
 		
-		foreach ( $keys as $key ) {
-			$user = get_user_by( 'id', $key );
+		foreach ( $keys as $id ) {
+			$user = get_user_by( 'id', $id );
 
-			$has_published_posts = in_array(
-				absint( $key ),
+			$user_has_published_posts = in_array(
+				absint( $id ),
 				$this->published_authors,
 				true 
 			);
 
-			$user->has_published_posts = $has_published_posts;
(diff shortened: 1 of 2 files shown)


src/Data/Loader/UserLoader.php
```diff
@@ -61,16 +68,78 @@ public function loadKeys( array $keys ) {
...
+        unset( $post_types[ 'attachment' ] );
+        unset( $post_types[ 'revision' ] );
+		
+		$stringified_ids        = implode( ',', $ids );
+		$stringified_post_types = '"' . implode( '", "', $post_types ) . '"';
+
+		$where = get_posts_by_author_sql( $post_types, true, null, false );
+		
+		global $wpdb;
+        $results = $wpdb->get_results(
```
jasonbahl: Spacing looks funky here. 

We'll also need to make sure to use `$wpdb->prefix` for the post tables as not all WP installs use `wp_` as the prefix, so we can't hardcode that.

TylerBarnes: Just pushed up a fix for this.

PR #972, opened 2019-10-14
Allow developers to short circuit potentially expensive is_private ch…

### LINKED ISSUE #961: Negative performance impact in the User Model

CodeProKid:
…eck in model layer

Makes the `graphql_data_is_private` filter a short circuit filter by using `null` as the default value. A callback could then override this with `true` or `false` and that will prevent the default `is_private()` check running for the model.

Fixes #961 

Any other comments?
-------------------
Technically this is a breaking change since the filter signature is changing, but I don't really consider it to be a breaking change since the previous `true` or `false` context passes as the first param from the `is_private()` method probably wasn't really useful in the callback anyways, other than returning it as the default value if certain conditions weren't met, and that functionality still works as it did.

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …


jasonbahl:
@CodeProKid this is breaking tests 🤔

src/Model/Model.php
```diff
@@ -209,17 +209,23 @@ public function get_visibility() {
...
 			 * @param null|int    $owner          The user ID for the owner of this piece of data
 			 * @param \WP_User    $current_user   The current user for the session
 			 *
-			 * @return bool
+			 * @return bool|null
 			 */
-			$is_private = apply_filters( 'graphql_data_is_private', $this->is_private(), $this->get_model_name(), $this->data, $this->visibility, $this->owner, $this->current_user );
+			$is_private = apply_filters( 'graphql_data_is_private', null, $this->get_model_name(), $this->data, $this->visibility, $this->owner, $this->current_user );
+
+			if ( null !== $is_private ) {
```
mwidmann: Am I completely stupid or should this be 

`if ( null === $is_private ) {`

because I want to be able to set $is_private through the filter, not unset it

esamattis:
WP core seems to use `pre_` style filters for this kind of short circuiting for the actually implemtations when needed.

Fox example here's how `option` filtering is implemented

Pre skip:

https://github.com/WordPress/WordPress/blob/c67763478840fabf689ecbd175dc682da1623069/wp-includes/option.php#L58-L62

Actual value filtering:

https://github.com/WordPress/WordPress/blob/c67763478840fabf689ecbd175dc682da1623069/wp-includes/option.php#L152


Maybe we should follow the same convention? Now this PR removes the possibility to reference the original private value.

jasonbahl (dismissed):

src/Model/Model.php
jasonbahl: The default value is `null`. If 3rd party code (plugin/theme) make use of this filter and return a value other than null (true/false, for example) that value will be used. If the value remains `null`, then default logic will be applied to the model's visibility.

jasonbahl (dismissed):

src/Model/Model.php
jasonbahl: So ya, it should be `===` to fallback to the Model's default is_private logic. 

Updated.

jasonbahl:
> WP core seems to use `pre_` style filters for this kind of short circuiting for the actually implemtations when needed.
> 
> Fox example here's how `option` filtering is implemented
> 
> Pre skip:
> 
> https://github.com/WordPress/WordPress/blob/c67763478840fabf689ecbd175dc682da1623069/wp-includes/option.php#L58-L62
> 
> Actual value filtering:
> 
> https://github.com/WordPress/WordPress/blob/c67763478840fabf689ecbd175dc682da1623069/wp-includes/option.php#L152
> 
> Maybe we should follow the same convention? Now this PR removes the possibility to reference the original private value.

jasonbahl:
@zacscott I got this PR updated and passing tests. 

This should help with the issue you were facing. @imranhsayed @chriszarate @benmay this should help with what ya'll were facing as well?

I still plan to fix the SQL check as well that checks for published authors (there are a few open PRs for that, but this should unblock you until I review those PRs and make some final decisions on that).

imranhsayed:
Thank you @jasonbahl 
Appreciate it 👍

zacscott:
Thank you @jasonbahl 🙌

PR #975, opened 2019-10-16
Feature/lint staged phpcs

TylerBarnes:

This PR adds lint-staged and husky to help with keeping the codebase clean during development.
Anytime someone tries to make a commit, husky will run it's precommit git hook which calls lint-staged to run a command against the staged files. I've configured it to run phpcs and it will output any errors and bail the commit if there are any.

This doesn't force anyone to lint though as you need to run `npm install` in the project before this takes effect, so it's more of a useful tool than a guard. As you can see from this GIF, I only added one error to the file I'm testing with but `phpcs` returns 16 errors and 1 warning. So this is really useful to have going forward!

lint-staged-php-2


TylerBarnes:
Oh, I nearly forgot. I actually had to delete the composer.lock in order to run composer install and get phpcs running for me. I'm not sure if that's going to cause any issues. Any ideas on that @jasonbahl ? It was having trouble resolving packages due to some conflict.

jasonbahl:
@TylerBarnes try with `composer install --no-dev` as it looks like it's trying to add the dev dependencies for versioning, which we don't want

jasonbahl:
when running dev scripts, you _do_ need the dev dependencies, but we don't want to ship them in the versioned code as they're not intended for execution by anything other than Travis-CI or devs running the dev scripts locally. 

To run the dev scripts, you can run `composer install` then run the script(s) you need to, then when committing for a PR, run `composer install --no-dev`

That _should_ probably be automated in some way though. . .or even better, we should just not version _any_ of the composer dependencies at all, and let automation bundle them into the distributed packages.

jasonbahl:
@TylerBarnes weird. Maybe a PHP version conflict? 🤔

TylerBarnes:
@jasonbahl it was a Catalina issue!
https://stackoverflow.com/a/58300437

TylerBarnes:
Ok, I added some additional instructions. I know this feature isn't a high priority but IMO code quality and cleanliness is super important and this should help with that!

Issue #846, opened 2019-05-09
Provide a way to clear the cached Schema.


kidunot89:
**Description**
In the event that a new type is created dynamically after the schema is established, for instance a new `term` or `post_type` is defined for use in a test, and function is needed to set `WPGraphQL::schema` to `null` before the next `graphql()` call.

**Solution**
I propose a `WPGraphQL::clear_schema()` be added. It should be static obviously.

jasonbahl:
@kidunot89 we both burned some time on this and got nowhere. 😭 

I think we have workarounds for this though re: testing bootstrapping. If this comes back up as a pain point, we can try and take another stab and re-open, or open another issue.

jasonbahl:
@kidunot89 re-opening. I got burned by this again working on some tests for the ACF plugin.

jasonbahl:
I also did a bit of work to dive in again and have some progress. 

I'm able to now register a post_type in the setUp method of a file, or within the test method directly, then query against it. 

However, the way I have this working breaks pretty much everything else in the plugin. 😂 

So, now that I understand it _can_ be done and that it's not a specific WP + Codeception issue, I can at least address it now. 

Hope to have this resolved tomorrow 🤞

jasonbahl:
Closed. This was resolved by v0.4.0

PR #979, opened 2019-10-18
Release/v0.4.0

jasonbahl:
# Release Summary

This release is centered around Issue #846. Plugins that extend the WPGraphQL Schema are very difficult to test because once the Schema is created, it can't be cleared or modified. This is because of the heavy use of Statics, which make testing very difficult. 

This Release changes many of the Static classes to classic objects that are instantiated and passed down using Dependency Injection. 

This makes it much easier to create instances of the Schema on the fly, which is essential for extension plugins to do when testing their extensions. It also helps the core plugin properly test API methods like `register_graphql_field`, `register_graphql_mutation`, etc. 

These changes make it easier for extensions to write and execute tests. 

Because of the statics, the WPGraphQL Schema could be generated just once for the entire test suite, which means to test things like registering custom fields in ACF will add fields to the WPGraphQL Schema would mean that _all_ fields would need to be added in the Bootstrap of the test suite, and _not_ within the individual test that's testing the fields. 

This is problematic to scaling tests. Ideally, we want to follow the AAA method of testing: 

- **Arrange**: setup what we need to test
- **Act**: execute what we want to test
- **Assert**: assert that the execution behaved as expected

What this means is that if we wanted to test something like adding an ACF Text field (doesn't have to be ACF, just an example) we would want a test that: 

- registers an ACF field group
- registers an ACF Field to that group
- creates a post
- updates meta to the ACF field of that post
- execute a GraphQL Query for the post and the field
- assert that the field registered in ACF added a field to the Schema and resolves as expected

We cannot do that today, but with this release, we can! Now, each test can "own" everything needed to test its behavior.

---

## Breaking Changes

The primary breaking change here is making the TypeRegistry class no longer a singleton, so the methods are not statically accessible anymore. Instead, the TypeRegistry is an object that is passed down in the context of the `graphql_register_types` action. 

If your code was previously making use of any static methods from the `TypeRegistry` class, such as `TypeRegistry::get_type( 'TypeName' )` you will need to refactor a bit. 

**Upgrading**

Below is an example from the [WPGraphQL for Gravity Forms](https://github.com/harness-software/wp-graphql-gravity-forms) plugin that needs to change: 

@kellenmace pointed out: 

> I am using TypeRegistry::get_type(), here: https://github.com/harness-software/wp-graphql-gravity-forms/blob/master/src/Types/Union/ObjectFieldUnion.php#L45
src/Types/Union/ObjectFieldUnion.php:45

To upgrade to v0.4.0, this code should be changed from: 

```
public function register_type() {
        $field_mappings = $this->get_field_type_mappings();
        register_graphql_union_type( self::TYPE, [
            'typeNames'   => array_values( $field_mappings ),
            'resolveType' => function( GF_Field $field ) use ( $field_mappings ) {
                if ( isset( $field_mappings[ $field['type'] ] ) ) {
                    return TypeRegistry::get_type( $field_mappings[ $field['type'] ] );
                }
                return null;
            },
        ] );
    }
```

to: 

```
public function register_type( $type_registry ) {
        $field_mappings = $this->get_field_type_mappings();
        register_graphql_union_type( self::TYPE, [
            'typeNames'   => array_values( $field_mappings ),
            'resolveType' => function( GF_Field $field ) use ( $field_mappings, $type_registry ) {
                if ( isset( $field_mappings[ $field['type'] ] ) ) {
                    return $type_registry->get_type( $field_mappings[ $field['type'] ] );
                }
                return null;
            },
        ] );
    }
```

Here we can see that the callback is passed the `$type_registry` as it's a callback of the `graphql_register_types` action. The `$type_registry` is the instance of the TypeRegistry, and now the `get_type` method can be used. We changed `TypeRegistry::get_type()` to `$type_registry->get_type()`. 

**NOTE:,** we are using a closure for the `resolveType` so we have to pass `$typeRegistry` through with the `use` statement. 

----

## Todo before release:

- [x] Write Release Summary (@jasonbahl)
- [x] Ensure Docs reflect current changes (@TylerBarnes)
- [x] Test/upgrade WPGraphQL Insights for compatibility (@jasonbahl)
   - Tested with [WPGraphQL Insights v0.3.1](https://github.com/wp-graphql/wp-graphql-insights/releases/tag/v0.3.1)
- [x] Test/upgrade WPGraphQL for ACF for compatibility (@jasonbahl)
   - See: https://github.com/wp-graphql/wp-graphql-acf/pull/75 
- [x] Test/upgrade WPGraphQL Tax Query for compatibility (@jasonbahl)
- [x] Test/upgrade WPGraphQL Meta Query for compatibility (@jasonbahl)
- [x] Test/upgrade WPGraphQL JWT Authentication for compatibility (@jasonbahl)
- [x] Test/upgrade WPGraphQL Dad Jokes for compatibility (@jasonbahl)
- [x] Test/upgrade WPGraphQL Custom Post Type UI for compatibility (@jasonbahl)
- [x] Test/upgrade WPGraphQL for WooCommerce for compatibility (@kidunot89)
- [x] Summarize breaking changes and upgrade path (@jasonbahl, with feedback from @kidunot89 and @TylerBarnes)


TylerBarnes:
The latest changes fix all the examples on the docs site! 🥇

src/Type/Union/PostObjectUnion.php
```diff
@@ -1,26 +1,63 @@
...
+
+	/**
+	 * Returns a list of possible types for the union
+	 *
+	 * @access public
+	 * @return array
+	 */
+	public static function get_possible_types() {
+		$possible_types     = [];
+		$allowed_post_types = \WPGraphQL::get_allowed_post_types();
```
kidunot89: Change `\WPGraphQL::get_allowed_post_types();` to `get_post_types( [ 'show_in_graphql' => true ] );` so custom post-types registered with a custom definition can still be resolved.

jasonbahl: @kidunot89 it looks like we use the `\WP_GraphQL::get_allowed_post_types()` several times throughout the plugin, and it calls `get_post_types( [ 'show_in_graphql' => true ])` then filters and applies another filter. Is there an issue with that? Is the static method problematic or something?

jasonbahl: Happy to change, just want to make sure I understand

Issue #901, opened 2019-07-11
Comments of private posts should also be private


jasonbahl:
When a user queries for all comments, comments that are on a private post should not be returned if the user doesn't have permission to "moderate_comments"

A query such as: 

```
query getComments{
  comments{
    edges{
      node{
        id
        content     
      }
    }
  }
}
```

Will return all comments, regardless of their relationship to a private post.

jasonbahl:
closed by #993

PR #993, opened 2019-11-02
Fix issue with model layer returning connected items on private posts

### LINKED ISSUE #992: Posts set to private show full taxonomy nodes instead of `null`
When a post is set to private, the `categories` (or similar taxonomy) list will return all terms. I would expect it to return simply `null` like the rest of the post fields. I was using WP-GraphQL v0.3.8, but I am also seeing this issue on the latest version, v0.4.0.

Example query:
```
query GetPostQuery($postId: Int) {
  postBy(postId: $postId) {
    status
    title
    categories {
      nodes {
        id
        name
        slug
      }
    }
  }
}
```

Returned data of a published post with category term of "Some Category 1" selected:
```
{
  "data": {
    "postBy": {
      "status": "publish",
      "title": "Some post",
      "categories": {
        "nodes": [
          {
            "id": "Y2F0ZWdvcnk6MTQ=",
            "name": "Some Category 1",
            "slug": "some-category-1"
          }
        ]
      }
    }
  }
}
```

Returned data of same post now set to private:
```
{
  "data": {
    "postBy": {
      "status": null,
      "title": null,
      "categories": {
        "nodes": [
          {
            "id": "Y2F0ZWdvcnk6MTQ=",
            "name": "Some Category 1",
            "slug": "some-category-1"
          },
          {
            "id": "Y2F0ZWdvcnk6MQ==",
            "name": "Uncategorized",
            "slug": "uncategorized"
          }
        ]
      }
    }
  }
}
```

My categories list is currently just "Some Category 1" and "Uncategorized", so the full list is showing here. Note that this was queried from the anonymous user. This issue does not occur when querying as an admin user.

This can be replicated on a fresh install of Wordpress and WP GraphQL. Create some category terms and a post with little to no categories, then query the post for its terms.

CodeProKid:

Makes sure we don't return all terms on taxonomy connections to private posts.

closes #992 

Any other comments?
-------------------
It's definitely debatable that this short circuit code is in the correct place. Having the test here is useful though, and we can have a conversation around this now.

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -6,6 +6,7 @@ use GraphQL\Error\UserError;
 use GraphQL\Type\Definition\ResolveInfo;
 use GraphQLRelay\Connection\ArrayConnection;
 use WPGraphQL\AppContext;
+use WPGraphQL\Model\Post;
 
 /**
  * Class AbstractConnectionResolver
@@ -106,6 +107,11 @@ abstract class AbstractConnectionResolver {
 	 * @throws \Exception
 	 */
 	public function __construct( $source, $args, $context, $info ) {
+
+		// Bail if the Post->ID is empty, as that indicates a private post.
+		if ( $source instanceof Post && empty( $source->ID ) ) {
+			return null;
+		}
 
 		/**
 		 * Set the source (the root object) for the resolver
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -110,7 +110,7 @@ abstract class AbstractConnectionResolver {
 
 		// Bail if the Post->ID is empty, as that indicates a private post.
 		if ( $source instanceof Post && empty( $source->ID ) ) {
-			return null;
+			$this->should_execute = false;
 		}
 
 		/**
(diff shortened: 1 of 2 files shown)


src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -107,6 +108,11 @@ abstract class AbstractConnectionResolver {
 	 */
 	public function __construct( $source, $args, $context, $info ) {
 
+		// Bail if the Post->ID is empty, as that indicates a private post.
+		if ( $source instanceof Post && empty( $source->ID ) ) {
+			return null;
```
jasonbahl:
```suggestion
			$this->should_execute = false;
```

CodeProKid:
@jasonbahl I don't think your change worked correctly: https://travis-ci.org/wp-graphql/wp-graphql/jobs/606779619#L2826

The test I added is failing.

src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -107,6 +108,11 @@ abstract class AbstractConnectionResolver {
 	 */
 	public function __construct( $source, $args, $context, $info ) {
 
+		// Bail if the Post->ID is empty, as that indicates a private post.
+		if ( $source instanceof Post && empty( $source->ID ) ) {
+			$this->should_execute = false;
```
jasonbahl:
```suggestion
			return null;
```

CodeProKid:
Chatted about this during bug scrub, I think the test is just failing because it's returning `null` instead of an empty array. Just need to re-implement the change and do some testing.

There's also another ticket open that references a similar issue: https://github.com/wp-graphql/wp-graphql/issues/901 this fix _should_ apply to that ticket as well, but we should confirm with a test.

jasonbahl:
Shoot. Didn't mean to merge this yet. 🤦‍♂

Issue #991, opened 2019-11-01
WPInterfaces aren't fully compatible with `register_graphql_field()` and `register_graphql_connection()`


kidunot89:
The have been a few issues related to Interfaces and they all point back to incorrect registration the schema. For example
1. Install and activate **WPGraphQL**, **WPGraphiQL**, **WooGraphQL**, and **WPGraphQL ACF**, along with their corresponding plugins.
2. Import the sample product data included in the **WooCommerce** plugin.
3. Create a coupon. Its properties aren't important.
4 Create an order. Again it's properties aren't important.
5. Create a custom field group *(e.g. fieldsTest)* with a test field *(e.g. testText)* with a default value and set rules to target the `Product`, and `Coupon` post-types.
6. Now navigate to **WPGraphiQL** and run the following query.
```
{
  coupons {
    edges {
      cursor
      node {
        id
        fieldsTest {
          fieldGroupName
          textTest
        }
      }
    }
  }
  products {
    edges {
      cursor
      node {
        id
      	fieldsTest {
          fieldGroupName
          textTest
        }
      }
    }
  }
}
```
And you should get a result like this
![query results](https://i.imgur.com/phE13Pr.png?1)

The coupons resolve the correctly, however the products don't even show the field. This is because `Product` is an Interface type and WPGraphQL ACF registers the field on the Interface but not it's child object types. And fields that registered to the Interface but not the child object type don't exist to that child object type. If you tried changing
```
 products {
    edges {
      cursor
      node {
        id
      	fieldsTest {
          fieldGroupName
          textTest
        }
      }
    }
  }
```
to
```
 products {
    edges {
      cursor
      node {
        id
        ... on SimpleProduct {
      	  fieldsTest {
            fieldGroupName
            textTest
          }
        }
      }
    }
  }
```
You would get this error.
```
{
  "errors": [
    {
      "message": "Cannot query field \"fieldsTest\" on type \"SimpleProduct\".",
      "category": "graphql",
      "locations": [
        {
          "line": 20,
          "column": 11
        }
      ]
    }
  ]
}
```

I personally feel this should be address at the `register_graphql_field()` level. And it should register the field on both the Interface and it's child object types.

The next part is `register_graphql_connection()`. Just as above connections aren't registered to the child object types, and manually registering the connection for the child object type creates a mix-match type issue. The return types will be a different for both Interface and the child object type resulting in a number of fragmenting errors with client-side libraries like Apollo.

[...]

PR #1005, opened 2019-11-04
interface field hook implemented

kidunot89:

- Creates a `graphql_interface_{$type_name}_fields` filter that is provided the WPInterface fields after all field registration has be completed.
- The `graphql_interface_{$type_name}_fields` filter is hooked into by any inheriting types and the WPInterface's fields are retrieved and merge with the inheriting types fields.
- The WPInterface's fields configs are overwrite by any existing field configs on the inheriting type.
- This allow the forgoing of any field defined in the Parent Interface. For example the `id` field definition was removed from CPT definitions in `src/Type/Post.php` and it's description was played on the `id` definition in `Node`.  

#991 
[WooGraphQL #166](https://github.com/wp-graphql/wp-graphql-woocommerce/issues/166)

Any other comments?
-------------------
…

Where has this been tested?
---------------------------

**WordPress Version:** 5.2.4

### DIFF
--- a/src/Type/Object/PostObject.php
+++ b/src/Type/Object/PostObject.php
@@ -209,12 +209,6 @@ class PostObject {
 	public static function get_post_object_fields( $post_type_object ) {
 		$single_name = $post_type_object->graphql_single_name;
 		$fields      = [
-			'id'                => [
-				'type'        => [
-					'non_null' => 'ID',
-				],
-				'description' => __( 'The globally unique ID for the object', 'wp-graphql' ),
-			],
 			$single_name . 'Id' => [
 				'type'        => [
 					'non_null' => 'Int',
--- a/src/Type/Object/PostObject.php
+++ b/src/Type/Object/PostObject.php
@@ -209,6 +209,13 @@ class PostObject {
 	public static function get_post_object_fields( $post_type_object ) {
 		$single_name = $post_type_object->graphql_single_name;
 		$fields      = [
+			'id'                => [
+				'description' => sprintf(
+					/* translators: %s: custom post-type name */
+					__( 'The globally unique identifier of the %s object.', 'wp-graphql' ),
+					$post_type_object->name
+				),
+			],
 			$single_name . 'Id' => [
 				'type'        => [
 					'non_null' => 'Int',
--- a/src/Type/WPInterfaceType.php
+++ b/src/Type/WPInterfaceType.php
@@ -98,6 +98,15 @@ class WPInterfaceType extends InterfaceType {
 		 * as it ensures it's not output in just random order
 		 */
 		ksort( $fields );
+
+		// Make field definitions available for inheriting objects.
+		add_filter(
+			"graphql_interface_{$type_name}_fields",
+			function( $other_interface_fields ) use ( $fields ) {
+				return array_merge( $other_interface_fields, $fields );
+			}
+		);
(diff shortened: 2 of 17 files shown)


src/Type/Object/PostObject.php
```diff
@@ -209,12 +209,6 @@ public static function register_post_object_types( $post_type_object ) {
 	public static function get_post_object_fields( $post_type_object ) {
 		$single_name = $post_type_object->graphql_single_name;
 		$fields      = [
-			'id'                => [
```
jasonbahl: So, you're deleting this because it is _exact_ same as the Interface definition and will inherit the interface definition of the field, but you're leaving the `id` field on `TermObject.php` because the description is a bit different?

src/Type/WPInterfaceType.php
```diff
@@ -98,6 +98,15 @@ public function prepare_fields( $fields, $type_name ) {
 		 * as it ensures it's not output in just random order
 		 */
 		ksort( $fields );
+
```
jasonbahl: Can you explain why this filter is being applied here? It appears to be merging fields with themselves. This seems unnecessary, but perhaps I'm missing something?

src/Type/WPObjectType.php
```diff
@@ -127,6 +126,22 @@ public function prepare_fields( $fields, $type_name ) {
 		 */
 		$fields = apply_filters( 'graphql_object_fields', $fields, $type_name );
 
+		// Get any parent interface fields.
+		$parent_fields = [];
+		if ( ! empty( $config['interfaces'] ) ) {
```
jasonbahl:
```suggestion
		if ( ! empty( $config['interfaces'] ) && is_array( $config['interfaces'] ) ) {
```

src/Type/WPInterfaceType.php
kidunot89: To catch all fields registered with `graphql_register_field()` on the Interface.

src/Type/WPObjectType.php
kidunot89: Can it be anything other than an `array`

src/Type/Object/PostObject.php
kidunot89: Yea. pretty much :man_shrugging:. We can change them all, or do a deep merge to simple overwrite the `description` instead. What do you think?

src/Type/WPObjectType.php
jasonbahl: it's not supposed to be, but we don't want to loop over it if it were a string or something else. It's a value 3rd party devs will be filling out, so we don't want to iterate over a non-array should they fill it out incorrect. 

Perhaps if it's not an array, we throw an error so they know they messed up. 🤔

kidunot89: Yea, that would probably be the best bet.

jasonbahl:
So, we were missing implementing the `node` interface on `menu`, `menuItem` and `userRole` so adding them fixes the tests. 

However, this makes me realize we don't properly have resolvers in place for _all_ nodes from the root `node` field. 

We should be able to pass any ID from any node into the `node` field to query nodes of _any_ type, for example:

```
{
  node( id: $nodeId ) { 
    __typename
    ...on Post { 
      id
      title
    } 
  } 
}
```

PR #1012, opened 2019-11-07
Bug/#1009 account for posts with invalid author

### LINKED ISSUE #1009: Account for posts whose author is no longer a user
Trying to run a `posts { ... }` or `pages{ ... }` query that includes posts/pages assigned to an author who is no longer a user on the site results in a 500 Internal Server Error.

## Steps to reproduce

1. On a WP site, create a user account and note the user ID.
1. Log in as that user and create a few posts.
1. Delete that user account and do not reassign their posts to another user upon deletion.
1. Run a GraphQL query such as this one, making sure that some of the deleted user's posts are in the range of posts being queried:

```
{
  posts(first: 10) {
    nodes {
      postId
      author {
        userId
        name
      }
    }
  }
}
```

1. Confirm that a 500 Internal Server error is returned to the client.

This happens because the user ID of the user that no longer exists gets passed in as the `$key` variable when these lines execute in `wp-graphql/src/Data/Loader/UserLoader.php`:

```
$user = get_user_by( 'id', $key );
$all_users[ $user->ID ] = new User( $user );
```

That results in this error being thrown:
`Argument 1 passed to WPGraphQL\Model\User::__construct() must be an instance of WP_User, bool given`

## Expected Behavior

The query should complete, with `null` for the value of author fields for posts/pages those author is no longer a user.

## Proposed Solution

@jasonbahl thinks updating those lines to something like this could solve the issue:

```
$user  = get_user_by( 'id', $key );
if ( $user instanceof WP_User ) {
  $all_users[ $user->ID ] = new User( $user );
} else {
  $all_users[ 0 ] = null;
}
```

^ That has not yet been tested, though. The result of that should be that the author fields in the query are set to `null` for posts/pages those author is no longer a user.

Version of WPGraphQL: 0.4.

kellenmace:
Account for posts whose author is no longer a user.
Add test to make sure posts queries that contain posts assigned to an author who is not a user on the current site still resolve, and that null is returned for the author field.

Closes #1009 

Example query and response with the fix applied:
https://cloudup.com/czeduU1ZIIG

Any other comments?
-------------------
n/a

Where has this been tested?
---------------------------
**Operating System:** Chrome on macOS Mojave

**WordPress Version:** 5.2.4


tests/wpunit/PostObjectQueriesTest.php
```diff
@@ -1659,4 +1659,53 @@ public function dataProviderRestrictedPosts() {
...
+	/**
+	 * Test the scenario where a post is assigned to an author
+	 * who is not a user on the site. This could happen for instance,
+	 * if the user was deleted, but their posts were never trashed
+	 * or assigned to another user.
+	 */
+	public function testQueryPostsWithOrphanedAuthorDoesntThrowErrors() {
+		global $wpdb;
+
+		$highest_user_id     = (int) $wpdb->get_var( "SELECT ID FROM {$wpdb->users} ORDER BY ID DESC limit 0,1" );
```
CodeProKid: Was this because of the inability to use `wp_delete_user()` ? I'd rather just include the core file that contains the function than do this.

src/Data/Loader/UserLoader.php
```diff
@@ -63,12 +63,18 @@ public function loadKeys( array $keys ) {
...
 		/**
-		 *
+		 * Populate array with User objects, or null if user does not exist.
 		 */
 		foreach ( $keys as $key ) {
-			$user                   = get_user_by( 'id', $key );
-			$all_users[ $user->ID ] = new User( $user );
+			$user = get_user_by( 'id', $key );
+
+			if ( $user instanceof \WP_User ) {
```
CodeProKid: Another option here is to do a `try{ } catch{ }` around `User( )` since it throws an exception when you pass it an invalid input. That would look like:
```php
try {
    $all_users[ $user->ID ? $user->ID : $key ] = new User( $user );
catch ( Exception $e ) {
    $all_users[ $key ] = null;
}
```
I think it's important to still return an entry for the `$key` as part of the returned array, since other things downstream will be looking for that, even if it's just a null value.

kellenmace: @CodeProKid Thanks for the feedback! Sounds good - I just updated the PR to match your code. So our try/catch will be responsible for catching exceptions for both of these, when `$user` is set to `false`:
- Trying to get property 'ID' of non-object
- `new User()` throwing an exception

Sound good?

tests/wpunit/PostObjectQueriesTest.php
kellenmace: @CodeProKid I wan't able to use `wp_delete_user()`, since it takes all the posts that were assigned to the  user being deleted and trashes them. So by using it, we don't end up with the "orphaned" post scenario we're trying to test.

With that in mind, I revamped my code to create a new user, insert a post with then as the author, then delete them from the DB without trashing their posts, like this:
https://gist.github.com/kellenmace/612411db9e1fa7f5d98d1d4862b5fe2f

Ultimately though, that seemed like a lot of code just to end up with a post assigned to a bogus user ID, so I switched to my current implementation of just getting the highest user ID in the `wp_users` table and incrementing it by 1 to guarantee that the post being inserted is being assigned to a non-existent user ID.

Which approach to you prefer? At the end of the day, both of them result it a post being inserted with a bogus user ID as the author, so I think either suits the purpose of the test.

src/Data/Loader/UserLoader.php
CodeProKid: Yeah, the ternary should deal with the `ID` on `$user` just fine, it's really just for catching the Exception from `User()`.

We **need** the ternary there because the `Trying to get property 'ID' of non-object` error is just that, an error, not an Exception, so the catch wouldn't catch it. 😄

tests/wpunit/PostObjectQueriesTest.php
CodeProKid: @kellenmace thanks for slogging through this ridiculous issue with `wp_delete_user()`, you're a rockstar.

I think the current implementation with finding the highest user ID and adding 1 to it is a bit overkill for this test. Let's just go back to basics and use `999999` for the user. The DB is rolled back periodically throughout the tests (between individual unit test methods, as well as classes), so the chances of the ID overlapping are so insanely slim.

If someone else has a problem with that they can take it up with me! 😆

kellenmace: @CodeProKid Cool! Sounds good. `999999` it is! I'll update it to that in a little bit.

src/Data/Loader/UserLoader.php
jasonbahl: So, with the code as above: 

```
$user = get_user_by( 'id', $key );
try {
  $all_users[ $user->ID ? $user->ID : $key ] = new User( $user );
} catch ( \Exception $e ) {
  $all_users[ $key ] = null;
}
```

I get a lot of errors. 

See: 

Screen Shot 2019-11-08 at 8 51 11 AM

---- 

But if I use the following code, I do not get said errors:

```
$user = get_user_by( 'id', $key );
if ( $user instanceof \WP_User ) {
  $all_users[ $user->ID ] = new User( $user );
} else if ( ! isset( $all_users[ 0 ] ) ) {
  $all_users[0] = null;
}
```

CodeProKid: ah, it's because we're using type hinting on the input of that class. That throws an error before it gets to the exception.

We'll have to either use the code @jasonbahl recommended or switch the catch to catch `Throwable`. That would look like:

```php
$user = get_user_by( 'id', $key );
try {
  $all_users[ $user->ID ? $user->ID : $key ] = new User( $user );
} catch ( \Throwable $e ) {
  $all_users[ $key ] = null;
}
```

jasonbahl: ^ This latest from @CodeProKid works for me

CodeProKid: Actually, now that I'm thinking about it, you can only catch `Throwable`'s in PHP 7+, so we should just go back to the original way.

jasonbahl: Ah! Good call.

jasonbahl:
Closes #1009

PR #1016, opened 2019-11-12
Revert "Fix issue with model layer returning connected items on private posts"

jasonbahl:
Reverts wp-graphql/wp-graphql#993

(Merged before I meant to) 🤦‍♂


jasonbahl:
- I'll make another forward PR instead of reverting

PR #1017, opened 2019-11-12
Feature/#974 revisions should be same type as parent

### LINKED ISSUE #974: Revisions should be the same type as what they are revising
Currently, Revisions in the Schema are treated as their own `Revision` Type. However, in real use cases, Revisions are the same Type (page, post, etc) as what they're revising. Their Type in the Schema should be the same. 

For example, this kind of query should work: 

```
{
  posts {
    nodes {
      id
      title
      someCustomFieldOnlyForPosts
      revisions {
          id
          title
          someCustomFieldOnlyForPosts
       }
    }
  }
}
```

We should be able to register things like custom fields to a Post Type, and revisions of that Post Type should expose those fields as well. 

A revision of a Post is still a Post. A revision of a Page is still a page. 

So, what I propose needs to change, is that all fields exposing revisions resolve to the same Type as the parent (for example `Page.revisions` should resolve to a `Page->Page` connection) and the `RootQuery.revisions` connection should resolve to a `PostObjectUnion` where the node can be one of _many_ Types. 

---

Additionally, a user should be able to fetch a single `page` or `post` (or CPT) using an ID of a revision. 

For example, if we had a revision of a Post with ID `5`, I should be able to do something like: 

```
{
  postBy( postId: 5 ) {
     id
     title
  }
}
```

If the ID is of a revision and the revision parent is a `Post`, then the revision should be resolved to the user, if the user has proper capabilities to view revisions.

jasonbahl:
Currently, Revisions are addd to WPGraphQL as their own Type in the Schema. This largely mimics WordPress implementation, in that `revisions` are a different Post Type. However, this doesn't mimic actual usage. 

Revisions of a Post, declaratively speaking, are still Posts. If I paint my house, it's still a house. If I change the title of a Post, it's still a Post. 

The intent of revisions is to show a thing (Post, Page, etc) as it exists over time. 

The Revision Type declares that the thing is in fact, a different thing, and the utility of Revisions is compromised. 

By changing Revisions in the WPGraphQL Schema to be of the same Type of the object being revised, clients can actually represent their objects as they've been revised over time. 

closes #974


mrclay:
Is there a sense of the remaining TODOs in case others could chip in? Exciting effort here!

jasonbahl:
@mrclay The biggest thing is documenting how folks using revisions need to upgrade, as this is a breaking change release. 

Working on that today. Hope to release today or tomorrow.

jasonbahl:
closing this PR as it's now part of the `v0.5.0` release branch

jasonbahl:
@mrclay check out the release in the works here: https://github.com/wp-graphql/wp-graphql/pull/1056

PR #1018, opened 2019-11-12
#992 - follow-up. 

### LINKED ISSUE #992: Posts set to private show full taxonomy nodes instead of `null`

jasonbahl:
This adjusts the logic for `should_execute` so the Abstract class can declare false and trump the extending class to centrally configure rules that should prevent execution, and expedite resolution.

---- 

Follow up to #993 by @CodeProKid and closes #992

### DIFF
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -110,7 +110,7 @@ abstract class AbstractConnectionResolver {
 
 		// Bail if the Post->ID is empty, as that indicates a private post.
 		if ( $source instanceof Post && empty( $source->ID ) ) {
-			return null;
+			$this->should_execute = false;
 		}
 
 		/**
@@ -562,6 +562,14 @@ abstract class AbstractConnectionResolver {
 	protected function execute_and_get_data() {
 
 		/**
+		 * If should_execute is explicitly set to false already, we can
+		 * prevent execution quickly. If it's not, we need to
+		 * call the should_execute() method to execute any situational logic
+		 * to determine if the connection query should execute or not
+		 */
+		$should_execute       = false === $this->should_execute ? false : $this->should_execute();
+
+		/**
 		 * Check if the connection should execute. If conditions are met that should prevent
 		 * the execution, we can bail from resolving early, before the query is executed.
 		 *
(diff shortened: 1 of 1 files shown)


src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -561,6 +561,14 @@ public function get_page_info() {
 
 	protected function execute_and_get_data() {
 
+		/**
+		 * If should_execute is explicitly set to false already, we can
+		 * prevent execution quickly. If it's not, we need to
+		 * call the should_execute() method to execute any situational logic
+		 * to determine if the connection query should execute or not
+		 */
+		$should_execute       = false === $this->should_execute ? false : $this->should_execute();
```
CodeProKid: spacing is off here.

jasonbahl: 🤦‍♂

Issue #1021, opened 2019-11-18
Router::is_graphql_request() returns false until the `init` action is fired.


kidunot89:
#### The Problem
I have WPGraphQL extension code that only wish to run during GraphQL requests before the `init` action is fired, however when I use `Router::is_graphql_request()` it always returns false.

#### The Expected Behavior
It should return if true on all GraphQL request at all steps in the request after the `Router` class is loaded.

#### Other Comments
The issue is the cause by the fact that the WP query vars aren't available until the `init` is fired. A alternate check should be used before the `do_action( `init` )` is called.

esamattis:
I hit this too. My use case was to conditionally define acf rule for  page_template 

```php
if (!Router::is_graphql_request()) {
  // acf-codifier
  $rule_group->add_rule('page_template', '==', 'some-form');
}
```

I needed to do this because the rule was always false in graphql context and thus not showing the field in the schema. But this workaround did not work because `is_graphql_request()` returns false on the top level where the acf definition was.

esamattis:
Also I originally contributed this method because I needed it for wp-graphql-polylang. But as it appears it does not work there neither because it needs it before the wp query vars are set.

My current hacky graphql context detection method is following: https://github.com/valu-digital/wp-graphql-polylang/blob/aac1be1657879d4590f2086c71794f3284315724/src/Loader.php#L108-L122

jasonbahl:
@kidunot89 can you provide more insight on what you're trying to do conditionally within a GraphQL request _before_ the `init` hook? Some more context will help me understand the best way to approach here.

jasonbahl:
If we're going to have conditionals like this, we have to also assume not everyone is using GraphQL via HTTP requests. 

The ACF example above, will only work when using ACF over HTTP, eh? Because the `Router::is_graphql_request()` is determined by the URI path. 

But if you wanted to use GraphQL within a PHP page template, for example, these wouldn't work because you'd be using the `graphql` or `do_graphql_request` method. 

We need to be careful about supporting these cases too.

jasonbahl:
I wonder if we rename this to `Router::is_graphql_http_request()` so that at least there's a clear expectation that using this conditional will only validate HTTP requests, and not _all_ GraphQL requests. 

There are valid cases where you need to know if it's a GraphQL request coming over HTTP, but there are also the cases where we need to know if it's a GraphQL request, regardless of coming through the `/graphql` HTTP endpoint. 

[...]

PR #1022, opened 2019-11-18
Adds more strict check to "is_graphql_request()"

kidunot89:

- Adds a more strict conditional to `Router::is_graphql_request()`.

#1021 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** Linux 19.2

**WordPress Version:** 5.2.3

### DIFF
--- a/src/Router.php
+++ b/src/Router.php
@@ -116,11 +116,24 @@ class Router {
 	 * @return boolean
 	 */
 	public static function is_graphql_request() {
-		if ( empty( $GLOBALS['wp']->query_vars ) || ! is_array( $GLOBALS['wp']->query_vars ) || ! array_key_exists( self::$route, $GLOBALS['wp']->query_vars ) ) {
-			return false;
+		// If 'init' fired, check query vars.
+		if ( did_action( 'init' ) || doing_action( 'init' ) ) {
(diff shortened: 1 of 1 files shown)


esamattis:
I don't think we'll need the "after init" check if we already have a check that works both before and after init. If we have both then we have a possibility for a bug where it returns different results before and after init.

kidunot89:
True, but I wasn't sure if using `$_SERVER['REQUEST_URI']` was the best practice for this, so I figured it would be better to only use it when the query bars aren't available.

jasonbahl:
Perhaps ya'll just need to check `if ( defined( 'GRAPHQL_REQUEST ) && true === GRAPHQL_REQUEST ) )`?

The Router is only used if you're using GraphQL at the `/graphql` endpoint, which isn't the case for all GraphQL requests. 

Do we need a more general `WPGraphQL::is_graphql_request()` method instead of one _just_ in the Router?

jasonbahl:
> Perhaps ya'll just need to check if ( defined( 'GRAPHQL_REQUEST ) && true === GRAPHQL_REQUEST ) )?

I guess that's also not defined before init. 

I think more context would go a long way for me.

kidunot89:
Just realized the conditional is invalid. I'll need to make some quick changes.

jasonbahl:
@kidunot89 @esamattis I'll merge this. I think there's value to it. I added a note for context explaining the purpose and that if you use this to check conditions, it will only work for HTTP requests, not _all_ requests such as internal `graphql()` calls.

PR #1027, opened 2019-11-19
Feature/#1024 graphql request condition

### LINKED ISSUE #1024: `GRAPHQL_REQUEST` should be dynamic
Opening new issue from https://github.com/wp-graphql/wp-graphql/issues/1021#issuecomment-555159476

There are multiple places where wp-graphql specific filters are active when the `GRAPHQL_REQUEST` is set to true.

For example in Config.php:

https://github.com/wp-graphql/wp-graphql/blob/4855753012815b13b4eb002e22d6ad407703ebfc/src/Data/Config.php#L250

https://github.com/wp-graphql/wp-graphql/blob/4855753012815b13b4eb002e22d6ad407703ebfc/src/Data/Config.php#L298

https://github.com/wp-graphql/wp-graphql/blob/4855753012815b13b4eb002e22d6ad407703ebfc/src/Data/Config.php#L223

https://github.com/wp-graphql/wp-graphql/blob/4855753012815b13b4eb002e22d6ad407703ebfc/src/Data/Config.php#L198

https://github.com/wp-graphql/wp-graphql/blob/4855753012815b13b4eb002e22d6ad407703ebfc/src/Data/Config.php#L179

https://github.com/wp-graphql/wp-graphql/blob/4855753012815b13b4eb002e22d6ad407703ebfc/src/Data/Config.php#L155

So when you do a graphql request from PHP using `graphql()` the constant `GRAPHQL_REQUEST` gets set to true and there for those filters are active even after the `graphql()` call which will break or change some normal WordPress functionality. 

`GRAPHQL_REQUEST` needs to be changed to something that can be set to false once graphql request has completed. Maybe `\WPGraphQL::is_graphql_http_request()` and `\WPGraphQL::is_graphql_request()`?

jasonbahl:
Introduces `is_graphql_request()` method which can be used to determine if a GraphQL request is in process. 

This will return true while a GraphQL request is executing, and is reset to false once execution is complete. 

This means conditions that filter WP_Query, etc can be targeted more narrowly to _just_ affect GraphQL, but _not_ affect queries outside of GraphQL requests. 

closes #1024


access-functions.php
```diff
@@ -202,3 +202,12 @@ function register_graphql_mutation( $mutation_name, $config ) {
 		$type_registry->register_mutation( $mutation_name, $config );
 	} );
 }
+
+/**
+ * Whether a GraphQL request is in action or not. Default false.
+ *
+ * @since 0.4.1
+ */
+function is_graphql_request() {
```
esamattis: Should `is_http_graphql_request()` to be exposed too? Although `is_graphql_request` should be generally preferred.

jasonbahl: Ah yes. I think both make sense to have. I do think there is value in being able to determine both cases

jasonbahl: added

PR #1056, opened 2019-12-03
Release/v0.5.0

### LINKED ISSUE #974: Revisions should be the same type as what they are revising

jasonbahl:
# Release Notes

## Breaking Changes

**This release is a breaking change release.**

The focus of this release is on how Revisions are handled.

Prior to this release, Revisions are addd to the WPGraphQL Schema as their own `Revision` Type. This largely mimics WordPress implementation, in that revisions are a different Post Type. However, this doesn't lend itself well to _actual_ use.

Revisions of a Post, declaratively speaking, are still Posts. If I paint my house, it's still a house, just with different properties. If I change the title of a Post, it's still a Post, it should be considered as much in the Graph.

The intent of revisions is to show a thing (Post, Page, etc) as it exists over time.

The `Revision` Type declares that the thing is in fact, a different thing, and the utility of Revisions is compromised.

By changing Revisions in the WPGraphQL Schema to be of the same Type of the object being revised, clients can actually represent their objects as they've been revised over time.

## Upgrading
The `Revision` Type no longer exists. If you were querying for Revisions your queries will need to be updated as the Type being returned will no longer be a `Revision`. 

### Querying Root Revisions

When querying revisions from the Root, the Type returned to the connection is a `ContentRevisionUnion`, which is comprised of any Types that support revisions (for example, Post and Page, or any Post Type that has registered support for `revisions`)

You can _now_ query revisions from the Root like so:

```
{
  revisions {
    nodes {
      __typename
      ... on Post {
        id
        title
        postId
      }
      ... on Page {
        id
        title
        pageId
      }
    }
  }
}
```

and you would get results similar to the following (dependent on your data set): 

```
{
  "data": {
    "revisions": {
      "nodes": [
        {
          "__typename": "Post",
          "id": "cmV2aXNpb246MTcxNQ==",
          "title": "Another Test"
          "postId":1715
        },
       {
          "__typename": "Page",
          "id": "cmV2aXNpb246MTY0Mg==",
          "title": "Test Page Revision",
          "pageId": 1642,
        }
        {
          "__typename": "Post",
          "id": "cmV2aXNpb246MTcxNA==",
          "title": "Test revision",
          "postId": 1543
        }
      ]
    }
  },
}
```

### Extending Revision Types

If you were extending the `Revision` Type, you will need to consider alternatives for extending the Schema for your use case. 

Querying revisions of a Post will return a connection to Posts. The WordPress database will still fetch posts of the "revision" post type, but the GraphQL Schema will represent the revisions of a Post as a Post. 

Querying revisions of a Page will return a connection to Pages. The WordPress database will still fetch posts of the "revision" post type, but the GraphQL Schema will represent the revisions of Pages as a Page.

Same goes for Custom Post Types.

Since revisions of a Post are always a Post, the connection is _not_ a Union as it is when querying Revisions from the RootQuery, so we can query like so (of course this will vary based on your data set):

```
{
  post(id: "cG9zdDoxNzEz") {
    __typename
    id
    title
    date
    revisions {
      nodes {
        __typename
        id
        title
        date
      }
    }
  }
}
```

and you might get results like so: 

```
{
  "data": {
    "post": {
      "__typename": "Post",
      "id": "cG9zdDoxNzEz",
      "title": "Test",
      "date": "2019-11-30T17:32:39",
      "revisions": {
        "nodes": [
          {
            "__typename": "Post",
            "id": "cmV2aXNpb246MTcxNQ==",
            "title": "Test",
            "date": "2019-11-30T17:33:10"
          },
          {
            "__typename": "Post",
            "id": "cmV2aXNpb246MTcxNA==",
            "title": "Test",
            "date": "2019-11-30T17:32:20"
          }
        ]
      }
    }
  }
```

Because Revisions of a Post are now represented as a Post, we can make use of some great GraphQL features, such as Query Fragments. We could refactor the above query like below, using the same fragment for the primary Post and the revisions of it, allowing for the same components to be used to render a Post and a revision of a Post. Powerful stuff!: 

```
{
  post(id: "cG9zdDoxNzEz") {
    __typename
    ...PostFields
    revisions {
      nodes {
        __typename
        ...PostFields
      }
    }
  }
}

fragment PostFields on Post {
  id
  title
  date
}
```

### Does this close any currently open issues?
closes #1017


jasonbahl:
Just checked, and this is working great with WPGraphQL for ACF! 

Since Advanced Custom Fields revises meta, we can now query for revisions and ACF fields come along for the ride for "free"

Screen Shot 2019-12-03 at 9 03 55 AM

TylerBarnes:
This is such an exciting release! Can't wait to add preview support for Gatsby!! Awesome work!

PR #1060, opened 2019-12-06
#944 - Query Multiple Post Types in PostObjectConnectionResolver

### LINKED ISSUE #944: Support querying multiple post types in PostObjectConnectionResolver
Currently, the `PostObjectConnectionResolver` is setup to expect to query just one post_type at a time, as seen here: 

https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Connection/PostObjectConnectionResolver.php#L44

You're expected to pass a post_type to the Class, which is set, then used to do validation, such as: 

https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Connection/PostObjectConnectionResolver.php#L91

The problem with this, is that for querying multiple post_types, say, for a `search` query, or another custom query that would return a `PostObjectUnion`, the resolver needs to be able to query many post types.

----

### Possible solution

change: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Connection/PostObjectConnectionResolver.php#L44

To set the value as an array.

Then:

When validating like so: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Connection/PostObjectConnectionResolver.php#L91

We could do something like: `if ( in_array( 'revision', $this->post_types ) ) {` and if `revision` is the _only_ post_type, we can prevent execution. If it's one of many, and the user doesn't have caps for it, we can just unset it from the array.

It gets tricky when trying to mix post_types here though: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Connection/PostObjectConnectionResolver.php#L177

Because images and revisions have a different status than most post_types. Will need to explore more with this one.

jasonbahl:
This allows for querying multiple post types when using the PostObjectConnectionResolver (as long as those post types are not `revision` or `attachment`, you must still query them individually, at least for the time being)

closes #944


src/Data/Connection/PostObjectConnectionResolver.php
```diff
@@ -41,7 +41,14 @@ public function __construct( $source, $args, $context, $info, $post_type ) {
 		/**
 		 * Set the post type for the resolver
 		 */
-		$this->post_type = $post_type;
+		if ( 'revision' === $post_type || 'attachment' === $post_type ) {
+			$this->post_type = $post_type;
+		} else {
+			$post_type = is_array( $post_type ) ? $post_type : [ $post_type ];
+			unset(  $post_type['attachment'] );
```
CodeProKid: extra space here.

src/Data/Connection/PostObjectConnectionResolver.php
```diff
@@ -41,7 +41,14 @@ public function __construct( $source, $args, $context, $info, $post_type ) {
 		/**
 		 * Set the post type for the resolver
 		 */
-		$this->post_type = $post_type;
+		if ( 'revision' === $post_type || 'attachment' === $post_type ) {
+			$this->post_type = $post_type;
```
CodeProKid: I'm confused... wouldn't you _not_ want to set this as the `post_type` for these two post types?

src/Data/Connection/PostObjectConnectionResolver.php
```diff
@@ -41,7 +41,14 @@ public function __construct( $source, $args, $context, $info, $post_type ) {
...
+		} else {
+			$post_type = is_array( $post_type ) ? $post_type : [ $post_type ];
+			unset(  $post_type['attachment'] );
+			unset( $post_type['revision'] );
+			$this->post_type = $post_type;
+		}
```
CodeProKid: An inline comment for this whole block would be good, it's a little confusing for why this is here.

src/Data/Connection/PostObjectConnectionResolver.php
jasonbahl: @CodeProKid no, you can still query these, but not as an array. 

For example, the `revisions` connections use this resolver and set the `post_type => 'revision'` so connections to singular Post Types, such as `revision` or `attachment` should still work. 

However, queries to plural post types should not include `revisions` or `attachments` as they're rules for resolving are different (they're always 'inherit' status).

src/Data/Connection/PostObjectConnectionResolver.php
jasonbahl: good call

src/Data/Connection/PostObjectConnectionResolver.php
CodeProKid: Ahhhh, OK. I get the logic here now. Such nuance. 🐶

jasonbahl: ya, I added a comment. It is kinda weird, but such is life when it comes to how WordPress handles a lot of things 😆

PR #1061, opened 2019-12-06
#1059 - postBy entry throws error if post can't be found

### LINKED ISSUE #1059: `postBy` and similar queries give an error when no post is found
The `postBy` and similar queries throw an error when no post is found.

Example query:
```
{
  postBy(uri: "foo-bar") {
    id
    title
  }
}
```

Example response:
```
{
  "errors": [
    {
      "message": "No post exists with this id: ",
      "category": "user",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "postBy"
      ]
    }
  ],
  "data": {
    "postBy": null
  }
}
```

Since nothing really went wrong here, I would not expect an error to be thrown; just a `null` response in `data.postBy`. While the error message is helpful, it ultimately leads to promises being falsely rejected.

Tested with WPGraphQL version 0.4.2

jasonbahl:
This updates the `{$postType}By` entry points to return null instead of throwing an error when a Post cannot be found. 

This makes it more consistent with other singe entry points in the Schema. 

There might be future follow-up as it seems like many APIs do indeed throw errors when individual resources cannot be found. See #1059 for more context on the discussion.

closes #1059

### DIFF
--- a/tests/wpunit/PostObjectQueriesTest.php
+++ b/tests/wpunit/PostObjectQueriesTest.php
@@ -1174,7 +1174,8 @@ class PostObjectQueriesTest extends \Codeception\TestCase\WPTestCase {
 		/**
 		 * This should return an error as we tried to query for a post using a Page ID
 		 */
-		$this->assertArrayHasKey( 'errors', $actual );
+		$this->assertArrayNotHasKey( 'errors', $actual );
+		$this->assertNull( $actual['data']['postBy'] );
 
 	}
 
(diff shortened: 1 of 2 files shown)


tests/wpunit/PostObjectQueriesTest.php
```diff
@@ -1174,7 +1174,8 @@ public function testPostByQueryWithIDForADifferentType() {
 		/**
 		 * This should return an error as we tried to query for a post using a Page ID
```
CodeProKid: Should update this comment since it should now _not_ throw an error.

jasonbahl: Good call. Thanks!

kidunot89: @CodeProKid @jasonbahl Why is bad to throw an error here? If you just return `null` the user won't know what went wrong.

PR #1068, opened 2019-12-11
Add simple tests for `is_graphql_http_request()`

esamattis:
You merged my PR too quickly 😂  

Anyway here's some very basic tests for `is_graphql_http_request()`. Codeception seem to automatically clear the `$_SERVER` and `$_GET` variables between the tests so it is safe to modify them like this.


jasonbahl:
haha! Thanks for the follow-up PR. Appreciate the tests very much!

PR #1072, opened 2019-12-12
WIP (DO NOT MERGE): Feature/#1066 content node interface

jasonbahl:
This adds a `ContentNode` Interface and has all PostObjectTypes (Page, Post, etc) implement it. 

The `ContentNode` interface contains fields common to _all_ Post Type Objects, and any PostTypeObject implementing the interface will inherit said fields. 

For Post Types that do not support fields, such as `content`, `title`, etc (see: https://developer.wordpress.org/reference/functions/post_type_supports/) those fields will be labeled as `deprecated` in the Schema, so they won't show in Docs for said Type, but they will still execute. 

This allows us to have a cleanly Documented Schema, but also allows for common fields to be queried across Types without breaking things. 

For example, if I were to register a post type and _only_ register support for the `title` field, the Docs for that Type in the GraphQL Schema should _not_ show fields such as `content`, `excerpt`, etc because the post type doesn't support them. The admin won't show them and users won't save data to them. 

However, we want these fields to be able to be queried as common fields across all types, so that common fallback templates can be created. 

For example, in a core WordPress theme, you would have an index.php theme, and it would output the title, content, date and author, even if the Post Type doesn't "support" those fields. It would render as empty, but it wouldn't error out trying to `get_the_content()`, etc. 

For WPGraphQL to allow fallback templates to be created, we need to be able to allow common fields to be queried, even if not supported by a Type. 

So, we should be able to query like so: 

```
{
  contentNodes {
    nodes {
      __typename
      id
      title
      date
      content
      ...on Post {
         somePostOnlyField
      }
      ...on Page {
         somePageOnlyField
      }
      ...on CustomType {
         someCustomTypeField
      }
    }
  }
}
```

If we were to only add fields such as `content` to Types that support content, then a query like this would error, because we're querying for `content` on _all_ Types of content nodes. 

By allowing the `content` field to be part of the `ContentNode` interface, but listing it as `deprecated` for Types that don't support it allows for the Schema Docs to be clean, but for execution of cross-type queries like this to still succeed.


jasonbahl:
@renatonascalves ya they haven’t been written yet. Opened this intentionally as a WIP to make sure folks had a chance to take a look in case it should go in a different directions. Tests coming.

jasonbahl:
> This would be a great time remove the postTypeBy query and add its functionality postType query.

@kidunot89 I agree! Will add that to the docket

jasonbahl:
@kidunot89 a thing of beauty. . .deprecate the old way so it's not in the docs, but still works, introduce the new way: 

Screen Shot 2019-12-17 at 2 50 49 PM

jasonbahl:
Closing in favor of: https://github.com/wp-graphql/wp-graphql/pull/1086

PR #1084, opened 2019-12-18
adds support for post_format as a taxonomy

### LINKED ISSUE #1083: Support needed for post formats
The `post_format` taxonomy is not currently supported as a taxonomy type in wp-graphql. Currently in order to support them, you need to use the `register_taxonomy_args` filter [as if they were custom taxonomies](https://docs.wpgraphql.com/getting-started/custom-taxonomies/#filtering-existing-taxonomies). But they're a core part of posts and should be supported.

aberkow:
This helps resolve #1083 to allow `post_format` as a taxonomy type.

#1083

### DIFF
--- a/wp-graphql.php
+++ b/wp-graphql.php
@@ -402,8 +402,8 @@ if ( ! class_exists( 'WPGraphQL' ) ) :
 			// Adds GraphQL support for post formats.
 			if ( isset( $wp_taxonomies['post_format'] ) ) {
 				$wp_taxonomies['post_format']->show_in_graphql = true;
-				$wp_taxonomies['post_format']->graphql_single_name = 'PostFormat';
-				$wp_taxonomies['post_format']->graphql_plural_name = 'PostFormats';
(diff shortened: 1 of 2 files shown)


wp-graphql.php
```diff
@@ -398,6 +398,13 @@ public static function show_in_graphql() {
 				$wp_taxonomies['post_tag']->graphql_single_name = 'tag';
 				$wp_taxonomies['post_tag']->graphql_plural_name = 'tags';
 			}
+
+			// Adds GraphQL support for post formats.
+			if ( isset( $wp_taxonomies['post_format'] ) ) {
+				$wp_taxonomies['post_format']->show_in_graphql = true;
+				$wp_taxonomies['post_format']->graphql_single_name = 'PostFormat';
```
esamattis: Probably should start with lower case `p` to match the convention on other taxonomies. Ex. `mediaItem` & `mediaItems`

aberkow: you're absolutely right. I'll fix that now

jasonbahl:
Thanks @aberkow!

Issue #263, opened 2017-10-16
Get Single Nodes (posts, terms, users, etc) by more than ID


jasonbahl:
We should be able to fetch a single term by it's slug, possibly other attributes?

Similar to how we can now get post objects by URI, ID and Global ID.

CodeProKid:
We should also think about what other objects we should be able to get by something other than ID. I think Users would probably be another good candidate. The way I'm thinking about it, is anything that would have it's own route in an SPA should be able to be retrieved by slug.

jasonbahl:
@CodeProKid agreed. Which also makes me think the singular queries for nodes just need to be more flexible. Right now we have `post( id: "someId" )` or `postBy()` which has the following args: `id`, `postId`, `slug`, `uri`. . .

Ideally, I'd like one singular entry point per node, instead of post(), and postBy(), especially if we're going to add it for each type of node. . .might be best to make a breaking change and make the existing singular entries more flexible where `id` is not a required input `$arg`. 

I think it would be good to instead support a single `where` field, similar to how we do connections, and can nest other inputs on it. . .like: 

```
query getNodes ( $postWhere: postWhere!, $userWhere: userWhere ) {  
  post( where: $postWhere ) {
    id
    title
  }
  user ( where: $userWhere ) {
    id
    firstName
    lastName
  }
}
```

```
$variables = {
  'postWhere': {
    id: null,
    slug: "some-post",
    uri: null,
    postId: null
  },
  'userWhere': {
      id: null,
      username: null,
      email: "email@something.com"
  }
};
```

This shape brings some consistency to the shape of the `args` for singular nodes and connections now, and it gives more flexibility to using variables individually or in combination to narrow the resulting node, and more flexibility to plugins to extend the shape with their own fields. 

I imagine all sorts of ways folks would want to get a users (or other nodes) by some unique value, like `user(where: {wooCommerceOrderNumber: 123} )` or something to that tune.  

Changing the shape of the root singular entries will be a breaking change, but I think now is the time to do it. 

Also, the `postBy()` field only makes semantic sense if there are no additional fields in the schema. . .for example, posts _can_ have passwords, so we will need to add a `password` input arg, and `postBy( password: "123", slug: "some-slug")` isn't as nice (to me) as `post( where: { slug: "some-slug" }, password: "123456" )`

chrisfullman:
Definitely looking forward to this one! I agree that it feels like a good time to introduce this breaking change ahead of 1.0.

[...]

PR #1086, opened 2019-12-19
Release/next

jasonbahl:
This release is focusing on implementing Interface support. 

More info to come. 

See: 
- LOTS of info here: https://github.com/wp-graphql/wp-graphql/issues/263#issuecomment-570091131 
- https://github.com/wp-graphql/wp-graphql/pull/1072
- https://github.com/wp-graphql/wp-graphql/milestone/20

----
Below is copied from my comment on Issue #263 
----

The next release I'm working on (#1086) addresses this (the ability to fetch single nodes by various forms of unique identifiers).

## Posts
We're deprecating the `$postType.'By'` entry points, and have added a new `idType` field and enum for single entry points. 

This means queries such as the following will still work, but will not show in the Schema documentation, and may be formally removed from the codebase at a later date:

### Deprecated Queries

```
{
  postBy( slug: "some-slug" ) {
    id
    title
  }
}
```
```
{
  pageBy( uri: "some-uri" ) {
    id
    title
  }
}
```

These are deprecated in favor of existing single entry points for each post type, and the newly added `idType` field on those entry points.

### New Singular Post Object Queries

With the added `idType` field on the singular post object entry points, below is a series of example queries now made possible:

#### Query Single Post By Database ID

```
{
  post(id: 1739, idType: DATABASE_ID) {
    id
    title
    uri
    slug
    postId
  }
}
```

Screen Shot 2019-12-31 at 3 34 28 PM

#### Query Single Post by URI

```
{
  post(id: "/test-5/", idType: URI) {
    id
    title
    uri
    slug
    postId
  }
}
```

Screen Shot 2019-12-31 at 3 35 31 PM

#### Query Single Post by Global Hashed ID

```
{
  post(id: "cG9zdDoxNzM5", idType: ID) {
    id
    title
    uri
    slug
    postId
  }
}
```
Screen Shot 2019-12-31 at 3 34 06 PM

or without specifying `idType` as the default is the hashed ID

```
{
  post(id: "cG9zdDoxNzM5") {
    id
    title
    uri
    slug
  }
}
```

Screen Shot 2019-12-31 at 3 33 53 PM

#### Query Post (non-hierarchical post type) by Slug

```
{
  post(id: "test-5", idType: SLUG) {
    id
    title
    uri
    slug
  }
}
```

Screen Shot 2019-12-31 at 3 36 46 PM

### ContentNodeInterface
This release also introduces the `ContentNode` Interface that is implemented by all Post Object types (posts, pages, custom post types). 

This Interface allows for new types of queries to be executed that span many post types. 

#### Content Nodes Connection
Here's an example of a new query: 

```
{
  contentNodes {
    nodes {
      __typename
      id
      title
      link
      uri
      isRevision
      ... on Page {
        isFrontPage
      }
    }
  }
}
```

Screen Shot 2019-12-31 at 3 49 33 PM

In the results above, we can see that a mix of Posts and Pages have been returned.

We can also pass arguments to do things like search across multiple post types: 

```
{
  contentNodes(where: {search: "test"}) {
    nodes {
      __typename
      id
      title
      link
      uri
      isRevision
      ... on Page {
        isFrontPage
      }
    }
  }
}
```

Screen Shot 2019-12-31 at 3 53 04 PM

#### ContentNode single entry point

We can now also fetch single Content (post) nodes like so: 

```
{
  contentNode(id: "cG9zdDoxNzM5") {
    __typename
    id
    title
    link
    uri
    isRevision
    ... on Page {
      isFrontPage
    }
  }
}
```

Screen Shot 2019-12-31 at 3 59 15 PM

But, we can fetch single nodes of any Post type by other unique identifiers as well, such as the URI: 

```
{
  contentNode(id: "/2019/12/05/test-5/", idType: URI) {
    __typename
    id
    title
    link
    uri
    isRevision
    ... on Page {
      isFrontPage
    }
  }
}
```
Screen Shot 2019-12-31 at 4 02 25 PM

or

```
{
  contentNode(id: "/test/", idType: URI) {
    __typename
    id
    title
    link
    uri
    isRevision
    ... on Page {
      isFrontPage
    }
  }
}
```

Screen Shot 2019-12-31 at 4 02 38 PM

This is a SUPER powerful feature as it allows for fragments to be more easily shared and re-used across queries and components.

## Terms 

Prior to this release, there wasn't really great ways of fetching singular term objects. You could do a query for a connection of terms and limit it to the `first: 1`, and in some cases that worked fine, but it wasn't always ideal. 

This release adds better support for fetching individual Term Nodes. 

### Example Term Node Queries

Instead of only fetching by the hashed global ID, we can now fetch by other unique identifiers using the `idType` field.

#### Query Tag by Hashed Global ID

```
{
  tag(id: "cG9zdF90YWc6Mw==") {
    id
    name
    slug
    tagId
  }
}
```

Screen Shot 2019-12-31 at 4 09 37 PM

#### Query Tag by Database ID

```
{
  tag(id: 3, idType: DATABASE_ID) {
    id
    name
    slug
    tagId
  }
}
```

Screen Shot 2019-12-31 at 4 12 14 PM

#### Query Tag by Name

```
{
  tag(id: "Another Test", idType: NAME) {
    id
    name
    slug
    tagId
  }
}
```

Screen Shot 2019-12-31 at 4 12 47 PM

#### Query Tag by Slug

```
{
  tag(id: "another-test", idType: SLUG) {
    id
    name
    slug
    tagId
  }
}
```

Screen Shot 2019-12-31 at 4 14 06 PM

#### Query Tag by URI

```
{
  tag(id: "tag/another-test/", idType: URI) {
    id
    name
    slug
    tagId
    uri
  }
}
```

Screen Shot 2019-12-31 at 4 14 46 PM

### TermNode Interface

This release also introduces the TermNode interface, which allows for querying terms across Taxonomy types. 

For example: 

#### Query TermNodes Connection

```
{
  terms {
    nodes {
      id
      __typename
      name
      uri
    }
  }
}
```
Screen Shot 2019-12-31 at 4 22 31 PM

This enables us to do things, like search terms across multiple taxonomies and get both Categories and Tags (and custom taxonomies) back in the same query: 

```
{
  terms(where: {search: "test"}) {
    nodes {
      id
      __typename
      name
      uri
      ... on Tag {
        tagId
      }
      ... on Category {
        categoryId
      }
    }
  }
}
```

Screen Shot 2019-12-31 at 4 24 38 PM

### Query Single Terms by Unique Identifiers
We can also query single term nodes by various identifiers, such as `Database ID`, `ID`, `Name`, `Slug` or `URI`: 

#### Query Single Term Node by (hashed) ID

```
{
  termNode(id: "cG9zdF90YWc6Mw==") {
    __typename
    id
    name
    link
    slug
    uri
  }
}
```

Screen Shot 2019-12-31 at 4 28 01 PM

#### Query Single Term Node by Database ID

```
{
  termNode(id: 3, idType: DATABASE_ID) {
    __typename
    id
    name
    link
    slug
    uri
    databaseId
  }
}
```

Screen Shot 2020-01-01 at 3 53 24 PM

#### Query Single Term Node by Name

```
{
  termNode(id: "Another Test", idType: NAME, taxonomy: TAG) {
    __typename
    id
    name
    link
    slug
    uri
    databaseId
  }
}
```
Screen Shot 2020-01-01 at 3 53 24 PM

#### Query Single Term Node by Slug

```
{
  termNode(id: "another-test", idType: SLUG, taxonomy: TAG) {
    __typename
    id
    name
    link
    slug
    uri
    databaseId
  }
}
```

Screen Shot 2020-01-01 at 3 56 32 PM

#### Query Single Term Node by URI

```
{
  termNode(id: "tag/another-test/", idType: URI) {
    __typename
    id
    name
    link
    slug
    uri
    databaseId
  }
}
```

Screen Shot 2020-01-01 at 3 57 44 PM

## Node By URI

Another thing this release provides is a way to get any node by URI, across types. This comes in handy because in many cases when a URI is provided, the type of content it is isn't always known ahead of time. By having an entry point for _all_ Types that implement the 1UniformResourceIdentifiable` Interface, we can create a main entry point for entire applications. 

Below is an example. This is showing many queries at once, but this could easily be one query with fragments for each Type, controlled by their appropriate component for rendering said type of content:

```
{
  page: nodeByUri(uri: "about/") {
    ...URI
  }
  post: nodeByUri(uri: "2019/12/05/test-5/") {
    ...URI
  }
  tag: nodeByUri(uri: "tag/8bit/") {
    ...URI
  }
  category: nodeByUri(uri: "category/alignment/") {
    ...URI
  }
  user: nodeByUri(uri: "author/jasonbahl/") {
    ...URI
  }
}

fragment URI on UniformResourceIdentifiable {
  __typename
  ... on Page {
    pageId
  }
  ... on Post {
    postId
  }
  ... on Category {
    categoryId
  }
  ... on Tag {
    tagId
  }
  ... on User {
    userId
  }
}
```

Screen Shot 2020-01-01 at 4 05 25 PM


src/Type/Object/PostObject.php
```diff
@@ -594,6 +209,85 @@ function() use ( $context, $ancestor_ids ) {
...
+
+		if ( ! post_type_supports( $post_type_object->name, 'editor' ) ) {
+			$fields['content']['isDeprecated']      = true;
+			$fields['content']['deprecationReason'] = __( 'This content type does not support the content editor', 'wp-graphql' );
+		}
+
+		if ( ! post_type_supports( $post_type_object->name, 'author' ) ) {
+			$fields['author']['isDeprecated']      = true;
+			$fields['author']['deprecationReason'] = __( 'This content type does not support authors', 'wp-graphql' );
+
```
CodeProKid: Extra space here.

src/Type/Object/RootQuery.php
```diff
@@ -177,27 +177,89 @@ public static function register_post_object_fields() {
...
+									$post    = DataSource::resolve_post_object( $post_id, $context );
+									break;
+								case 'global_id':
+								default:
+									$id_components = Relay::fromGlobalId( $args['id'] );
+									if ( ! isset( $id_components['id'] ) || ! absint( $id_components['id'] ) ) {
+										throw new UserError( __( 'The ID input is invalid. Make sure you set the proper idType for your input.', 'wp-graphql' ) );
+									}
+									$post_id = absint( $id_components['id'] );
+									$post    = DataSource::resolve_post_object( $post_id, $context );
```
CodeProKid: Can move this outside of the switch statement so it only has to be done once.

src/Type/Union/ContentRevisionUnion.php
```diff
@@ -7,35 +7,40 @@
...
-
-					$type = 'Post';
-					$parent = get_post( $object->parentId );
-					if ( ! empty( $parent ) && isset( $parent->post_type ) ) {
-						$parent_post_type_object = get_post_type_object( $parent->post_type );
-						if ( isset( $parent_post_type_object->graphql_single_name ) ) {
-							$type = $type_registry->get_type( $parent_post_type_object->graphql_single_name );
+			$type_names = array_map(
+				function( $post_type ) {
+						$post_type_object = get_post_type_object( $post_type );
```
CodeProKid: extra indentation here.

src/Type/WPObjectType.php
```diff
@@ -227,4 +232,17 @@ public function prepare_fields( $fields, $type_name, $config ) {
...
 
+	/**
+	 * Simple getter function so "$type_registry" can be used in hooks
+	 *
+	 * @param string $name  Member name.
+	 * @return TypeRegistry|null
+	 */
+	public function __get( $name ) {
+		if ( 'type_registry' === $name ) {
+			return $this->type_registry;
```
CodeProKid: I don't think this will ever get invoked. The magic method only gets called if there's not a matching method / property in the class, but you're returning a property with the same name you're checking against here. Unless the property is private, but in that case I would just change the property visibility instead of doing this.

src/Utils/Utils.php
```diff
@@ -0,0 +1,75 @@
...
+	 * Maps new input query args and sanitizes the input
+	 *
+	 * @param array $args The raw query args from the GraphQL query
+	 * @param array $map  The mapping of where each of the args should go
+	 *
+	 * @since  0.5.0
+	 * @return array
+	 * @access public
+	 */
+	public static function map_input( $args, $map ) {
```
CodeProKid: I think we should approach this a little differently where we recursively call sanitization methods so we can sanitize multiple levels of input. Currently, this only supports a single level of input, and only sanitizes strings. Another issue is that we're doing the type checking based on the incoming value rather than config. So if an input is expecting an integer, but a client sends a number typeset as a string, it will fail. Instead we could look up what type the arg is _supposed_ to be from config, and then sanitize the value based off that.

src/Type/InterfaceType/Uri.php
```diff
@@ -0,0 +1,19 @@
+<?php
+namespace WPGraphQL\Type\InterfaceType;
+
+class Uri {
+	public static function register_type() {
+		register_graphql_interface_type(
+			'Uri',
```
CodeProKid: I'm not sure how I feel about `Uri` as an interface name. Seems a little nebulous to me as to what it's trying to define. Perhaps something like `UriObject` or even `SluggableObject` / `ObjectWithUri`. I just think `Uri` sounds too much like a field name than an interface.

src/Connection/PostObjects.php
```diff
@@ -359,20 +316,163 @@ public static function get_connection_args( $args = [] ) {
...
 				'type'        => 'MimeTypeEnum',
 				'description' => __( 'Get objects with a specific mimeType property', 'wp-graphql' ),
 			],
-		], $args );
+		];
+
+		/**
+		 * If the connection is to a single post type
+		 */
+		if ( isset( $post_type_object ) && $post_type_object instanceof \WP_Post_Type ) {
```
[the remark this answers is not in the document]
jasonbahl: They're not being re-defined here. This is the only definition of them. I moved them here to be applied only on connections to singular post types, and _not_ applied to connections that go to many post types.

I've added a comment with a lot of context.

src/Registry/TypeRegistry.php
```diff
@@ -335,7 +341,10 @@ function( TypeRegistry $type_registry ) {
 				 */
 				do_action( 'graphql_register_types_late', $type_registry );
 
-			}, 1, 1 );
+			},
+			1,
+			1
```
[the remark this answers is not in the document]
jasonbahl: > omg, this makes my eyes want to bleed

😂 

Fixed.

src/Type/InterfaceType/ContentNode.php
```diff
@@ -0,0 +1,436 @@
...
+						],
+						'resolve'     => function( $source, $args ) {
+							if ( isset( $args['format'] ) && 'raw' === $args['format'] ) {
+								// @codingStandardsIgnoreLine.
+								return $source->contentRaw;
+							}
+
+							// @codingStandardsIgnoreLine.
+							return $source->contentRendered;
+						},
```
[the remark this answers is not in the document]
jasonbahl: @CodeProKid so, if a Type implements an interface, the resolver of the field defined at the interface level will be invoked unless the Type defines a resolver for a field, and in that case the Type resolver will override the resolver of the Interface. 

This allows us to define common resolvers in the Interface level, but specific (overridden) resolvers at the Type level.

src/Type/InterfaceType/ContentNode.php
```diff
@@ -0,0 +1,436 @@
...
+						$post_type_object = get_post_type_object( $post->post_type );
+						if ( isset( $post_type_object->graphql_single_name ) ) {
+							$type = $type_registry->get_type( $post_type_object->graphql_single_name );
+						}
+					}
+
+					return ! empty( $type ) ? $type : null;
+
+				},
+				'fields'      => [
```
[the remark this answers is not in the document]
jasonbahl: 🤔 I'll have to think through this a bit. At the moment we don't really have multiple interfaces implementing the same fields (other than this one and the Node interface both declaring the "id" field), so sharing fields across many interfaces and types hasn't come up as a pattern to look for and abstract, yet. 

But I can see it becoming maybe more relevant the more we implement interfaces. . .the GraphQL Spec is looking at introducing Interfaces that can implement other Interfaces though, so that might solve the issue of sharing fields across many interfaces too.

We could even provide that functionality on our own via the registry, so we wouldn't have to re-define fields. 

Like the `ContentNode` Interface could `implement` the `Node` Interface which would add the ID field to it so that we didn't have to re-define the `ID` field again for `contentNode`

🤔

src/Type/WPObjectType.php
jasonbahl: @kidunot89 added this to address some things he filters for WPGraphQL for WooCommerce. 

the property has changed to `public` so we can now get it without a method like this anyway.

src/Utils/Utils.php
jasonbahl: Probably worth a different issue for addressing this. This function itself didn't change here, just `phpcbf` formatting, I believe

src/Type/InterfaceType/Uri.php
jasonbahl: Naming things is tough indeed.

It's ok to have fields and Interfaces with the same name. The `node` field and `Node` Interface are great examples of this. 

----

After researching a bit, I found that Github has an interface named `UniformResourceLocatable` with 2 fields: `resourcePath` and `url` which are of a custom `URI` Scalar type, described as: `An RFC 3986, RFC 3987, and RFC 6570 (level 4) compliant URI string.`

See below: 

Screen Shot 2019-12-30 at 4 36 36 PM

jasonbahl: @CodeProKid ^ In that Spirit, I think perhaps we can name it `UniformResourceIdentifier` interface?

jasonbahl: Update to `UniformResourceIdentifiable`

jasonbahl:
Closing in favor of #1112 (which is properly opened against Master)

PR #1098, opened 2019-12-30
Feature/#263 get single nodes by uri

### LINKED ISSUE #263: Get Single Nodes (posts, terms, users, etc) by more than ID

jasonbahl:
This adds nodeByUri field to the RootQuery allowing for any node that implements the URI interface to be resolved from a single entry point. 

See the following example, where we use the same field to query a Tag, Category, Page, Post and User: 

Screen Shot 2019-12-23 at 4 57 09 PM

closes #263


jasonbahl:
I have some cleanup I want to do still, but wanted to get this open so folks could check it out.

Issue #594, opened 2018-12-03
hasNextPage / hasPreviousPage not respecting the Relay spec fully


jasonbahl:
The GraphQL Relay spec states the following for `hasPreviousPage`:

HasPreviousPage(allEdges, before, after, first, last)
- If last is set:
   - Let edges be the result of calling ApplyCursorsToEdges(allEdges, before, after).
   - If edges contains more than last elements return true, otherwise false.
- If after is set:
   - If the server can efficiently determine that elements exist prior to after, return true.
- Return false.

## Currently, WPGraphQL is skipping:

```
- If after is set:
   - If the server can efficiently determine that elements exist prior to after, return true.
```

I think we should be able to efficiently determine if there are previous items so that we can handle this case appropriately. 

The same is the case with `hasNextPage`:

HasNextPage(allEdges, before, after, first, last)
- If first is set:
   - Let edges be the result of calling ApplyCursorsToEdges(allEdges, before, after).
   - If edges contains more than first elements return true, otherwise false.
- If before is set:
   - If the server can efficiently determine that elements exist following before, return true.
- Return false.

Currently, we're not respecting:

```
- If before is set:
   - If the server can efficiently determine that elements exist following before, return true.
```

-----

Thanks for pointing this out @MoOx!

MoOx:
Tell me if I can help (even if I didn't do any php for ages).
I could also be a beta tester if necessary :)

jasonbahl:
@MoOx even if you don't want to write tests, you _could_ write some client code, in something like CodeSandbox.io, write it up as you'd expect it to work, and we can use that as the "test", then someone with more PHP experience can convert that into actual Unit/Integration tests

agyimr:
I am also struggling to get over this. I wanted to use hasNextPage and hasPreviousPage to allow the user to press the buttons "Next" and "Previous". But currently if I set "last" or "first" the other one (in the opposite direction I am paginating) resolves to false and the user can not go back. 

Any suggested workarounds until this gets officially supported? :) 

Keep up the good work!

let00:
I can take a look at this. My current thoughts are to get the offset from the after cursor, and then check if the offset corresponds to a valid object ID. Assuming that the offset of the cursor is the ID of the node at the tail of the edge. Would this be a good approach?

jasonbahl:
@let00 ya, I think that should work.

jasonbahl:
closed by #1111

PR #1111, opened 2020-01-12
Fix for hasNextPage/hasPreviousPage not fully respecting Relay spec

let00:

If you query with `after` arg, `hasPreviousPage` will always return `false`. Similarly, if you query with `before` arg, `hasNextPage` will always return false. This PR fixes this for all connections supporting pagination.

#594 

Any other comments?
-------------------
`hasNextPage`/`hasPreviousPage` will always return false if you query `userRoles` with `before`/`after` queries respectively. This PR leaves that behavior as is, as pagination is yet to be implement for `userRoles`. This will need to be fixed in the future.

Where has this been tested?
---------------------------
wpgraphql docker test script

### DIFF
--- a/src/Data/Connection/CommentConnectionResolver.php
+++ b/src/Data/Connection/CommentConnectionResolver.php
@@ -253,7 +253,13 @@ class CommentConnectionResolver extends AbstractConnectionResolver {
 	 * @return bool
 	 */
 	public function is_valid_offset( $offset ) {
-		return ! empty( get_comment( $offset ) );
+		global $wpdb;
+
+		if ( ! empty( wp_cache_get( $offset, 'comment' ) ) ) {
+			return true;
+		}
+
+		return $wpdb->get_var( $wpdb->prepare( "SELECT EXISTS (SELECT 1 FROM $wpdb->comments WHERE comment_ID = %d)", $offset ) );
 	}
 
 }
--- a/src/Data/Connection/CommentConnectionResolver.php
+++ b/src/Data/Connection/CommentConnectionResolver.php
@@ -245,4 +245,15 @@ class CommentConnectionResolver extends AbstractConnectionResolver {
 
 	}
 
+	/**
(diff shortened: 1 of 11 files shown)


jasonbahl:
@let00 I like the sound of this! Checking this out right now! Thanks so much for working through this!

let00:
@jasonbahl Sure thing. 

https://github.com/wp-graphql/wp-graphql/blob/bf5f5bc9b96882c46ca549c9c1852aee9585108b/src/Data/Connection/PostObjectConnectionResolver.php#L500-L502
I am getting the whole post here, just to validate if the post exists. Seems like overkill to me, but not sure how much this affects performance.

jasonbahl:
This looks good. I ran through it locally and it seems to be working as expected. It's quite late for me though, so I want to run through it again tomorrow before I approve. 

----

I am noticing some funky stuff with user pagination. Not related to this PR, but just making note of it while fresh in my head. The default ordering seems to be opposite of what I'd expect. WordPress typically orders users by username in Ascending order, but we're defaulting to order by username in Descending order 🤔 

Screen Shot 2020-01-11 at 11 19 19 PM
Screen Shot 2020-01-11 at 11 19 41 PM

src/Data/Connection/CommentConnectionResolver.php
```diff
@@ -253,7 +253,13 @@ public function sanitize_input_fields( array $args ) {
 	 * @return bool
 	 */
 	public function is_valid_offset( $offset ) {
-		return ! empty( get_comment( $offset ) );
+		global $wpdb;
```
jasonbahl: @let00 I'm going to revert this particular commit. The queries already get the cursor objects up front when doing the cursor pagination, so they'll already be cached, so calling `get_comment`, `get_post`, etc will actually be more efficient because the objects will be in memory. 

I guess I missed this commit being added before I merged.

Issue #1101, opened 2020-01-03
PHP 7.4 has test failures


jasonbahl:
Need to look into PHP 7.4 support. 

@renatonascalves reported issues when testing WPGraphQL for BuddyPress on PHP 7.4, and when testing updating Travis tests to test on PHP 7.4 there are 5 failing tests. 

It appears that updating WordPress from 5.2.3 to 5.3.2 also causes one test failure. 

see: https://travis-ci.org/wp-graphql/wp-graphql/builds/632351510?utm_medium=notification&utm_source=github_status

jasonbahl:
closed by #1102

Issue #1096, opened 2019-12-28
Throw Exception when WPGraphQL is used on a PHP server below PHP 5.0


jasonbahl:
WPGraphQL should return an error when it's used on a PHP server running a PHP version lower than 7.0 so users can better identify why it's not working in such environments.

Issue #1088, opened 2019-12-20
Add `list_of` and `non_null` to the TypeRegistry


jasonbahl:
CC: @kidunot89 

When filtering Type definitions, it's currently not possible to add `non_null` and `list_of` types because they're not properly in the Type Registry. 

For example: https://github.com/kidunot89/wp-graphql-jwt-authentication/blob/feature/wp-graphql-v0.6.0-support/src/ManageTokens.php#L130

Here, we're filtering the `graphql_user_fields` and we can't do `type => [ 'list_of' => 'non_null' ]` because it's too late in the Schema generation and the prepare fields won't run against that Type again. 

We should support something to this tune: 

```
$type_registry->non_null( 'String' );
$type_registry->list_of( 'String' );
```

PR #1112, opened 2020-01-13
v0.6.0

jasonbahl:
# Release Notes

## Summary

This release focused primarily on adding Interfaces for Post and Term objects.

Interfaces in GraphQL allow for common fields to be registered and shared across Types. Similar to Unions, they allow for multiple types to be queried from the same field. 

By having Post Objects (posts, pages, etc) and Terms (Tags, Categories, etc) implement common interfaces, various Types can be queried for from the same entry point. So, instead of having to query _only_ Posts or _only_ Pages, interfaces make it possible to query for Post Objects of any type and have predictable results. 

See below for examples of such queries.

## Breaking Changes

- updated formatting of URI resolvers across types to be consistent. May be a breaking change depending on your current use of `uri` fields when querying nodes
- placed many fields of PostObject mutations behind conditionals to only allow inputs on Post Types that support said fields. For example, a Post Type that doesn't support author won't have an `author` input on its mutation anymore. Or a non-hierarchical Post Type doesn't have a `parent` input field by default in its mutation anymore. Might be a breaking change depending on your use of mutations and fields that a Post Type doesn't support
- fixed a bug with settings with dashes not properly being mapped to the schema. This fix in behavior could cause a breaking change depending on what types of settings you had registered and were querying from the schema.
- Made many fields of Post Object types (Post, Page, etc) part of Interfaces instead of the PostObject type itself, and the Types now implement the interfaces if the Post Type supports said feature. This might cause breaking changes if you were querying for fields on a Post Object Type that the Post Type doesn't actually support. For example, if you registered a post type without support for `author`, you could have queried for the `author` on that node, and got null. Now the author field will not exist on that Type and your query would not work. 
- remove `__get` methods from WPObjectType, WPUnionType, etc in favor of making the type_registry a public property (@kidunot89 this will affect WPGraphQL for WooCommerce)
- Deprecate Types.php static methods
- changed `parent` field to `revisionOf` field for Types that support revisions. If the node is a revision, the `revisionOf` field will return the node that it is a revision of. This is applied only to post types that support revisions via the `NodeWithRevisions` Interface
- added `abstract public function is_valid_offset( $offset );` to AbstractConnectionResolver. If you extend the AbstractConnectionResolver in your code, you will need to have your extending class implement this method. It is used for pagination to respect the Relay spec.

## New Features
- add support for the `post format` taxonomy - thanks @aberkow!
- update testing matrix for Travis CI to cover PHP 7.1 - 7.4
- misc updates to documentation - Thanks @aberkow, @jacobarriola, @craigfay 
- better support for respecting the Relay spec for pagination. (#1111) Thanks @let00!
- RootQuery.contentNodes connection added, allowing for lists of Posts of any Post Type to be queried from a single Root entry point
- All Post Object Types (Post, Page, etc) implement the ContentNode interface
- Most fields defined in PostObjects.php have been moved into interfaces to be defined
- Add new `RootQuery.terms` connection to query terms of any Taxonomy type from the root of the graph
- Add new NodeResolver class that resolves any node (Post, Term, User) from a URI. This allows for users to pass a URI (path of the resource) to fetch nodes by
- Update UserMutation.php to use string definition of types, instead of the Types static definitions
- added `pageTemplate` field to the Post model layer
- added uri field to the User model
- Added the following Interfaces: 
  - ContentNode
  - ContentTemplate
  - HierarchicalContentNode
  - NodeWithAuthor
  - NodeWithComments
  - NodeWithContentEditor
  - NodeWithExcerpt
  - NodeWithFeaturedImage
  - NodeWithRevisions
  - NodeWithTitle
  - NodeWithTrackbacks
  - TermNode
  - UniformResourceIdentifiable
- Added the following Enums
  - ContentNodeIdTypeEnum
  - TermNodeIdTypeEnum
  - UserNodeIdTypeEnum
- added ContentTemplateUnion, allowing for templates to be queried for on Post objects
- added description to various fields missing descriptions
- deprecated the $postType.by fields. They will still work, but they won't show in Schema docs and may be removed from the codebase in the future. This means that you can no longer query like so: `postBy( slug: "slug" ) { ...postFields }`, but are encouraged to query by the single entry point with the newly added `idType` input instead, like so: `post( id: "slug" idType: SLUG ) { ...postFields }` 
- added  RootQuery.contentNode field for querying an individual content node from the root of the graph
- added RootQuery.nodeByUri field for querying any node by it's resource path. 
- added RootQuery.termNode field for querying any type of term node from the root of the graph
- updated single resource entry points by adding `idType` inputs. Now single resources can be queried for using various types of unique identifiers such as database id, slug, name, uri. The type of ID available depends on the type of node. 

## Bugfixes
- Fixed a bug causing use of the `graphql()` function in PHP to break the admin bar
- throw an exception if WPGraphQL is being used on a server with a PHP version below what WPGraphQL supports (#1099)

## Issues Closed

- #405, #263, #278, #1086, #1066, #899, #1088, #944, #1101, #1100, #1096, #1083, #1063, #1059 

# Examples of new functionality

## Posts
We're deprecating the `$postType.'By'` entry points, and have added a new `idType` field and enum for single entry points. 

This means queries such as the following will still work, but will not show in the Schema documentation, and may be formally removed from the codebase at a later date:

### Deprecated Queries

```
{
  postBy( slug: "some-slug" ) {
    id
    title
  }
}
```
```
{
  pageBy( uri: "some-uri" ) {
    id
    title
  }
}
```

These are deprecated in favor of existing single entry points for each post type, and the newly added `idType` field on those entry points.

### New Singular Post Object Queries

With the added `idType` field on the singular post object entry points, below is a series of example queries now made possible:

#### Query Single Post By Database ID

```
{
  post(id: 1739, idType: DATABASE_ID) {
    id
    title
    uri
    slug
    postId
  }
}
```

Screen Shot 2019-12-31 at 3 34 28 PM

#### Query Single Post by URI

```
{
  post(id: "/test-5/", idType: URI) {
    id
    title
    uri
    slug
    postId
  }
}
```

Screen Shot 2019-12-31 at 3 35 31 PM

#### Query Single Post by Global Hashed ID

```
{
  post(id: "cG9zdDoxNzM5", idType: ID) {
    id
    title
    uri
    slug
    postId
  }
}
```
Screen Shot 2019-12-31 at 3 34 06 PM

or without specifying `idType` as the default is the hashed ID

```
{
  post(id: "cG9zdDoxNzM5") {
    id
    title
    uri
    slug
  }
}
```

Screen Shot 2019-12-31 at 3 33 53 PM

#### Query Post (non-hierarchical post type) by Slug

```
{
  post(id: "test-5", idType: SLUG) {
    id
    title
    uri
    slug
  }
}
```

Screen Shot 2019-12-31 at 3 36 46 PM

### ContentNodeInterface
This release also introduces the `ContentNode` Interface that is implemented by all Post Object types (posts, pages, custom post types). 

This Interface allows for new types of queries to be executed that span many post types. 

#### Content Nodes Connection
Here's an example of a new query: 

```
{
  contentNodes {
    nodes {
      __typename
      id
      title
      link
      uri
      isRevision
      ... on Page {
        isFrontPage
      }
    }
  }
}
```

Screen Shot 2019-12-31 at 3 49 33 PM

In the results above, we can see that a mix of Posts and Pages have been returned.

We can also pass arguments to do things like search across multiple post types: 

```
{
  contentNodes(where: {search: "test"}) {
    nodes {
      __typename
      id
      title
      link
      uri
      isRevision
      ... on Page {
        isFrontPage
      }
    }
  }
}
```

Screen Shot 2019-12-31 at 3 53 04 PM

#### ContentNode single entry point

We can now also fetch single Content (post) nodes like so: 

```
{
  contentNode(id: "cG9zdDoxNzM5") {
    __typename
    id
    title
    link
    uri
    isRevision
    ... on Page {
      isFrontPage
    }
  }
}
```

Screen Shot 2019-12-31 at 3 59 15 PM

But, we can fetch single nodes of any Post type by other unique identifiers as well, such as the URI: 

```
{
  contentNode(id: "/2019/12/05/test-5/", idType: URI) {
    __typename
    id
    title
    link
    uri
    isRevision
    ... on Page {
      isFrontPage
    }
  }
}
```
Screen Shot 2019-12-31 at 4 02 25 PM

or

```
{
  contentNode(id: "/test/", idType: URI) {
    __typename
    id
    title
    link
    uri
    isRevision
    ... on Page {
      isFrontPage
    }
  }
}
```

Screen Shot 2019-12-31 at 4 02 38 PM

This is a SUPER powerful feature as it allows for fragments to be more easily shared and re-used across queries and components.

## Terms 

Prior to this release, there wasn't really great ways of fetching singular term objects. You could do a query for a connection of terms and limit it to the `first: 1`, and in some cases that worked fine, but it wasn't always ideal. 

This release adds better support for fetching individual Term Nodes. 

### Example Term Node Queries

Instead of only fetching by the hashed global ID, we can now fetch by other unique identifiers using the `idType` field.

#### Query Tag by Hashed Global ID

```
{
  tag(id: "cG9zdF90YWc6Mw==") {
    id
    name
    slug
    tagId
  }
}
```

Screen Shot 2019-12-31 at 4 09 37 PM

#### Query Tag by Database ID

```
{
  tag(id: 3, idType: DATABASE_ID) {
    id
    name
    slug
    tagId
  }
}
```

Screen Shot 2019-12-31 at 4 12 14 PM

#### Query Tag by Name

```
{
  tag(id: "Another Test", idType: NAME) {
    id
    name
    slug
    tagId
  }
}
```

Screen Shot 2019-12-31 at 4 12 47 PM

#### Query Tag by Slug

```
{
  tag(id: "another-test", idType: SLUG) {
    id
    name
    slug
    tagId
  }
}
```

Screen Shot 2019-12-31 at 4 14 06 PM

#### Query Tag by URI

```
{
  tag(id: "tag/another-test/", idType: URI) {
    id
    name
    slug
    tagId
    uri
  }
}
```

Screen Shot 2019-12-31 at 4 14 46 PM

### TermNode Interface

This release also introduces the TermNode interface, which allows for querying terms across Taxonomy types. 

For example: 

#### Query TermNodes Connection

```
{
  terms {
    nodes {
      id
      __typename
      name
      uri
    }
  }
}
```
Screen Shot 2019-12-31 at 4 22 31 PM

This enables us to do things, like search terms across multiple taxonomies and get both Categories and Tags (and custom taxonomies) back in the same query: 

```
{
  terms(where: {search: "test"}) {
    nodes {
      id
      __typename
      name
      uri
      ... on Tag {
        tagId
      }
      ... on Category {
        categoryId
      }
    }
  }
}
```

Screen Shot 2019-12-31 at 4 24 38 PM

### Query Single Terms by Unique Identifiers
We can also query single term nodes by various identifiers, such as `Database ID`, `ID`, `Name`, `Slug` or `URI`: 

#### Query Single Term Node by (hashed) ID

```
{
  termNode(id: "cG9zdF90YWc6Mw==") {
    __typename
    id
    name
    link
    slug
    uri
  }
}
```

Screen Shot 2019-12-31 at 4 28 01 PM

#### Query Single Term Node by Database ID

```
{
  termNode(id: 3, idType: DATABASE_ID) {
    __typename
    id
    name
    link
    slug
    uri
    databaseId
  }
}
```

Screen Shot 2020-01-01 at 3 53 24 PM

#### Query Single Term Node by Name

```
{
  termNode(id: "Another Test", idType: NAME, taxonomy: TAG) {
    __typename
    id
    name
    link
    slug
    uri
    databaseId
  }
}
```
Screen Shot 2020-01-01 at 3 53 24 PM

#### Query Single Term Node by Slug

```
{
  termNode(id: "another-test", idType: SLUG, taxonomy: TAG) {
    __typename
    id
    name
    link
    slug
    uri
    databaseId
  }
}
```

Screen Shot 2020-01-01 at 3 56 32 PM

#### Query Single Term Node by URI

```
{
  termNode(id: "tag/another-test/", idType: URI) {
    __typename
    id
    name
    link
    slug
    uri
    databaseId
  }
}
```

Screen Shot 2020-01-01 at 3 57 44 PM

## Node By URI

Another thing this release provides is a way to get any node by URI, across types. This comes in handy because in many cases when a URI is provided, the type of content it is isn't always known ahead of time. By having an entry point for _all_ Types that implement the 1UniformResourceIdentifiable` Interface, we can create a main entry point for entire applications. 

Below is an example. This is showing many queries at once, but this could easily be one query with fragments for each Type, controlled by their appropriate component for rendering said type of content:

```
{
  page: nodeByUri(uri: "about/") {
    ...URI
  }
  post: nodeByUri(uri: "2019/12/05/test-5/") {
    ...URI
  }
  tag: nodeByUri(uri: "tag/8bit/") {
    ...URI
  }
  category: nodeByUri(uri: "category/alignment/") {
    ...URI
  }
  user: nodeByUri(uri: "author/jasonbahl/") {
    ...URI
  }
}

fragment URI on UniformResourceIdentifiable {
  __typename
  ... on Page {
    pageId
  }
  ... on Post {
    postId
  }
  ... on Category {
    categoryId
  }
  ... on Tag {
    tagId
  }
  ... on User {
    userId
  }
}
```

Screen Shot 2020-01-01 at 4 05 25 PM


jasonbahl:
@kidunot89 I have it called out in the notes:

> added `abstract public function is_valid_offset( $offset );` to AbstractConnectionResolver. If you extend the AbstractConnectionResolver in your code, you will need to have your extending class implement this method. It is used for pagination to respect the Relay spec.

src/Connection/PostObjects.php
```diff
@@ -19,6 +19,34 @@ class PostObjects {
...
+				'connectionArgs' => self::get_connection_args(
+					[
+						'contentTypes' => [
+							'type'        => [ 'list_of' => 'PostTypeEnum' ],
+							'description' => __( 'The Types of content to filter', 'wp-graphql' ),
+						],
+					],
+					null
+				),
+				'resolve'        => function( $source, $args, $context, $info ) {
```
renatonascalves: Suggestion: Could this be moved together with the other connection wrappers in the `DataSource` class?

src/Data/Connection/CommentConnectionResolver.php
```diff
@@ -245,4 +245,21 @@ public function sanitize_input_fields( array $args ) {
...
+	 * @return bool
+	 */
+	public function is_valid_offset( $offset ) {
+		global $wpdb;
+
+		if ( ! empty( wp_cache_get( $offset, 'comment' ) ) ) {
+			return true;
+		}
+
+		return $wpdb->get_var( $wpdb->prepare( "SELECT EXISTS (SELECT 1 FROM $wpdb->comments WHERE comment_ID = %d)", $offset ) );
```
renatonascalves: Is this multisite supported? Besides, where are you caching the offset?

src/Data/Connection/PostObjectConnectionResolver.php
```diff
@@ -492,4 +491,21 @@ function( $status ) use ( $post_type_objects ) {
...
+	 * @return bool
+	 */
+	public function is_valid_offset( $offset ) {
+		global $wpdb;
+
+		if ( ! empty( wp_cache_get( $offset, 'posts' ) ) ) {
+			return true;
+		}
+
+		return $wpdb->get_var( $wpdb->prepare( "SELECT EXISTS (SELECT 1 FROM $wpdb->posts WHERE ID = %d)", $offset ) );
```
renatonascalves: Same question here!

src/Data/DataSource.php
```diff
@@ -899,4 +893,24 @@ public static function get_registered_nav_menu_locations() {
...
+	 * @param string      $uri     The URI to fetch a resource from
+	 * @param AppContext  $context The AppContext passed through the GraphQL Resolve Tree
+	 * @param ResolveInfo $info    The ResolveInfo passed through the GraphQL Resolve tree
+	 *
+	 * @return mixed
+	 * @throws \Exception
+	 */
+	public static function resolve_resource_by_uri( $uri, $context, $info ) {
+
+		$node_resolver = new NodeResolver();
```
renatonascalves: Suggestion: `return ( new NodeResolver )->resolve_uri( $uri );`

src/Data/NodeResolver.php
```diff
@@ -0,0 +1,388 @@
...
+namespace WPGraphQL\Data;
+
+use GraphQL\Error\UserError;
+use WPGraphQL\Model\Post;
+use WPGraphQL\Model\Term;
+use WPGraphQL\Model\User;
+
+class NodeResolver {
+
+	protected $wp;
```
renatonascalves: Missing phpDoc.

src/Data/Connection/CommentConnectionResolver.php
jasonbahl: @renatonascalves ah! I actually reviewed the PR this was part of before this commit came in, and I guess I overlooked this commit's changes. (https://github.com/wp-graphql/wp-graphql/pull/1111/commits/25e5812876359b428882f10af2161eb5867db3a3#r366142346)

Thanks for pointing this out. I've reverted these changes as they actually won't be more performant at this point because the cursor objects are cached when the query first starts, so just asking for the objects again will be faster than another raw sql query. And, as you point out, the cache wasn't being set anyway.

src/Connection/PostObjects.php
jasonbahl: Ah yes. I originally pulled it in here because I thought I was going to need to customize and use `$resolver->setQueryArg()`, but that ended up not being the case. I reverted to use the existing `DataSource::resolve_post_objects_connection();`

Good call! 😄

PR #1122, opened 2020-01-18
feat: add seopress plugin to docs

moonmeister:
Adding docs for a new plugin

https://github.com/moonmeister/wp-graphql-seopress

### DIFF
--- a/docs/gatsby-config.js
+++ b/docs/gatsby-config.js
@@ -133,6 +133,7 @@ module.exports = {
             'extensions/wpgraphiql',
             'extensions/wpgraphql-content-blocks',
             'extensions/wpgraphql-gutenberg',
+            'extensions/wp-graphql-seopress',
           ],
         },
       },
--- a/docs/source/extensions/wpgraphql-seopress.mdx
+++ b/docs/source/extensions/wpgraphql-seopress.mdx
@@ -1,0 +1,9 @@
+---
+title: WPGraphQL for SEOPress 
+description: Expose SEOPress data in WPGraphQL Schema
+---
+
+
+# WPGraphQL SEOPress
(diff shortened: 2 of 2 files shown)


jasonbahl:
@moonmeister good stuff with the extension!

To have this show up in the docs, you'll also need to add it to the navigation by listing it here: https://github.com/wp-graphql/wp-graphql/blob/develop/docs/gatsby-config.js#L125-L135

renatonascalves:
I'd recommend setting some guidelines on what to include for each extension. The amount of information here is low.

At least the order could be:

* What it is // Getting started
* What it does // How to install
* Examples / Playground
* Etc.

moonmeister:
@renatonascalves Hey, yes I thought about this when creating the page. /The reason I only included a link to the repo is two fold. First, most the other docs pages don't have any link to a repository, I've found this quite frustrating, second, I didn't want to maintain documentation for the plugin in two locations, it's mostly already in the other repository. 

One solution for this is probably doing what Gatsby has, and using a package tag and the (in their case NPM) packagist API to fetch all wpgraphql packages automatically, and using their `readme.md` files as the contents of the docs page. 

Obliviously this doesn't exist, So Ii thought the best option was linking to the repo for the project where all the docs/code were.

jasonbahl:
I've been talking with @henrikwirth about using the README's of extensions as their docs as well, and probably come up with a specific template for extensions that link to their repos, etc.

I created this issue earlier to track that: https://github.com/wp-graphql/wp-graphql/issues/1124

jasonbahl:
@moonmeister thanks for adding the menu link! Until we refactor the docs section to make it better reflect the extensions, I'm ok with something short linking to the repo.

Issue #1141, opened 2020-02-03
context is not defined


spencersmb:
Tried to publish a new static site build from Gatsby this morning with minor CSS changes and this error appeared. 2 days ago I was able to publish just fine. I've seen others with the same error on slack. I can't even query basic posts anymore under wpgraphql.

Running:
"gatsby-source-graphql": "^2.1.33",
WP GraphQL v.6.0

Here is the error output:
```
{
  "errors": [
    {
      "message": "context is not defined",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "wpgraphql",
        "posts"
      ],
      "stack": [
        "ReferenceError: context is not defined",
        "    at buildDelegationTransforms (/Users/spencerbigum/Documents/github/etShop-gatsby/node_modules/graphql-tools-fork/dist/stitching/delegateToSchema.js:48:98)",
        "    at delegateRequest (/Users/spencerbigum/Documents/github/etShop-gatsby/node_modules/graphql-tools-fork/dist/stitching/delegateToSchema.js:75:32)",
        "    at Object.delegateToSchema [as default] (/Users/spencerbigum/Documents/github/etShop-gatsby/node_modules/graphql-tools-fork/dist/stitching/delegateToSchema.js:40:12)",
        "    at /Users/spencerbigum/Documents/github/etShop-gatsby/node_modules/graphql-tools-fork/dist/stitching/resolvers.js:57:86",
        "    at resolveFieldValueOrError (/Users/spencerbigum/Documents/github/etShop-gatsby/node_modules/graphql/execution/execute.js:467:18)",
        "    at resolveField (/Users/spencerbigum/Documents/github/etShop-gatsby/node_modules/graphql/execution/execute.js:434:16)",
        "    at executeFields (/Users/spencerbigum/Documents/github/etShop-gatsby/node_modules/graphql/execution/execute.js:275:18)",
        "    at collectAndExecuteSubfields (/Users/spencerbigum/Documents/github/etShop-gatsby/node_modules/graphql/execution/execute.js:713:10)",
        "    at completeObjectValue (/Users/spencerbigum/Documents/github/etShop-gatsby/node_modules/graphql/execution/execute.js:703:10)",
        "    at completeValue (/Users/spencerbigum/Documents/github/etShop-gatsby/node_modules/graphql/execution/execute.js:591:12)",
[...]

PR #1147, opened 2020-02-14
#1144 - Prevent model layer from attempting to execute cached values

### LINKED ISSUE #1144: User/Author fields value treated as function
Hi,

Bug found in core wp-graphql

Initially thought it was related to ACF therefore created an issue here https://github.com/wp-graphql/wp-graphql-acf/issues/105

If any user fields eg, firstName, lastName, have a value that happens to also be a function name eg. count, sizeof, the function is actually run, resulting in a null value.

See link above for more details and how to replicate

jasonbahl:
If the Model has already processed a field, and it's a scalar, array or non-callable object, return it. Otherwise call the function to process and cache the field for next run.	

closes #1144 

**Before**
Screen Shot 2020-02-14 at 12 59 22 PM

**After**
Screen Shot 2020-02-14 at 12 58 28 PM

**Before**
Screen Shot 2020-02-14 at 12 59 35 PM

**After**
Screen Shot 2020-02-14 at 12 58 03 PM

### DIFF
--- a/src/Model/Model.php
+++ b/src/Model/Model.php
@@ -145,7 +145,15 @@ abstract class Model {
 	 */
 	public function __get( $key ) {
 		if ( ! empty( $this->fields[ $key ] ) ) {
-			if ( is_callable( $this->fields[ $key ] ) ) {
+			/**
+			 * If the property has already been processed and cached to the model
+			 * return the processed value.
+			 *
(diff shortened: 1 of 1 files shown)


jasonbahl:
@iamlili This pull request addresses #1141

PR #1154, opened 2020-02-20
Bug/#1138 resolve nodes earlier

### LINKED ISSUE #1138: Connection Nodes should resolve earlier
Nodes are resolved too late by connection resolvers. 

What I mean by this, is that currently connection resolvers query for IDs of nodes, and pass them down through edges, to nodes, and at the node level is when the ID is passed to a Deferred resolver to resolve the node. The deferred resolver allows for many nodes to be resolved at once because of the DataLoader pattern. 

The problem with this, is that the nodes are being resolved too late to make decisions on their privacy. 

Take a query like the following:

```
{
  users {
    nodes {
       id
       name
    }
  }
}
```

A public user should get a list of users. Only users that have published posts are considered public, but that's determined after the User has been queried for. We take that user, check if it has published posts, if it does, it's public, if not, it's not public and a public user shouldn't have said user in the connection. 

Because we resolve the nodes so late, we end up with errors, something like the following: 

`Cannot return null for non-nullable field User.id.`

Also, when registering edge fields, the `$root` that is passed down to the resolver doesn't have the node as context, just the ID of the node. This can make it difficult to resolve edge fields properly when actual node data is often necessary, not just an ID. 

## Proposal
The nodes of a connection should be resolved and passed through the Model Layer _before_ the list of edges/nodes is  passed down to the next level of the tree. This will allow for the nodes that are private or otherwise considered not-visible for the requesting user to be stripped from the results prior to passing down to edges/nodes. 

This will solve the `id cannot be null` errors while also providing more rich context to edge fields, and performance will not be negatively impacted as we can still make use of deferred resolvers, just at a higher point in the resolve tree.

jasonbahl:
Heading to dinner. . .will update with more info, but wanted to get this in.

-----------------------------------------
closes #1138 

### DIFF
--- a/src/Connection/Comments.php
+++ b/src/Connection/Comments.php
@@ -80,9 +80,6 @@ class Comments {
 			'toType'         => 'Comment',
 			'fromFieldName'  => 'comments',
 			'connectionArgs' => self::get_connection_args(),
-//			'resolveNode'    => function( $id, $args, $context, $info ) {
-//				return DataSource::resolve_comment( $id, $context );
-//			},
 			'resolve'        => function( $root, $args, $context, $info ) {
 				return DataSource::resolve_comments_connection( $root, $args, $context, $info );
 			},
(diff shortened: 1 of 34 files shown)


jasonbahl:
Closing, as this has now become this PR: https://github.com/wp-graphql/wp-graphql/pull/1163

PR #1197, opened 2020-03-21
AbstractDataLoader::get() added.

kidunot89:

Adds a `get` function to the `AbstractDataLoader` class. This function is for the purpose of retrieving object cached by the loader, but loads the object on spot if not found in the cache.
```
public function get( $key ) {
	if ( ! empty( $this->cached[ $key ] ) ) {
		return $this->cached[ $key ];
	} else {
		return $this->load( $key );
	}
}
```
`AbstractConnectionResolver::get_node_by_id()` is no longer abstract.
```
public function get_node_by_id( $id ) {
	return $this->loader->get( $id );
}
```
The following connection resolver classes have had the `get_node_by_id()` function removed.
- CommentConnectionResolver
- MenuConnectionResolver
- TermObjectConnectionResolver
- UserConnectionResolver

The **PostObjectConnectionResolver** and **MenuItemConnectionResolver** use the **PostObjectLoader** which doesn't seem to be compatible with the `AbstractDataLoader::get()` most like due to this **Deferred** [here](https://github.com/kidunot89/wp-graphql/blob/feature/abstract-loader-get-function/src/Data/Loader/PostObjectLoader.php#L128-L148). So `get_node_by_id()` has been left unchanged in those two classes.

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------

**WordPress Version:** 5.2.3

### DIFF
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -332,7 +332,9 @@ abstract class AbstractConnectionResolver {
 	 *
 	 * @return mixed|Model|null
 	 */
-	abstract public function get_node_by_id( $id );
+	public function get_node_by_id( $id ) {
+		return $this->loader->get( $id );
+	}
 
 	/**
 	 * get_query_amount
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -333,7 +333,7 @@ abstract class AbstractConnectionResolver {
 	 * @return mixed|Model|null
 	 */
 	public function get_node_by_id( $id ) {
-		return $this->loader->get( $id );
+		return $this->loader->load( $id );
 	}
 
 	/**
--- a/src/Data/Loader/AbstractDataLoader.php
+++ b/src/Data/Loader/AbstractDataLoader.php
@@ -285,22 +285,6 @@ abstract class AbstractDataLoader {
 	}
 
 	/**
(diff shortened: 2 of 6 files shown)


src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -332,7 +332,9 @@ public function setQueryArg( $key, $value ) {
 	 *
 	 * @return mixed|Model|null
 	 */
-	abstract public function get_node_by_id( $id );
+	public function get_node_by_id( $id ) {
+		return $this->loader->get( $id );
```
jasonbahl: Good call here. We should actually be able to do `$this->loader->load( $key );` and not add a new `get` method

src/Data/Loader/AbstractDataLoader.php
```diff
@@ -284,6 +284,22 @@ protected function normalizeEntry( $entry, $key ) {
...
+	 * @return mixed
+	 */
+	public function get( $key ) {
+		if ( ! empty( $this->cached[ $key ] ) ) {
+			return $this->cached[ $key ];
+		} else {
+			return $this->load( $key );
+		}
+	}
+
```
jasonbahl: The `load( $key )` method should work

```suggestion
```

src/Data/Connection/AbstractConnectionResolver.php
jasonbahl:
```suggestion
		return $this->loader->load( $id );
```

PR #1198, opened 2020-03-22
WPCS compliance

kidunot89:

Checklist
- [x] Connection
- [x] Data
- [ ] Model
- [ ] Mutation
- [ ] Registry
- [ ] Server
- [ ] Type
- [ ] Utils
- [ ] everything else :rofl: 

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------

**WordPress Version:** 5.2.3


src/connection/class-term-objects.php
```diff
@@ -38,10 +37,10 @@ public static function register_connections() {
...
 				),
 				'resolve'        => function ( $source, $args, $context, $info ) {
-					$taxonomies = isset( $args['where']['taxonomies'] ) && is_array( $args['where']['taxonomies'] ) ? $args['where']['taxonomies'] : \WPGraphQL::get_allowed_taxonomies();
-					$resolver   = new TermObjectConnectionResolver( $source, $args, $context, $info, array_values( $taxonomies ) );
-					$connection = $resolver->get_connection();
-					return $connection;
+					$taxonomies = isset( $args['where']['taxonomies'] ) && is_array( $args['where']['taxonomies'] )
+						? $args['where']['taxonomies']
+						: \WPGraphQL::get_allowed_taxonomies();
+					return DataSource::resolve_term_objects_connection( $source, $args, $context, $info, array_values( $taxonomies ) );
```
jasonbahl: I'm hesitant to make these changes. 

The new ConnectionResolver classes are flexible and intended to allow extension authors the ability to use them to resolve custom connections more easily. 

For example, let's say a custom connection were created to display a connection from Posts of one Post Type to Posts of another Post Type, and the IDs of the connected posts were stored in meta. 

A custom connection could be registered like so: 

```
register_graphql_connection([
  'fromType' => 'Post',
  'toType' => 'Post',
  'fromFieldName' => 'relatedPosts',
  'connectionTypeName' => 'RelatedPosts', 
  'resolve' => function( $source, $args, $context, $info ) {
      $related_ids = get_post_meta( $source->ID, 'related_posts', true );
      if ( ! empty( $related_ids ) ) {
          $resolver = new Post_Object_Connection_Resolver( $source, $args, $context, $info );
          $resolver->setQueryArg( 'post__in', $related_posts );
      }
      return null;
  }
]);
```

I'm actually inclined to make _more_ resolvers throughout the graph look like _this_ and cleanup some deep internals within connection resolvers. 

For example: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Connection/PostObjectConnectionResolver.php#L244-L266

Instead of this switch statement living within the ConnectionResolver, it could be more explicitly defined for the resolvers that need this context. 

Like here, for example: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Connection/PostObjectConnectionResolver.php#L246

This line sets the `post_parent` argument if the source of the connection is a Post. 

This screws up connections like the custom one above, where it would be a Post->Post connection, but not a parent/child connection. If this argument `post_parent=>0` were defined only in the connection resolvers that _need_ it, it would allow custom connections to be registered and resolved with less friction. 

Resolving with `DataSource::resolve_**_connection()` calls doesn't provide this flexibility, nor does it document by example for extension authors how they can utilize the connection resolver classes to "easily" resolve their custom connections.

kidunot89: Then why not make the `DataSource::resolve_**_connection()` return an instance of the resolver class instead? The naming convention can even change to `DataSource::**_connections()`. This way you can do something like this
```
register_graphql_connection([
  'fromType' => 'Post',
  'toType' => 'Post',
  'fromFieldName' => 'relatedPosts',
  'connectionTypeName' => 'RelatedPosts', 
  'resolve' => function( $source, $args, $context, $info ) {
      $related_ids = get_post_meta( $source->ID, 'related_posts', true );
      if ( ! empty( $related_ids ) ) {
          $resolver = DataSource::post_object_connections( $source, $args, $context, $info );
          $resolver->setQueryArg( 'post__in', $related_posts );
      }
      return null;
  }
]);
```

tests/wpunit/MenuConnectionQueriesTest.php
```diff
@@ -32,6 +32,7 @@ public function testMenusQueryById() {
 
 		$actual = do_graphql_request( $query );
 
+		// use --debug to view
```
CodeProKid, line 35: I feel like we should add this to the readme somewhere, instead of adding this to individual `codecept_debug()` calls.

src/AppContext.php
```diff
@@ -3,11 +3,11 @@
...
-use WPGraphQL\Data\Loader\CommentLoader;
-use WPGraphQL\Data\Loader\MenuItemLoader;
-use WPGraphQL\Data\Loader\PostObjectLoader;
-use WPGraphQL\Data\Loader\TermObjectLoader;
-use WPGraphQL\Data\Loader\UserLoader;
+use WPGraphQL\Data\Loader\Comment_Loader;
+use WPGraphQL\Data\Loader\Menu_Item_Loader;
+use WPGraphQL\Data\Loader\Post_Object_Loader;
+use WPGraphQL\Data\Loader\Term_Object_Loader;
+use WPGraphQL\Data\Loader\User_Loader;
```
CodeProKid, line 10: I don't think these changes are valuable to the plugin. We aren't currently considering merging the plugin into core, which is the only instance that I think these changes would make sense. These are going to cause breaking changes for people extending the plugin which will be a poor developer experience without much to gain.

phpcs.xml
```diff
@@ -1,27 +1,36 @@
 <?xml version="1.0"?>
-<ruleset name="Coding standards for the WP GraphQL plugin">
-
+<ruleset name="Coding standards for WPGraphQL">
```
CodeProKid, line 2: Quick tip, you can use `<arg name="parallel" value="20"/>` within the ruleset to have the lints run in parallel, which makes them run much faster! I just found out about this the other day, so I just figured I'd pass it along! 😄

src/data/class-datasource.php
```diff
@@ -160,38 +153,36 @@ public static function resolve_plugin( $info ) {
...
+	 * @param \WP_Post    $source  WP_Post object.
+	 * @param array       $args    Array of arguments to pass to reolve method.
+	 * @param AppContext  $context AppContext object passed down.
+	 * @param ResolveInfo $info    The ResolveInfo object.
+	 *
+	 * @return array
 	 */
 	public static function resolve_plugins_connection( $source, array $args, AppContext $context, ResolveInfo $info ) {
-		return PluginConnectionResolver::resolve( $source, $args, $context, $info );
+		return \WPGraphQL\Data\Connection\Plugin_Connection_Resolver::resolve( $source, $args, $context, $info );
```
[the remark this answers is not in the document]
kidunot89: It's only used once in the file so I don't see a reason to use a `use` statement for that :thinking-face:

src/data/class-datasource.php
```diff
@@ -648,6 +712,7 @@ public static function resolve_node_type( $node ) {
 					$type = get_post_type_object( $node->post_type )->graphql_single_name;
 					break;
 				case $node instanceof Term:
+					// phpcs:ignore WordPress.NamingConventions.ValidVariableName
```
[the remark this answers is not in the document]
kidunot89: Maybe, but the use of the statement has been minimized to like 4 times in all of WooGraphQL by simply changing the naming conventions of fields on the Models that aren't exposed to the GraphQL schema to be snake case. Like arrays of IDs only used to resolve connected posts in connections. Usages of those fields in connection resolvers are where commonly you'll find these ignore statements, I think. 🤔

src/data/class-node-resolver.php
```diff
@@ -138,7 +148,7 @@ public function resolve_uri( $uri, $extra_query_vars = '' ) {
 
 							$post_status_obj = get_post_status_object( $page->post_status );
 							if ( ! $post_status_obj->public && ! $post_status_obj->protected
-								 && ! $post_status_obj->private && $post_status_obj->exclude_from_search ) {
+								&& ! $post_status_obj->private && $post_status_obj->exclude_from_search ) {
```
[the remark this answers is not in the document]
kidunot89: I agree. I didn't want to do too much to the class because I wasn't completely sure how it worked yet.

src/AppContext.php
kidunot89: I hadn't actually gotten to the files in the `src` directory root yet. So I only made enough changes to make the files in the `connection` and `data` directory work.

jasonbahl:
Closing this PR in favor of #1235 

After the meeting with @CodeProKid, @renatonascalves and @kidunot89 we agreed it would be best to proceed with more clearly defining the rules in the phpcs.xml file, opting out of the ones we're breaking in bulk right now, and then opting-in and fixing them one at a time.

PR #1235 updates the phpcs.xml file and composer dependencies as needed to allow tests and sniffs to work. It also includes some of the formatting changes that phpcbf could fix automatically as well as some manual changes. 

We can opt-in to any of the ignored rules one at a time now in smaller pull requests.

PR #1202, opened 2020-03-26
Feature/1201 prevent password changed emails on user registration

### LINKED ISSUE #1201: Prevent "Password Changed" Emails on User Registration
When you register a new user using WPGraphQL's `registerUser` mutation and set the user's password at the same time, the registered user will receive a "Password Changed" email.

This happens because WPGraphQL calls WordPress' `register_new_user()` function to save the new user's username and email, then calls WP's `wp_update_user()` function to save everything else (firstName, lastName, password, etc.). When WP runs a check inside of `wp_update_user()` to see if the new password matches the one on file previously (which it doesn't), it will trigger sending of the "Password Changed" email.

In my mind, this makes for a confusing experience for users, though. They'll likely think:

> I just registered as a user on this website, then I immediately got an email saying my password has been _changed_ -- what's the deal with that?

## Steps to Reproduce
Fire off this mutation and confirm that a "Password Changed" email is sent out:

```graphql
mutation {
  registerUser(
      input: {
        clientMutationId: "123"
        username: "123@example.com"
        firstName: "Test"
        lastName: "McTestUser"
        email: "123@example.com"
        password: "password"
      }
    ) {
      user {
        userId
      }
    }
}
```

## Solution
I propose we prevent "Password Changed" emails from being sent out during the user registration process.

We can disable the "Password Changed" emails immediately before WPGraphQL calls `wp_update_user()` inside of `UserRegister.php`, then re-enable it immediately afterward.

This will prevent "Password Changed" emails from being sent out when a user is registered via the `registerUser` mutation, but won't have any affect on those emails being sent out otherwise (such as when the user actually does change their password via a password reset flow).

 PR incoming!

kellenmace:
Disabled "Password Changed" emails during the user registration work done in the `registerUser` mutation resolver.

Closes #1201

Any other comments?
-------------------
WPGRAPHQL RULZ!

Where has this been tested?
---------------------------
**Operating System:** macOS 10.15.3
**WordPress Version:** Latest


src/Mutation/UserRegister.php
```diff
@@ -133,11 +133,21 @@ public static function mutate_and_get_payload() {
...
+
 			/**
 			 * Update the registered user with the additional input (firstName, lastName, etc) from the mutation
 			 */
 			wp_update_user( $user_args );
 
+			/**
+			 * Re-enable "Password Changed" emails.
+			 */
+			remove_filter( 'send_password_change_email', '__return_false' );
```
jasonbahl: @kellenmace I think your point in the issue about removing the filter is valid. 

What if we do something like this: 

```
// Apply the filter to get the current filtered value
$filter_value_before_update = apply_filters( 'send_password_change_email', '__return_false' );

// Add filter to make it explicitly false
add_filter( 'send_password_change_email', '__return_false' );

// Update the user, which will call the function with the filter
wp_update_user( $user_args );

// Reset the filter to the value it was before this function
add_filter( 'send_password_change_email', $filter_value_before_update );
```

jasonbahl:
@kellenmace looks like your latest changes have breaking tests. 

I think for static methods instead of `[ self, 'method' ]` you want just `\WPGraphQL\Mutation\UserRegister\return_false` as the filter callback

src/Mutation/UserRegister.php
```diff
@@ -133,11 +133,21 @@ public static function mutate_and_get_payload() {
 			 */
 			unset( $user_args['role'] );
 
+			/**
+			 * Prevent "Password Changed" emails from being sent.
+			 */
+			add_filter( 'send_password_change_email', [ self, 'return_false' ] );
```
jasonbahl, line 139:
```suggestion
			add_filter( 'send_password_change_email', `WPGraphQL\Mutation\UserRegister\return_false` );
```

src/Mutation/UserRegister.php
```diff
@@ -133,11 +133,21 @@ public static function mutate_and_get_payload() {
...
+
 			/**
 			 * Update the registered user with the additional input (firstName, lastName, etc) from the mutation
 			 */
 			wp_update_user( $user_args );
 
+			/**
+			 * Remove filter preventing "Password Changed" emails.
+			 */
+			remove_filter( 'send_password_change_email', [ self, 'return_false' ] );
```
jasonbahl, line 149:
```suggestion
			remove_filter( 'send_password_change_email', `WPGraphQL\Mutation\UserRegister\return_false` );
```

src/Mutation/UserRegister.php
```diff
@@ -133,11 +133,21 @@ public static function mutate_and_get_payload() {
 			 */
 			unset( $user_args['role'] );
 
+			/**
+			 * Prevent "Password Changed" emails from being sent.
+			 */
+			add_filter( 'send_password_change_email', `WPGraphQL\Mutation\UserRegister\return_false` );
```
jasonbahl, line 139:
```suggestion
			add_filter( 'send_password_change_email', 'WPGraphQL\Mutation\UserRegister\return_false' );
```

src/Mutation/UserRegister.php
```diff
@@ -133,11 +133,21 @@ public static function mutate_and_get_payload() {
...
+
 			/**
 			 * Update the registered user with the additional input (firstName, lastName, etc) from the mutation
 			 */
 			wp_update_user( $user_args );
 
+			/**
+			 * Remove filter preventing "Password Changed" emails.
+			 */
+			remove_filter( 'send_password_change_email', `WPGraphQL\Mutation\UserRegister\return_false` );
```
jasonbahl, line 149:
```suggestion
			remove_filter( 'send_password_change_email', 'WPGraphQL\Mutation\UserRegister\return_false' );
```

kellenmace:
@jasonbahl Yeah, I pulled the trigger a little too quickly on that first commit – I meant to use the `__CLASS__` magic constant instead of `self`. 😀

I fixed that, and also wrote a test to confirm that the new solution does in fact prevent the "Password Changed" emails from going out when a user is registered via the `registerUser` mutation.

I initially tried capturing the initial value of that filter, then resetting it as you had suggested. It worked, but resulted a lot of extra clutter being added to the resolver, in my opinion. Have a hook:

```
/**
 * Apply filter to get the current filtered value.
 */
$filter_value_before_update = apply_filters( 'send_password_change_email', true );

/**
 * Disable "Password Changed" emails from being sent.
 */
add_filter( 'send_password_change_email', '__return_false' );

/**
 * Update the registered user with the additional input (firstName, lastName, etc) from the mutation
 */
wp_update_user( $user_args );

/**
 * Reset the filter to the value it was before this function.
 */
add_filter( 'send_password_change_email', function() use ( $filter_value_before_update ) {
	return $filter_value_before_update;
} );
```

So I instead opted for adding a `return_false()` method to the `UserRegister` class that gets hooked before our call to `wp_update_user()`, then unhooked right after it. I think that's a simpler approach. Please let me know what you think.

Also: I notice the tests above are failing, but the issues don't appear to be related to my changes.

Thanks!

tests/wpunit/UserObjectMutationsTest.php
```diff
@@ -969,4 +973,24 @@ public function runSendPasswordResetEmailSentTest( $mutation_arg ) {
...
+		$this->registerUserMutation( [
+			'username' => 'password-changed-email-test-user',
+			'email'    => 'password-changed-email-test-user@example.com',
+			'password' => 'password-changed-email-test-user-password'
+		] );
+
+		/**
+		 * Assert that the "Password Changed" email was not sent when the user was registered.
+		 */
+		$this->assertFalse( $was_password_change_email_sent );
```
jasonbahl, line 994: I think this will always be false? The `$was_password_change_email_sent` variable in the filter callback is a variable scoped to that callback function, but this assertion is asserting the variable outside of that callback which is set to `false` and never changed.

jasonbahl: Is there an action that fires when an email is sent? we might be able to assert whether `did_action( 'email_sent_or_whatever' );` is true / false?

kellenmace: @jasonbahl The variable inside of the callback is not scoped to that callback function, since I'm passing it in by reference with `use ( &$was_password_change_email_sent )`. So when the `$was_password_change_email_sent = true;` line runs, it updates the value of the original `$was_password_change_email_sent` variable in the outer scope to `true`.

If you comment out our new `add_filter()` line in `UserRegister.php`, then fire off a `registerUser` mutation, `$was_password_change_email_sent` will get set to `true` and the email will be sent. Alternatively, if you leave our new `add_filter()` line in `UserRegister.php`, and fire off the mutation, `$was_password_change_email_sent` will NOT get set to `true`, and the email will not be sent.

According to this trac ticket (https://core.trac.wordpress.org/ticket/35357), you can't use `did_action()` to determine whether a filter has been applied. I was thinking along those same lines, too. And there isn't a `do_action()` hook anywhere in `wp_update_user()` or `wp_mail()` that we can use.

The only other filter hook we could use would be `wp_mail`, here: https://core.trac.wordpress.org/browser/tags/5.4/src/wp-includes/pluggable.php#L181. We would need to add a filter to it and check whether the email header contains the string "Password Changed". I think my implementation is cleaner though. `wp_mail` fires for every email that gets sent, which is why we'd need to check the header text, whereas the `password_change_email` filter I'm using only fires when the changed password email is about to be sent.

jasonbahl: @kellenmace I overlooked `&$was_password_change_email_sent` when reviewing. 

I walked through this locally, and you've done it properly. 

Good stuff! And thanks for the great follow up response!

kellenmace: @jasonbahl Sweet! Thanks for getting this one merged in. 🙌🏼

Issue #143, opened 2017-05-26
Abstract mutation permission checks


jasonbahl:
We should abstract mutation permission checks. 

see: https://github.com/wp-graphql/wp-graphql/pull/140#discussion_r118592548

CodeProKid:
@jasonbahl was this added in #140 ? Can't tell because the PR is gigantic and my browser won't load it 😆

jasonbahl:
Was not. This issue was created 3 days after that PR, I believe in response to some non-DRY things happening in that PR

jasonbahl:
I'm thinking that we should _probably_ think about moving the actual logic of mutations into the Model Layer as well. 

This will allow us to have more granular mutations that still pass through the same cap checks

For example, we have `updatePost` but let's pretend we also wented a very granular `setPostTitle` mutation, if the logic for writing to a post was centralized in a Model Layer, then smaller mutations could exist and make use of the central capability checks, etc. . .much like on the query side.

renatonascalves:
Hi guys! Something about the members mutations for the BuddyPress API that I want to discuss.

BuddyPress has some members actions, delete and update, that we can use the same mutations from the plugin.

But the current implementation has no way for me to filter the permissions check. For example. In BuddyPress we check if the user has the `edit_user` cap to update his own account or the `delete_user` to delete his own account.

Also we have admins or mods that have a specific cap `bp_moderate`.

The problem is that I can't hook into the caps and verify those credentials. 

Either by moving the code to the model or adding a hook here: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Mutation/UserDelete.php#L81

I think this is a nice addition not only to this mutation but to all of the other mutation. This will bring more flexibility to plugin authors.

At the BP REST API, we added in a way one could hook into the permission check and change that: https://github.com/buddypress/BP-REST/blob/64a675a0a23cde1254a073f531bb6091d3e96c6a/includes/bp-activity/classes/class-bp-rest-activity-endpoint.php#L332

CodeProKid:
I definitely agree that we should centralize the cap checks for mutations, but I don't think the model layer is the correct place to do so. The model layer is to model WordPress data to conform to the shape of the Graph. The checks for mutations are fundamentally different in that they don't affect the shape of objects in any way, but rather enforce rules around what actions a user can take.

[...]

PR #1231, opened 2020-04-09
Delete User Mutation

renatonascalves:
Passing the user ID to the `current_user_can` function in the `deleteUser` mutation so that extensions and plugins can properly extend it.

More background information here: #143


renatonascalves:
@jasonbahl Is it okay to merge this:

PR #1239, opened 2020-04-14
Replace DateQueryInput with strings for the dateQuery before and afte…

### LINKED ISSUE #1234: Provide a way to query for posts before or after a specific time (not just date)
In my app, I need the ability to fire off a request every few minutes to the WP backend to get any posts that have beed added or edited since the last time I checked. Currently though, WPGraphQL only supports doing date queries using day, month and year. You can't get more specific than that and narrow it down to a certain hour/minute/second. Example:

```graphql
query getPosts {
  posts(
    first: 10,
    where: {
      dateQuery: {
        column: MODIFIED
        after: {
          # Need to add hour, minute & second here.
          day: 13
          month: 4
          year: 2020
        }
      }
    }
  ) {
    nodes {
      postId
      modifiedGmt
    }
  }
}
```

`WP_Query()` does provide that ability by accepting a `strtotime()` compatible string as the `before`/`after` argument. Documentation: https://developer.wordpress.org/reference/classes/wp_query/#date-parameters

Even if we want to avoid accepting an arbitrary string like `WP_Query()` does, I think this can still be accomplished by accepting `hour`, `minute`, and `second` input args for `before` and `after`. Then apps like mine that need to query for published/modified posts on a frequent interval would be able to do so.

Who's with me?

kellenmace:
…r fields so that strtotime()-compatible strings can be passed in

Replaces `DateQueryInput` with strings for the `dateQuery` `before` and `after` fields so that `strtotime()`-compatible strings can be passed in.

Test query:

```graphql
query {
  posts(first: 10, where: {
    dateQuery: {
      after: "2020-03-13T19:05:30"
    }
  }) {
    nodes {
      title
      date
      dateGmt
    }
  }
}
```

Closes #1234

Any other comments?
-------------------
WPGRAPHQL RULZ!

Where has this been tested?
---------------------------
**Operating System:** macOS 10.15.3
**WordPress Version:** 5.4


jasonbahl:
@kellenmace thanks for the PR. I'll try and review shortly and provide my thoughs. 🙏

src/Type/Input/DateQueryInput.php
```diff
@@ -38,12 +38,12 @@ public static function register_type() {
 						'description' => __( 'Second (0 to 59)', 'wp-graphql' ),
 					],
 					'after'     => [
-						'type'        => 'DateInput',
-						'description' => __( 'Nodes should be returned after this date', 'wp-graphql' ),
+						'type'        => 'String',
```
CodeProKid, line 41: I think we need tighter validation here. I think it would be better to introduce a new DateTime scalar: https://webonyx.github.io/graphql-php/type-system/scalar-types/#writing-custom-scalar-types and use that here instead.

src/Type/Input/DateInput.php
```diff
@@ -1,28 +0,0 @@
-<?php
-namespace WPGraphQL\Type\Input;
-
-class DateInput {
-	public static function register_type() {
```
CodeProKid, line 5: Since this is probably a pretty widely used input arg, I'd rather see us deprecate the existing field and introduce the new field separately. We could then make the breaking change in a future release.

jasonbahl:
@kellenmace I've added `register_graphql_scalar()` as an API and I think we should probably have a maintainers meeting sooner than later to discuss what scalars might make sense to introduce and how we should introduce them. If we go the route of adding a DATETIME scalar, it would be perfect for this.

humet:
This would be very useful for us.

rgevrek:
This is a very useful PR, are there any plans to finish and merge? Else many inefficient calls and responses are needed on our side

PR #1261, opened 2020-04-23
Make `menuItems` connection flat  and other enhancements

### LINKED ISSUE #1260: Make `menuItems` flat by default
This is a breaking change I will be working on. We discussed this with @jasonbahl in depth. 

Currently the only way to get the menu hierarchy is to manually nest the `childItems` field as many times as needed. It is also a really slow way to do it because every time `childItems` is called it will get all the menu items and use the `array_filter()` PHP function to get correct ones for the nesting level.

So getting the items as a flat list will be much faster. We also need to add parent ids to the nodes so it is possible to build menu tree in the consumers (JS).

So the plan is to do following:

- Default `menuItems` connection to a flat list
- Add `where: { parentId: $id, parentDatabaseId: $dbid }` input where args where 0 returns the root items
- Add `parentId` and `parentDatabaseId` to the menu item itself

So if you need to need previous behaviour you just need to turn

```graphql
{
	menuItems {
      ...
    }
}
```

to

```graphql
{
	menuItems(where: { parentId: "0" }) {
      ...
    }
}
```

esamattis:
Closes #1260

This PR makes the `menuItems` connection flat by default for the reasons mentioned in #1260: Speed and ability request arbitrary deep menu structures without manually nesting  `childItems`.

Here's other changes in the PR

- Add `parentId` and `parentDatabaseId` where args to `menuItems` connection
- Add `parentId` and `parentDatabaseId` fields to the `MenuItem` type
- Add `order` field to `MenuItem`
- Add `databaseId` and deprecate `menuItemId` to match other types in WPGraphQL
- Adds tests for all these changes

I believe @chriszarate might be interested in these changes.


esamattis:
The get the root menu items only `parentId: "0"` can be used.

So to migrate existing code this is the required change:


```graphql
{
	menuItems {
      ...
    }
}
```

to

```graphql
{
	menuItems(where: { parentId: "0" }) {
      ...
    }
}
```

Issue #998, opened 2019-11-03
Update Resolvers for all ID inputs to behave consistently


jasonbahl:
See issue #997 

### Mutations
- [x] Comment mutations ( #2328 )
- [x] MediaItem mutations ( #2331 )
- [x] PostObject mutations ( #2338 )
- [x] TermObject mutations ( #2336 )
- [x] User mutations ( #2337 )
### Connections
- [x] Comment connections ( #2522 )
- [x] MenuItem connections ( #2523 )
- [x] PostObject connections (#2525 )
- [x] TermObject connections ( #2524 )

justlevine:
## To do
### Mutations
- [ ] Comment mutations ( #2328 )
- [ ] MediaItem mutations ( #2331 )
- [ ] PostObject mutations
- [ ] TermObject mutations
- [ ] User mutations
### Connections
- [ ] Comment connections
- [ ] MenuItem connections
- [ ] PostObject connections
- [ ] TermObject connections

justlevine:
@jasonbahl mutation prs are ready for review and merge.

[As mentioned](https://github.com/wp-graphql/wp-graphql/pull/2328#issuecomment-1089395266), connections are more complicated - I'll begin playing around with different ideas soon.

justlevine:
PRS for connection where args are up and ready for merge. Once that happens, we can close this issue _and_ I believe #1095

PR #1262, opened 2020-04-24
Add `parentId` and `parentDatabaseId` to `HierarchicalContentNode` interface

esamattis:
To match the new fields for `MenuItem` in #1261 

This was supposed to be a trivial addition but it appears that the post model used `parentId` internally as the parent database id which collided with the new resolver definition so I had to change it to `parentDatabaseId`.

For graphql api users no breaking changes but this *might* affect some plugins.

Intentionally sending the codesniff fixes in a separate commit so the actually change is more clear in its commit.

### DIFF
--- a/src/Model/Post.php
+++ b/src/Model/Post.php
@@ -318,86 +318,86 @@ class Post extends Model {
 			};
 
 			$this->fields = [
-				'ID'              => function() {
+				'ID'               => function() {
 					return $this->data->ID;
 				},
-				'post_author'     => function() {
+				'post_author'      => function() {
 					return ! empty( $this->data->post_author ) ? $this->data->post_author : null;
 				},
-				'id'              => function() {
+				'id'               => function() {
 					return ( ! empty( $this->data->post_type ) && ! empty( $this->data->ID ) ) ? Relay::toGlobalId( $this->data->post_type, $this->data->ID ) : null;
 				},
-				'post_type'       => function() {
+				'post_type'        => function() {
 					return isset( $this->data->post_type ) ? $this->data->post_type : null;
 				},
-				'authorId'        => function() {
+				'authorId'         => function() {
 					return isset( $this->data->post_author ) ? $this->data->post_author : null;
 				},
-				'date'            => function() {
+				'date'             => function() {
 					return ! empty( $this->data->post_date ) && '0000-00-00 00:00:00' !== $this->data->post_date ? Types::prepare_date_response( null, $this->data->post_date ) : null;
 				},
-				'dateGmt'         => function() {
+				'dateGmt'          => function() {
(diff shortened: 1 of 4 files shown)


src/Model/Post.php
```diff
@@ -406,49 +407,52 @@ protected function init() {
...
+				'modified'         => function() {
 					return ! empty( $this->data->post_modified ) && '0000-00-00 00:00:00' !== $this->data->post_modified ? $this->data->post_modified : null;
 				},
-				'modifiedGmt'     => function() {
+				'modifiedGmt'      => function() {
 					return ! empty( $this->data->post_modified_gmt ) ? Types::prepare_date_response( $this->data->post_modified_gmt ) : null;
 				},
-				'parentId'        => function() {
+				'parentId'         => function() {
+					return ( ! empty( $this->data->post_type ) && ! empty( $this->data->post_parent ) ) ? Relay::toGlobalId( $this->data->post_type, $this->data->post_parent ) : null;
```
CodeProKid, line 423: Hmmm 🤔 so this is going to cause a breaking change, since consumers would be expecting the numeric DB ID from this field previously. I think it's OK to make this breaking change if that's the direction we're going to take with:  https://github.com/wp-graphql/wp-graphql/issues/998.

src/Type/InterfaceType/NodeWithRevisions.php
```diff
@@ -26,12 +26,12 @@ public static function register_type( TypeRegistry $type_registry ) {
...
 						'resolve'     => function( Post $post, $args, AppContext $context, ResolveInfo $info ) {
 							// @codingStandardsIgnoreLine.
-							if ( 'revision' !== $post->post_type || ! isset( $post->parentId ) || ! absint( $post->parentId ) ) {
+							if ( 'revision' !== $post->post_type || ! isset( $post->parentDatabaseId ) || ! absint( $post->parentDatabaseId ) ) {
 								return null;
 							}
 
 							// @codingStandardsIgnoreLine.
-							return DataSource::resolve_post_object( $post->parentId, $context );
+							return DataSource::resolve_post_object( $post->parentDatabaseId, $context );
```
CodeProKid, line 34: FWIW, you could get around adding this field to the model by doing `$post->data->parent` here. Not saying that's the right way forward here, just throwing it out as a possible solution.

jasonbahl: @CodeProKid we don't load the whole object for the Parent in the Model though, eh? You're suggesting that we do load it in the Model?

CodeProKid: @jasonbahl nope, that'll return the parent DB ID though.

PR #1267, opened 2020-04-27
Removes the type-hinting from the Post model constructor

kidunot89:

🚨Please review the [guidelines for contributing](CONTRIBUTING.md) to this repository.

Removes the `\WP_Post` parameter type-hinting from the `Post` model constructor.

It seems like a simply change, but I don't know if this is here for a reason beyond aesthetics :man_shrugging: 

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Model/Post.php
+++ b/src/Model/Post.php
@@ -79,13 +79,13 @@ class Post extends Model {
 	/**
 	 * Post constructor.
 	 *
-	 * @param \WP_Post $post      The incoming WP_Post object that needs modeling.
-	 * @param integer  $owner_id  Owner of incoming WP_Post object.
+	 * @param \WP_Post|mixed $post      The incoming WP_Post object that needs modeling.
+	 * @param integer        $owner_id  Owner of incoming WP_Post object.
 	 *
 	 * @throws \Exception
 	 * @return void
 	 */
-	public function __construct( \WP_Post $post, $owner_id = null ) {
+	public function __construct( $post, $owner_id = null ) {
 
 		/**
 		 * Set the data as the Post object


src/Model/Post.php
```diff
@@ -79,13 +79,13 @@ class Post extends Model {
 	/**
 	 * Post constructor.
 	 *
-	 * @param \WP_Post $post      The incoming WP_Post object that needs modeling.
-	 * @param integer  $owner_id  Owner of incoming WP_Post object.
+	 * @param \WP_Post|mixed $post      The incoming WP_Post object that needs modeling.
+	 * @param integer        $owner_id  Owner of incoming WP_Post object.
```
jasonbahl, line 83: @kidunot89 I imagine this is somehow causing issues with WooGraphQL?

Can you expand so I can have more context on the problem at hand? 

I imagine WooCommerce doesn't use WP_Post objects? Might be good  to have a deeper understanding on _why_ and _what_ they're doing different here.

kidunot89: WooCommerce uses data stores that distributes objects that wraps the WP_Post object for the purpose adding an extra layer logic, decorators, hooks, etc that wrap the WP_Post object. Although they contain the WP_Post object data they don't extend the WP_Post class. Those are the objects that I provide as the `$data`.

CodeProKid: I'm pretty hesitant to change this... Without this tightly restricted, we can't make the same types of assumptions about the incoming object to be modeled.

jasonbahl: I recall when looking at some of the DataStores for WooCommerce, they load a lot of unnecessary data anytime you instantiate one of their classes. 

I'm curious if this would actually force revisiting some of those things and doing some things more efficiently. 

For example for each Product that's instantiated, it calls this method: 

https://github.com/woocommerce/woocommerce/blob/master/includes/data-stores/class-wc-product-data-store-cpt.php#L186

That then calls `get_the_terms()` for each product. 

This is a source of the `n+1` problem that WPGraphQL + Deferred resolvers attempts to reduce. 

We could, instead, load the `product_visibility` terms for _all_ Products at once with a deferred resolver and it could save some SQL queries and keep things loading efficiently. 

I'm sure there are even more cases of this `n+1`. I know the core WP REST API is littered with it.

kidunot89: The DataStores are the heart of WooCommerce, and those n+1 problems can probably be fixed with by overriding the individual DataStores in the [WC_Data_Store](https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-data-store.php#L35-L56) class and overloading the `read()` in a child class of the parent data store class.

jasonbahl:
Closing as duplicate of https://github.com/wp-graphql/wp-graphql/pull/1281/files

Issue #1268, opened 2020-04-27
Fix schema linting on builds


CodeProKid:
Currently, the schema linting is breaking on all of the builds. The issue is we're not doing all of the docker setup before trying to install the DB and running the command to generate the schema here: https://github.com/wp-graphql/wp-graphql/blob/d8bbb4ca0e58c51bd3525009efe2d105fb10855d/.github/workflows/schema-linter.yml#L35

We should probably just move this check into one of the other build steps rather than setting up docker again.

CodeProKid:
@jasonbahl I noticed you took a stab at this in #1278 but I think we should take a different approach.

I think we should add a `GENERATE_SCHEMA` env flag, and then add that to the matrix for a single build, like how we generate codecov. Then have _that_ run the CLI command while the environment is bootstrapped, and save the file as an artifact. Then we can reference the persisted artifact in the schema linting job. This way we don't have to bootstrap the entire environment again. Some GH docs on persisting artifacts during jobs: https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts

jasonbahl:
Ah nice. Ya, this is good stuff. I'll give it another go.

PR #1278, opened 2020-04-29
Bug/#1268 fix schema linting

jasonbahl:
Attempts to fix the Schema Linting in the Github Workflow


CodeProKid:
@jasonbahl looks like this still isn't working. I added some notes to the issue on how I think this should be fixed here: https://github.com/wp-graphql/wp-graphql/issues/1268#issuecomment-621928867

jasonbahl:
@CodeProKid 🤔 no. . .

I _did_ mean to merge this one: https://github.com/wp-graphql/wp-graphql/pull/1280

And maybe I screwed something up in the process 🤔 

There was a merge conflict so I checked out locally, made the fix, pushed to develop and Github marked this as merged

PR #1279, opened 2020-04-29
Add user meta to User Model

### LINKED ISSUE #479: Add core user meta to the User Type
Currently the core User type in the Schema doesn't include the user meta that is used by core, such as the fields here: https://github.com/WordPress/WordPress/blob/5c291d49de756db1dbfddfe3fd69709595800622/wp-includes/author-template.php#L116-L186

The User Type should include these core fields by default and support them in queries and `createUser` and `updateUser` mutations.

bahiirwa:
This PR add - status, activation key, login, pass user meta to the User Model and Object. As a follow up to Close #479 

Screenshots
Screenshot 2020-04-29 at 16 22 19

Where has this been tested?
---------------------------
**Operating System:** WIndows 10, Mac Os 10.15
**WordPress Version:** WP 5.4


src/Type/Object/User.php
```diff
@@ -92,6 +82,22 @@ public static function register_type() {
 						'type'        => 'String',
 						'description' => __( 'The nicename for the user. This field is equivalent to WP_User->user_nicename', 'wp-graphql' ),
 					],
+					'status'          => [
+						'type'        => 'String',
+						'description' => __( 'The status for the user. This field is equivalent to WP_User->user_status', 'wp-graphql' ),
+					],
+					'activation_key'          => [
```
jasonbahl, line 89: camelCase. Shoiuld be `activationKey`

src/Type/Object/User.php
```diff
@@ -92,6 +82,22 @@ public static function register_type() {
...
+						'type'        => 'String',
+						'description' => __( 'The activation_key for the user. This field is equivalent to WP_User->user_activation_key', 'wp-graphql' ),
+					],
+					'pass'          => [
```
jasonbahl, line 93: `Pass` is not descriptive to a consumer. It's an abbreviation, but the word "Pass" also has many meanings. We should consider being more explicit and name this field `password`.

src/Model/User.php
```diff
@@ -168,6 +172,18 @@ function( $cap ) {
 				'nicename'          => function() {
 					return ! empty( $this->data->user_nicename ) ? $this->data->user_nicename : null;
 				},
+				'status'          => function() {
+					return ! empty( $this->data->user_status ) ? $this->data->user_status : null;
+				},
+				'activation_key'          => function() {
+					return ! empty( $this->data->user_activation_key ) ? $this->data->user_activation_key : null;
+				},
+				'pass'          => function() {
```
jasonbahl, line 181: So we will need to make sure we're handling cap checks properly on all these fields. 

For example, the password should only be exposed to the current user (or users with proper caps to see it. . .if that's a thing? I don't think anyone _but_ the current user _can_ see a Password, eh? Even admins?

src/Type/Object/User.php
```diff
@@ -1,20 +1,10 @@
...
 
 use WPGraphQL\Data\DataSource;
 
-/**
- * Class User
- *
- * @package WPGraphQL\Type\Object
- */
 class User {
-
```
jasonbahl, line 14: why are these docblocks are being removed?

src/Type/Object/User.php
bahiirwa: I agree. However, I was distracted by the way WP names this in the WP_user class. Do we move away from that?

src/Model/User.php
bahiirwa: Funny thing is WP does not store actual passwords. They hash the information.

> This is why WordPress passwords are hashed: The only reason the stored value needs to be used for is to compare whether the password that the user enters on next login matches what was originally stored as the password. WordPress never displays the password anywhere, so it never needs to be able to decrypt it. And because there is no need for that, it’s safer to use a hashing function which basically makes it irreversible.
Ref: https://felix-arntz.me/blog/storing-confidential-data-in-wordpress/

Should we still work to query this?

src/Type/Object/User.php
bahiirwa: That's weird. Will fix it.

src/Model/User.php
jasonbahl: Ya, I'm actually thinking we don't need to expose passwords at all. 

My initial thought was that you could build a UI to edit your user profile and you'd be able to see your current password in a `<input type="password" value="{user.password}" />` field, but it looks like even WordPress core doesn't show a password field on your profile, the only option is to generate a new password. 

So ya, I think we can leave password out altogether.

src/Type/Object/User.php
jasonbahl: ya, so WordPress internals are `snake_case` but GraphQL uses `camelCase` for field names and `PascalCase` for Type names, so at the layers that define the GraphQL schema, we are using `camelCase` for fields. 

We don't want to enforce this centrally. If you _do_ want to use other cases in your field names for custom plugins, etc, you can, but in general we want to follow common practices in the greater GraphQL ecosystem.

src/Model/User.php
CodeProKid: I also think passwords shouldn't be exposed in the schema. FWIW we explicitly remove the user_pass from the user object passed in to the model as an extra security measure. This way the data isn't being passed around in filters and such https://github.com/wp-graphql/wp-graphql/blob/dbdf14f5b6adb12d8343f9d3b69cb7a96c7a3b3c/src/Model/User.php#L51-L52

bahiirwa:
With mock data, this works fine as screenshot shows. 

I need help:
- Any more pointers on how to add the "pseudo admin"to return proper data?
- Writing tests once the code is done.

Screenshot 2020-05-05 at 18 29 43

PR #1280, opened 2020-04-29
Upgrade graphql php

jasonbahl:
This upgrades the underlying GraphQL PHP library from `0.12.6` to `0.13.8`. 

With this upgrade WPGraphQL is now also updating the minimum PHP requirement to PHP 7.3, as recommended by WordPress.org (https://wordpress.org/about/requirements/)

This PR includes changes in the main plugin file, README and tests to adjust to the new 7.3+ requirement.


CodeProKid:
@jasonbahl I think the PHP minimum requirement should only be bumped to 7.1, since that is what is required by the PHP GraphQL lib. The 7.3 minimum from core is just a _recommendation_. Core actually supports down to 5.6 currently.

I think if we can reasonably support down to PHP 7.1, we should do so.

jasonbahl:
PHP 7.3+ is what PHP itself supports. 

I think the official active supported versions of PHP is what we should follow (and that's what WordPress recommendations are based on). 

That way we can also commit to a schedule for when we make the bumps ourselves instead of arbitrarily bumping because a dependency made the bump.

jasonbahl:
@CodeProKid I think maybe short term we could jump to 7.1, put a date on jumping to 7.3, then follow the PHP active support lifecycle after that?

esamattis:
> then follow the PHP active support lifecycle after that?

That's probably not enough. For example Ubuntu Bionic LTS which is supported by Canonical until 2023 has PHP 7.2 which is really popular. The next LTS was just released a week ago.

I'd recommend we just support whatever `graphql-php` supports now which is 7.1. Should not be too much effort. Anything older would be painful since we would get left behind with graphql-php security updates and would have to backport them.

CodeProKid:
Yeah, I agree with @esamattis these older versions of PHP are often used and supported for a long time after PHP has EOL'd them. I think we should bump our required version when it makes sense to, like in this case to support a new version of an important dependency. Supporting 7.1 & 7.2 wouldn't add a significant burden on us, so if we can reasonably support those versions, I say we do it.

jasonbahl:
aight. I'll update the PR. We can revisit later.

PR #1281, opened 2020-04-29
Make model layer more flexible

CodeProKid:

🚨Please review the [guidelines for contributing](CONTRIBUTING.md) to this repository.

This makes the model layer a little more flexible so that _any_ data can be pushed through the model classes for modeling. I also added a param to a filter that was needed per some convo's with @jasonbahl and @kidunot89 

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Model/Model.php
+++ b/src/Model/Model.php
@@ -45,6 +45,13 @@ abstract class Model {
 	protected $owner;
 
 	/**
+	 * Stores the result of whether the owner of this object is the current user.
+	 *
+	 * @var bool $owner_matches_current_user
+	 */
+	protected $owner_matches_current_user;
+
+	/**
 	 * Stores the WP_User object for the current user in the session
 	 *
 	 * @var \WP_User $current_user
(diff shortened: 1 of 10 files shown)


src/Model/Model.php
```diff
@@ -436,10 +436,11 @@ protected function prepare_fields() {
 		 * @param string   $visibility   The visibility setting for this piece of data
 		 * @param null|int $owner        The user ID for the owner of this piece of data
 		 * @param \WP_User $current_user The current user for the session
+		 * @param object   $data         The raw data passed into the model
 		 *
 		 * @return array
 		 */
-		$this->fields = apply_filters( 'graphql_return_modeled_data', $this->fields, $this->get_model_name(), $this->visibility, $this->owner, $this->current_user );
+		$this->fields = apply_filters( 'graphql_return_modeled_data', $this->fields, $this->get_model_name(), $this->visibility, $this->owner, $this->current_user, $this->data );
```
jasonbahl, line 443: I believe we still need a few more filters in this class, eh?

Like `$this->owner` I think needs to be filterable. For example, Orders in WooCommerce are not "owned" by the `post->post_author` so the `owner` needs to be filters for `Posts` of `$this->data->post_type === 'Order'`

CodeProKid: I don't think so. I think people will just want to use the filters in the `get_visibility` method: https://github.com/wp-graphql/wp-graphql/blob/dbdf14f5b6adb12d8343f9d3b69cb7a96c7a3b3c/src/Model/Model.php#L187-L242

jasonbahl: @CodeProKid I think it probably still should be filtered early when it's set. 

`$this->owner` gets passed to a lot of filters, and if it's not accurate (like the case of WooCommerce orders), how should these other filters use it?

If the `__construct` did something like: 

```
$this->owner = apply_filters( 'graphql_model_owner', $owner, ...$other_args );
```

Then something like WooGraphQL could filter the owner based on the `post_type === order`, and all other filters that get `$owner` as an argument would have it already properly filtered for use within their filtering. 

There could easily be a case where filtering fields might need to take into granular details about the `$owner` to decide how to resolve a field, for example.

CodeProKid: @jasonbahl I'm a little hesitant to add a filter in the `__construct` since a lot of the data hasn't been setup by then, so I'm not sure how useful it would be. I ended up adding a filter to the `owner_matches_current_user` method in this commit: https://github.com/wp-graphql/wp-graphql/pull/1281/commits/ea0376a53ee9c4ab61bff9124c77e80abac03db7 LMK if you think that still wouldn't suffice and we can revisit.

PR #1286, opened 2020-05-01
Bug/#acf-130-field-names-should-follow-spec

jasonbahl:
- This formats Field Names and Type Names to follow the GraphQL Spec and ensure there are no spaces or unallowed characters in the names.

Closes https://github.com/wp-graphql/wp-graphql-acf/issues/130


jasonbahl:
Reverting, as this causes breaking changes to many ecosystem extensions.

PR #1302, opened 2020-05-08
Add `graphql_ast` filters

### LINKED ISSUE #112: Filter the Query to allow for caching and using cached ASTs
See: https://github.com/graphql/graphql-js/issues/158#issuecomment-135158371

To enable named/cached queries, we should filter the `$query` to allow for a $_REQUEST to send the key to a cached AST and use that AST to run the query. 

Lee Byron points out some of the benefits of doing this in that github thread.

But this would allow for clients to send just a small key that is a reference to a query, and that key would be used to get the full query AST from the cache and process the query. 

So instead of a client sending something like: 

```
{
  comments(first: 5) {
    pageInfo {
      startCursor
      endCursor
      hasNextPage
      hasPreviousPage
    }
    edges {
      cursor
      node {
        id
        content
        commentedOn {
          __typename
        }
      }
    }
  }
  posts(first: 5) {
    pageInfo {
      startCursor
      endCursor
      hasNextPage
      hasPreviousPage
    }
    edges {
      cursor
      node {
        id
        link
        title
        date
      }
    }
  }
}
```

They could just send something like: 

```
id_of_the_cached_query
```

So the upload time by a network to pass the query to the server would be extremely minimal compared to sending the data for the full query string.

esamattis:

This add two filters `graphql_ast` and `graphql_pre_ast` which allow AST of the GraphQL queries to be inspected, transformed and cached. There are more detailed docs in the doc blocks.

Closes #112

### DIFF
--- a/src/Data/Connection/PostObjectConnectionResolver.php
+++ b/src/Data/Connection/PostObjectConnectionResolver.php
@@ -87,7 +87,7 @@ class PostObjectConnectionResolver extends AbstractConnectionResolver {
 
 		$query = new \WP_Query( $this->query_args );
 
-		if ( isset( $query->query_vars['suppress_filters'] ) && true == $query->query_vars['suppress_filters'] ) {
+		if ( isset( $query->query_vars['suppress_filters'] ) && false !== (bool) $query->query_vars['suppress_filters'] ) {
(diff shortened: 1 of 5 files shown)


esamattis:
This now ready for review but not merging since this actually requires update to webonyx/graphql-php I PRed here

https://github.com/webonyx/graphql-php/pull/658


I also added tests for the filter and extended the error reporting tests.

PR #1316, opened 2020-05-20
make PostObjectLoader.loadKeys robust against non-existent keys

rburgst:

We have a complex setup with wpgraphql + acf. In our situation `PostObjectLoader.loadKeys()` is called with post ids that no longer exists. 
In such a case the loaded `post_object` will be empty which will lead to errors when trying to dereference attributes of this object.

- did not find any

```
[Wed May 20 14:34:54.489400 2020] [php7:notice] [pid 798] [client 10.42.1.0:56692] PHP Notice:  Trying to get property 'post_type' of non-object in /var/www/html/wp-content/plugins/wp-graphql-0.8.4/src/Data/Loader/PostObjectLoader.php on line 114,
```

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** linux

**WordPress Version:** 5.4.1

### DIFF
--- a/src/Data/Loader/PostObjectLoader.php
+++ b/src/Data/Loader/PostObjectLoader.php
@@ -93,70 +93,73 @@ class PostObjectLoader extends AbstractDataLoader {
 				$loaded_posts[ $key ] = null;
 			}
 
-			/**
-			 * If there's a Post Author connected to the post, we need to resolve the
-			 * user as it gets set in the globals via `setup_post_data()` and doing it this way
-			 * will batch the loading so when `setup_post_data()` is called the user
-			 * is already in the cache.
-			 */
-			$context     = $this->context;
-			$user_id     = null;
-			$post_parent = null;
+            if (!empty($post_object)) {
+                /**
+                 * If there's a Post Author connected to the post, we need to resolve the
+                 * user as it gets set in the globals via `setup_post_data()` and doing it this way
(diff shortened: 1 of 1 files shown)


src/Data/Loader/PostObjectLoader.php
```diff
@@ -93,70 +93,73 @@ function( $split, \WP_Query $query ) {
...
-			$loaded_posts[ $key ] = $load_dependencies->then(
-				function() use ( $post_object ) {
-					$post = new Post( $post_object );
-					if ( ! isset( $post->fields ) || empty( $post->fields ) ) {
-						return null;
-					}
-					return $post;
-				}
-			);
+            if (!empty($post_object)) {
```
jasonbahl, line 96: @rburgst we have a check above that sets the `$loaded[ $key ] = null`. 

Perhaps we just change that to an if/else statement. 

```php
if ( ! $post_object instanceof \WP_Post ) {
   $loaded_posts[ $key ] = null;
} else {
   // all the other stuff here
}
```

jasonbahl:
@rburgst thanks for the PR! Really appreciate it. I left a comment about a small change.

src/Data/Loader/PostObjectLoader.php
rburgst: good point, done

jasonbahl: Thanks!

Issue #1319, opened 2020-05-21
With the #1290 change, WordPress no longer routes to `/graphql` when running on a TCP port other than normal (80,443).


sawanoboly:
`HTTP_HOST`, changed in #1290, contained the port number, but `SERVER_NAME` does not contain the port number.
Therefore, the subsequent string length comparison may not work well in some cases. It was working fine before this change.

checked environments

- graphql 0.9.0
- php 7.3, 7.4
- php-fpm, litespeed 1.6(lsphp)

For example, `/graphql` does not work with the following URLs for the purpose of development or etc.

- `http://127.0.0.1:8080`
- `https://127.0.0.1:8443`

It can be reproduced with WordPress in the following repository.

- https://github.com/getshifter/shifter-local

First, if you run the container with `docker-compose up`, the
WordPress will be available at `https://127.0.0.1:8443`.

Then you can install wp-graphql 0.9.0 from the container host by following the steps below.

```
$ docker-compose exec wp git clone https://github.com/wp-graphql/wp-graphql -b v0.9.0 /var/www/html/web/wp/wp-content/plugins/wp-graphql
```

sawanoboly:
As another example, here is `$_SERVER` when accessed with Openlitespeed.

```
array(28) {
  ["PATH"]=>
  string(13) "/bin:/usr/bin"
  ["HTTP_ACCEPT"]=>
  string(63) "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
  ["HTTP_ACCEPT_ENCODING"]=>
  string(13) "gzip, deflate"
  ["HTTP_ACCEPT_LANGUAGE"]=>
  string(5) "ja-jp"
  ["HTTP_CONNECTION"]=>
  string(10) "keep-alive"
  ["HTTP_COOKIE"]=>
  string(293) "****"
  ["HTTP_HOST"]=>
  string(14) "127.0.0.1:8088"
  ["HTTP_USER_AGENT"]=>
  string(117) "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_4) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.1 Safari/605.1.15"
  ["HTTP_UPGRADE_INSECURE_REQUESTS"]=>
  string(1) "1"
  ["DOCUMENT_ROOT"]=>
  string(29) "/usr/local/lsws/wordpress/app"
  ["REMOTE_ADDR"]=>
[...]

PR #1320, opened 2020-05-21
use SERVER_PROTOCOL to check request via http

sawanoboly:
It's one of the ideas that solves the problem in #1319.

Where has this been tested?
---------------------------
**Operating System:** …

- alpine linux 3.11
- PHP 7.3.16

**WordPress Version:** …

5.4.1

### DIFF
--- a/src/Router.php
+++ b/src/Router.php
@@ -121,16 +121,8 @@ class Router {
 		}
 
 		// If before 'init' check $_SERVER.
-		if ( isset( $_SERVER['SERVER_NAME'] ) && isset( $_SERVER['REQUEST_URI'] ) ) {
-			$haystack = wp_unslash( $_SERVER['SERVER_NAME'] )
-				. wp_unslash( $_SERVER['REQUEST_URI'] );
-			$needle   = site_url( self::$route );
-
-			// Strip protocol.
-			$haystack = preg_replace( '#^(http(s)?://)#', '', $haystack );
-			$needle   = preg_replace( '#^(http(s)?://)#', '', $needle );
-			$len      = strlen( $needle );
-			return ( substr( $haystack, 0, $len ) === $needle );
+		if ( isset( $_SERVER['SERVER_NAME'] ) && isset( $_SERVER['SERVER_PROTOCOL'] ) ) {
+			return ( strpos($_SERVER['SERVER_PROTOCOL'], 'HTTP') !== false );
 		}
 
 		return false;


src/Router.php
```diff
@@ -121,16 +121,8 @@ public static function is_graphql_http_request() {
 		}
 
 		// If before 'init' check $_SERVER.
-		if ( isset( $_SERVER['SERVER_NAME'] ) && isset( $_SERVER['REQUEST_URI'] ) ) {
-			$haystack = wp_unslash( $_SERVER['SERVER_NAME'] )
-				. wp_unslash( $_SERVER['REQUEST_URI'] );
-			$needle   = site_url( self::$route );
```
jasonbahl, line 127: This conditional checks to see if the `/graphql` route (or whatever route GraphQL is set to resolve at) is being requested so that plugin authors can do things like: 

```
if ( is_graphql_http_request() ) {
  // do something just for graphql
} else {
  // do something for all other WP requests
}
```

This PR takes away the check to see if the route is being called, so this conditional wouldn't be doing what it's intended to be doing. 

I'm open to changing this because things clearly broke with the release, but we have to make sure the function still does what it's intended to do and check that the route is indeed the graphql route.

jasonbahl:
Thanks for opening the issue and the PR to resolve it! Much appreciated. 

I left a note on a change we need to make.

jasonbahl:
@sawanoboly I'm thinking we revert to what we had (it's been largely working for quite some time), but add a filter to allow folks to return true/false if they have other conditions they'd like to check to ensure `is_graphql_http_request()` should be true or false.

src/Router.php
```diff
@@ -121,16 +121,8 @@ public static function is_graphql_http_request() {
...
-
-			// Strip protocol.
-			$haystack = preg_replace( '#^(http(s)?://)#', '', $haystack );
-			$needle   = preg_replace( '#^(http(s)?://)#', '', $needle );
-			$len      = strlen( $needle );
-			return ( substr( $haystack, 0, $len ) === $needle );
+		if ( isset( $_SERVER['SERVER_NAME'] ) && isset( $_SERVER['SERVER_PROTOCOL'] ) ) {
+			return ( strpos($_SERVER['SERVER_PROTOCOL'], 'HTTP') !== false );
```
jasonbahl, line 125: @sawanoboly @WybrenKoelmans 

Let me know if this solution works for ya'll:

```
	/**
	 * Returns true when the current request is a GraphQL request coming from the HTTP
	 *
	 * NOTE: This will only indicate whether the GraphQL Request is an HTTP request. Many features
	 * need to affect _all_ GraphQL requests, including internal requests using the `graphql()`
	 * function, so be careful how you use this to check your conditions.
	 *
	 * @return boolean
	 */
	public static function is_graphql_http_request() {
		
		/**
                  *Default value is false
                  *
                  * @param boolean $is_graphql_http_request Whether the request is a graphql HTTP request. Default false.
                  */
		$is_graphql_http_request = apply_filters( 'is_graphql_http_request', false );
		
		// Support wp-graphiql style request to /index.php?graphql.
		if ( true !== $is_graphql_http_request && isset( $_GET[ self::$route ] ) ) {
			$is_graphql_http_request = true;
		}

		// If before 'init' check $_SERVER.
		if ( true !== $is_graphql_http_request && isset( $_SERVER['HTTP_HOST'] ) && isset( $_SERVER['REQUEST_URI'] ) ) {
			$haystack = wp_unslash( $_SERVER['HTTP_HOST'] )
				. wp_unslash( $_SERVER['REQUEST_URI'] );
			$needle   = site_url( self::$route );

			// Strip protocol.
			$haystack = preg_replace( '#^(http(s)?://)#', '', $haystack );
			$needle   = preg_replace( '#^(http(s)?://)#', '', $needle );
			$len      = strlen( $needle );
			$is_graphql_http_request = ( substr( $haystack, 0, $len ) === $needle );
		}

		return $is_graphql_http_request;
	}

```

jasonbahl:
Closing in favor of: https://github.com/wp-graphql/wp-graphql/pull/1323

PR #1330, opened 2020-05-25
Add tests for draft pages in menus

esamattis:
Found a bug where wp-graphql would crash when fetching menu items list that has draft posts connected to it. But I realized too late that I was running an old version of wp-graphql and it was already fixed in the latest one 🤦‍♂️.

But before realizing it I already managed to write tests for it. Sending them anyway because there doesn't seem to be auth tests for menus ATM.


jasonbahl:
Really glad I fixed the bug. Not so glad we don't have tests for this yet. Really glad you've changed that. 🙌

PR #1337, opened 2020-06-03
Preview support. . .and so much more

### LINKED ISSUE #347: Nest one-to-one relationships
Right now we have several one-to-one relationships in the Schema, such as 

```
post( id:"...") { 
   featuredImage { 
      id
      caption 
   } 
}
```

But often time there is edge data, that's useful in the space between the 2 connections. 

For example, we have some "featured image settings" in a metabox on our post edit screen. From a Schema perspective, the settings make sense under: 

```
post(...) { 
   featuredImage { 
       settings { 
           someField 
       } 
   } 
}
```

But that's not really easy to resolve with the current architecture where `featuredImage` returns a `MediaItem` type, and we don't want `settings` on that Type, as it's really only relevant in the context of the `post<->featuredImage` connection. . .it's "edge" data.

So, after chatting with Lee Byron at GraphQL Summit, he actually proposed one-to-one relations like this have a structure like: 

```
{
  post() {
     featuredImage {
         settings {
             someEdgeField
         }
         node { 
            id
            caption
         }
     }
  }
}
```

### LINKED ISSUE #410: Fields should be consistently unescaped
WordPress generally follows the pattern "sanitize data going into the database, escape (late) on output." However, WP_User and WP_Term fields break this pattern. The simple reason is that `sanitize_user_field` and `sanitize_term_field` call `esc_*` functions, so they go into the database escaped:

https://github.com/WordPress/WordPress/blob/6e665d1f1523a0aeec1769d243ced9c3ff0b1434/wp-includes/user.php#L1256

https://github.com/WordPress/WordPress/blob/6e665d1f1523a0aeec1769d243ced9c3ff0b1434/wp-includes/taxonomy.php#L1538

As far as I can tell, this is a historical quirk/mistake that has been preserved for compatibility reasons. Both the REST API and current versions of WP-GraphQL output field data consistently—which is to say, they preserve the _inconsistency_ of WP Core and output unescaped data for most fields and escaped data for User and and Term fields.

I'd like to suggest that WP-GraphQL correct the inconsistency of WP Core and _unescape_ User and Term field data. That way the consuming app can treat `String` fields of all types consistently, avoiding the need to conditionally escape. Interested in thoughts / counter-arguments.

### LINKED ISSUE #887: No sub-categories showing on posts if also getting featuredImage
This is a strange bug, if you try to to get all posts, and do a nested categories query, and also ask for the featuredImage on the post, then the categories come back with no sub-categories. If you remove the featured image query, then you get all the sub-categories!

See the two screen shots, the only difference is the `featuredImage`.

Screen Shot 2019-06-23 at 6 20 52 PM
Screen Shot 2019-06-23 at 6 21 01 PM

### LINKED ISSUE #928: Introduce showInFlatList to Pages, Posts, or Custom Post Types
So the default behavior of querying a post or page is to get all root elements, but not their children. Like so:

```
query GET_PAGES {
  pages {
    nodes {
      title
      childPages {
        nodes {
          title
        }
      }
    }
  }
}
```

But what if you want t get all elements, even the children? As @jasonbahl told me, in `categories` the `showInFlatList` argument got introduced. So I suggest the same for this.

Also there is an example here: https://github.com/progital/gatsby-wpgraphql-blog-example/blob/f7695fe41ee40ec98638949d954422d8ad3be743/gatsby/createPages.js#L4-L26

This actually gets you all the elements, even the child-elements. Which doesn't make sense. So the where filter should be checked with this too. I'll split it up in another ticket: #929

### LINKED ISSUE #1035: WPGraphQL should call `html_entity_decode()` for all content fields
By default WordPress encodes HTML entities before they are saved to database. This can be convinient when templating HTML strings in PHP because developers don't have to think about it.

But with WPGraphQL this is a burden when content is rendered via React or any other modern tool that has HTML entity encoding build-in because you'll get double html entity encoding.

There are tools that allow HTML entity decoding in JS but they neither intentionally incomplete like [unescape][] in Lodash or too large like [he][] or [ent][] to be bundled into the browser bundles. But even if we had good tools to do it, it is very cumbersome because developers would have to remember to use them basically every single time when content is being rendered from WP. This is a common issue we have faced when building real world headless sites and apps using React and React Native with WPGraphQL.

So we end up building HTML entity decoding as WPGrapQL filters on the PHP side:
https://gist.github.com/esamattis/54abd801b078056f9e646d63d47ccc4c

I believe tools like React are the major consumers of WPGraphQL and so **I would like to propose that WPGraphQL would add build-in HTML entity decoding** for the content fields.

By content fields I mean text content which is added by the content creators. Such are title, post content, excerpts, ACF text fields etc. Examples of non-content fields would be url, slugs, id where decoding could break something if they where to contain strings that match HTML entities.

[unescape]: https://lodash.com/docs/4.17.11#unescape
[he]: https://bundlephobia.com/result?p=he@1.2.0
[ent]: https://bundlephobia.com/result?p=ent@2.2.0

### LINKED ISSUE #1041: Custom post type should get connected to users if it supports authors.
When registering a custom post type, we add connections to things like terms and featured media if they support those things, but we don't add a connection on users to that post type if the post type supports `authors`. Let's add that connection dynamically when the post type is registered.

### LINKED ISSUE #1078: childPages vs parent
I'm noticing that there's a discrepancy between how child and parent nodes are handled. For parents we have a PostObjectUnion type but children are the same type as the current node. Seems like they should work in the same way, though I may be missing something! The naming is also different `childPages` vs `parent`, should it also be `parentPage`?

```gql
{
  pages {
    nodes {
      title
      childPages {
        nodes {
          title
          parent {
            __typename
            ... on Page {
              title
            }
          }
        }
      }
    }
  }
}
```

```json
{
  "data": {
    "pages": {
      "nodes": [
        {
          "title": "Sample Page",
          "childPages": {
            "nodes": [
              {
                "title": "Child page",
                "parent": {
                  "__typename": "Page",
                  "title": "Sample Page"
                }
              }
            ]
          }
        }
      ]
    }
  }
}
```

### LINKED ISSUE #1106: Revisions should use parent post's categories
```
query {
  revisions {
    nodes {
      ... on Post {
        categories(where: {shouldOutputInFlatList: true}) {
          nodes {
            id
          }
        }
      }
    }
  }
}

```

The response:
```
{
  "data": {
    "revisions": {
      "nodes": [
        {
          "categories": {
            "nodes": []
          }
        }
      ]
    }
  }
}
```

This query will not return categories, as Wordpress Core does not support revisions for categories. I think we should still return the parent post's categories, rather return an empty array of categories. Seems like this should be an easy change in the [TermObjectConnectionResolver](https://github.com/wp-graphql/wp-graphql/blob/dd11205fc9ef572d0cbb5ddcc32d9d58b95475cc/src/Data/Connection/TermObjectConnectionResolver.php#L130), if we replace the `source ID` with the `parent ID` for revision post types.

### When reporting a _bug_, please be sure to include the following:
- [x] A descriptive title
- [x] An *isolated* way to reproduce the behavior (example: GitHub repository with code isolated to the issue that anyone can clone to observe the problem)
- [x] What version of `WPGraphQL` you're using, and the platform(s) you're running it on
WPGraphQL 0.5.0, OSX Mojave
- [x] What other plugins you're using
WPGraphQL ACF
- [x] The behavior you expect to see, and the actual behavior

### When you open an _issue_ for a feature request, please add as much detail as possible:
- [x] A descriptive title
- [x] A description of the problem you're trying to solve, including *why* you think this is a problem
- [x] An overview of the suggested solution
- [x] If the feature changes current behavior, reasons why your solution is better

### LINKED ISSUE #1162: Preview revisions don't have any ACF data
WordPress: Version 5.3.2
WPGraphQL: Version 0.6.3
WPGraphQL ACF: Version 0.3.1
Advanced Custom Fields Pro: Version 5.8.7

I'm currently trying to query revisions to set up WP previews. When I query normal revisions they have all of the data I expect, but the preview revision doesn't have any acf data at all. All acf fields are null or just an empty array for flexible content. It seems like normal WP fields work fine. I just tested `title` and `content` here. 

My test query:
```
{
  revisions(first: 2) {
    nodes {
      __typename
      ... on Page {
        title
        content
        id
        databaseId
        acfFields {
          text
          repeater {
            item
          }
          flexibleContent {
            ... on Page_Acffields_FlexibleContent_Text {
              text
            }
            ... on Page_Acffields_FlexibleContent_Image {
              image {
                uri
              }
            }
          }
        }
      }
    }
  }
}
```
_The first item in the response is the preview revision and the second item is a regular revision._

response:
```
{
  "data": {
    "revisions": {
      "nodes": [
        {
          "__typename": "Page",
          "title": "Sample PREVIEW Page",
          "content": "\n<p>This is the preview revision of my sample page!!!!! (please work)</p>\n",
          "id": "cmV2aXNpb246MTY=",
          "databaseId": 16,
          "acfFields": {
            "text": null,
            "repeater": null,
            "flexibleContent": []
          }
        },
        {
          "__typename": "Page",
          "title": "Sample Page",
          "content": "\n<p>This is my sample page!</p>\n",
          "id": "cmV2aXNpb246MTU=",
          "databaseId": 15,
          "acfFields": {
            "text": "This is some regular text",
            "repeater": [
              {
                "item": "repeater item 1"
              },
              {
                "item": "repeater item 2"
              },
              {
                "item": "repeater item 3"
              }
            ],
            "flexibleContent": [
              {
                "text": "this is flexible content text~!!"
              },
              {
                "image": {
                  "uri": "sample-page/aditya-chinchure-2yrocbpcvey-unsplash/"
                }
              }
            ]
          }
        }
      ]
    }
  }
}
```

You can reproduce this by adding some acf fields to your page, clicking the preview button to generate the preview revision, then querying for your page revisions.

I just put together a very minimal reproduction of this. Here is the acf field group import json file.
[acf-export-2020-02-22.json.zip](https://github.com/wp-graphql/wp-graphql/files/4238542/acf-export-2020-02-22.json.zip)

### LINKED ISSUE #1169: Decode object titles
Task: Make sure all the object titles are decoded.

Bug first mentioned here: https://github.com/wp-graphql/wp-graphql-woocommerce/issues/244

### LINKED ISSUE #1287: CommentAuthor.email should be limited to users with `moderate_comments` capability
The `email` field on the `CommentAuthor` model should be limited to users with the `moderate_comments` capability as the email address is not considered a public field in WordPress core.

### LINKED ISSUE #1292: PostObjectUnion vs ContentNode Interface
Is there a reason to keep the PostObjectUnion type around? Doesn't ContentNode achieve the same thing but in a better way?

For example if you want the uri of the parent post you have to do:
```graphql
      parent {
        ... on Post {
          uri
        }
        ... on Page {
          uri
        }
        ... on TeamMember {
          uri
        }
        # etc
      }
```

This would be alot nicer:
```graphql
      parent {
        uri
      }
```

### LINKED ISSUE #1312: Add `asPreview` input argument to single post object RootQuery entry points in the graph
In order to preview a Post, we should add an `asPreview` input argument on RootQuery entry points for single content nodes. 

For example: 

```
{
  post( id: 123, idType: DATABASE_ID, asPreview: true ) {
    id
    title
    content
    ...NodeWithPreview {
       isPreview # <-- new boolean field
    }
  }
}
```

or

```
{
  contentNode( id: 123, idType: DATABASE_ID, asPreview: true ) {
    ...on Post {
      ...PostFragment
    }
  }
}
```

This would allow us to set the global `$wp_query->is_preview` state so that underlying resolution of previews could execute properly.

jasonbahl:
## Issues Addressed:

- closes #1312 (see [details](https://github.com/wp-graphql/wp-graphql/issues/1312#issuecomment-638882448))
- closes #1292 (see [details](https://github.com/wp-graphql/wp-graphql/issues/1292#issuecomment-638885009))
- closes #1287 (see [details](https://github.com/wp-graphql/wp-graphql/issues/1287#issuecomment-638915789))
- closes #1169 (see [details](https://github.com/wp-graphql/wp-graphql/issues/1169#issuecomment-638920549))
- closes #1162 (see [details](https://github.com/wp-graphql/wp-graphql/issues/1312#issuecomment-638882448))
- closes #1106 (see [details](https://github.com/wp-graphql/wp-graphql/issues/1312#issuecomment-638882448))
- closes #1078 (see [details](https://github.com/wp-graphql/wp-graphql/issues/1078#issuecomment-639045437))
- closes #1041 (see [details](https://github.com/wp-graphql/wp-graphql/issues/1041#issuecomment-639053863))
- closes #1035 (see [details](https://github.com/wp-graphql/wp-graphql/issues/1035#issuecomment-638327666))
- closes #928 (see [details](https://github.com/wp-graphql/wp-graphql/issues/928#issuecomment-638230255))
- closes #887 (see [details](https://github.com/wp-graphql/wp-graphql/issues/887#issuecomment-639067361))
- closes #410 (see [details](https://github.com/wp-graphql/wp-graphql/issues/1035#issuecomment-638327666))
- closes #347 (see [details](https://github.com/wp-graphql/wp-graphql/issues/347#issuecomment-639071772))

## Breaking Changes
- `Comment.author` no longer returns a CommentAuthorUnion, it's a one-to-one connection returning an edge/node where the node is of the Commenter Interface type (which is shared by User and CommentAuthor) types.
- Make `Comment.parent` field a formal one-to-one connection returning an edge/node instead of the parent directly.
- Change `Comment.children` field to `Comment.replies`.
- Makes connections from `NodeWithFeaturedImage->MediaItem` a formal one-to-one connection returning an edge/node where the node is of MediaItem type.
- Makes `Comment.commentedOn` field a formal one-to-one connection returning an edge/node where the node is a ContentNode Interface
- Makes `ContentNode.revisionOf` field a formal one-to-one connection returning an edge/node where the node is a ContentNode interface
- Makes `HierarchicalContentNode.parent` a formal one-to-one connection returning an edge/node where the node is of ContentNode Interface
- Makes `ContentNode.terms` field a formal connection to the TermNode type. Previously was a listOf.
- Changes `PostObject.editLock` field to `ContentNode.editingLockedBy` and is a one-to-one connection which returns an edge/node. The node is of type User, and there's an edge field `lockTimestamp`
- Changes `PostObject.editLast` field to `ContentNode.lastEditedBy` and is a one-to-one connection retuning an edge/node where the Node is of type User.
- Changes the `PostObject.author` field to be a formal one-to-one connection between a Post and the Author, returning an edge/node where the Node is of type User.
- Remove `shouldOnlyIncludeConnectedItems` and `shouldOutputInFlatList` arguments for hierarchical term queries. The default now is a flat list and that can be overridden by passing an explicit parent in the where args.
- Remove EditLock Type from the Schema
- Removed `PostObject.terms`, `PostObject.termNames`, `PostObject.termSlugs` fields in favor of formal connections.
- Removed `Taxonomy.connectedPostTypeNames` in favor of `Taxonomy.connectedPostTypes` connection. Converted `Taxonomy.connectedPostTypes` to formal connection instead of listOf.

### New Features
- Formal support for one-to-one connections. Can declare `oneToOne => true` when registering a connection, and use `$resolver->oneToOne()->get_connection()` to resolve using any resolvers extending the AbstractConnectionResolver class.
- Introduce Commenter Interface. An Interface shared by Commenters (CommentAuthor and User) when querying for the author of a Comment
- Support resolution of PostObject->Comments connections when queried from revision nodes (will resolve using the parent node's connections)
- Adds Taxonomy.connectedContentTypes field as a formal connection from Taxonomy->ContentType
- Adds `ContentType.contentNodes` field that returns a connection of content nodes of that post type
- Adds a PostObject.preview field as a one-to-one connection returning the Preview node for the Post
- Add support to AbstractConnectionResolver to support resolution of one-to-one connections. Ex: $resolver->one_to_one()->get_connection();`. This returns the `edge/node` shape to fulfill the connection.
- Deprecate AbstractConnectionResolver->setQueryArg for ->set_query_arg. And set_query_arg() returns instance of the resolver.
- Added `databaseId` field to `Comment` type
- Filter revisions to use meta from parent node by default.
- Add filter "graphql_resolve_revision_meta_from_parent" to allow certain meta fields (ACF, for example) to bypass using the parent node meta for resolution and use the revision node's meta
- Various fields in the Post model now resolve differently to accommodate previews/revisions.
- Changed `authorId` on Post model to return global ID
- added Post.authorDatabaseId to the Post model to return an Integer ID
- Pass Post.titleRendered through html_entity_decode before returning
- Pass Post.excerptRendered through html_entity_decode before returning
- Change Post.featuredImageId on Post Model to Global Relay ID
- Add Post.featuredImageDatabaseId to Post Model as Int
- Add Post.isRevision field to Post model
- Add Post.previewRevisionDatabaseId as field to Post model
- Add Post.PreviewRevisionId as field to Post model
- Add Post.isPreview field to Post model
- Add Post.PreviewDatabaseId to Post model
- Pass Term.description and Term.name through html_entity_decode before returning
- Add `ContentNode.isPreview`, `ContentNode.previewRevisionDatabaseId` and `ContentNode.previewRevisionId` fields to ContentNode interface
- Introduced `HierarchicalTermNode` Interface
- Added `asPreview` argument to singular RootQuery entry points for Post objects. This allows for the Preview node to be returned instead of the published/draft node.
- Apply `HierarchicalTermNode` to Terms of hierarchical taxonomies (Categories, custom hierarchical taxonomy types)
- Apply `Commenter` Interface to the User Type
- Deprecate User.userId in favor of user.databaseId

## Bugfixes

- Remove default `parent=>0` logic from connections. For the cases where top-level items should be queried, `parent=>0` can be used as an argument, but the default will be a flat-list of connected items
- Fix how the parent ID is resolved in MediaItem mutations
- Remove query args based on source from connection resolvers. These were causing issues. Instead, each connection that needs specific args set now uses set_query_args in the resolve function.
- Pass comment content through `html_entity_decode()` before returning
- Updated instances of deprecated Types::prepare_date_response to use Utils::prepare_date_response
- Add `wp_check_post_lock()` before returning editLock data


TylerBarnes:
omg dude hahaha 🔥

jasonbahl:
@CodeProKid sorry for the heart attack on this one

TylerBarnes:
2 commits 🤣

CodeProKid:
omg I can't even open this PR without it freezing up my browser.

CodeProKid:
> 2 commits 🤣

@TylerBarnes and one of them has a commit message:"no message"! 😆

renatonascalves:
He wake up, sat in his chair and thought: "shit, gonna fix all the bugs the plugin have". =P

jasonbahl:
@renatonascalves gets me

esamattis:
omg 😅

jasonbahl:
Yoooo.... It's not as bad as it looked. I accidentally committed without `composer install --no-dev`

Fixed. Now the diff is: 

Screen Shot 2020-06-04 at 8 05 39 AM

Instead of what it was before: 

Screen Shot 2020-06-04 at 8 07 18 AM

hashimwarren:
Ka-pow! 💎

TylerBarnes:
Sounds great :D I was hoping that's the answer. I would think almost always you want the entities unescaped in a headless context so that's good to hear 👍

TylerBarnes:
wow just realized the id for this was L33T 😂 seems fitting somehow

PR #1340, opened 2020-06-05
Fix resolving of node type when node is a revision

### LINKED ISSUE #1339: Cannot query node when post is a revision
When querying for a node type and the node is a `revision` the query fails with error above.

```graphql
{
  node(id: "cG9zdDoxMTc1Ng==") {
    __typename
  }
}
```

```json
{
  "errors": [
    {
      "debugMessage": "Type loader is expected to return valid type \"\", but it returned null",
      "message": "Internal server error",
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "node"
      ]
    }
  ],
  "data": {
    "node": null
  }
}
```

pristas-peter:
Closes #1339

### DIFF
--- a/src/Data/DataSource.php
+++ b/src/Data/DataSource.php
@@ -602,8 +602,8 @@ class DataSource {
 			switch ( true ) {
 				case $node instanceof Post:
 					$post_type = $node->post_type;
-					if ('revision' === $post_type) {
-						$post_type = get_post_type( get_post($node->ID)->post_parent );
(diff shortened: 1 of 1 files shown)


jasonbahl:
@pristas-peter this is addressed in #1337 (along with a gazillion other things)

pristas-peter:
Nice PR :)

PR #1342, opened 2020-06-06
Remove vendor packages and Fix Schema Linter

szepeviktor:
Please see individual commits :)


szepeviktor:
You should run `composer install --no-dev` just before release in a release branch.

See how WooCommerce does this.
https://github.com/woocommerce/woocommerce/blob/master/bin/build-zip.sh

szepeviktor:
@jasonbahl All tests are 🍏 

The GraphQL schema needs to be fixed,
please see https://github.com/wp-graphql/wp-graphql/pull/1342/checks?check_run_id=745369165

jasonbahl:
@szepeviktor ok, so we'll need a custom build script for this then to make sure the dependencies are properly included in the .zip file for folks downloading the plugin. 

Following your link for WooCommerce, it looks like it has the following build script: https://github.com/woocommerce/woocommerce/blob/eab8920a07396ead39095d3cec0fb98b833de7b9/.github/workflows/build-release.yml

szepeviktor:
@jasonbahl Yes. That is essentially https://github.com/woocommerce/action-build/blob/master/entrypoint.sh

Could we use the same Action?

szepeviktor:
After that stop we could automate WP.org releases https://github.com/10up/action-wordpress-plugin-deploy

szepeviktor:
@jasonbahl Could you help/advise on fixing the Schema?

jasonbahl:
@szepeviktor the schema linter has been broken for quite some time. Not related to this PR. 

I just got approved this week to publish to WordPress.org under the `wp-graphql` slug so we can start moving in that direction. 

I want to do some communicating with the community that the plugin will now live on WordPress.org, because as soon as it does folks will start seeing the "Update plugin" notification in the plugins page, and if you're using Composer or some other workflow to manage your plugins, clicking update could cause issues for you.

szepeviktor:
> if you're using Composer or some other workflow to manage your plugins, clicking update could cause issues for you.

The released plugin will contain those packages. The build process should add them.
This PR only removes them from the repository.

szepeviktor:
@jasonbahl May I clean up the repo from vendor code and automate the release process?

jasonbahl:
Yes please! Would be great. We're getting a security audit scheduled before we go to 1.0 and we're not deploying to the .org repo until after that's complete, but we can automate things like adding the vendor dependencies back to the zip you download from Github, etc. 

Very much appreciate help on this! 🙌

jasonbahl:
@szepeviktor ok, sorry for going quite on this one. The security audit just came back. I'm working on some issues around that, then will be publishing the plugin to the WordPress.org repo and circling back on PRs like this to cleanup things that don't need to be versioned. Shouldn't be too long before we can re-visit this.

szepeviktor:
@jasonbahl Do you have the intention to fix the schema linter?

jasonbahl:
@szepeviktor haha. At some point! But it requires some CI work in the Docker container to generate the static schema and save it as a github artifact or something like that. . .and that's not my area of expertise, so I keep ignoring it. 

Might be better to just remove it for now and have an open issue should someone who knows that type of stuff better than me can help, or when I have time to learn it and dig in.

jasonbahl:
@szepeviktor now that the plugin is on the .org repo and deploying there, I think we can finish up this PR and stop versioning the vendor dependencies. . .which I imagine is why you're commenting on this issue?

szepeviktor:
@jasonbahl What do we lack here?

szepeviktor:
It is done! 🚀 
https://github.com/wp-graphql/wp-graphql/commit/72dcb63c6ec62d67d3cef170c31790262da79e7a#diff-630ba09448af522154f38ef7685ef1f44b0f3e9430f80829a03ce24f400f3754

jasonbahl:
Ah yes, should have commented when I closed this. 🤦‍♂️

PR #1345, opened 2020-06-17
Fix user ID variable

0aveRyan:
The Extending Fields document has a PHP snippet showing how to add a field to a User. This fixes the PHP variable name to `$user->ID` from `$user->userId`. userId is the WPGraphQL field 😃 

Reference [WP_User](https://developer.wordpress.org/reference/classes/wp_user/#public-properties) public properties

Nope

Any other comments?
-------------------
👋  Howdy

Where has this been tested?
---------------------------
**Operating System:** N/A

**WordPress Version:** N/A


0aveRyan:
Well, mea culpa this is $user->userId. So this was wrong. But, I'll open another PR because it'd be great to clarify :)

jasonbahl:
@0aveRyan ya, the Model Layer isn't exactly the most clearly documented thing. Could for sure use more docs around how it works, and some of the differences between using internal objects like WP_Post, etc

PR #1358, opened 2020-07-01
Fix for tax query term to post connection by wrapping it an extra array

dpnolte:

What does this implement/fix?
When adding taxonomy parameters as query arg, it takes an array of arrays. (see: 
https://developer.wordpress.org/reference/classes/wp_query/#taxonomy-parameters )
In the connection from term to post, the tax query arg value was not wrapped in array. This PR adds the additional array. 
Without this fix, unexpected posts were returned when using this connection.

Where has this been tested?
---------------------------
**WordPress Version:** 5.2.6


jasonbahl:
Oof. Good catch. I'll merge, then add a test for this case, then release. 

thanks for the PR!

jasonbahl:
Thanks for the PR and the test! 🙌

jasonbahl:
closes #1360

Issue #957, opened 2019-09-17
How to add custom input fields to existing mutation?


safouane-ahmed-salah:
I would like to add custom input fields to createUser Mutation?

Thanks

safwan39:
I think there should be filter for config mutation in function register_graphql_mutation.

`function register_graphql_mutation( $mutation_name, $config ) {
       $config = apply_filter("graphql_mutation_{$mutation_name}_config", $config);
	\WPGraphQL\TypeRegistry::register_mutation( $mutation_name, $config );
}
`

CodeProKid:
Related: #965

johnlund:
👍

jonshipman:
At the very least a filter around the get_input_fields return in src/Mutation/PostObjectCreate.php

maximilianschmidt:
any update on this issue? i am looking for a way to add a custom field to the `RegisterUserInput`-Type

matepaiva:
any update on this? I think it's a very important issue, and it's easy to solve it with @jonshipman suggestion.

SimeonGriggs:
Is this an actual limitation at the moment? I can't do a createPost mutation including ACF Fields/post meta because I can't find a way to extend the default input object.

maximilianschmidt:
there is a way, @hsimah figured it out and was kind enough to give me an code snippet for this. basically you make use of the [graphql_input_fields](https://www.wpgraphql.com/filters/graphql_input_fields/) filter. Here is an example for adding a custom field "Company" to the UserRegistration-Mutation:

```
add_filter('graphql_input_fields', function($input_fields, $type_name) {
  if ($type_name === "RegisterUserInput") {
    $input_fields['company'] = [
      'type' => 'String',
      'description' => __('A string containing the user\'s company.', 'wp-graphql'),
    ];
  }

  return $input_fields;
}, 10, 2);
```

Wali-ur-Rehman:
> there is a way, @hsimah figured it out and was kind enough to give me a code snippet for this. basically you make use of the [graphql_input_fields](https://www.wpgraphql.com/filters/graphql_input_fields/) filter. Here is an example for adding a custom field "Company" to the UserRegistration-Mutation:
> 
> ```
> add_filter('graphql_input_fields', function($input_fields, $type_name) {
>   if ($type_name === "RegisterUserInput") {
>     $input_fields['company'] = [
>       'type' => 'String',
[...]

Issue #965, opened 2019-09-29
Filters for different mutation inputFields


manzoorwanijk:
A header in a response told me something 👇
![image](https://user-images.githubusercontent.com/18226415/65830912-85a90680-e2d1-11e9-9ad6-077583910b7d.png)

So, here I am 😁

WP GraphQL has different mutations for Posts, Comments, Users, Taxonomies etc. but we cannot add our own input fields (custom fields) to the corresponding object input. I think, having filters for those input fields would be great!

For example, for `PostObjectCreate` mutation, we could have `graphql_post_object_create_input_fields` filter inside `PostObjectCreate::get_input_fields()`. So, the return statement would change from `return $fields;` to 
```PHP
/**
 * Filter the input fields.
 *
 * @param array         $fields The array of input fields.
 * @param \WP_Post_Type $post_type_object The post_type_object.
 */
return apply_filters( 'graphql_post_object_create_input_fields', $fields, $post_type_object );
```

We could even add post type specific filter like `"graphql_post_{$post_type}_object_create_input_fields"`

Would happily make PR.

jasonbahl:
@manzoorwanijk sorry for the delayed response. I was out for 2 weeks with almost no internet access. 

I've been thinking about this for a while. 

I think we need a filter applied to _all_ fields that make it easy to do something with the input of an argument during resolution of a field. 

There are 2 ways that arguments typically affect resolution: 
- Modify arguments used to interact with the database
- Modify results that have been fetched from (or written to) the database. 

For example, if you wanted to add a new input field to the `createPage` mutation, you might want that field to modify the input _before_ it is sent to `wp_insert_post()` _or_ you may want it to modify something between the post being inserted and the post being returned in the GraphQL response, like store some meta on the post. 

I'm thinking we should be able to add 2 filters that are applied to all resolvers. One before primary resolution, and one after primary resolution but before returning. 

This would then allow us to do something like: 

```
register_graphql_argument( $type_name, $field_name, [
  'name' => 'MyCustomArgument',
  'type' => 'String',
  'filterArgs' => function( $this_arguments_value, $all_field_args, $context, $info ) {
     // Not sold on the naming convention, but this callback would be a filter of args, allowing for this input to somehow filter the args for the field
  },
  'filterResolver' => function( $resolved_object, $this_arguments_value, $all_field_args, $context, $info ) {
[...]

PR #1362, opened 2020-07-02
wrapped filters around the return for get_input_fields

jonshipman:
Wraps the PostObjectCreate and PostObjectUpdate with filter hooks.

https://github.com/wp-graphql/wp-graphql/issues/957
I wouldn't say it closes it, but adds a path towards fixing the issue.

### DIFF
--- a/src/Mutation/PostObjectCreate.php
+++ b/src/Mutation/PostObjectCreate.php
@@ -192,7 +192,7 @@ class PostObjectCreate {
 			}
 		}
 
-		return $fields;
+		return apply_filters( 'graphql_post_object_mutation_input_fields', $fields, $post_type_object );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


jasonbahl:
@jonshipman Thanks for the PR! 

I think we want to take this in a more centralized direction. Something along these lines: https://github.com/wp-graphql/wp-graphql/issues/965#issuecomment-539719493

This way we'd have an API that could work for input arguments anywhere in the Graph, not _just_ for one specific mutation. 

This is an important feature to get nailed down sooner than later. 

If you're interested in helping drive it to the finish line, I'd even be interested in pairing with you to make it happen.

jonshipman:
That doesn't sound like bad idea. I could take a swing at it. It would also be good to integrate the solution into the https://github.com/wp-graphql/wp-graphql-acf so that ACF field data can be a part of the input as well.

jonshipman:
Closing pull request; see https://github.com/wp-graphql/wp-graphql/issues/965#issuecomment-539719493
#965 #957

PR #1399, opened 2020-07-24
Use get_allowed_post_types() in documentation.

sboerrigter:
`\WPGraphQL::$allowed_post_types;` returns `null`. I think it should be `\WPGraphQL::get_allowed_post_types();`


jasonbahl:
Thanks!

PR #1401, opened 2020-07-24
Allow ContentType to be returned as a possible type for homepage

### LINKED ISSUE #937: Feature Request: Add homepage display
Wordpress allows selecting which homepage should be shown in the "Reading Settings"

Screenshot 2019-08-12 at 15 03 39

This information is missing by the `allSettings` query. Others, like `readingSettingsPostsPerPage` are included. It would be great to have this piece of information available through the API.

❤️

### LINKED ISSUE #1365: Expose archive page urls
This is necessary to be able to programatically create archive pages using the url pattern specified in WP. I guess this should exist for the `Taxonomy` and `ContentType` types and possibly other types?

### LINKED ISSUE #1390: ContentType.hasArchive always returns false
In a brand new WP instance with only WPGraphQL and wp-graphiql installed, hasArchive returns false for all content types including the built in Post type which I believe should have an archive by default:

Screen Shot 2020-07-15 at 11 42 15 AM

jasonbahl:
This PR addresses a few issues related to Post Type Archives and how WordPress handles the homepage. 

## Breaking Change
This PR does include a breaking change. It changes the `uri` field on the `UniformResourceIdentifiable` Interface to be nullable instead of non-null. 

This likely will not affect many consumers, but is technically a breaking change to the Schema. 

The reason for the change is because not everything that _can_ have a URI in WordPress is guaranteed to have one. For example Post Types and Taxonomies _can_ have a URI (archive page), but sometimes they don't and we need to provide for that flexibility in the Schema.

## closes #1390 (ContentType.hasArchive always returns false)

This PR closes #1390 by returning true for post types that do indeed have a public archive url. Even though the `post` post_type has `has_archive => false` behind the scenes in the Post Type registry, it _does_ indeed actually have an archive. It's archive is either the homepage, or a designated page, as set in the Customizer under the settings "show_on_front" and "page_for_posts".

This PR fixes this bug by appropriately showing true/false for the field `ContentType.hasArchive`

For example, given the following code to register 2 new post types (one with archives and one without): 

```

	register_post_type( 'archive_test', [
		'label' => 'Archive Test',
		'show_in_graphql' => true,
		'graphql_single_name' => 'ArchiveTest',
		'graphql_plural_name' => 'ArchiveTests',
		'public' => true,
		'has_archive' => true,
	] );

	register_post_type( 'no_archive_test', [
		'label' => 'Archive Test',
		'show_in_graphql' => true,
		'graphql_single_name' => 'NoArchiveTest',
		'graphql_plural_name' => 'NoArchiveTests',
	] );

} );
```

Screen Shot 2020-07-24 at 10 57 50 AM

We can see that all Post Types that do have archives have `hasArchive: true` and those that do not have public archives have `hasArchive:false`.

## closes #1365 (expose post type archive urls)

This PR closes #1365 by implementing the `UniformResourceIdentifiable` Interface on the `ContentType` type. This adds a `uri` field to ContentType type. 

So now we can ask for the `uri` of a ContentType, but _also_ fetch a ContentType by it's URI. 

For example, we can take the results of this query: 

Screen Shot 2020-07-24 at 10 57 50 AM

And use the uri to fetch Content Type nodes from their URI, like so: 

Screen Shot 2020-07-24 at 11 01 48 AM

## closes #937 (homepage display) 

In WordPress there are 2 options for the homepage. The homepage by default is the Post post type archive, a list of posts.

But you can change the settings to be a specific Page. 

With this PR, if you query a node by URI, and ask for the `/` URI, you can now get the `post` ContentType returned _or_ the page that's assigned to home. 

Here are some examples: 

#### With Default Settings

Screen Shot 2020-07-24 at 11 04 47 AM

We can query for the node by `/` uri and get the "post" `ContentType` in response: 

Screen Shot 2020-07-24 at 11 06 43 AM

### With Homepage set to a specific page

With the homepage set to a specific page, in this case the "About" page, we get a "Page" returned instead of a "ContentType"

Screen Shot 2020-07-24 at 11 10 53 AM

So now, we can request the homepage by URI, and get either a Page and it's fields, or the Post ContentType and it's respected fields.


TylerBarnes:
Is this just for the homepage? What about `nodeByUri("/some_archive_slug/")`?

jasonbahl:
@TylerBarnes yes, other archives work as well. Let me update the descriptions to show that.

jasonbahl:
@TylerBarnes actually, I included an example of fetching a custom post type archive by uri: 

Screen Shot 2020-07-24 at 11 01 48 AM

jasonbahl:
@TylerBarnes I also updated the description to mention the breaking change in this PR

PR #1415, opened 2020-08-03
Adds Caching to guides

### LINKED ISSUE #1414: [Docs] Add Caching to Guides
Since there were some interesting conversations about caching on Slack, I'd like to make sure the info doesn't get lost and add it to the Documentation Guides.

**Jason:**
There are different types of cache.
For client side cache when using GraphQL, I recommend using Apollo. I believe I read that you're using Angular, and Apollo has an Angular integration you can read more about here: https://www.apollographql.com/docs/angular/
For server-side cache, there are many many options for this.
A lot of managed WP hosts (such as WP Engine) take care of at least persistent object caching for you without needing any extra plugins. Persistent Object caching means that objects such as Posts, Users, etc are placed in a cache that persists across sessions, so many requests from different users for Post A will get a cached copy of Post A instead of talking to the SQL database.
If you're not on a managed host that provides object caching, there are many plugins that offer this. I don't have a specific recommendation. If I'm not mistaken, believe WordPress.com VIP uses this plugin or a variation of it: https://wordpress.org/plugins/batcache/
Object cache invalidation is generally straightforward. Whenever an object (Post, User, etc) is created/updated/deleted, update the cache for said object.
Query caching is another beast that's less straight-forward and can get more complicated more quickly, and there's currently no "one size fits all" plugin that solves this.
Even if you have objects in a cache, you still have to run queries to determine what objects to return.
You can store queries in WordPress transients, but you have to figure out how to properly invalidate the queries. This can get pretty complicated, because many variables can affect queries.
A lot of times, I see queries being cached for a random amount of time. For example, 3 minutes. Meaning, anytime a user asks for said query for 3 minutes, they'll get the same results, then at 3 minutes and 1 second, the SQL will re-execute and get new results and cache them for the next 3 minutes.
TTL (time to live) caches based on a timestamp are the easiest to invalidate, but can lead to users being served stale data, so depending on your use case they may or may not be worth the tradeoff.
You could trigger various actions to invalidate caches as well. For example, if you have a query that returns posts, you could just automatically invalidate the cache when certain actions like wp_insert_post / wp_delete_post, etc are run.

**Seagyn Davis**
If you're a Redis fanboy like me then you can use https://wordpress.org/plugins/redis-cache/

**codeprokid**
Yeah there are some good docs on how the WP object cache works here: https://codex.wordpress.org/Class_Reference/WP_Object_Cache I would start by getting that setup and see how far that takes you, since the setup is pretty simple and gets you some nice wins. Custom caching implementations are usually pretty complicated to get setup, and the invalidation logic is a lot of overhead to maintain, so I would only go that route as a last resort.
You could also use something like persisted queries if you go the Apollo client route https://github.com/Quartz/wp-graphql-persisted-queries

I'll make a PR and will attempt to bundle that information somehow.

henrikwirth:
Closes #1414 

Gives a starting point for the Caching Guide in the docs.

Probably could need some more info in the future.

### DIFF
--- a/docs/gatsby-config.js
+++ b/docs/gatsby-config.js
@@ -120,6 +120,7 @@ module.exports = {
             'guides/the-graphql-query-language',
             'guides/relay-spec',
             'guides/connections',
+            'guides/caching',
             'guides/anatomy-of-a-graphql-request',
             'guides/upgrading',
             'guides/authentication-and-authorization',
--- a/docs/source/guides/caching.mdx
+++ b/docs/source/guides/caching.mdx
@@ -1,0 +1,45 @@
+---
+title: Caching
+description: Learn about what kind of caching methods you can facilitate with WPGraphQL
+---
+
+Caching is a broad subject and there is many levels on which you can implement caching.
+
+*@todo This article is just a quick draft to hold on to 
+information shared on the Slackchat. More info is more than welcome.*
+
+
+## Server-Side Caching
+
+For server-side caching, there are many options.
+A lot of managed WP hosts (such as WP Engine) take care of at least persistent object caching for you without needing any extra plugins.
+Persistent Object caching means that objects such as Posts, Users, etc are placed in a cache that persists across sessions.
+Many requests from different users for Post A will get a cached copy of Post A instead of talking to the SQL database.
+If you're not on a managed host that provides object caching, there are many plugins that offer this. Checkout the WordPress references for it.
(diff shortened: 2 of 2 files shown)


jasonbahl:
I may change my mind about having this live in documentation, but for now I've moved it over to a Github Discussion:

https://github.com/wp-graphql/wp-graphql/discussions/1478

PR #1418, opened 2020-08-04
#1417 - Introduce DatabaseIdentifier Interface

### LINKED ISSUE #1417: Move databaseId field to its own interface
In v0.11 the `UniformResourceIdentifiable` Interface was applied to the `ContentType` type. The `UniformResourceIdentifiable` Interface defines the field `databaseId` as a `nonNull Int`., but the `ContentType` type doesn't have a DatabaseId, so when querying for the `databaseId` field on a ContentType we get an error.

## Propose Solution:
The field `databaseId` should be moved out of the `UniformResourceIdentifiable` Interface and into its own `DatabaseId` interface and applied to Types that actually do have a DatabaseId

jasonbahl:
## Problem:
Currently the `UniformResourceIdentifiable` Interface includes the field `databaseId`, but not all entities that have a URI have a databaseId, for example Post Type archives. Post Types have a uri, but not a database ID.

- This moves the field `databaseId` out of the URI Interface and into a new DatabaseIdentifier Interface
- This Implements the `DatabaseIdentifier` Interface on the Types that have a Database ID (Posts, Pages, Custom Post Types, Tags, Categories, Custom Taxonomy Terms, Menus, MenuItems, Comments, Users)
- Update test to reflect breaking change (databaseId can no longer be queried as part of `...on UniformResourceIdentifiable` fragments)

----

closes #1417


TylerBarnes:
@jasonbahl graphql-js complains at me with this: `Interface field WpDatabaseIdentifier.databaseId expects type Int! but WpMenuItem.databaseId is type Int.`

TylerBarnes:
If I patch this locally to be non null it works 👍

jasonbahl:
Oh word. Will fix. Silly mistake 🙃

PR #1486, opened 2020-10-03
Post Type schema settings implemented

### LINKED ISSUE #1443: [Feature Request] Use the function parameter for each "WPGraphQL::get_allowed_post_types()" call.
[`WPGraphQL::get_allowed_post_types()`](https://github.com/wp-graphql/wp-graphql/blob/develop/wp-graphql.php#L444) function accepts array parameter to be a used to filter the custom post types returned by the `get_post()` function. 
Currently, the function is called throughout the register custom post types to the schema at different point the GraphQL tree. However, I believe that these calls could be made more flexible by adding more post-type properties to be used as filters for these calls.
For example, take the call in the [TypeRegistry]() class, if it was change like so.
```
$allowed_post_types = \WPGraphQL::get_allowed_post_types( [ 'skip_graphql_core_registration' => true ] );
```

This would allow for the picking and choosing of specific WPGraphQL features.

kidunot89:

Adds more Post-type settings related to WPGraphQL and using them to allows post-types to opt-in or out of special points of registration in the GraphQL tree. See the example below to see how this works with the `Product` type of WooGraphQL.
```
/**
 * Registers WooCommerce post-types to be used in GraphQL schema
 *
 * @param array  $args      - allowed post-types.
 * @param string $post_type - name of taxonomy being checked.
 *
 * @return array
 */
function register_post_types( $args, $post_type ) {
		$args['show_in_graphql']                 = true;
		$args['graphql_single_name']             = 'Product';
		$args['graphql_plural_name']             = 'Products';
		$args['register_schema_type']            = false;
		$args['register_root_query']             = false;
		$args['register_root_connection']        = false;
		$args['connect_to_comments']             = false;
		$args['connect_to_taxonomies']           = false;
		$args['use_post_object_loader']          = false;
	}

        return $args;
}
add_filter( 'register_post_type_args', 'register_post_types', 10, 2 );
```

Closes #1443 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Connection/Comments.php
+++ b/src/Connection/Comments.php
@@ -73,7 +73,7 @@ class Comments {
 		/**
 		 * Register Connections from all existing PostObject Types to Comments
 		 */
-		$allowed_post_types = \WPGraphQL::get_allowed_post_types();
+		$allowed_post_types = \WPGraphQL::get_allowed_post_types( [ 'connect_to_comments' => true ] );
 		if ( ! empty( $allowed_post_types ) && is_array( $allowed_post_types ) ) {
 			foreach ( $allowed_post_types as $post_type ) {
 				$post_type_object = get_post_type_object( $post_type );
--- a/wp-graphql.php
+++ b/wp-graphql.php
@@ -306,7 +306,7 @@ if ( ! class_exists( 'WPGraphQL' ) ) :
 			add_action( 'after_setup_theme', [ $this, 'init_admin' ] );
 
 			// Add GraphQL default post-type settings.
-			add_filter( 'register_post_type_args', [ __CLASS__, 'add_post_type_settings'], 99, 2 );
+			add_filter( 'register_post_type_args', [ __CLASS__, 'add_post_type_settings' ], 99, 2 );
 
 			$tracing = new \WPGraphQL\Utils\Tracing();
 			$tracing->init();
@@ -720,7 +720,7 @@ if ( ! class_exists( 'WPGraphQL' ) ) :
 
 			if ( is_a( $args, \WP_Post_Type::class ) && $args->show_in_graphql ) {
 				// Add settings to the global WP_Post_Type instance.
-				foreach( $post_type_settings as $setting => $default_value ) {
+				foreach ( $post_type_settings as $setting => $default_value ) {
 					$wp_post_types[ $post_type ]->{$setting} = $args->{$setting} ?? $default_value;
 				}
 			} else if ( is_array( $args ) && ! empty( $args['show_in_graphql'] ) && $args['show_in_graphql'] ) {
--- a/wp-graphql.php
+++ b/wp-graphql.php
@@ -305,6 +305,9 @@ if ( ! class_exists( 'WPGraphQL' ) ) :
 			// Initialize Admin functionality
 			add_action( 'after_setup_theme', [ $this, 'init_admin' ] );
 
+			// Add GraphQL default post-type settings.
+			add_filter( 'register_post_type_args', [ __CLASS__, 'add_post_type_settings'], 99, 2 );
+
 			$tracing = new \WPGraphQL\Utils\Tracing();
 			$tracing->init();
 
@@ -419,6 +422,7 @@ if ( ! class_exists( 'WPGraphQL' ) ) :
 				$wp_post_types['attachment']->show_in_graphql     = true;
 				$wp_post_types['attachment']->graphql_single_name = 'mediaItem';
 				$wp_post_types['attachment']->graphql_plural_name = 'mediaItems';
+				self::add_post_type_settings( $wp_post_types['attachment'], 'attachment' );
 			}
 
 			// Adds GraphQL support for pages.
@@ -426,6 +430,7 @@ if ( ! class_exists( 'WPGraphQL' ) ) :
 				$wp_post_types['page']->show_in_graphql     = true;
 				$wp_post_types['page']->graphql_single_name = 'page';
 				$wp_post_types['page']->graphql_plural_name = 'pages';
+				self::add_post_type_settings( $wp_post_types['page'], 'page' );
 			}
 
 			// Adds GraphQL support for posts.
@@ -433,6 +438,7 @@ if ( ! class_exists( 'WPGraphQL' ) ) :
 				$wp_post_types['post']->show_in_graphql     = true;
 				$wp_post_types['post']->graphql_single_name = 'post';
 				$wp_post_types['post']->graphql_plural_name = 'posts';
+				self::add_post_type_settings( $wp_post_types['post'], 'post' );
 			}
 
 			// Adds GraphQL support for categories.
@@ -683,6 +689,45 @@ if ( ! class_exists( 'WPGraphQL' ) ) :
 			$app_context->request  = ! empty( $_REQUEST ) ? $_REQUEST : null; // phpcs:ignore
 
 			return $app_context;
+		}
+
+		/**
+		 * Adds default GraphQL post-types settings
+		 *
+		 * @param WP_Post_Type|array $args       Post type arguments.
+		 * @param string             $post_type  Post type
+		 * 
+		 * @return array
+		 */
+		public static function add_post_type_settings( $args, $post_type ) {
+			global $wp_post_types;
+
+			$post_type_settings = [
+				'show_in_graphql'                 => true,
+				'graphql_single_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ),
+				'graphql_plural_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ) . 's',
+				'register_schema_type'            => true,
+				'register_root_query'             => true,
+				'register_root_connection'        => true,
+				'connect_to_comments'             => true,
+				'connect_to_taxonomies'           => true,
+				'supports_content_node'           => true,
+				'supports_post_object_union'      => true,
+				'supports_menu_item_object_union' => true,
+				'resolve_from_uri'                => true,
+				'use_post_object_loader'          => true,
+			];
+
+			if ( is_a( $args, \WP_Post_Type::class ) && $args->show_in_graphql ) {
+				// Add settings to the global WP_Post_Type instance.
+				foreach( $post_type_settings as $setting => $default_value ) {
+					$wp_post_types[ $post_type ]->{$setting} = $args->{$setting} ?? $default_value;
+				}
+			} else if ( is_array( $args ) && ! empty( $args['show_in_graphql'] ) && $args['show_in_graphql'] ) {
+				$args = wp_parse_args( $args, $post_type_settings );
+			}
+
+			return $args;
 		}
 	}
 endif;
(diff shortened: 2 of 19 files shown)


src/Connection/Comments.php
```diff
@@ -73,7 +73,7 @@ public static function register_connections() {
 		/**
 		 * Register Connections from all existing PostObject Types to Comments
 		 */
-		$allowed_post_types = \WPGraphQL::get_allowed_post_types();
+		$allowed_post_types = \WPGraphQL::get_allowed_post_types( [ 'connect_to_comments' => true ] );
```
jasonbahl, line 76: How is this different than what we're already doing here: https://github.com/wp-graphql/wp-graphql/blob/2a349fdf8f4355590ddd2694e8ad4910d0ea07fe/src/Connection/Comments.php#L80

What if you set `connect_to_comments => true` but the post type doesn't register support for comments?

Seems kinda like an akward rule when we already have: `show_in_graphql` and `supports => [ 'comments' ]`

wp-graphql.php
```diff
@@ -684,6 +690,45 @@ public static function get_app_context() {
...
+		 *
+		 * @param WP_Post_Type|array $args       Post type arguments.
+		 * @param string             $post_type  Post type
+		 * 
+		 * @return array
+		 */
+		public static function add_post_type_settings( $args, $post_type ) {
+			global $wp_post_types;
+
+			$post_type_settings = [
```
jasonbahl, line 705: I think we should nest these rules under a graphql key. 

like: 

```
'graphql' => [
  'use_post_object_loader' => true,
  ...other args
]
```

src/Connection/Comments.php
kidunot89: @jasonbahl Not necessarily, currently I manual register the comments connection for the `Product` type in WooGraphQL so I can add the comment meta `rating` as an edge field.

wp-graphql.php
```diff
@@ -684,6 +690,45 @@ public static function get_app_context() {
...
+				'register_schema_type'            => true,
+				'register_root_query'             => true,
+				'register_root_connection'        => true,
+				'connect_to_comments'             => true,
+				'connect_to_taxonomies'           => true,
+				'supports_content_node'           => true,
+				'supports_post_object_union'      => true,
+				'supports_menu_item_object_union' => true,
+				'resolve_from_uri'                => true,
+				'use_post_object_loader'          => true,
```
jasonbahl, line 718: I wonder if instead of a true/false here it should just be the name of the loader to use?

wp-graphql.php
```diff
@@ -684,6 +690,45 @@ public static function get_app_context() {
...
+				'graphql_plural_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ) . 's',
+				'register_schema_type'            => true,
+				'register_root_query'             => true,
+				'register_root_connection'        => true,
+				'connect_to_comments'             => true,
+				'connect_to_taxonomies'           => true,
+				'supports_content_node'           => true,
+				'supports_post_object_union'      => true,
+				'supports_menu_item_object_union' => true,
+				'resolve_from_uri'                => true,
```
jasonbahl, line 717: WordPress already has rules for when objects can/cannot resolve from a URI. If it's a public post type and the post object is published or in some other publicly visible state, it has a URI. The goal is to be able to use queries like `nodeByUri( $uri: '/same-url-you-could-use-in-browser` )` and be able to get the same node the browser would return.

wp-graphql.php
```diff
@@ -684,6 +690,45 @@ public static function get_app_context() {
...
+				'graphql_single_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ),
+				'graphql_plural_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ) . 's',
+				'register_schema_type'            => true,
+				'register_root_query'             => true,
+				'register_root_connection'        => true,
+				'connect_to_comments'             => true,
+				'connect_to_taxonomies'           => true,
+				'supports_content_node'           => true,
+				'supports_post_object_union'      => true,
+				'supports_menu_item_object_union' => true,
```
jasonbahl, line 716: WordPress also has rules for this. 

`show_in_nav_menus`: https://developer.wordpress.org/reference/functions/register_post_type/#show_in_nav_menus

If a CPT is set to show in Nav Menus, WPGraphQL should support Nav Menu Items returning that type.

wp-graphql.php
```diff
@@ -684,6 +690,45 @@ public static function get_app_context() {
...
+				'show_in_graphql'                 => true,
+				'graphql_single_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ),
+				'graphql_plural_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ) . 's',
+				'register_schema_type'            => true,
+				'register_root_query'             => true,
+				'register_root_connection'        => true,
+				'connect_to_comments'             => true,
+				'connect_to_taxonomies'           => true,
+				'supports_content_node'           => true,
+				'supports_post_object_union'      => true,
```
jasonbahl, line 715: Hmm. If you use `register_post_type()` and `show_in_graphql` your post type is already agreeing to this. 🤔

wp-graphql.php
```diff
@@ -684,6 +690,45 @@ public static function get_app_context() {
...
+			$post_type_settings = [
+				'show_in_graphql'                 => true,
+				'graphql_single_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ),
+				'graphql_plural_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ) . 's',
+				'register_schema_type'            => true,
+				'register_root_query'             => true,
+				'register_root_connection'        => true,
+				'connect_to_comments'             => true,
+				'connect_to_taxonomies'           => true,
+				'supports_content_node'           => true,
```
jasonbahl, line 714: `ContentNode` is essentially equal to `This thing is a post_type in WordPress`.

wp-graphql.php
```diff
@@ -684,6 +690,45 @@ public static function get_app_context() {
...
+
+			$post_type_settings = [
+				'show_in_graphql'                 => true,
+				'graphql_single_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ),
+				'graphql_plural_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ) . 's',
+				'register_schema_type'            => true,
+				'register_root_query'             => true,
+				'register_root_connection'        => true,
+				'connect_to_comments'             => true,
+				'connect_to_taxonomies'           => true,
```
jasonbahl, line 713: This case is already handled by registering a post type and taxonomy to be connected to each other and having both set to `show_in_graphql`. If you set one or the other, or both to not show_in_graphql that decision is respected.

wp-graphql.php
```diff
@@ -684,6 +690,45 @@ public static function get_app_context() {
...
+			global $wp_post_types;
+
+			$post_type_settings = [
+				'show_in_graphql'                 => true,
+				'graphql_single_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ),
+				'graphql_plural_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ) . 's',
+				'register_schema_type'            => true,
+				'register_root_query'             => true,
+				'register_root_connection'        => true,
+				'connect_to_comments'             => true,
```
jasonbahl, line 712: See: https://github.com/wp-graphql/wp-graphql/pull/1486/files#r499819148

wp-graphql.php
```diff
@@ -684,6 +690,45 @@ public static function get_app_context() {
...
+		public static function add_post_type_settings( $args, $post_type ) {
+			global $wp_post_types;
+
+			$post_type_settings = [
+				'show_in_graphql'                 => true,
+				'graphql_single_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ),
+				'graphql_plural_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ) . 's',
+				'register_schema_type'            => true,
+				'register_root_query'             => true,
+				'register_root_connection'        => true,
```
jasonbahl, line 711: what's the use case for this so I can understand more?

wp-graphql.php
```diff
@@ -684,6 +690,45 @@ public static function get_app_context() {
...
+		 * @return array
+		 */
+		public static function add_post_type_settings( $args, $post_type ) {
+			global $wp_post_types;
+
+			$post_type_settings = [
+				'show_in_graphql'                 => true,
+				'graphql_single_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ),
+				'graphql_plural_name'             => str_replace( '_', '', ucwords( strtolower( $post_type ), '_' ) ) . 's',
+				'register_schema_type'            => true,
```
jasonbahl, line 709: what's the use case for this so I can understand more?

wp-graphql.php
```diff
@@ -684,6 +690,45 @@ public static function get_app_context() {
...
+				'register_root_query'             => true,
```
jasonbahl, line 710: what's the use case for this so I can understand more?

jasonbahl (changes requested):

wp-graphql.php
kidunot89: Good idea :+1: I'll see what I can wire up.

wp-graphql.php
kidunot89: I know, but this is just determine if it's displayed that way in the schema tree.

src/Connection/Comments.php
jasonbahl: The Disqus use case is interesting. 🤔 

You would want the post_type to support comments in general, but would want to be able to disable the connection in GraphQL because the actual comments are handled outside of WordPress. 🤔

wp-graphql.php
kidunot89: Not necessarily, I'm forced to inject the `PostObjectUnion` using the `graphql_union_resolve_type` filter just to make it work with `Product` child types, if this is toggled to true.

wp-graphql.php
kidunot89: The root queries in WooGraphQL are actually rolled for the time being actually :sweat_smile:.

wp-graphql.php
kidunot89: Same thing as the `register_root_query`. Most of WooGraphQL connections resolve using custom AbstractConnectionResolver classes and not the PostObjectConnectionResolver class. This is due to the fact that they use a WooCommerce Query class and not WP_Query

wp-graphql.php
kidunot89: This would replace the primary use of the `skip-post-type-registry` WooGraphQL has been using to opt out of using the core PostObject types definition for WooCommerce post-type so it can roll it's own.

kidunot89: See the example in the PR summary above to see how it will be use for `Products` in WooGraphQL. :point_up:

src/Connection/Comments.php
kidunot89: that or in a custom table.

justlevine:
superseded by #2399

PR #1488, opened 2020-10-05
Change isset to !empty on the ordering to match WordPress's behavior.

jonshipman:
When passing an empty array to ['where']['in'], wp-graphql will return posts back in ASC order starting from the oldest. WP_Query's default behavior is returning the set in DESC order from the newest. Changing this function will use WordPress's behavior.


jasonbahl:
Thanks!

PR #1510, opened 2020-10-16
Make comments attached to private posts private

smthomas:
Comments that are attached to private posts are publicly available. This PR checks to see if a comment is attached to a private post and if so, it marks the comment as private.

This PR also adds a test to verify that admin users can view comments attached to private posts but anonymous users can not view comments attached to private posts.

Where has this been tested?
---------------------------
**Operating System:** macOS Catalina Version 10.15.7

**WordPress Version:** 5.5.1

### DIFF
--- a/src/Model/Comment.php
+++ b/src/Model/Comment.php
@@ -74,6 +74,11 @@ class Comment extends Model {
 	 */
 	protected function is_private() {
 
+		// A comment is considered private if it is attached to a private post.
+		if ( ! empty( $this->data->comment_post_ID ) && ( new Post( get_post( $this->data->comment_post_ID ) ) )->is_private() ) {
+			return true;
+		}
+
 		// NOTE: Do a non-strict check here, as the return is a `1` or `0`.
 		// phpcs:disable WordPress.PHP.StrictComparisons.LooseComparison
 		if ( true != $this->data->comment_approved && ! current_user_can( 'moderate_comments' ) ) {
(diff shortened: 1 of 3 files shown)


src/Model/Comment.php
```diff
@@ -74,6 +74,11 @@ public function __construct( \WP_Comment $comment ) {
 	 */
 	protected function is_private() {
 
+		// A comment is considered private if it is attached to a private post.
+		if ( ! empty( $this->data->comment_post_ID ) && ( new Post( get_post( $this->data->comment_post_ID ) ) )->is_private() ) {
```
jasonbahl, line 78: I think we need to actually change this line. 

if the Post a comment was added to was deleted, then `get_post( $this-data->comment_post_ID )` will return null, and we can't instantiate `new Post( null )`, so we need to make sure `get_post()` returns a `WP_Post` before instantiating `new Post()`


Something like: 

```
// Comments not associated with a post ID should be considered private, we can return true here
if ( empty( $this->data->comment_post_ID ) {
  return true;
}

$commented_on = get_post( $this->data->comment_post_ID );

// If the commented on post doesn't exist, or the Post Model is private, this comment should be private too
if ( empty( $commented_on ) || new Post( get_post( $this->data->comment_post_ID ) ) )->is_private() ) {
  return true;
}
```

smthomas: After further testing it looks like WordPress is preventing orphaned comments when posts are deleted. As long as users are using WordPress in a standard way this is a case we should not need to worry about.

jasonbahl (approved):

PR #1516, opened 2020-10-19
#1332: Adjust the behavior of the filters for is_graphql_http_request()

### LINKED ISSUE #1332: The result of graphql_is_graphql_http_request filter should be respected if it returns false
Each request is parsed by the Router to determine if it is mean to be be processed by the graphql endpoint.

This functionality can be overriden by a filter here:
https://github.com/wp-graphql/wp-graphql/blob/135cf0fa5278fce06c651ab9d5d9b8d34e5552c1/src/Router.php#L139

After applying the filter, it will check if that returned `true`, else it will just continue:
https://github.com/wp-graphql/wp-graphql/blob/135cf0fa5278fce06c651ab9d5d9b8d34e5552c1/src/Router.php#L146:L148

Next, it will continue with the default parsing, thereby overriding any previous value for`is_graphql_http_request` here:
https://github.com/wp-graphql/wp-graphql/blob/135cf0fa5278fce06c651ab9d5d9b8d34e5552c1/src/Router.php#L160

Adding a filter for `graphql_is_graphql_http_request` should allow to alter the outcome for that check to return `false` for the default /graphql route.

This could be solved either by initializing as `null` as @jasonbahl pointed out, and just skip the internal check if some filter set the value, or do the internal check first and apply the custom filter afterwards.

jasonbahl:
- This adds a new `graphql_pre_is_graphql_http_request` filter with a default of null. If this filter is applied, the function uses that value and returns before executing default behavior, otherwise it executes as normal and then applies `graphql_is_graphql_http_request' filter to the final results. This allows overriding default logic early, or letting it execute and overriding late.

closes #1332

### DIFF
--- a/src/Router.php
+++ b/src/Router.php
@@ -118,6 +118,24 @@ class Router {
 	 */
 	public static function is_graphql_http_request() {
 
+		/**
+		 * Filter whether the request is a GraphQL HTTP Request. Default is null, as the majority
+		 * of WordPress requests are NOT GraphQL requests (at least today that's true 😆).
+		 *
+		 * If this filter returns anything other than null, the function will return now and skip the
+		 * default checks.
+		 *
+		 * @param boolean $is_graphql_http_request Whether the request is a GraphQL HTTP Request. Default false.
+		 */
+		$pre_is_graphql_http_request = apply_filters( 'graphql_pre_is_graphql_http_request', null );
+
+		/**
+		 * If the filter has been applied, return now before executing default checks
+		 */
+		if ( null !== $pre_is_graphql_http_request ) {
+			return (bool) $pre_is_graphql_http_request;
+		}
+
 		// Default is false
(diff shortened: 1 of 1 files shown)


smthomas (commented):
On [Line 152](https://github.com/wp-graphql/wp-graphql/blob/fd13a36883b7d6b6d8b83aeafd5ddcd65cf76ac0/src/Router.php#L152) there is a check to see if `$is_graphql_http_request` is `true`. However, the only way it appears that value could be true is on [Line 144](https://github.com/wp-graphql/wp-graphql/blob/fd13a36883b7d6b6d8b83aeafd5ddcd65cf76ac0/src/Router.php#L144). If this is the case, we can get rid of this additional conditional and just return at that point.

jasonbahl:
@smthomas ah yes. Good call. Was moving too quick. Thanks for catching this!

PR #1529, opened 2020-10-23
WordPress Application Password Support

jasonbahl:
This adds initial support for the upcoming WordPress Application Passwords feature by telling WordPress core that GraphQL HTTP Requests are API requests and that application passwords should set the current user for WPGraphQL requests the same way it does for REST / XML-RPC requests. 

Here's a screenshot testing the functionality:

## Public request, no password:

This is a public request asking for the viewer, and getting nothing back.

Screen Shot 2020-10-22 at 3 50 05 PM (1)

## Request with Application Password in the header:

This request passes an application password in the headers of the GraphQL request and in response the viewer data is returned. 

Screen Shot 2020-10-22 at 3 50 17 PM

## Application Passwords

You can verify this functionality using the Application Passwords plugin: https://wordpress.org/plugins/application-passwords/

And you can learn more about Application Passwords in this helpful guide: https://maheshwaghmare.com/doc/application-passwords/

### DIFF
--- a/src/Router.php
+++ b/src/Router.php
@@ -72,6 +72,11 @@ class Router {
 		 */
 		add_action( 'parse_request', [ $this, 'resolve_http_request' ], 10 );
 
+		/**
+		 * Adds support for application passwords
+		 */
+		add_filter( 'application_password_is_api_request', [ $this, 'is_api_request' ] );
+
 	}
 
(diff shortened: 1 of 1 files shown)


smthomas (approved):
This look great to me.

PR #1538, opened 2020-10-28
Add filters for connection args

### LINKED ISSUE #1537: Allow filtering of post and user connection args
Currently there is no easy way to add custom query args to posts and users queries. For example, I want to query users based on some custom field, but there is no filter to modify `RootQueryToUserConnectionWhereArgs`.

I think every connection's args should have filters to allow custom args to be added.

manzoorwanijk:
This PR:
- Adds filters for connection args of `User` and `PostObject` connections
- Adds an explicit filter for `UserConnectionResolver` query args - `graphql_user_connection_query_args`

Closes #1537 

Example usage:

**To filter users based on their Language (`locale`):**

```php
// register the field in connection where args
add_filter( 'graphql_users_connection_args', function ( $connection_args ) {
	$connection_args['localeIn'] = array(
		'type'        => array(
			'list_of' => 'String', // or may be an enum
		),
		'description' => __( 'An array of locales. Matched users must have at least one of these locales.' ),
	);

	return $connection_args;
}, 10, 1 );

// use the field in user query
add_filter( 'graphql_user_connection_query_args', function ( $query_args ) {
	// `localeIn` should already be in $query_args
	if ( isset( $query_args['localeIn'] ) ) {
		// it is already sanitized by `UserConnectionResolver`
		$localeIn = $query_args['localeIn'];
		// remove `localeIn` from the array to avoid it going to WP_User_Query
		unset( $query_args['localeIn'] );

		// locale meta query
		$meta_query = array(
			array(
				'key'     => 'locale',
				'value'   => $localeIn,
				'compare' => 'IN',
			),
		);
		// if there is already a meta_query
		if ( ! empty( $query_args['meta_query'] ) ) {
			$meta_query[] = $query_args['meta_query'];
		}
		// add the meta query
		$query_args['meta_query'] = $meta_query;
	}

	return $query_args;
}, 10, 1 );
```

GQL Query
```gql
query GET_USERS {
  users( where: { localeIn:[ "en_US" ] } ) {
    nodes {
      id
      name
    }
  }
}
```

### DIFF
--- a/src/Connection/PostObjects.php
+++ b/src/Connection/PostObjects.php
@@ -670,6 +670,15 @@ class PostObjects {
 			}
 		}
 
-		return array_merge( $fields, $args );
+		$connection_args = array_merge( $fields, $args );
+
+		/**
+		 * Filter the $connection_args args to allow custom query args
+		 *
+		 * @param array         $connection_args  The connection args
+		 * @param \WP_Post_Type $post_type_object The post type the connection is going to
+		 * @param array         $args             The defaults args
+		 */
+		return apply_filters( 'graphql_post_objects_connection_args', $connection_args, $post_type_object, $args );
 	}
 }
--- a/src/Connection/PostObjects.php
+++ b/src/Connection/PostObjects.php
@@ -675,10 +675,20 @@ class PostObjects {
 		/**
 		 * Filter the $connection_args args to allow custom query args
 		 *
-		 * @param array         $connection_args  The connection args
-		 * @param \WP_Post_Type $post_type_object The post type the connection is going to
-		 * @param array         $args             The defaults args
+		 * @param array         $connection_args  The connection args.
+		 * @param \WP_Post_Type $post_type_object The post type the connection is going to.
+		 * @param array         $args             The defaults arguments.
 		 */
-		return apply_filters( 'graphql_post_objects_connection_args', $connection_args, $post_type_object, $args );
+		$connection_args = apply_filters( 'graphql_post_objects_connection_args', $connection_args, $post_type_object, $args );
+
+		/**
+		 * Filter the $connection_args args to allow custom query args.
+		 * Dynamic part of the hook being the post type
+		 *
+		 * @param array         $connection_args  The connection args.
+		 * @param \WP_Post_Type $post_type_object The post type the connection is going to.
+		 * @param array         $args             The defaults arguments.
+		 */
+		return apply_filters( "graphql_{$post_type_object->graphql_single_name}_connection_args", $connection_args, $post_type_object, $args );
 	}
 }
--- a/src/Connection/Users.php
+++ b/src/Connection/Users.php
@@ -196,9 +196,9 @@ class Users {
 		];
 
 		/**
-		 * Filter the $connection_args args to allow custom query args
+		 * Filter the $connection_args args to allow custom query args.
 		 *
-		 * @param array $connection_args The connection args
+		 * @param array $connection_args The connection arguments.
 		 */
 		return apply_filters( 'graphql_users_connection_args', $connection_args );
 	}
--- a/src/Connection/Users.php
+++ b/src/Connection/Users.php
@@ -110,7 +110,7 @@ class Users {
 	 * @return array
 	 */
 	public static function get_connection_args() {
-		return [
+		$connection_args = [
 			'role'              => [
 				'type'        => 'UserRoleEnum',
 				'description' => __( 'An array of role names that users must match to be included in results. Note that this is an inclusive list: users must match *each* role.', 'wp-graphql' ),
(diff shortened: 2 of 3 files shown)


src/Connection/PostObjects.php
```diff
@@ -670,6 +670,15 @@ public static function get_connection_args( $args = [], $post_type_object = null
...
+		$connection_args = array_merge( $fields, $args );
+
+		/**
+		 * Filter the $connection_args args to allow custom query args
+		 *
+		 * @param array         $connection_args  The connection args
+		 * @param \WP_Post_Type $post_type_object The post type the connection is going to
+		 * @param array         $args             The defaults args
+		 */
+		return apply_filters( 'graphql_post_objects_connection_args', $connection_args, $post_type_object, $args );
```
renatonascalves, line 682: Would be better to make the hook name dynamic to the current post type?

src/Connection/PostObjects.php
```diff
@@ -670,6 +670,15 @@ public static function get_connection_args( $args = [], $post_type_object = null
...
-		return array_merge( $fields, $args );
+		$connection_args = array_merge( $fields, $args );
+
+		/**
+		 * Filter the $connection_args args to allow custom query args
+		 *
+		 * @param array         $connection_args  The connection args
+		 * @param \WP_Post_Type $post_type_object The post type the connection is going to
+		 * @param array         $args             The defaults args
+		 */
```
renatonascalves, line 681:
```suggestion
		/**
		 * Filter the $connection_args args to allow custom query args
		 *
		 * @param array         $connection_args  The connection args.
		 * @param \WP_Post_Type $post_type_object The post type the connection is going to.
		 * @param array         $args             The defaults arguments.
		 */
```

src/Connection/Users.php
```diff
@@ -194,6 +194,13 @@ public static function get_connection_args() {
 				'description' => __( 'What paramater to use to order the objects by.', 'wp-graphql' ),
 			],
 		];
+
+		/**
+		 * Filter the $connection_args args to allow custom query args
+		 *
+		 * @param array $connection_args The connection args
```
renatonascalves, line 201:
```suggestion
		 * Filter the $connection_args args to allow custom query args.
		 *
		 * @param array $connection_args The connection args.
```

src/Data/Connection/UserConnectionResolver.php
```diff
@@ -141,7 +141,16 @@ public function get_query_args() {
...
-		return $query_args;
+		/**
+		 * Filter the $query args to allow folks to customize queries programmatically
+		 *
+		 * @param array       $query_args The args that will be passed to the WP_User_Query
+		 * @param mixed       $source     The source that's passed down the GraphQL queries
+		 * @param array       $args       The inputArgs on the field
+		 * @param AppContext  $context    The AppContext passed down the GraphQL tree
+		 * @param ResolveInfo $info       The ResolveInfo passed down the GraphQL tree
+		 */
```
renatonascalves, line 152:
```suggestion
		 * Filter the $query args to allow folks to customize queries programmatically.
		 *
		 * @param array       $query_args The args that will be passed to the WP_User_Query
		 * @param mixed       $source     The source that's passed down the GraphQL queries
		 * @param array       $args       The inputArgs on the field
		 * @param AppContext  $context    The AppContext passed down the GraphQL tree
		 * @param ResolveInfo $info       The ResolveInfo passed down the GraphQL tree
		 */
```

renatonascalves (changes requested):
A few suggestions but I'm not 100% sure this is even necessary. You could easily hook into `graphql_connection_query_args` and apply your changes.

manzoorwanijk:
> You could easily hook into `graphql_connection_query_args` and apply your changes.

"connection args" and "connection query args" are different. Former being the args supplied by GraphQL client while as later being the args passed to `WP_Query` class inside connection resolver. `graphql_connection_query_args` filter is of less use without being able to add custom args to connection that a GraphQL client can pass.

src/Connection/PostObjects.php
manzoorwanijk: Done, thanks

src/Connection/PostObjects.php
manzoorwanijk: Done, thanks

src/Connection/Users.php
manzoorwanijk: Done, thanks

src/Data/Connection/UserConnectionResolver.php
manzoorwanijk: Done, thanks

akshaykatale99:
Not able to get field in users where fields. i am using graphql_users_connection_args filter to add field for user meta query

image

Using this filter 

    add_filter( 'graphql_users_connection_args', function ( $connection_args ) {
	    $connection_args['localeIn'] = array(
			    'type'        => array(
			    'list_of' => 'String', // or may be an enum
		    ),
		    'description' => __( 'An array of locales. Matched users must have at least one of these locales.' 
		    ),
	    );
	    return $connection_args;
    }, 10, 1 );

PR #1540, opened 2020-10-28
Fix UsersConnectionSearchColumnEnum

### LINKED ISSUE #1539: UsersConnectionSearchColumnEnum is wrong
`UsersConnectionSearchColumnEnum` should be an enum of user columns (`'ID', 'user_login', 'user_nicename', 'user_email', 'user_url'`) but it's actually an enum of user roles.

manzoorwanijk:
This PR fixes the `UsersConnectionSearchColumnEnum` which was wrongly showing the roles instead of user columns.

Closes #1539

### DIFF
--- a/src/Type/Enum/UsersConnectionSearchColumnEnum.php
+++ b/src/Type/Enum/UsersConnectionSearchColumnEnum.php
@@ -6,30 +6,34 @@ use WPGraphQL\Type\WPEnumType;
 
 class UsersConnectionSearchColumnEnum {
 	public static function register_type() {
-		global $wp_roles;
-		$all_roles      = $wp_roles->roles;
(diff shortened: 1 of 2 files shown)


src/Type/Enum/UsersConnectionSearchColumnEnum.php
```diff
@@ -6,30 +6,34 @@
 
 class UsersConnectionSearchColumnEnum {
 	public static function register_type() {
-		global $wp_roles;
```
renatonascalves, line 9: I wonder if we should add this in another enum type. Probably a simpler path for migration from folks using the incorrect way.

renatonascalves (changes requested):
One suggestion! :)

Issue #1545, opened 2020-11-02
Meta value not getting cast correctly during pagination (cursor)


jacobarriola:
**Description**
Somehow my query is not executing correctly when trying to sort by the WooCommerce `PRICE` enum when paginating. I ran my query and logged out the SQL.

**How to reproduce**
The following raw SQL query results from my GraphQL query:
```sql
SELECT
	SQL_CALC_FOUND_ROWS wp_posts.ID
FROM
	wp_posts
	LEFT JOIN wp_term_relationships ON (wp_posts.ID = wp_term_relationships.object_id)
	INNER JOIN wp_postmeta ON (wp_posts.ID = wp_postmeta.post_id)
WHERE
	1 = 1
	AND((wp_term_relationships.term_taxonomy_id IN(634)))
	AND(wp_postmeta.meta_key = '_price')
	AND wp_posts.post_type = 'product'
	AND((wp_posts.post_status <> 'trash'
		AND wp_posts.post_status <> 'auto-draft'))
	AND wp_postmeta.meta_value <= '85.00'
	AND(wp_postmeta.meta_value < '85.00'
		OR(wp_posts.ID < 496032))
GROUP BY
	wp_posts.ID
ORDER BY
	wp_postmeta.meta_value + 0 DESC,
	wp_posts.ID DESC
LIMIT 0,
4
```

**Expected results**
The next 4 IDs (the first 3 will return in my graphql query, the 4th will be set up for the `endCursor` field) should return.

**Actual results**
* 4 ids return
* The first one is the first result of the entire query, not matter what page/endCursor I'm in, the other three are the expected IDs.

In other words, no matter what cursors I pass, the first ID result is always the same ID, which is the first product in the original query (ie the highest priced product).

The issue is in the `WHERE` clause:
```sql
AND wp_postmeta.meta_value <= '85.00'
	AND(wp_postmeta.meta_value < '85.00'
		OR(wp_posts.ID < 496032))
```

Specifically, the price values are not get cast correctly.

👏 This gives me the expected results:
```sql
AND wp_postmeta.meta_value <= CAST('85.00' AS FLOAT)
	AND(wp_postmeta.meta_value < CAST('85.00' AS FLOAT)
		OR(wp_posts.ID < 496032))
```

👏 This ALSO gives me the expected results:
```sql
AND wp_postmeta.meta_value <= 85.00
	AND(wp_postmeta.meta_value < 85.00
		OR(wp_posts.ID < 496032))
```

[...]

PR #1546, opened 2020-11-03
Add new filter for meta field order

jacobarriola:
Adds a new filter so that consumer can handle types before the SQL query is set. There are instances where a float is being passed but the type value is set to null. This creates unexpected SQL results.

An example implementation would be something like this:

```php
/**
 * Ensure meta field type is set to decimal (float)
 */
add_filter( 'wpgraphql_cursor_ordering_field', function ( $fields, $builder, $object_cursor ) {
	// Bail if not a float
	if ( empty( strpos( $fields['value'], '.' ) ) ) {
		return $fields;
	}
	
	
	$fields['type'] = 'DECIMAL(10,' . wc_get_price_decimals() . ')';
	
	return $fields;
}, 10, 3 );
```

#1545 

See issue.

### DIFF
--- a/src/Data/Cursor/CursorBuilder.php
+++ b/src/Data/Cursor/CursorBuilder.php
@@ -43,7 +43,7 @@ class CursorBuilder {
 		 * @param null | PostObjectCursor $object_cursor The PostObjectCursor class
 		 */
 		$field = apply_filters(
-			'wpgraphql_cursor_ordering_field',
+			'graphql_cursor_ordering_field',
 			[
 				'key'   => esc_sql( $key ),
 				'value' => esc_sql( $value ),
--- a/src/Data/Cursor/CursorBuilder.php
+++ b/src/Data/Cursor/CursorBuilder.php
@@ -21,28 +21,34 @@ class CursorBuilder {
 		$this->compare = $compare;
 		$this->fields  = [];
 	}
-
+	
 	/**
 	 * Add ordering field. The order you call this method matters. First field
 	 * will be the primary field and latters ones will be used if the primary
 	 * field has duplicate values
 	 *
-	 * @param string $key   database colum
-	 * @param string $value value from the current cursor
-	 * @param string $type  type cast
-	 * @param string $order custom order
+	 * @param string                  $key           database column
+	 * @param string                  $value         value from the current cursor
+	 * @param null                    $type          type cast
+	 * @param null                    $order         custom order
+	 * @param null | PostObjectCursor $object_cursor The PostObjectCursor class
 	 */
-	public function add_field( $key, $value, $type = null, $order = null ) {
+	public function add_field( $key, $value, $type = null, $order = null, $object_cursor = null ) {
 		/**
 		 * This only input for variables which are used in the SQL generation. So
 		 * escape them here.
 		 */
-		$this->fields[] = [
-			'key'   => esc_sql( $key ),
-			'value' => esc_sql( $value ),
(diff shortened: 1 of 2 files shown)


jacobarriola:
@jasonbahl Do you need me to address some of these failing tests? Feels like issues outside the scope of this PR. LMK.

src/Data/Cursor/CursorBuilder.php
```diff
@@ -21,28 +21,34 @@ public function __construct( $compare = '>' ) {
...
 		 * escape them here.
 		 */
-		$this->fields[] = [
-			'key'   => esc_sql( $key ),
-			'value' => esc_sql( $value ),
-			'type'  => esc_sql( $type ),
-			'order' => esc_sql( $order ),
-		];
+		$this->fields[] = apply_filters(
+			'wpgraphql_cursor_ordering_field',
```
jasonbahl, line 42: @jacobarriola can we get a docblock over the cursor explaining the params?

something like: 

```
/**
 * Filters the field used for ordering when cursors are used for pagination
 * 
 * @param array $field The field key, value, type and order 
 * @param CursorBuilder $this The CursorBuilder class
 * @param mixed null | PostObjectCursor $cursor The PostObjectCursor class
 */

**Review by jasonbahl — CHANGES_REQUESTED** [review]
see: https://github.com/wp-graphql/wp-graphql/pull/1546/files#r518813086

**Comment by jasonbahl** [comment]
@jacobarriola it looks like the tests are failing because of some composer dependency that needs updated. I'm working on getting that fixed now.

**Comment by jasonbahl** [comment]
@jacobarriola I fixed the composer issue that was messing up tests and merged it to your branch so tests are re-running now.

File: src/Data/Cursor/CursorBuilder.php (hunk 1 above)
Comment (jasonbahl) [reply src/Data/Cursor/CursorBuilder.php:42]: @jacobarriola upon looking at this a bit more, we also need to make sure we're escaping _after_ the filter. 

So, I think we need to change this a bit more too. 

Something like: 

```php
$field = [
  'key'   => esc_sql( $key ),
  'value' => esc_sql( $value ),
  'type'  => esc_sql( $type ),
  'order' => esc_sql( $order ),
];

$filtered_field = apply_filters( 'wpgraphql_cursor_ordering_field', $field, $this, $object_cursor );
$clean_fields = [];
if ( ! empty( $field ) && is_array( $field ) ) {
  foreach( $field as $key => $value ) {
     $clean_fields[ $key ] = esc_sql( $value );
  }
}

$this->fields = $clean_fields;

```

jasonbahl: ^ untested, just wrote here in Github as pseudo code, but might be legit 🤷‍♂️

jasonbahl (dismissed):

src/Data/Cursor/CursorBuilder.php
```diff
@@ -21,28 +21,57 @@ public function __construct( $compare = '>' ) {
...
+		 * @param null | PostObjectCursor $object_cursor The PostObjectCursor class
 		 */
-		$this->fields[] = [
-			'key'   => esc_sql( $key ),
-			'value' => esc_sql( $value ),
-			'type'  => esc_sql( $type ),
-			'order' => esc_sql( $order ),
-		];
+		$field = apply_filters(
+			'wpgraphql_cursor_ordering_field',
```
jasonbahl, line 46:
```suggestion
			'graphql_cursor_ordering_field',
```

PR #1550, opened 2020-11-06
Disable html_entity_decoding by default, allow it to be filtered back on if needed

### LINKED ISSUE #1549: Allow html_entity_decode to be enabled/disabled by default via filter
some fields, including `post_content` are passed through `html_entity_decode()` during resolution. This has solved _some_ problems, but has caused others. 

Many users of React and other JS libraries have expressed bugs with the way entities are decoded by PHP before resolving. However, some users have also expressed that it's solved issues for them.

I'm proposing we have one function that handles the decoding and has a filter to enable/disable `html_entity_decode()` so the server can decide whether to apply it by default or not. 

We can _also_ consider a client side field argument to let the client decide, but that's a separate issue to consider.

jasonbahl:
## BREAKING CHANGE

Some fields, such as `post.content`, `post.title`, `post.excerpt`, `term.name`, `term.description`, and `comment.content` pass through `html_entity_decode()` before returning. 

This has been problematic for many clients (myself included).

This PR **changes the default behavior** to _not_ pass these fields through `html_entity_decode`, but provides a filter `graphql_html_entity_decoding_enabled` that can be used to re-enable for sites that benefit from the decoding. 

You can use the filter like so to re-enable decoding: 

```php
add_filter( 'graphql_html_entity_decoding_enabled', '__return_true' );
```

Or you could even target a specific field of a specific model like so: 

```php
add_filter( 'graphql_html_entity_decoding_enabled', function( $enabled, $string, $field_name, $model ) {

	if ( $model instanceof \WPGraphQL\Model\Post && 'content' === $field_name ) {
		return true;
	}
	
	return $enabled;

}, 10, 4 );
```

- adds new static html_entity_decode utility function with "graphql_html_entity_decoding_enabled" filter. Sets it to false by default. 
- update comment contentRendered to use new html_entity_decode utility
- update Post contentRendered, titleRendered and excerptRendered to use the new html_entity_decode utility
- update Term description and name to use the new html_entity_decode utility

Other changes (should be another PR, but whatever 🤷‍♂️ ): 

- update action names in SettingsRegistry.php
- add new graphql_insert_post_object action to replace the dynamically named action
- add new graphql_update_term action to replace the dynamically named action
- change name of graphql_union_possible_types filter to replace dynamically named filter

----

closes #1549


TylerBarnes:
@jasonbahl this is sensible. Would it be confusing to only disable it for post content and make a separate filter for it? (or for each field). I think it's unlikely that folks will run into these issues for other fields. At the same time I've seen too many React apps with encoded characters in titles 😆

esamattis:
What kind of problems has risen from enabling it by default?

jasonbahl:
@esamattis there seems to be primarily issues with nested HTML/JSX, specifically in pre/code tags. 

For example: 

## Current behavior 

Below is a screenshot from a blog post that includes React code snippets. 

Screen Shot 2020-11-06 at 6 16 34 AM

## With undecoded content

Below is the same screenshot, but without using `html_entity_decode` on the content field. The code snippets that contain tags come through and are rendered properly.

Screen Shot 2020-11-06 at 6 17 26 AM

jasonbahl:
@TylerBarnes that's an interesting point. It does seem to primarily be an issue with nested HTML which is much more rare in title fields. 

I think `post.content`, `post.excerpt`, `term.description` and `comment.content` fields are the most likely culprits to have issues here. 

Where the `post.title` and `term.name` are much less likely to have this issue. 🤔

TylerBarnes:
@esamattis, To add to what @jasonbahl said, it's also problematic for any encoded characters that are not text inside an html tag. escaped JSON in a data attribute is a big problem because once it's decoded there's no way to properly parse that tag anymore as the change becomes irreversible.

esamattis:
Oh crap. This is such an awful feature in WordPress. It seems it's impossible to get all the situations right.

PR #1562, opened 2020-11-07
Fix composer config

szepeviktor:
1. remove redundant class loader
1. calling scripts the Composer way https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands

💡 This needs lock file update!!


jasonbahl (approved):

jasonbahl:
@szepeviktor when I pull this locally and update the `composer.lock` there are no changes to the lock file.

szepeviktor:
Okay! Then this is a noop update :)

PR #1576, opened 2020-11-13
Vip coding standard fixes

### LINKED ISSUE #1535: Introduce WordPress VIP coding standards and fixes for existing errors
WordPress VIP has pretty high coding standards and everyone who will work with VIP in the future will face the same problems as our team right now. It would be great if we could get coding standards up and running with VIP's standards.

Ideally we could setup PHP_Codesniffer with all the configurations of them and let a bot check run through it before merging PRs.

We will attempt to setup everything and also fix major errors, so that we will be able to use WPGraphQL in the future. It would be great if some maintainers can share some of their attention on this, so we don't end up breaking stuff.

Also this is obviously open for discussion, we would just use that PR for our purposes anyways, but would greatly appreciate if it would make it into the main repos best practices.

**References:** 
- https://wpvip.com/documentation/how-to-install-php-code-sniffer-for-wordpress-vip/
- https://github.com/Automattic/VIP-Coding-Standards/tree/master/WordPress-VIP-Go

henrikwirth:
#1535

So here we go. Thats what we needed for them to approve.

There is a vendor/ fix in there. So be careful with that.


jasonbahl (approved):

jasonbahl:
@henrikwirth thanks for the PR! 

I made a couple of minor changes. There was one issue in the DataSource.php file that was causing a breaking change to the settings portion of the WPGraphQL Schema so I adjusted the array to be set before the foreach loop instead of within the foreach loop. 

Then, I had to leave out the change to the underlying graphql-php file as that is pulled in as a dependency. We'll need to investigate other options for addressing that particular issue. 

Thanks so much to you and @oriooctopus for your work on this! It's a great help to me personally and the greater WPGraphQL community 🙏

PR #1577, opened 2020-11-14
WordPress.org deploy workflow

### LINKED ISSUE #224: Add to WordPress.org Plugin Repo
This plugin should be available on the WordPress.org plugins repository. 

In order for that to happen, we should have an automated build process that prepares the plugin for SVN. 

Ideally, the vendor dependencies _should not_ be versioned in Git like they are now, but they _will_ need to be versioned in SVN. 

So, I'd like to have some sort of automated process around releases that prepares the plugin for SVN. . .runs composer install, moves files to the proper /trunk directory (or whatever needs to happen for SVN. . .been a while since I used it), makes sure tags/version numbers are updated properly, etc. . .

Could draw some inspiration from a script like: https://github.com/thenbrent/multisite-user-management/blob/master/deploy.sh

jasonbahl:
This PR sets up a Github workflow to deploy WPGraphQL to the WordPress.org repository whenever WPGraphQL is tagged with a new release.

This adds an initial Readme.txt, and some initial artwork for display on WordPress.org.

---- 

closes #224


.gitattributes
```diff
@@ -13,3 +13,4 @@
 codeception.yml export-ignore
 coveralls.yml export-ignore
 netlify.toml export-ignore
+node_modules
```
TylerBarnes, line 16: Did you mean to put this in .gitignore?

readme.txt
```diff
@@ -1,13 +1,40 @@
 === WPGraphQL ===
-Contributors: jasonbahl, ryankanner, chopinbach, hughdevore
-Tags: GraphQL, API, HTTP, Remote, Query Language
+Contributors: jasonbahl, ryankanner, chopinbach, hughdevore, tylerbarnes1
```
TylerBarnes, line 2: 🎉

TylerBarnes (changes requested):
Looking great :D

PR #1589, opened 2020-11-17
Fix invalid array keys in WPObjectType

szepeviktor:
Objects are invalid array keys.


szepeviktor:
Please consider working for 2 hour to reach error-free PHPStan Level 4.

PR #1611, opened 2020-12-04
#1610 Fix Author Comparison in PostObjectUpdate Mutation

### LINKED ISSUE #1610: Author Check in PostObjectUpdate always fails
In the `PostObjectUpdate.php` a comparison is made to check if the current Author is allowed to Update a Post by checking it's User ID against the ID of the original Post Author. In the current code the User ID from the entity requesting the Update is retrieved by calling `get_current_user_id()` which returns an `Integer`. The original Author's ID of the Post is retrieved by calling `$existing_post->post_author` which returns a `String`, thus the check always returns `true`.

So even when both User IDs are the same, the code tries to compare a `String` with an `Integer` (e.g. `"35" !== 35`)

https://github.com/wp-graphql/wp-graphql/blob/9881dd3066191bc028fd8eb2a05c9040cec7ba47/src/Mutation/PostObjectUpdate.php#L101

maximilianschmidt:
- Cast `$existing_post->post_author` as `Integer` (by default it's a `String`) to make sure the comparison with `get_current_user_id()` (which is `Integer`) doesn't always return true
- Add missing word in the comparisons `UserError`

Closes #1610

### DIFF
--- a/src/Mutation/PostObjectUpdate.php
+++ b/src/Mutation/PostObjectUpdate.php
@@ -94,11 +94,12 @@ class PostObjectUpdate {
 				// translators: the $post_type_object->graphql_single_name placeholder is the name of the object being mutated
 				throw new UserError( sprintf( __( 'Sorry, you are not allowed to update a %1$s', 'wp-graphql' ), $post_type_object->graphql_single_name ) );
 			}
-
+			
 			/**
 			 * If the existing post was authored by another author, ensure the requesting user has permission to edit it
 			 */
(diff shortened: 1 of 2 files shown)


jasonbahl (approved):

jasonbahl:
@maximilianschmidt Thanks for the PR!

maximilianschmidt:
You are very welcome!

jasonbahl:
@maximilianschmidt I added a test case for this. Thanks again for the PR. Merging.

PR #1620, opened 2020-12-15
Release/v1.0.2

jasonbahl:
# Release Notes

## Bugfixes

- fix issue with using the count() function on potentially not-countable value
- fix bug where post_status was being checked instead of comment_status
- fix error message when restoring a comment doesn't work
- (#1610) fix check to see if current user has permission to update another Author's post. Thanks @maximilianschmidt!

## New Features

- (#1608) move connections from each post type->contentType to be ContentNode->ContentType. Thanks @jeanfredrik!
- pass status code through as a param of the `graphql_process_http_request_response` action
- add test for mutating other authors posts


jasonbahl:
Closed. Meant to open PR to master 🤦‍♂️

PR #1624, opened 2020-12-15
Bug/#1622 menu item labels not filtered for entity decoding

### LINKED ISSUE #1622: Normalize html entities
Currentlyl, when you use Gatsby and WP Graphql you get content with escaped html entities. 

You get this
image
Instead of this
image

@TylerBarnes mentioned in that issue https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/issues/19#issuecomment-642113522 that there are plans to handle it on WP Graphql side. Do you have plans for this or it should stay as a part of normalizers on a Gatsby side?

jasonbahl:
This passes Menu Item labels through the `$model->html_entity_decode` method, with default behavior of decoding entities for the field, but passing through a filter to allow folks to opt-out on the server side when necessary.

closes #1622


lnikell:
Thank you for this PR! It will be really helpful.  Just out of curiosity, how could I pass through that filter fields from Yoast. It seems it will be necessary to have more precise normalization on Gatsby wordpress filters side, for all different WP plugin specific fields.

jasonbahl:
For data coming from Yoast to take advantage of these filters, you'd need to wrap the data coming from Yoast in the method like I did in this PR.

Issue #1629, opened 2020-12-17
Regression with previews


jasonbahl:
v1.0.3 introduced a regression with previews. 

A bugfix in v1.0.3 now validates that the post_type of a post matches the post_type of the singular entry point, but this breaks for previews as the post_type behind the scenes is revision. 

For preview we need to validate the post_type of the parent post, not the post itself.

jasonbahl:
closed by #1630

PR #1630, opened 2020-12-17
#1629 - Fix regression to previews

jasonbahl:
This fixes the validation of single entry points for posts to allow previews to return if their parent is of the same post type as the single entry field is intended to return.


TylerBarnes (approved):
I don't understand this change but it fixes previews for me 😅

Issue #1653, opened 2021-01-01
Users query does not follow count_user_posts() in case a user is not logged in


jmartinhoj:
When making a "users" graphql query, the expected result is that the returned list of users contain only the users that the current user has the capability to see. In case it is an unauthenticated request, only the users that have published content must be listed. If I use a filter to  count_user_posts(), the expected wpgraphql behaviour is such that it follows the  count_user_posts()'s new behaviour. However, this is not happening.

This way, if I use a third party plugin that filter count_user_posts() (such as the co-authors plugin), if a user is co-author of some posts, but never the original author, it will never be listed by the wpgraphql users query.

jasonbahl:
closed by #1654

PR #1654, opened 2021-01-01
Fix issue #1653: removed user query arg, as it was redundant

jmartinhoj:

The way users are being filtered in case the user is unauthenticated was redundant: firstly, the WPQuery has the `has_published_posts` arg set to true: `$query_args['has_published_posts'] = true;` and after, the users are filtered by the `is_private()` method of the User class.
Removing the query arg, the query then follows the count_user_posts() function behaviour and third-party plugins that filter the count_user_posts() (like co-authors plugin) are now supported

#1653

### DIFF
--- a/src/Data/Connection/UserConnectionResolver.php
+++ b/src/Data/Connection/UserConnectionResolver.php
@@ -91,14 +91,6 @@ class UserConnectionResolver extends AbstractConnectionResolver {
 		$query_args['fields'] = 'ID';
 
 		/**
-		 * If the request is not authenticated, limit the query to users that have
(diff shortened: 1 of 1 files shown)


jasonbahl:
@jmartinhoj thanks for the PR and the detailed explanation! 🙏

jasonbahl:
@jmartinhoj 👋  unfortunately this change has actually cause regressions and we'll need to reverse it. 

I'll open a new issue explaining and we can see if we can come up with a good solution that works for your use cases with Co Authors Plus and for other use cases.

PR #1670, opened 2021-01-07
#1641 - Fix regression with page by uri queries

### LINKED ISSUE #1641: isFrontPage, isPostsPage throw unexpected errors
We're encountering issues related to single-page queries when including `isFrontPage` and/or `isPostsPage`.

### Expected behavior

Querying page data including `isFrontPage` should return boolean true or false depending on whether the queried page is set as the front page in WP.

### Actual behavior

If the queried page is set as the front page, the query works, returning true for `isFrontPage`; however, if the queried page is _not_ set as the front page, an error is thrown:

```
     "debugMessage": "Cannot return null for non-nullable field \"Page.isFrontPage\".",
      "message": "Internal server error",
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 14,
          "column": 5
        }
      ],
      "path": [
        "page",
        "isFrontPage"
      ]
```

### Expected behavior

Querying page data including `isPostsPage` should return boolean true or false depending on whether the queried page is set as the posts page in WP.

### Actual behavior

a) If the queried page is _not_ set as the posts page, the query works, returning false for `isPostsPage`, when queried by either URI or database ID.

b) If the queried page is set as the posts page _and_ the page is queried by database ID, the query works, returning true for `isPostsPage`.

c) If the queried page is set as the posts page _and_ the page is queried by URI, the query fails with an error:

```
      "debugMessage": "Cannot return null for non-nullable field \"Page.databaseId\".",
      "message": "Internal server error",
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "page",
        "databaseId"
      ]
```

The same query, on the same page, queried by URI, works fine and returns false for `isPostsPage` as soon as the page is unset as the posts page in WP. This appears potentially related to #1639, except the page in question _does_ exist in this case.

Using WP GraphQL v1.0.3

jasonbahl:
- Fixes a regression with querying for pages by URI. Since the homepage can be either a page or a ContentType (post archive) there were some gaps in the logic for what to return if a page queries for the homepage, but the options are set to post archive, etc. This adds a test to help prevent regressions.

## Before

Screen Shot 2021-01-07 at 2 35 24 PM

## After

Screen Shot 2021-01-07 at 2 36 07 PM

---- 

closes #1641

### DIFF
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -6,6 +6,8 @@ use Exception;
 use GraphQL\Error\UserError;
 use WP;
 use WPGraphQL\AppContext;
+use WPGraphQL\Model\Post;
+use WPGraphQL\Model\PostType;
 
 class NodeResolver {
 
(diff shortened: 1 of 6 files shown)


smthomas (approved):
This looks good to me!

PR #1676, opened 2021-01-08
Bug/#1675 menu item ids inconsistent

### LINKED ISSUE #1675: Menu Item Ids are inconsistent
Menu Item IDs are not always returned with the same encoded value. 

Menu Items use the `post` loader, so IDs should always be encoded as `post:$database_id` so that if passed back to the `node( id: $id )` query they could be loaded properly.

jasonbahl:
- Add a `nav_menu_item` loader to allow old menu item IDs to work properly with WPGraphQL should they be passed to a node query
- Update cases of menu item IDs to be `post:$id` instead of `nav_menu_item:$id`
- Update tests to test that both the old nav_menu_item:$id and post:$id work for nav menu item node queries to support previously issued IDs

closes #1675 

Any other comments?
-------------------
This was reported by @TylerBarnes and should be confirmed to play nice with Gatsby Source WordPress Experimental before release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,16 @@
 # Changelog
 
+## Upcoming
+
+### Bugfixes
+
+- Add a `nav_menu_item` loader to allow previous menu item IDs to work properly with WPGraphQL should they be passed to a node query (like, if the ID were persisted somewhere already)
+- Update cases of menu item IDs to be `post:$id` instead of `nav_menu_item:$id`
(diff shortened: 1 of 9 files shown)


smthomas (approved):
This looks good

jasonbahl:
@TylerBarnes Waiting for you to confirm compatibility before merging this one.

TylerBarnes (approved):
Looks like the source plugin patch that was fixing this is fully compatible with your change here :) So this can be released and then I can just make a release which removes my temporary code from before 👍

PR #1684, opened 2021-01-17
Resolve user visibility in UserLoader

chriszarate:

There has been a long-standing performance issue with the is_private method of the User model, because it attempts to count the number of published posts for each user, a potentially very expensive operation. And since the operation is uncached, this can very quickly overwhelm the database.

More discussion of the issue:
https://github.com/wp-graphql/wp-graphql/issues/961

At Quartz, this has prevented us from upgrading WPGraphQL, and it appears to have affected other users as well. A few different PRs exist to address this issue:

https://github.com/wp-graphql/wp-graphql/pull/962
https://github.com/wp-graphql/wp-graphql/pull/972
https://github.com/wp-graphql/wp-graphql/pull/1677

This PR is my attempt to fix this issue the way that Jason and others have described as ideal: using DataLoader.

Instead of creating a new loader (e.g., "UserVisibilityLoader"), I am including this in UserLoader. The reason is that they seem tightly coupled and making them separate would just introduce unnecessary complexity. You'd need to remember to use UserVisibilityLoader everywhere you used UserLoader or risk losing the performance benefit it provides.

My fix simply plugs into the existing "loadKeys" implementation and produces a single query that determines visibility for each user. ~I'd love feedback on the efficiency of this query; I hope that this is good enough to sidestep the need for caching.~ **This is not an efficient query, apologies. Working on a better one.** If its not possible to generate a single efficient query, generating an extra query for each user along the lines of `SELECT id FROM wp_posts WHERE post_author = X LIMIT 1` seems like the next best approach.

Finally, we need a way to provide the visibility to the User model. Luckily, WP_User is not a final class and provides a setter:

https://developer.wordpress.org/reference/classes/wp_user/__set/

This allows us to set an "is_private" property on the WP_User instance, which is then available to the model via "$this->data". This seems to be done widely, but we could implement a wrapper around WP_User if we wanted to avoid mutating WP_User instances.

From there, it's a very simple matter of having the User model check this property.

Looking forward to getting this bug squashed and happy to jump on any feedback. Also paging @abhijitrakas and @zacscott since they seem to have been negatively impacted by this issue as well.

Fixes https://github.com/wp-graphql/wp-graphql/issues/961

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** Linux

**WordPress Version:** 5.3.2

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -23,5 +23,6 @@ parameters:
         - "#Access to an undefined property WP_Taxonomy::\\$graphql_single_name.#"
         - "#Access to an undefined property WP_Taxonomy::\\$graphql_plural_name.#"
         - "#Access to an undefined property WP_Taxonomy::\\$show_in_graphql.#"
+        - "#Access to an undefined property WP_User::\\$is_private.#"
         - "#^Parameter \\#1 \\$taxonomy of function get_taxonomy expects string, WP_Taxonomy given.#"
         - "#^Parameter \\#1 \\$post_type of function get_post_type_object expects string, string|WP_Taxonomy given.#"
--- a/src/Data/Loader/UserLoader.php
+++ b/src/Data/Loader/UserLoader.php
@@ -41,7 +41,7 @@ class UserLoader extends AbstractDataLoader {
 	 * In this example, user 1 is not public (has no published posts) and is
 	 * omitted from the returned array.
 	 *
-	 * @param array[int] $keys Array of author IDs.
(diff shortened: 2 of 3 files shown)


jasonbahl:
@chriszarate I want to get this tackled. Went to try and get this branch updated to current version of WPGraphQL. Typically I can push to forks of WPGraphQL but seems like Quartz fork is locked down. 

I pushed up the branch to this repo and opened #2031 where we can iterate and get this resolved. 

I'll close this PR and we can use that PR/branch to get this over the finish line.

PR #1685, opened 2021-01-17
Add the  filter to User::is_private() so users can choose to bypass o…

zacscott:
…r modify this functionality (relates to issue #961)

- [X] Make sure you are making a pull request against the **develop branch** (left side). Also you should start *your branch* off *our develop*.
- [X] Make sure you are requesting to pull request from a **topic/feature/bugfix branch** (right side). Don't pull request from your master!

This change adds a filter `graphql_user_model_is_private` to the `User::is_private()` method.

Importantly this filter is BEFORE the normal code of the method is executed and can be used to bypass the rest of the function.

Ordinarily this method does a very expensive query via the use of the count_user_posts() function. This is called multiple times for some GraphQL queries adding considerable load as documented in the issue - https://github.com/wp-graphql/wp-graphql/issues/961

This filter can be used to bypass this count_user_posts() at the developers discretion, like so:

```
add_filter( 'graphql_user_model_is_private', __NAMESPACE__ .'\\graphql_user_model_is_private' );

function graphql_user_model_is_private( $is_private ) {
	return false;
}
```

This addresses this issue - https://github.com/wp-graphql/wp-graphql/issues/961

N/A / everything is noted on the issue - https://github.com/wp-graphql/wp-graphql/issues/961

Any other comments?
-------------------

This was based on this declined PR https://github.com/wp-graphql/wp-graphql/pull/962

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04

**WordPress Version:** 5.6

### DIFF
--- a/src/Model/User.php
+++ b/src/Model/User.php
@@ -143,6 +143,26 @@ class User extends Model {
 	 */
 	protected function is_private() {
 
+		/**
+		 * Filter whether the user is private.
+		 *
+		 * If true, the user will be considered private without further checks.
+		 *
+		 * @param mixed|null|boolean $is_private If set, return without executing the manual query
+		 *                                       to see if the user has published posts.
+		 * @param User               $this       Instance of the User Model
+		 */
+		$is_private = apply_filters( 'graphql_user_model_is_private', null, $this );
+
+		/**
+		 * If the filter returns true, we can return now as the user is considered private.
+		 *
+		 * Otherwise we need to continue and check if the User should be considered private or not
+		 */
+		if ( null !== $is_private ) {
+			return (bool) $is_private;
+		}
+
 		if ( ! current_user_can( 'list_users' ) && false === $this->owner_matches_current_user() ) {
 
 			/**


src/Model/User.php
```diff
@@ -143,6 +143,26 @@ public function tear_down() {
...
+		/**
+		 * Filter whether the user is private.
+		 *
+		 * If true, the user will be considered private without further checks.
+		 *
+		 * @param mixed|null|boolean $is_private If set, return without executing the manual query
+		 *                                       to see if the user has published posts.
+		 * @param User               $this       Instance of the User Model
+		 */
+		$is_private = apply_filters( 'graphql_user_model_is_private', null, $this );
```
CodeProKid, line 155: I think we should probably handle this in the `Model` abstract so all models get this. That's how I handled it here: https://github.com/wp-graphql/wp-graphql/pull/972.

jasonbahl: I've made updates to @CodeProKid original #972 PR to get it to pass tests, and this now allows the filter to run on all Models. 

If 3rd party code uses that filter, it will be used. If not, it will fall back to the individual Model's `is_private()` method. 

I'll get that merged and released. 

Closing this PR in favor of that one.

Issue #1689, opened 2021-01-19
application_password_is_api_request filter set up too late in some cases


therealgilles:
### Issue
A GraphQL users POST request with basic authentication does not return any users.

### System Information
WordPress 5.6
WP-GraphQL 1.1.2
The Events Calendar by Modern Tribe 5.3.1.1
PHP 7.4.14

### How to reproduce
Create an admin user application password on the WP site with The Events Calendar plugin installed and issue the following query request using basic authentication from a tool like Postman:

Request: POST
URI: <insert site url>/<insert graphql endpoint>
Headers:
Authentication 'Basic <insert `echo -n USERNAME:APP_PASSWORD | base64` here>'
Body:
```GRAPHQL
query  {
  users {
    nodes {
      name
    }
  }
}
```

### Explanation
The POST request is not identified as a valid API request by WP in wp_authenticate_application_password, because the WP-GraphQL application_password_is_api_request filter has not set yet when user authentication happens. The trace looks like this:

```PHP
=> wp_validate_application_password() / .../wp-includes/class-wp-hook.php, line 287
=> apply_filters() / .../wp-includes/plugin.php, line 212
=> apply_filters() / .../wp-includes/user.php, line 3005
=> _wp_get_current_user() / .../wp-includes/pluggable.php, line 70
=> wp_get_current_user() / .../wp-includes/pluggable.php, line 2198
=> wp_create_nonce() / .../wp-content/plugins/the-events-calendar/common/src/Tribe/Admin/Help_Page.php, line 55
=> register_assets() / .../wp-content/plugins/the-events-calendar/common/src/Tribe/Main.php, line 264
=> load_assets() / .../wp-includes/class-wp-hook.php, line 287
=> apply_filters() / .../wp-includes/class-wp-hook.php, line 311
=> do_action() / .../wp-includes/plugin.php, line 484
=> do_action() / .../wp-content/plugins/the-events-calendar/common/src/Tribe/Main.php, line 107
=> plugins_loaded() / .../wp-includes/class-wp-hook.php, line 287
=> apply_filters() / .../wp-includes/class-wp-hook.php, line 311
=> do_action() / .../wp-includes/plugin.php, line 484
=> do_action() / .../wp-settings.php, line 420
=> require_once() / .../wp-config.php, line 133
=> require_once() / .../wp-load.php, line 37
=> require_once() / .../wp-blog-header.php, line 13
=> require() / .../index.php, line 17
```

[...]

PR #1690, opened 2021-01-19
fixes wp-graphql/wp-graphql#1689 (not the right fix)

therealgilles:
• Move Router setup during plugin initialization so that the application_password_is_api_request filter is always added before any user authentication.

Fixes wp-graphql/wp-graphql#1689

Where has this been tested?
---------------------------
**Operating System:**
Ubuntu 20.04.1 LTS
**WordPress Version:**
5.6

Any other comments?
----------------------
Let me know if there is a better to make this change, or if you have any change suggestions.

### DIFF
--- a/wp-graphql.php
+++ b/wp-graphql.php
@@ -287,6 +287,11 @@ if ( ! class_exists( 'WPGraphQL' ) ) :
 				$tracker->delete_timestamp();
 			} );
 
+      /**
+       * Set up router actions/filters as early as possible
+       */
+			new \WPGraphQL\Router();
+
 			/**
 			 * Init WPGraphQL after themes have been setup,
 			 * allowing for both plugins and themes to register
@@ -297,7 +302,6 @@ if ( ! class_exists( 'WPGraphQL' ) ) :
 				function() {
 
 					new \WPGraphQL\Data\Config();
-					new \WPGraphQL\Router();
 
 					/**
 					 * Fire off init action


wp-graphql.php
```diff
@@ -287,6 +287,11 @@ private function actions() {
 				$tracker->delete_timestamp();
 			} );
 
+      /**
+       * Set up router actions/filters as early as possible
+       */
+			new \WPGraphQL\Router();
```
jasonbahl, line 293: So, the primary reason this is hooked in to `after_setup_theme` is for themes functions.php to be able to add functions that modify behavior if needed. If the router is setup before this hook, then filters and actions in the Router construct (https://github.com/wp-graphql/wp-graphql/blob/develop/src/Router.php#L44-L82) cannot be hooked into from themes.

therealgilles: Could you describe the scenario you have mind? I'm not quite following your reasoning (which means I must be missing something). As I see it, the Router constructor is setting up the endpoint, most likely set through settings, and a few filters and actions. If functions.php is loaded after this is set up, these filters and actions have not run yet so they can still be hooked into.

Separate question: do you think there is an issue with The Events Calendar plugin in that it's forcing user authentication before after_setup_theme has run?

jasonbahl: This change will instantiate the Router class as soon as WPGraphQL is instantiated, which happens when the plugin is loaded, prior to themes being loaded. 

This means that any code inside a themes functions.php (or other files included by it) would not hav executed yet. 

So, if a theme had a filter such as: 

```
add_filter( 'graphql_endpoint', function() {
    return 'my_custom_endpoint';
});
```

Their code would not work, because the filter (`graphql_endpoint`) would have already been called at WPGraphQL instantiation when the Router is instantiated, and not after themes have been loaded. 

I think there might be a path forward to moving some of the data that's currently in the Router around a bit so we can make both cases work (auth being respected _and_ filters being hookable by themes). 

I'll need to think on this a bit, but the information you've provided and this PR helps get things close for sure!

therealgilles: Thanks for taking the time to explain, appreciated and makes sense.

I also submitted a ticket with The Events Calendar plugin support just to let them know of the issue. Not sure if they could easily change their code. Wish WP was a bit more opinionated about what happens when.

therealgilles: I keep turning this in my head. I'm thinking that a similar issue would happen if one were to try to do a REST API request with a rest_url_prefix filter in functions.php (which I have). Let me investigate that, will report my findings.

therealgilles: Finally was able to pinpoint why it works with the REST API. They do this:

```PHP
  public function serve_request( $path = null ) {
    /* @var WP_User|null $current_user */
    global $current_user;

    if ( $current_user instanceof WP_User && ! $current_user->exists() ) {
      /*
       * If there is no current user authenticated via other means, clear
       * the cached lack of user, so that an authenticate check can set it
       * properly.
       *
       * This is done because for authentications such as Application
       * Passwords, we don't want it to be accepted unless the current HTTP
       * request is a REST API request, which can't always be identified early
       * enough in evaluation.
       */
      $current_user = null;
    }
```

I think this is needed for WP-GraphQL too.

therealgilles: New PR in wp-graphql/wp-graphql#1693.

jasonbahl: @therealgilles ooh! Good find. Shall we close this PR in favor of the other then?

jasonbahl:
@therealgilles closing in favor of #1693.

wp-graphql.php
therealgilles: Yes. I opted to create a new PR instead of trying to fix up this one, seemed cleaner.

PR #1693, opened 2021-01-20
fixes wp-graphql/wp-graphql#1689

therealgilles:
Clear global $current_user in process_http_request if there is no current user authenticated, the same way it is done for the REST API serve_request in wp-includes/rest-api/class-wp-rest-server.php.

fixes wp-graphql/wp-graphql#1689

Where has this been tested?
---------------------------
**Operating System:**
Ubuntu 20.04.1 LTS
**WordPress Version:**
5.6

### DIFF
--- a/src/Router.php
+++ b/src/Router.php
@@ -408,11 +408,30 @@ class Router {
 	 * This processes the graphql requests that come into the /graphql endpoint via an HTTP request
 	 *
 	 * @since  0.0.1
+	 * @global WP_User $current_user The currently authenticated user.
 	 * @return mixed
 	 * @throws \Exception Throws Exception.
 	 * @throws \Throwable Throws Exception.
 	 */
 	public static function process_http_request() {
+		/* @var WP_User|null $current_user */
+		global $current_user;
+
+		if ( $current_user instanceof WP_User && ! $current_user->exists() ) {
+			/*
+			 * If there is no current user authenticated via other means, clear
+			 * the cached lack of user, so that an authenticate check can set it
+			 * properly.
+			 *
+			 * This is done because for authentications such as Application
+			 * Passwords, we don't want it to be accepted unless the current HTTP
+			 * request is a GraphQL API request, which can't always be identified early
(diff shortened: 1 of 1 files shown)


src/Router.php
```diff
@@ -408,11 +408,30 @@ public static function get_raw_data() {
...
 	 *
 	 * @since  0.0.1
+	 * @global WP_User $current_user The currently authenticated user.
 	 * @return mixed
 	 * @throws \Exception Throws Exception.
 	 * @throws \Throwable Throws Exception.
 	 */
 	public static function process_http_request() {
+		/* @var WP_User|null $current_user */
+		global $current_user;
```
jasonbahl, line 418: @therealgilles I like this because it only affects GraphQL HTTP Requests, but won't affect GraphQL operations that are done within PHP (such as https://www.wpgraphql.com/2017/09/20/using-graphql-queries-in-php-within-your-wordpress-theme-or-plugin/)

jasonbahl (approved):

jasonbahl:
@therealgilles Looks like 2 of the test suites are failing, but I'm not sure that it's related to your code changes. It almost seems like it was some sort of docker/composer issue setting up the test environment. Will report back shortly.

jasonbahl:
> @therealgilles Looks like 2 of the test suites are failing, but I'm not sure that it's related to your code changes. It almost seems like it was some sort of docker/composer issue setting up the test environment. Will report back shortly.

Ok, ya, this was a blip in the setup of the testing environment. Re-running tests and we're good now.

src/Router.php
therealgilles: The REST API server_request code was added recently to WP when they integrated Application Passwords in version 5.6. The documented reasoning is a bit different than the problem I encountered. This may actually solve two problems at once: (1) It ensures authentication is done correctly once everything is setup, or (2) re-done in my case when it could not be done properly earlier. A tiny bit surprised the WP developers did not create some kind of _wp_reset_current_user_if_zero function.

PR #1704, opened 2021-02-01
Replace closing '/>' by '>' for html input tags

therealgilles:
Replace closing '/>' by '>' for html input tags, to follow the html specification which specifies no end tag.
See [https://www.w3.org/TR/html52/sec-forms.html#the-input-element](https://www.w3.org/TR/html52/sec-forms.html#the-input-element).

This was causing the disabled tag to be ignored by Chrome 88.0.4324.96 on Mac OS 11.2 in the following:
`<input type="checkbox" ... checked="checked" disabled/>`

Where has this been tested?
---------------------------
**Operating System:**
Ubuntu 20.04.2 LTS
**WordPress Version:**
5.6
** Browser:**
Chrome 88.0.4324.96 on Mac OS 11.2

### DIFF
--- a/src/Admin/Settings/SettingsRegistry.php
+++ b/src/Admin/Settings/SettingsRegistry.php
@@ -253,7 +253,7 @@ class SettingsRegistry {
 		$type        = isset( $args['type'] ) ? $args['type'] : 'text';
 		$placeholder = empty( $args['placeholder'] ) ? '' : ' placeholder="' . $args['placeholder'] . '"';
 		$disabled    = isset( $args['disabled'] ) && true === $args['disabled'] ? 'disabled' : null;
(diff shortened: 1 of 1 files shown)


jasonbahl (approved):
Thanks @therealgilles!

szepeviktor:
@jasonbahl Oh noooooooooooooooooooooo! When mixing HTML with PHP you always end up being the looser.
https://github.com/szepeviktor/Toolkit4WP/blob/8d36e7a168725069102a647553bf360f0f52176e/src/helpers.php#L17-L28

```php
$html = tag(
    'input',
    [
		'type' => bla...
        'class' => foo...
        'disabled' => null
	]
);
```

PR #1706, opened 2021-02-02
Remove public condition

ljanecek:
I think, there should be an option to retrieve non-public post types.


jasonbahl:
@ljanecek Ya, good catch. I agree, the Enum should represent all possible Content Types available in the Schema. They should be allowed to be asked for, and the resolver/loader/model layer should determine whether or not the data from that type can be returned.

jasonbahl (approved):

PR #1725, opened 2021-02-18
Remove redundant CI combinations

szepeviktor:
These are already generated by the matrix.

### DIFF
--- a/.github/workflows/testing-integration.yml
+++ b/.github/workflows/testing-integration.yml
@@ -23,16 +23,6 @@ jobs:
           - php: '7.4'
             wordpress: '5.6'
             coverage: 1
-          - php: '7.4'
-            wordpress: '5.5.3'
-          - php: '7.4'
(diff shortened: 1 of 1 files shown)


szepeviktor:
Very strange. Where do these come from?

kép

jasonbahl:
Looks like this PR removes the tests altogether 🤔 

I think there's a setting in the repo that's adding the redundant settings. Checking it out 👀

szepeviktor:
This prevented CI from starting:

https://github.com/wp-graphql/wp-graphql/blob/a355df661f119e84ff29291afdea2a0e918a697a/.github/workflows/testing-integration.yml#L8-L13

jasonbahl:
Ah! Thanks for pointing that out. Got the other PR merged as I confirmed it ran the tests. I also updated the setting in Github to not require the tests that were not configured to run.

PR #1760, opened 2021-03-05
Fix #1758 - Improve sizes on mediaItems to not return null when size not found

### LINKED ISSUE #1758: Improve srcSet and sizes on mediaItems to not return null when size not found
OK, so I have my theme setup with this sizes. These cover most common screen sizes under 4k.

```
	function custom_image_sizes(){
		add_theme_support( 'post-thumbnails' );
		set_post_thumbnail_size( 960, 540, false );
		add_image_size( 'social-preview', 600, 315, true ); // Square thumbnail used by facebook

		// You may want to change these, but these defaults cover most use cases
		add_image_size( 'small-preview', 375, 0, false );
		add_image_size( 'medium-preview', 960, 0, false );
		add_image_size( 'large-preview', 1280, 0, false );
		add_image_size( 'fullscreen-small', 1920, 0, false );
		add_image_size( 'fullscreen', 2560, 0, false );
	}
	add_action( 'after_setup_theme', 'custom_image_sizes' );
```

Upload an image (attached), and then you do a query like this:

```
query MyQuery {
  mediaItems {
    nodes {
      id
      sizes(size: FULLSCREEN)
      srcSet(size: FULLSCREEN)
    }
  }
}
```

Which gets this:

```
          "sizes": null,
          "srcSet": "https://exile-edit-api.flywheelsites.com/wp-content/uploads/2021/02/grass-outdoor-man-person-sport-water-sport-jumping-1.jpg 1920w, https://exile-edit-api.flywheelsites.com/wp-content/uploads/2021/02/grass-outdoor-man-person-sport-water-sport-jumping-1-300x169.jpg 300w, https://exile-edit-api.flywheelsites.com/wp-content/uploads/2021/02/grass-outdoor-man-person-sport-water-sport-jumping-1-1024x576.jpg 1024w, https://exile-edit-api.flywheelsites.com/wp-content/uploads/2021/02/grass-outdoor-man-person-sport-water-sport-jumping-1-768x432.jpg 768w, https://exile-edit-api.flywheelsites.com/wp-content/uploads/2021/02/grass-outdoor-man-person-sport-water-sport-jumping-1-1536x864.jpg 1536w, https://exile-edit-api.flywheelsites.com/wp-content/uploads/2021/02/grass-outdoor-man-person-sport-water-sport-jumping-1-960x540.jpg 960w, https://exile-edit-api.flywheelsites.com/wp-content/uploads/2021/02/grass-outdoor-man-person-sport-water-sport-jumping-1-375x211.jpg 375w, https://exile-edit-api.flywheelsites.com/wp-content/uploads/2021/02/grass-outdoor-man-person-sport-water-sport-jumping-1-1280x720.jpg 1280w"
```

You'll see that `srcSet` returns a correct value, but `sizes` returns null. This means that any responsive images won't work very well. I'm not sure what happens when `sizes` isn't set, but the result is on mobile screens an image at 100vw is really blurry (it's showing a low res one).

I think because the image size `FULLSCREEN` isn't found for this image (it's too small to trigger that size) that WP-GQL should return `sizes` based on `FULL` (the WP default size for original image).

But I confess, I find the `sizes` and `srcSet` attributes to be confusing, so perhaps I'm understanding this wrong. But regardless, I can't see a scenario where you'd want `sizes` to return `null`.

grass-outdoor-man-person-sport-water-sport-jumping-1

rsm0128:
Fix #1758
Before, `mediaItem`'s `sizes` value returned null if there is no matched size for the image. @drewbaker explained about it in #1758 in more detail.
Now, it returns original image size if there is no matched size. For example, if the original width of the image is 1920px, and user requested 2560px image, the `sizes` value has `"(max-width: 1920px) 100vw, 1920px"` now. (before it was `null`)

#1758 

No.

Any other comments?
-------------------
No.

Where has this been tested?
---------------------------
**Operating System:** Windows10

**WordPress Version:** 5.6.1


jasonbahl:
@rsm0128 can we get some tests to cover this?

rsm0128:
@jasonbahl I added a wpunit test `testMediaItemNotExistingSizeQuery`
Please let me know your opinion. Thanks!

tests/wpunit/MediaItemQueriesTest.php
```diff
@@ -488,4 +487,87 @@ public function testQueryMediaItemBySourceUrl() {
...
+				'iso' => 0,
+				'shutter_speed' => 0,
+				'title' => '',
+				'orientation' => '1',
+				'keywords' => [],
+			],
+		];
+
+		update_post_meta( $attachment_id, '_wp_attachment_metadata', $meta_data );
+		update_post_meta( $attachment_id, '_wp_attached_file', 'example.jpg' );
```
jasonbahl, line 537: I believe we should be able to change this to: 

```php
$filename      = ( WPGRAPHQL_PLUGIN_DIR . '/tests/_data/images/test.png' );
$attachment_id = $this->factory()->attachment->create_upload_object( $filename );
```

This will create the image upload with associated meta in the database the way WordPress does so that we're using "actual" data as WordPress would create it and not fully mocking the data. 

I think the change will break the tests where you're asserting string matches though.

rsm0128: @jasonbahl Thanks, I updated the test code. Please check :)

jasonbahl (approved):

jasonbahl:
@rsm0128 Good stuff! Pulled this locally and walked through it. Seems to be working as expected. Appreciate the test to go with it! 🙌

Issue #1714, opened 2021-02-17
Pagination with startCursor and endCursor mixed up after a few rounds


andyvu92:
The issue happens on the pagination demo too: https://codesandbox.io/s/wpgraphql-apollo-forward-backward-pagination-example-spq53?from-embed

Here is a short video: https://www.loom.com/share/fbb7360a94944b11b27d0a0462c02142

When the page is first loaded
![original](https://user-images.githubusercontent.com/29296255/108202617-81ef7b80-7175-11eb-8a15-e30c198a4dfa.png)

After clicking first next
![First next](https://user-images.githubusercontent.com/29296255/108202707-a64b5800-7175-11eb-9c69-97af36df4514.png)

Then click first prev
![First prev](https://user-images.githubusercontent.com/29296255/108202787-c0853600-7175-11eb-826e-9b888063cd80.png)

After click next again
![Second Next](https://user-images.githubusercontent.com/29296255/108203735-0393d900-7177-11eb-9719-9751ad6b2d26.png)

Then click prev again
![Second prev](https://user-images.githubusercontent.com/29296255/108203795-17d7d600-7177-11eb-85d2-301e6b223003.png)

andyvu92:
@jasonbahl @kidunot89 Could you pls look into this? I'm running a project building an e-commerce web application and it's close to the launch day. It would be much appreciated if you guys can look into this asap as I believe it will affect many others. Thanks.

andyvu92:
I figured out the issue. When you use "before" and "last", the return arrays is reversed. So what I do is to reverse that array again and use the cursor coming from edges instead of pageInfo. Like this: 

`
{ after: currentProducts[currentProducts.length - 1].cursor, first: 9 }
{ before: currentProducts[0].cursor, last: 9 }
`

ardiewen:
Unfortunately, this been reported on the following open issues already:

#1552 
#1440 
#1411 

My current workaround is to reverse the order (e.g. from ASC to DESC) when paginating backwards. It's definitely not how it should work according to relay spec, though.

If your use case is eCommerce, you may need to keep track of the sort order in state, as paginating backwards on a sorted list has the reverse implication:

e.g. What happens when...

1. User sorts by Price (high to low)
2. User paginates forwards
3. User changes sort to Price (low to high)
4. User paginates backwards

jasonbahl:
@ardiewen @andyvu92 Please see #1773 which addresses this and includes tests with forward and backward pagination. 

[...]

Issue #1552, opened 2020-11-06
Forwards cursor pagination works as expected, endCursor and startCursor gets swapped on backwards.


jonshipman:
#1411 might be the same issue, however, as I'm on 0.14 it could be an entirely different problem and fix.

Video illustrating the problem: https://youtu.be/ETBp_wXnIDQ

Forwards, all seems golden. However going backwards swaps the endCursor and startCursor causing different results to pull in.

Installed:
```
WordPress              5.5.3
wp-graphql-yoast-seo   4.8.0
advanced-custom-fields 5.9.3
custom-post-type-ui    1.8.1
wp-graphql             0.14.0
wp-graphql-custom-post-type-ui 1.1
wp-graphql-acf         0.3.4
wp-graphql-woocommerce 0.6.1
wordpress-seo          15.2.1
```

ardiewen:
@jasonbahl Can we get this looked at? It's the same issue as the one I reported in #1411, and I can confirm it's still happening in WP GraphQL 1.0. For those of us using WP GraphQL without the experimental source plugin (or not using Gatsby at all), this prevents us from doing meaningful on-page pagination.

Please let me know if you'd like an email invite to a minimal repro wp-admin demonstrating the issue. This can be 100% replicated using the built in GraphiQL IDE and is frontend agnostic.

ardiewen:
@jonshipman if you read through my original issue, my workaround currently is to invert the "order" (e.g. ASC to DESC) when paginating backwards. You will get your intended backwards pagination results this way.

However, be forewarned as this does not appear to be the intent of the Relay pagination spec described here: https://relay.dev/graphql/connections.htm

etoler1841:
Running into this issue as well. When going flipping between an `after`-based query and a `before`-based query, as @ardiewen said, you need to flip the order of the sort. But once you get a response, if this is a `before` query, you also have to reverse the order of the results and switch the startCursor and endCursor. This works, but it is definitely not the intended implementation.

IvanCoronado:
Still happening with v1.1.1 published 3 days ago

hallee:
I'm seeing this too. Our solution was to swap the cursors in `graphql_connection_page_info`:

```php
/**
  * Swap the startCursor and endCursor for reverse pagination.
  * See: https://github.com/wp-graphql/wp-graphql/issues/1552
  */
 add_filter( 'graphql_connection_page_info', function( $page_info, $resolver ) {
[...]

Issue #1440, opened 2020-08-26
Pagination: hasPreviousPage unexpectedly returns true


jacobarriola:
**Repro steps:**

1. Run a search query to get first two results

2. Run a search query to get next two results (ie page 2)

3. Run a search query to get previous two results (ie back to initial results)

**Expected results**
* on query 3, `hasPreviousPage` should return `false`

**Actual results**
* on query 3, `hasPreviousPage` returns `true`

**Notes**
* everything works as expected when the `search` parameters are not included in the query

**Versions:**
WP-GraphQL: 0.10.3
WordPress: 5.4.2

#### Query 1 (initial search)
```graphql
query FirstTwoPosts {
  posts(first: 2, where: {search: "security"}) {
    pageInfo {
      endCursor
      startCursor
      hasPreviousPage
      hasNextPage
    }
    edges {
      cursor
    }
  }
}
```

#### Query 1 result 🆗 
```json
{
  "data": {
    "posts": {
      "pageInfo": {
        "endCursor": "YXJyYXljb25uZWN0aW9uOjg5MTI=",
        "startCursor": "YXJyYXljb25uZWN0aW9uOjYxMjIz",
        "hasPreviousPage": false,
        "hasNextPage": true
      },
      "edges": [
        {
          "cursor": "YXJyYXljb25uZWN0aW9uOjYxMjIz"
        },
        {
          "cursor": "YXJyYXljb25uZWN0aW9uOjg5MTI="
        }
      ]
    }
  }
}
```

#### Query 2 (page 2)
```graphql
query SecondTwoPosts {
  posts(first: 2, after: "YXJyYXljb25uZWN0aW9uOjg5MTI=", where: {search: "security"}) {
    pageInfo {
      endCursor
      startCursor
      hasPreviousPage
      hasNextPage
    }
    edges {
      cursor
    }
  }
}
```

#### Query 2 results 🆗 
```json
{
  "data": {
    "posts": {
      "pageInfo": {
        "endCursor": "YXJyYXljb25uZWN0aW9uOjU0MTQz",
        "startCursor": "YXJyYXljb25uZWN0aW9uOjIyNzI=",
        "hasPreviousPage": true,
        "hasNextPage": true
      },
      "edges": [
        {
          "cursor": "YXJyYXljb25uZWN0aW9uOjIyNzI="
        },
        {
          "cursor": "YXJyYXljb25uZWN0aW9uOjU0MTQz"
        }
      ]
    }
  }
}
[...]

Issue #1411, opened 2020-07-30
Unexpected interaction between "orderby" argument and backwards pagination


ardiewen:
Hi all, wondering if this is intended behaviour or possibly a bug?

### Issue Summary

When paginating forward on an ordered post list (e.g. alphabetical by title) using `first` + `after` arguments, everything works as expected.

However, running the same query backwards, swapping `first` + `after` to `before` + `last` causes unexpected behaviour, assuming everything else is equal: **doesnt return the right posts, nor the right order.**

Changing the `order` argument to the opposite (e.g. "ASC" to "DESC") during backwards pagination seems to fix this issue, but is this the intended behaviour? Reference: [Relay GraphQL Cursor Connections Specification,  4.3 - Edge Order](https://relay.dev/graphql/connections.htm)

### Expected Behaviour

1. Specify a query with `{orderby: { field: "TITLE", order: "ASC" } }` and forward pagination arguments
2. Paginate in a forwards direction.
3. Change only the `first` + `after` arguments to `last` + `before`.
4. Paginate in a backwards direction and see the **exact same pages that forward pagination showed previously**.

### Actual Behaviour / Steps to Replicate

Let's say I have 13 posts with the following titles:
`['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M']`

1. Paginating in a forwards directions works as expected:

- first arguments are `first: 5`, `after: null`, `last: null`, `before: null`, `field: "TITLE"`, `order: "ASC"` and correctly returns posts **A, B, C, D, E,** and the appropriate **pageInfo**.

- second arguments are `first: 5`, `after: ${firstResult.pageInfo.endCursor}`, `last: null`, `before: null`, `field: "TITLE"`, `order: "ASC"` and correctly returns posts **F, G, H, I, J** and the appropriate **pageInfo**.

2. Paginating backwards doesn't work, or at least works in an unexpected way:

- third arguments (paginating backwards) are `first: null`, `after: null`, `last: 5`, `before: ${secondResult.pageInfo.startCursor}`, `field: "TITLE"`, `order: "ASC"` and incorrectly returns **K, J, I, H, G** with the corresponding **pageInfo**. The correct response here should have been the first page: **A, B, C, D, E**.

[...]

PR #1773, opened 2021-03-12
Updates to pagination

jasonbahl:
This addresses the following issues related to pagination stability: #1411, #1440, #1714, #1552 

The following changes were made to address this:

- Update AbstractConnectionResolver.php to reverse the order of the IDs when paging backward
- Update order cardinality in `PostObjectConnectionResolver` (ASC/DESC) to be inverse when paging backwards
- Update Cursor Builder to use `sprintf()` instead of concatenated string
- Add new PostConnectionPaginationTest suite to test several pages of forward and backward pagination, with and without custom order
- Update PostObjectSearchTest to properly reflect the results of the backward pagination fix
- Add new UserConnectionPaginationTest to test forward and backward pagination for users


src/Data/Cursor/CursorBuilder.php
```diff
@@ -104,6 +105,9 @@ public function has_fields() {
 	 * @return string
 	 */
 	public function to_sql( $fields = null ) {
+
+		global $wpdb;
```
markkelnar, line 109: Maybe unused in this function? wpdb

jasonbahl: ah good catch. Fixed!

PR #1774, opened 2021-03-12
fix: terms field was not registered in custom post types

### LINKED ISSUE #1771: Missing terms field in custom post types
Hello everybody,

I created two custom post types `cpt1` and` cpt2` and assigned a custom taxonomy to `cpt2`.
I noticed that there is no `term` field in `cpt2` so I went to look at the wp-graphql code and found something strange at the line 180 in:
https://github.com/wp-graphql/wp-graphql/blob/develop/src/Connection/TermObjects.php#L180

There is a `return` statement but I think it should be replaced with a `continue` statement otherwise in some cases the field `term` will not be registered for all subsequent post types.

Is this a mistake or intentional? Can I open a pull request?

Thanks for the great job you're doing with wp-graphql 👍

toriphes:
This commit prevents the terms field from not being registered in custom post types.

Closes #1771

### DIFF
--- a/src/Connection/TermObjects.php
+++ b/src/Connection/TermObjects.php
@@ -177,7 +177,7 @@ class TermObjects {
 				$post_type_object = get_post_type_object( $allowed_post_type );
 
 				if ( empty( get_object_taxonomies( $allowed_post_type ) ) ) {
-					return;
+					continue;
 				}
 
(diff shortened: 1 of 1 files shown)


jasonbahl:
Thanks for the PR! 🙌

PR #1781, opened 2021-03-15
Bug/#1780 pagitation for post  in

### LINKED ISSUE #1780: Backward pagination not working when using `where: { in: $ids }`
The following query is not working with backward pagination:

```graphql
query GetPosts($first: Int, $after: String $last:Int $before:String ) {
  posts(first: $first, after: $after, last:$last before:$before where: {in: [1374, 1372, 1370, 6, 1, 1031, 1027]}) {
    pageInfo {
      startCursor
      endCursor
      hasPreviousPage
      hasNextPage
    }
    nodes {
      databaseId
      title
    }
  }
}
```

Going forward with `first: 2` I properly get: 

- **Page 1**: `1374`, `1372`
- **Page 2**: `1370`, `6`
- **Page 3**: `1`, `1031`
- **Page 4**: `1027`

The startCursor for page 4 is: `YXJyYXljb25uZWN0aW9uOjEwMjc=`

So we can go backward with `last: 2, before: "YXJyYXljb25uZWN0aW9uOjEwMjc="`

## Expected Behavior

The expectation is that this should return nodes `1` and `1031` (like page 3 of forward pagination), but this is not happening. 

## Actual Behavior

Instead, I get nodes: `1372` and `1374`, the first 2 nodes in the `where: {in: $ids}` list.

jasonbahl:
This adjusts the PostObjectConnectionResolver to handle backward pagination properly for queries with `post__in` arguments. 

closes #1780 

## Before

Given the following list of Post IDs: `1374, 1372, 1370, 6, 1, 1031, 1027`

I can use the following query: 

```graphql
query GetPosts($first: Int, $after: String, $last: Int, $before: String $in:[ID]) {
  posts(first: $first, after: $after, last: $last, before: $before, where: {in: $in}) {
    pageInfo {
      startCursor
      endCursor
      hasPreviousPage
      hasNextPage
    }
    nodes {
      databaseId
      title
    }
  }
}
```

And get the first 2 items with the following variables: 

```json
{
  "first":  2,
  "in": [1374, 1372, 1370, 6, 1, 1031, 1027],
  "after": null,
  "last":null,
  "before": null
}
```

I properly get nodes `1374` and `1372`. 

I can get the next page with the following variables: 

```json
{
  "first":  2,
  "in": [1374, 1372, 1370, 6, 1, 1031, 1027],
  "after": "YXJyYXljb25uZWN0aW9uOjEzNzI=",
  "last":null,
  "before": null
}
```

And I get nodes `1370` and `6` as expected. 

I can keep paginating using the end cursor until the last page where the only node is `1027`. 

I can then paginate backward with the following variables: 

```json
{
  "first":  null,
  "in": [1374, 1372, 1370, 6, 1, 1031, 1027],
  "after": null,
  "last":2,
  "before": "YXJyYXljb25uZWN0aW9uOjEwMjc="
}
```

**BUG:** Instead of getting nodes the nodes in the 2 nodes in the list _before_ `10271, nodes `1` and `1031` as expected, I get nodes `1374` and `1372`. 

Screen Shot 2021-03-15 at 1 10 34 PM

## After

With this PR in place I can paginate forward the same, but when I paginate backward from node `1027`, I properly get nodes `1` and `1031` and can continue paginating backward properly. 

The following variables (`YXJyYXljb25uZWN0aW9uOjEwMjc=` being the cursor for node `1027`): 

```json
{
  "first":  null,
  "in": [1374, 1372, 1370, 6, 1, 1031, 1027],
  "after": null,
  "last":2,
  "before": "YXJyYXljb25uZWN0aW9uOjEwMjc=" 
}
```

Properly returns nodes `1` and `1031`.

Screen Shot 2021-03-15 at 1 12 16 PM

Using the startCursor from that response (`YXJyYXljb25uZWN0aW9uOjE=`) to go back one more page: 

```json
{
  "first":  null,
  "in": [1374, 1372, 1370, 6, 1, 1031, 1027],
  "after": null,
  "last":2,
  "before": "YXJyYXljb25uZWN0aW9uOjE="
}
```

Properly returns nodes `1370` and `6`

Screen Shot 2021-03-15 at 1 13 23 PM


hallee:
👋 I'm trying this locally and reverse pagination is still broken from the first page to the second page. This PR fixes the 2nd page onward.

I think the issue is relying on a `before` argument to determine pagination direction instead of `last`. `before` is missing when fetching the first page of results.

Using this PR, fetching the second page, I get all the same results as the first page but shifted down by one, and in reverse order.

src/Data/Connection/PostObjectConnectionResolver.php
```diff
@@ -269,20 +270,19 @@ public function get_query_args() {
 				$ids = array_map( function( $id ) {
 					return absint( $id );
 				}, $ids );
+
+				if ( ! empty( $this->args['before'] ) ) {
+					$ids = array_reverse( $ids );
```
chriszarate, line 275: Passing on feedback from @hallee: The array is only reversed if a `before` cursor has been passed. This won't be the case on the first page. Seems like we additionally need to look at the `order` or comparison operator.

jasonbahl:
@hallee @chriszarate ah yes! Good catch. Will update. Thanks! 🙏

jasonbahl:
@hallee @chriszarate I've updated the PR and included some additional tests. 

Please confirm that things are working for your use cases. 🙏

hallee:
@jasonbahl I'm still seeing the offset window issue from the first to second pages when paginating backwards — first page is 4 posts, second page includes 3 posts from the first page plus one new one (instead of 4 fresh posts).

I think your tests are missing this case because the backwards tests are using the `startCursor` from a forward-paginated result. This should work the same as using a `startCursor` from a backward-paginated result but in practice it seems like it doesn't with the current implementation — maybe because the current implementation is flipping the order?

I also see other places in the codebase that use the before `before` arg to determine pagination direction, like AbstractConnectionResolver. It might we worth a more thorough audit to move to `first` / `last` as directionality indicators instead of the cursor argument?

src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -593,7 +593,7 @@ public function get_nodes() {
 		$ids = array_slice( $ids, 0, $this->query_amount, true );
 
 		// If pagination is going backwards, revers the array of IDs
-		$ids = ! empty( $this->args['before'] ) ? array_reverse( $ids ) : $ids;
+		$ids = ! empty( $this->args['last'] ) ? array_reverse( $ids ) : $ids;
```
markkelnar, line 596: Do you also need `! empty( $this->args['before'] )` here? Can 'before' be specified for backward pagination without 'last'?

jasonbahl: @markkelnar No. It was a mistake for me to have reversed order by `before` in the first place. 

Let's say we have a list like so: `[ a, b, c, d, e, f, g, h, i, j ]`

If I ask for `first: 3`

I should get `a, b, c`

If I ask for `first: 3, before: h`, I would get `a,b,c` because they are the `first 3` items in the list, and they are `before h` in the list.

If I ask for `first: 3, after: h` I should get: `i, j` as they are the first (up to) 3 items in the list, and they are `after h`

If I ask for `last: 3` the list would come to us (from the database) as `[j, i, h, g, f, e, d, c, b, a]` so we need to reverse as it, as the items we need to return are `i, j`. So, we slice the array, to `j, i`, then reverse it to `i, j` to adhere to the Relay spec that needs the startCursor (`i`) to be the first item. 

Now, we can page backward: `last: 2, before: i` and we slice the array as `h, g`, reverse it, and return `g, h` with `g` as the startCursor so we can keep going backward. 

Drafting this response helped me with writing some tests for this PR! 😄

hallee:
@jasonbahl Just tried the latest commit and it looks good to me 👍

PR #1787, opened 2021-03-17
#1786 - Add contentTypes arg to nested PostObject connections

### LINKED ISSUE #1786: contentTypes arg not available on nested PostObject connections
Could not find a way to filter by `contentType` when querying `contentNodes` in nested objects (ex. specific post_type in a taxonomy term). The `contentTypes` arg exists on `RootQuery`'s `contentNodes`, but not on other instances.

WPGraphQL Version 1.1.3

plong0:
This adds the `contentTypes` arg to nested PostObject connections.  Previously the `contentTypes` filter was only available on `RootQuery`'s `contentNodes`.

Fixes #1786 - contentTypes arg not available on nested PostObject connections

Any other comments?
-------------------
Credit to @jasonbahl for guidance on this in the Slack channel


src/Connection/PostObjects.php
```diff
@@ -678,6 +678,20 @@ public static function get_connection_args( $args = [], $post_type_object = null
...
+			/**
+			 * Handle cases when the connection is for many post types
+			 */
+
+			/**
+			 * Content Type $args
+			 *
+			 * @see   : https://developer.wordpress.org/reference/classes/wp_query/#post-type-parameters
+			 */
+			$args['contentTypes'] = [
```
jasonbahl, line 691: I believe we should be able to remove the args from here as well: https://github.com/wp-graphql/wp-graphql/blob/c8f855834f8459d40664b123487f5d9f325a8d0c/src/Connection/PostObjects.php#L94-L101

jasonbahl (changes requested):

src/Connection/PostObjects.php
```diff
@@ -678,6 +678,20 @@ public static function get_connection_args( $args = [], $post_type_object = null
 					'description' => __( 'Array of tag slugs, used to exclude objects in specified tags', 'wp-graphql' ),
 				];
 			}
+		} else {
```
jasonbahl, line 681: Ok, so thinking more about this, I'm curious if we want to expose a `list_of => 'ContentTypeEnum'` to all of these connections. 

If you register a custom taxonomy that is connected to exactly one post type, then the `contentNodes` connection should only allow you to filter by that one type, not all types. 

For example: 

```php
register_post_type( 'test_cpt', [
  'show_in_graphql' => true,
  'graphql_single_name' => 'TestType',
  'graphql_plural_name' => 'TestTypes',
] );

register_taxonomy( 'test_taxonomy', 'test_cpt', [
  'show_in_graphql' => true,
  'graphql_single_name' => 'TestTaxonomy',
  'graphql_plural_name' => 'TestTaxonomies',
] );
```

This sets up a `test_tax` Taxonomy that is connected to _only_ the `test_cpt` Custom Post Type, yet the Schema would now allow:

```graphql
{
  testTaxonomies {
    nodes {
       id
       name
       contentNodes( where: { contentTypes: [ POST, PAGE ] } ) {
         nodes {
            __typename
            id
         }
       }
    }
  }
}
```

And that feels a wrong to me the more I think about it. 

We know ahead of time (when the post type and taxonomies are registered) that we can't actually resolve this, so exposing it to consumers as a possibility in the Schema feels misleading and inaccurate.

plong0: That makes sense.  Could it work to detect if `$post_type_object instanceof WP_Taxonomy` and use its `object_type` property to generate an Enum for the arg type?

Also wondering about the case where there is only one CPT associated with the taxonomy, should it use the `WP_Post_Type` for that and proceed to load the type-specific args (as in the normal connection to a single post)?

jasonbahl: Ya, I think we probably need something like a `ContentTypesOf{$taxonomyName}` Enum for each Taxonomy. 

So for my example, we would have a `ContentTypesOfTestTaxonomyEnum`

```graphql
enum ContentTypesOfTestTaxonomyEnum {
  TEST_TYPE
}
```
Then for that connection we'd be able to filter like so: 

```graphql
{
  testTaxonomies {
    nodes {
       id
       name
       contentNodes( where: { contentTypes: [ TEST_TYPE ] } ) {
         nodes {
            __typename
            id
         }
       }
    }
  }
}
```

Then, for a Taxonomy that has multiple post type connections, such as the Post post_type it would look like: 

```graphql
enum ContentTypesOfCategory {
  POST
}
```

And that could be queried like so: 

```graphql
{
  categories {
    nodes {
       id
       name
       contentNodes( where: { contentTypes: [ POST ] } ) {
         nodes {
            __typename
            id
         }
       }
    }
  }
  tags {
    nodes {
       id
       name
       contentNodes( where: { contentTypes: [ POST ] } ) {
         nodes {
            __typename
            id
         }
       }
    }
  }
}
```

We should make the following a separate issue, but I'm thinking we need the inverse as well: `TaxonomiesOf{$ContentType}Enum`

For example: 

```graphql
enum TaxonomiesOfPost {
  CATEGORY
  TAG
}
```

Then we can add a filter like so (this needs more thought, but the available enum will help): 

```graphql
{
  posts {
    nodes {
       id
       name
       contentNodes( where: { someTaxonomyFilter: [ TAG, CATEGORY ] } ) {
         nodes {
            __typename
            id
         }
       }
    }
  }
```

plong0: Thanks for the detailed notes on this, and apologies for my delayed follow up.  I've finally been able to cycle back to filtering in our API.  Looking forward to picking this back up.

plong0: Latest push cleans up RootQuery's connectionArgs and adds the ContentTypesOf{$taxonomyName} Enum for each Taxonomy.

PR #1812, opened 2021-03-22
Bug/#1810 fix regression to user connection resolver

### LINKED ISSUE #1605: Upgrading from 0.13.3 to 1.0 causes idType: URI to return null
Upgraded to 0.13.3 today and ran into a pesky one that I haven't been able to deeply trouble shoot yet, but I can, if it'd be helpful. 

Previously, I used the `idType: URI` in combination with the slug of a post or custom post type to return the desired element. 
Example:
```
posts { ... } 
```
returns all fields correctly, but 
```
post(id: 'foo-bar-is-my-slug', idType: URI) { ... } 
```
Does not (where it did previously).

Upgrading to 0.14.0 (and onwards) causes this to fail for all post types, default and custom, but not pages. 

If this is a wontfix because 1.0 Works Fine™, how should I change my slug-based queries to conform to the new style, as I can't find any documentation around that. If this is an issue, let me know and I can go through testing on a clean env and provide some stack traces and whatnot

### LINKED ISSUE #1810: Fix regression to UserConnectionResolver / querying unpublished users
## Overview of the issue

In #1654 the following code was removed from the `UserConnectionResolver`:

```php
if ( ! is_user_logged_in() ) {
  $query_args['has_published_posts'] = true;
}
```

The intent of this code is to only ask the database for Users that have published posts before attempting to resolve the list of users, when the request is from a non-authenticated user.

The explanation in #1654, when the code was removed, was that this was redundant as the Model Layer also checks to see if users are published authors. 

While there is _some_ redundancy, they serve different purposes. 

If we know the user making the request is a public user and they are asking for a list of users, we know they should only be given a list of users with published posts. Since we can efficiently ask SQL for this, we should do so. 

The Model Layer does _also_ ensure we don't return users that have not published posts, but this helps protect against all the entry points in the Graph. 

Since Users can be fetched by ID directly, or in connections that have their own resolvers, etc we can't rely _only_ on the UserConnectionResolver query args and SQL to prevent unpublished authors from being exposed when they shouldn't be. 

So while we try to optimize in the Connection Resolver with a preventative SQL query, the Model Layer has the final say on what Models should / should not be returned. 

## Regressions

With this code removed, querying for users generates many more queries. 

For example, I have a WordPress install with 109 users. Only 9 of them are published authors and 100 of them are users with no published content. 

When I make the following query: 

```graphql
{
  users(last: 100) {
    nodes {
      id
      name
    }
  }
}
```

This generates 118 SQL queries, and returns 0 users (even though I have 9 published users). 

Screen Shot 2021-03-22 at 11 52 40 AM

The expectation is that it should return my 9 users that do have published posts. 

With the removed code added back, I get the 9 users that have published posts, and only 25 total SQL queries (not 118) are generated. 

Screen Shot 2021-03-22 at 11 52 15 AM

Screen Shot 2021-03-22 at 11 52 03 AM

jasonbahl:
This updates the `UserConnectionResolver` to include the `has_published_posts` argument passed to WP_User_Query when 
 the query is made as part of a non-authenticated request. 
 
 ### Overview of the issue: 

Let's say we have the following list of users, conveniently named with one letter of the alphabet: 

```json
A published
B published
C published
D published
E published
F published
G published
H published
I published
J published
K published
L not-published
M not-published
N not-published
O not-published
P not-published
Q not-published
R not-published
S not-published
T not-published
U not-published
V not-published
W not-published
X not-published
Y not-published
Z not-published
```

The bug right now, is that SQL queries the number of items that were asked for (+1 for cursor pagination), then the Model Layer determines if the object can be returned. 

So, asking for `first:10` will return authors `A-K`, slice `K` off the list (and set `hasNextPage` as true) and the Model Layer will see that the authors are all published and can be returned so we will get authors `A-J` returned. 

Now, if we ask for `last:10` the bug can be seen. 

SQL will ask for authors `P-Z`, slice off `P` (and set `hasPreviousPage` as true) then pass the rest to the Model Layer. Since none of them are published authors, the Model Layer will prevent them from being returned to a non-authenticated user and it will appear as if there are no users at all. 

What should happen, though, is that a public request for `last: 10` should return the `last 10 users that the requesting user has permission to see`. 

So the expectation is that the public user should get users `B-K` returned when asking for `{ users(last:10) { ... } }`

Closes #1810 
Closes #1605 

Any other comments?
-------------------
This unfortunately reverses the change made by @jmartinhoj in #1654. 

@jmartinhoj feel free to open another issue with more specifics about your use case so we can work toward a solution that works for both cases. I'm confident that with a bit more information on what you were experiencing with Co Authors Plus we can find a solution that works all around.

### DIFF
--- a/tests/wpunit/UserObjectQueriesTest.php
+++ b/tests/wpunit/UserObjectQueriesTest.php
@@ -1553,4 +1553,154 @@ class UserObjectQueriesTest extends \Codeception\TestCase\WPTestCase {
 
 	}
 
+	public function testQueryUsersAsPublicUserShouldReturnOnlyPublishedAuthors() {
+
+		$alphabet = range( 'A', 'Z' );
+		$published_users = [];
+		$unpublished_users = [];
+		foreach ( $alphabet as $letter ) {
+			$unpublished_users[] = $this->factory()->user->create([
+				'user_login' => 'unpublished_' . $letter,
+				'user_email' => $letter . '_unpublishded@example.com'
+			]);
+			$author_id = $this->factory()->user->create([
+				'user_login' => 'published_' . $letter,
+				'user_email' => $letter . '_published@example.com',
+				'role' => 'administrator'
+			]);
+
+			$published_users[] = $author_id;
+			$this->factory()->post->create([
+				'post_status' => 'publish',
+				'post_title' => $letter . '_Post',
+				'post_author' => $author_id,
+				'role' => 'administrator'
+			]);
+		}
+
+		$query = '
+		{
+		  users(first:100) {
+		    nodes {
+		      databaseId
+		    }
+		  }
+		}
+		';
+
+		$actual = graphql([
+			'query' => $query,
+		]);
+
+		codecept_debug( $actual );
+
+		$ids = wp_list_pluck( $actual['data']['users']['nodes'], 'databaseId' );
+
+		// There should be 26 users. One published user for each letter of the alphabet.
+		$this->assertEquals( 26, count( $ids ) );
+
+		foreach ($ids as $id ) {
+			$this->assertTrue( in_array( $id, $published_users, true  ) );
+			$this->assertTrue( ! in_array( $id, $unpublished_users, true  ) );
+		}
+
+		$query = '
+		{
+		  users(last:100) {
+		    nodes {
+		      databaseId
+		    }
+		  }
+		}
+		';
+
+		$actual = graphql([
+			'query' => $query,
+		]);
+
+		codecept_debug( $actual );
+
+		$ids = wp_list_pluck( $actual['data']['users']['nodes'], 'databaseId' );
+
+		// There should be 26 users. One published user for each letter of the alphabet.
+		$this->assertEquals( 26, count( $ids ) );
+		foreach ($ids as $id ) {
+			$this->assertTrue( in_array( $id, $published_users, true  ) );
+			$this->assertTrue( ! in_array( $id, $unpublished_users, true  ) );
+		}
+
+		$query = '
+		{
+		  users(last:10) {
+		    nodes {
+		      databaseId
+		    }
+		  }
+		}
+		';
+
+		$actual = graphql([
+			'query' => $query,
+		]);
+
+		codecept_debug( $actual );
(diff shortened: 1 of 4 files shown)


tests/wpunit/UserObjectQueriesTest.php
```diff
@@ -1553,4 +1553,154 @@ public function testGetUserByIdTypesAsSubscriber() {
...
+			$published_users[] = $author_id;
+			$this->factory()->post->create([
+				'post_status' => 'publish',
+				'post_title' => $letter . '_Post',
+				'post_author' => $author_id,
+				'role' => 'administrator'
+			]);
+		}
+
+		$query = '
```
markkelnar, line 1581: Why query 2 or 3 different ways? You are only checking the ids match the correct type of author/publisher.

jasonbahl: Ah yes. I should explain more in the PR description. 

Let's say we have the following list of users, conveniently named with one letter of the alphabet: 

```json
A published
B published
C published
D published
E published
F published
G published
H published
I published
J published
K published
L not-published
M not-published
N not-published
O not-published
P not-published
Q not-published
R not-published
S not-published
T not-published
U not-published
V not-published
W not-published
X not-published
Y not-published
Z not-published
```

The bug right now, is that SQL queries the number of items that were asked for (+1 for cursor pagination), then the Model Layer determines if the object can be returned. 

So, asking for `first:10` will return authors `A-K`, slice `K` off the list (and set `hasNextPage` as true) and the Model Layer will see that the authors are all published and can be returned so we will get authors `A-J` returned. 

Now, if we ask for `last:10` the bug can be seen. 

SQL will ask for authors `P-Z`, slice off `P` (and set `hasPreviousPage` as true) then pass the rest to the Model Layer. Since none of them are published authors, the Model Layer will prevent them from being returned to a non-authenticated user and it will appear as if there are no users at all. 

What should happen, though, is that a public request for `last: 10` should return the `last 10 users that the requesting user has permission to see`. 

So the expectation is that the public user should get users `B-K` returned when asking for `{ users(last:10) { ... } }`

Issue #1761, opened 2021-03-06
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to WPGraphQL\\Type\\WPEnumType::get_safe_name() must be of the type string, null given


KarlBaumann:
![image](https://user-images.githubusercontent.com/11026930/110222131-56d48e00-7ed0-11eb-8f5c-9e002cce6589.png)

Looks like not all roles have a name.

KarlBaumann:
If somebody else is having the same problem, here's a quickfix:

`add_filter(
    'editable_roles',
    function ($roles) {
        foreach ($roles as $roleName => $roleData) {
            if (!$roleData['name']) {
                $roles[$roleName]['name'] = $roleName;
            }
        }

        return $roles;
    }
);`

PR #1816, opened 2021-03-24
Use key for user type when name not defined

markkelnar:
When name field not present for a user role, use the key value as the name string.

#1761 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Type/Enum/UserRoleEnum.php
+++ b/src/Type/Enum/UserRoleEnum.php
@@ -20,7 +20,7 @@ class UserRoleEnum {
 		if ( ! empty( $editable_roles ) && is_array( $editable_roles ) ) {
 			foreach ( $editable_roles as $key => $role ) {
 
-				$formatted_role = WPEnumType::get_safe_name( $role['name'] );
+				$formatted_role = WPEnumType::get_safe_name( isset( $role['name'] ) ? $role['name'] : $key );
 
 				$roles[ $formatted_role ] = [
 					'description' => __( 'User role with specific capabilities', 'wp-graphql' ),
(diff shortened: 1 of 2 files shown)


src/Type/Enum/UserRoleEnum.php
```diff
@@ -20,7 +20,7 @@ public static function register_type() {
 		if ( ! empty( $editable_roles ) && is_array( $editable_roles ) ) {
 			foreach ( $editable_roles as $key => $role ) {
 
-				$formatted_role = WPEnumType::get_safe_name( $role['name'] );
+				$formatted_role = WPEnumType::get_safe_name( isset( $role['name'] ) ? $role['name'] : $key );
```
jasonbahl, line 23: Did you mean this to be: 

```php
$formatted_role = isset( $role['name'] ) ? WPEnumType::get_safe_name(  $role['name'] ) : WPEnumType::get_safe_name ( $key ) ;
```

jasonbahl (changes requested):
Left one question. The ternary reads wrong to me. We want to check if `isset( $role['name'] )` before using it in the `get_safe_name()` method, otherwise fallback to the key, right?

src/Type/Enum/UserRoleEnum.php
markkelnar: Both ways should return the same safe_name. I can change if it's more readable the other way.

jasonbahl: Oh wow. I totally read this as: 

```
$formatted_role = WPEnumType::get_safe_name( isset( $role['name'] ) ) ? $role['name'] : $key;
```

I missed the ternary _inside_ of `get_safe_name` 🤦‍♂️.

jasonbahl: I'm ok with it as-is. Just need me some reading glasses or something 🤓

jasonbahl (approved):

PR #1819, opened 2021-03-24
Fix incorrect references to $query_args['search']

### LINKED ISSUE #1818: Cursor pagination doesn't work with search queries
Cursor-based pagination combined with search queries generates SQL queries that do not correctly paginate search results. Investigating SQL Query Logs shows this is because cursor-based pagination translates to post date comparisons, which depends on queries being ordered by by post date. However, search queries are ordered by `wp_posts.post_title LIKE '%query%'` first, therefore causing the start/end cursors to reference posts that are not in chronological order.

During my investigation I found logic that was already put in place to prevent this, right here:
https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Connection/PostObjectConnectionResolver.php#L262

```
/**
  * If the query contains search default the results to
  */
if ( isset( $query_args['search'] ) && ! empty( $query_args['search'] ) ) {
  /**
    * Don't order search results by title (causes funky issues with cursors)
    */

  $query_args['search_orderby_title'] = false;
  $query_args['orderby']              = 'date';
  $query_args['order']                = isset( $last ) ? 'ASC' : 'DESC';
}
```

However this block is never executing because it looks like the search query is at `$query_args['s']` not `$query_args['search']`.

**Example query:**

```graphql
query SearchPosts($after: String, $before: String, $query: String!) {
  posts(after: $after, before: $before, first: 10, last: null, where: {search: $query}) {
    pageInfo {
      startCursor
      endCursor
      hasNextPage
      hasPreviousPage
    }
    edges {
      cursor
    }
    nodes {
      databaseId
      title
      date
    }
  }
}
```

Will translate into the following SQL query:

`SELECT wp_posts.ID FROM wp_posts  WHERE 1=1  AND (((wp_posts.post_title LIKE '%example%') OR (wp_posts.post_excerpt LIKE '%example%') OR (wp_posts.post_content LIKE '%example%')))  AND wp_posts.post_type = 'post' AND ((wp_posts.post_status = 'publish')) AND  CAST( wp_posts.post_date as DATETIME ) <= CAST( '2014-06-03 12:19:21' as DATETIME ) AND ( CAST( wp_posts.post_date as DATETIME ) < CAST( '2014-06-03 12:19:21' as DATETIME ) OR (  wp_posts.ID < 10768 ) ) ORDER BY wp_posts.post_title LIKE '%example%' DESC, wp_posts.post_date DESC, wp_posts.ID DESC  LIMIT 0, 6
`

WPGraphQL version 1.3.3

benknight:
Fixes #1818

### DIFF
--- a/src/Data/Connection/PostObjectConnectionResolver.php
+++ b/src/Data/Connection/PostObjectConnectionResolver.php
@@ -250,14 +250,14 @@ class PostObjectConnectionResolver extends AbstractConnectionResolver {
 		 * explicitly set in the query, unset the $query_args['post_parent'] so the search
 		 * can search all posts, not just top level posts.
 		 */
-		if ( ! $this->source instanceof \WP_Post && isset( $query_args['search'] ) && ! isset( $input_fields['parent'] ) ) {
+		if ( ! $this->source instanceof \WP_Post && isset( $query_args['s'] ) && ! isset( $input_fields['parent'] ) ) {
 			unset( $query_args['post_parent'] );
 		}
 
 		/**
 		 * If the query contains search default the results to
 		 */
-		if ( isset( $query_args['search'] ) && ! empty( $query_args['search'] ) ) {
+		if ( isset( $query_args['s'] ) && ! empty( $query_args['s'] ) ) {
 			/**
 			 * Don't order search results by title (causes funky issues with cursors)
 			 */


jasonbahl:
Oh nice! 

@benknight do you have interest in writing a test for this case to make sure we don't have a regression like this again?

benknight:
@jasonbahl actually while this change does fix a bug I'm not sure I would recommend merging it, because it's arguably likely to break search for a number of users -- basically in the current form, cursor pagination is broken, but search results are still (mostly) in order of relevance, especially for a single large query / when grabbing the first page. Merging this would cause search results to be ALL posts where the query "LIKE" matches the post title or post content, in a straight-up chronological order (so the most relevant result could be buried if it's old). The decision comes down to what is better - broken pagination or broken relevance.

jasonbahl:
@benknight ok, ya, I think we need to get both working. 

I think the codebase is in a better spot now to support both. 

I'll see what I can work up!

PR #1824, opened 2021-03-26
fix: non-logged in users could not see the tracing or query log

toriphes:
When the value 'any' is set in the settings of wp graphql both in the Tracing role and in the Query log role none of these are returned.


jasonbahl (approved):

jasonbahl:
Ah! Good catch @toriphes 👏

PR #1830, opened 2021-03-31
"rename_graphql_field()" implemented and tested.

kidunot89:
- `rename_graphql_field( $type, $field_name, $new_field_name )` implemented.
  - Use to replace conflicting field name.
- **AccessFunctionTest** updated to **WPGraphQLTestCase**

Any other comments?
-------------------
This allows users to replace a field name on a Type. 

For example, let's say we want to rename the root `user` field to `wpUser`. 

We could do that like so: 

```php
  rename_graphql_field( 'RootQuery', 'user', 'wpUser' );
} );
```

And now we can see the `wpUser` field in the Schema, instead of the `user` field. 

rename_graphql_field


tests/wpunit/AccessFunctionsTest.php
```diff
@@ -297,23 +272,32 @@ public function testFilterInputFields() {
...
 		codecept_debug( $field_names );
 
-		$this->assertArrayNotHasKey( 'errors', $actual );
+		$this->assertArrayNotHasKey( 'errors', $response );
 		$this->assertContains( 'testInputField', $field_names );
 
 	}
 
+	public function testReplaceGraphQLFieldName() {
+		replace_graphql_field_name( 'RootQuery', 'user', 'wpUser' );
```
jasonbahl, line 293: Ok, so I think we should assert that `user` exists, and `wpUser` does not. Then use the `replace_graphql_field_name()` then assert that `wpUser` does exist and `user` does not.

jasonbahl (approved):

Issue #1737, opened 2021-02-26
documentation about using docker with xdebug is out of date


rburgst:
The page 
https://github.com/wp-graphql/wp-graphql/blob/master/documentation/UsingDocker.md

references files which are no longer in the repository such as : `run-docker-local-app-xdebug.sh`.

Therefore, its no longer clear how to debug wp-graphql. It would be nice to update the documentation so that people can debug wp-graphql using docker.

markkelnar:
I wonder if these links can be replaced by links to according pages at wpgraphql.com/docs and fill in anything that's missing.

PR #1833, opened 2021-03-31
Simplify use of env variable SUITES to run specific tests.

markkelnar:
Follow up work to issue #1737 

Allow environment variable SUITES to pass into `composter run-test` command.
Use composer packages from dist, not source to speed up update and install commands. As this pulls package versions and caches them instead of pulling from github .git repo.
Increase php file upload limit.

Updates to documentation at https://www.wpgraphql.com/docs/testing/ to come soon.

### DIFF
--- a/docker/testing.entrypoint.sh
+++ b/docker/testing.entrypoint.sh
@@ -76,7 +76,7 @@ if version_gt $PHP_VERSION 7.0 && [[ -n "$COVERAGE" ]] && [[ -z "$USING_XDEBUG" 
     echo "pcov.directory = /var/www/html/wp-content/plugins/wp-graphql" >> /usr/local/etc/php/conf.d/docker-php-ext-pcov.ini
     COMPOSER_MEMORY_LIMIT=-1 composer require pcov/clobber --dev
     vendor/bin/pcov clobber
-elif [[ -n "$COVERAGE" ]] && [[ -z "$USING_XDEBUG" ]]; then
+elif [[ -n "$COVERAGE" ]] && [[ -n "$USING_XDEBUG" ]]; then
     echo "Using XDebug for codecoverage"
 fi
 
--- a/docker/testing.entrypoint.sh
+++ b/docker/testing.entrypoint.sh
@@ -2,7 +2,6 @@
 
 # Processes parameters and runs Codeception.
 run_tests() {
-    echo "Running Tests"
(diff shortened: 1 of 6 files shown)


jasonbahl:
@markkelnar ok, so do I understand that this will work like so? 

`SUITES=tests/wpunit/MyTest.php:someTest composer run-test`?

jasonbahl:
One of the test environments failed. Re-running the tests to see if it was a hiccup (happens sometimes) or if it's a legit issue.

markkelnar:
And could list multiple tests to run like so
export SUITES="tests/wpunit/TypesTest.php tests/wpunit/WPEnumTypeTest.php" ; composer run-test

docker/testing.entrypoint.sh
```diff
@@ -63,13 +62,11 @@ if [ ! -f "$PROJECT_DIR/c3.php" ]; then
     curl -L 'https://raw.github.com/Codeception/c3/2.0/c3.php' > "$PROJECT_DIR/c3.php"
 fi
 
-if [[ -n "$LOWEST" ]]; then
-    PREFER_LOWEST="--prefer-source"
-fi
-
 # Install dependencies
-COMPOSER_MEMORY_LIMIT=-1 composer update --prefer-source ${PREFER_LOWEST}
```
markkelnar, line 71: There is plenty of google/stackoverflow about composer with --prefer-source vs --prefer-dist.

docker/testing.entrypoint.sh
```diff
@@ -2,17 +2,16 @@
...
         local coverage="--coverage --coverage-xml"
     fi
     if [[ -n "$DEBUG" ]]; then
         local debug="--debug"
     fi
 
-    local suites=${1:-" ;"}
-    IFS=';' read -ra target_suites <<< "$suites"
-    for suite in "${target_suites[@]}"; do
+    local suites=${1:-" "}
```
markkelnar, line 12: Slice and dice the SUITES list which are separated by white space.

docker/testing.entrypoint.sh
markkelnar: If we ever think we will use the --prefer-lowest of packages versions, let me know ;) otherwise I removed the LOWEST flag.

jasonbahl:
@markkelnar these changes look suite! 🍬

PR #1834, opened 2021-03-31
"rename_graphql_type()" implemented and tested.

kidunot89:
- `rename_graphql_type( $typename, $new_typename )`
- Test added to `AccessFunctionsTest`

Any other comments?
-------------------
This allows Types in the Schema to be renamed. 

Take the `User` type, for example. 

If we wanted to rename the `User` type in the Schema, we could do the following: 

```php
  rename_graphql_type( 'User', 'WPUser' );
});
```

Now, if we refresh and look at the Schema in GraphiQL, we will see the fields that used to reference the `User` type all reference the `WPUser` Type: 

WPUser-in-schema

We can also register new fields that reference the `User` Type or the newly named `WPUser` type, and they will both correctly reference the renamed Type in the Schema: 

```php
  
  // Rename the User type to WPUser
  rename_graphql_type( 'User', 'WPUser' );
  
  // Register a field with type 'User'
  register_graphql_field( 'RootQuery', 'testUserField', [
    'type' => 'User',
  ] );
  
  // Register a field with type 'WPUser'
  register_graphql_field( 'RootQuery', 'testWpUserField', [
    'type' => 'WPUser',
  ]);
  
});
```

And now we can see both of these fields, returning the renamed `WPUser` Type in the Schema. 

WPUser-as-field-type


src/Type/WPEnumType.php
```diff
@@ -13,12 +13,14 @@
...
 	/**
-	 * WPInputObjectType constructor.
+	 * WPEnumType constructor.
 	 *
 	 * @param array $config
 	 */
 	public function __construct( $config ) {
-		$config['name']   = ucfirst( $config['name'] );
+		$name           = ucfirst( $config['name'] );
+		$config['name'] = apply_filters( "graphql_typename_$name", $name, $config, $this );
```
jasonbahl, line 22: I've been trying to limit the number of dynamic variable names. It's more difficult to document and harder to debug. 

I'd prefer a static filter name such as `graphql_type_name` and then use the arguments to determine whether the callback should be applied to the filter.

access-functions.php
```diff
@@ -295,6 +295,23 @@ function( \WPGraphQL\Registry\TypeRegistry $type_registry ) use ( $type_name, $f
...
+ * Renames GraphQL fields.
+ *
+ * @param string $typename      Type being renamed.
+ * @param string $new_typename  New typename.
+ *
+ * @return void
+ */
+function rename_graphql_type( string $typename, string $new_typename ) {
+	add_filter(
+		"graphql_typename_{$typename}",
```
jasonbahl, line 308: along with my other comment, I think this should be `graphql_type_name`, then the callback should include something like: 

```php
function( $existing_typename ) use ( $typename, $new_typename ) {
  if ( $existing_typename !== $typename ) {
    return $new_typename;
  }
  return $existing_typename;
}
```

jasonbahl (approved):

Issue #997, opened 2019-11-03
All inputs that expect an ID (global or integer) should be of Type ID in the Schema


jasonbahl:
We're breaking #109 into 2 issues. 

This issue should address making all input Types that expect an ID of some sort to be of Type ID. 

Resolution can still be inconsistent (for now), but the Schema itself should all be of Type ID so that we can make resolution consistent in the future.

moonmeister:
`commenOn` for Comment Mutations currently requires `Int`

jasonbahl:
Status and plan for the rest of this. We've chipped away at the ID typing here over the years, applying what we could without breaking changes, and what's left are the changes that are breaking, so they need to land in a major.

Splitting the remaining work along the where-args boundary:

- The inputs that are not connection where args, for example the `commentOn` input on the createComment mutation, will be addressed as part of the 3.0 type-safety work tracked in #3902.
- The connection where args (the posts `where.id`/`where.author`, `categoryId`, `tagId`, and so on) will be addressed as part of the connection filter and sort redesign in #1385. Since that work deprecates and replaces the `where` args, it doesn't make sense to introduce a breaking type change to them now, right before they're deprecated. The new filter and sort args will be designed with correct `ID` typing from the start.

Keeping this open as the home for ID typing, with those two tracks carrying the actual changes.

PR #1849, opened 2021-04-07
fix: various issues on createComment Mutation

moonmeister:
1. Changed the `commentOn` type to by ID as it seems it should be. Given this is potentially a breaking change I can revert it if needed. 
2. Modified the Schema to correctly reflect what fields are required. 
3. Check for name/email when they are required in the discussion settings.

1 addresses #997

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …
Windows 10 /WSL
**WordPress Version:** …

5.7

I'll try my hand at tests here soon. I'll also see if I should modify related mutations to match.

### DIFF
--- a/src/Mutation/CommentCreate.php
+++ b/src/Mutation/CommentCreate.php
@@ -41,11 +41,11 @@ class CommentCreate {
 				'description' => __( 'The ID of the post object the comment belongs to.', 'wp-graphql' ),
 			],
 			'author'      => [
-				'type'        => 'String',
+				'type'        => boolval( get_option( 'require_name_email' ) ) ? [ 'non_null' => 'String' ] : 'String',
 				'description' => __( 'The name of the comment\'s author.', 'wp-graphql' ),
 			],
 			'authorEmail' => [
-				'type'        => 'String',
+				'type'        => boolval( get_option( 'require_name_email' ) ) ? [ 'non_null' => 'String' ] : 'String',
 				'description' => __( 'The email of the comment\'s author.', 'wp-graphql' ),
 			],
 			'authorUrl'   => [
@@ -146,8 +146,8 @@ class CommentCreate {
 			}
 
 			if ( '1' === get_option( 'require_name_email' ) && ( empty( $input['author'] ) || empty( $input['authorEmail'] ) ) ) {
-				throw new UserError( __( "This site requires you to provide a name and email address leave a comment", 'wp-graphql' ) );
-			} 
+				throw new UserError( __( 'This site requires you to provide a name and email address leave a comment', 'wp-graphql' ) );
+			}
 
 			/**
 			 * Map all of the args from GraphQL to WordPress friendly args array
--- a/src/Mutation/CommentCreate.php
+++ b/src/Mutation/CommentCreate.php
@@ -143,6 +143,10 @@ class CommentCreate {
 				throw new UserError( __( 'This site requires you to be logged in to leave a comment', 'wp-graphql' ) );
 			}
 
+			if ( '1' === get_option( 'require_name_email' ) && ( empty( $input['author'] ) || empty( $input['authorEmail'] ) ) ) {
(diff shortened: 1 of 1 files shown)


src/Mutation/CommentCreate.php
```diff
@@ -35,15 +35,17 @@ public static function register_mutation() {
...
 			'commentOn'   => [
-				'type'        => 'Int',
+				'type'        => [
+					'non_null' => 'ID',
+				],
 				'description' => __( 'The ID of the post object the comment belongs to.', 'wp-graphql' ),
 			],
 			'author'      => [
-				'type'        => 'String',
+				'type'        => boolval( get_option( 'require_name_email' ) ) ? [ 'non_null' => 'String' ] : 'String',
```
jasonbahl, line 44: If an option can make this nullable, we should probably leave it nullable in the Schema. 

Otherwise setting/unsetting this option in the admin would be a breaking change for clients.

moonmeister: That doesn't make sense to me. 

When this option is enabled on the server any mutations not passing this data will start breaking either way, right?. The only difference is whether it's a schema error or a server error (not sure if those are the correct terms 🤷 ). If this remains, then when this option is enabled the schema will correctly indicate to devs what the server requires, which seems extremely helpful. 

If this option is disabled, It can still accept name/email but it isn't required...i don't see what would break...am I missing something?

justlevine: @moonmeister the difference as I see it is that a schema change can _break_ the frontend (possibly even preventing it from compiling) whereas if it fails validation on the server, an existing frontend will just handle it like any other error response.

moonmeister: Personally if someone flips that switch and causes something to break I'd rather my build break than finding out in production.

justlevine: What exactly do you fear would break? The UserError you added will ensure that sites which require the param submit it, while keeping the schema non-breaking.

moonmeister: Okay, We don't want to change this. Keep in `nullable`

src/Mutation/CommentCreate.php
```diff
@@ -35,15 +35,17 @@ public static function register_mutation() {
...
 				'description' => __( 'The name of the comment\'s author.', 'wp-graphql' ),
 			],
 			'authorEmail' => [
-				'type'        => 'String',
+				'type'        => boolval( get_option( 'require_name_email' ) ) ? [ 'non_null' => 'String' ] : 'String',
```
moonmeister, line 48: Okay, We don't want to change this. Keep in nullable

src/Mutation/CommentCreate.php
```diff
@@ -35,15 +35,17 @@ public static function register_mutation() {
 	public static function get_input_fields() {
 		return [
 			'commentOn'   => [
-				'type'        => 'Int',
+				'type'        => [
+					'non_null' => 'ID',
```
moonmeister, line 39: Revert this.

justlevine (changes requested):
(just to make the status change more visible)

moonmeister:
@justlevine Got this updated. Looks like I might need to format. Otherwise, we should probably add tests. I can try, just point me in the right direction.

src/Mutation/CommentCreate.php
```diff
@@ -146,6 +146,10 @@ public static function mutate_and_get_payload() {
 				throw new UserError( esc_html__( 'This site requires you to be logged in to leave a comment', 'wp-graphql' ) );
 			}
 
+			if ( '1' === get_option( 'require_name_email' ) && ( empty( $input['author'] ) || empty( $input['authorEmail'] ) ) ) {
+				throw new UserError( __( 'This site requires you to provide a name and email address leave a comment', 'wp-graphql' ) );
```
justlevine, line 150:
```suggestion
				throw new UserError( esc_html__( 'This site requires you to provide a name and email address leave a comment', 'wp-graphql' ) );
```

PR #1872, opened 2021-04-27
Feature/#1869 restrict requests to authenticated users

### LINKED ISSUE #1869: Add setting to restrict GraphQL requests to authenticated users
WPGraphQL can be restricted in many ways. One way users have commonly asked to restrict the endpoint is to limit queries to authenticated users, as seen in this blog post: https://www.wpgraphql.com/2019/01/30/preventing-unauthenticated-requests-to-your-wpgraphql-api/

**Action:**
WPGraphQL should provide an option to restrict execution of API requests to authenticated users.

jasonbahl:

This adds a new setting to the GraphQL Settings page that allows WordPress Administrators to limit GraphQL requests to authenticated users. 

Screen Shot 2021-04-27 at 4 39 46 PM

By checking this setting (which is "off" by default) the GraphQL http endpoint becomes restricted to _only_ authenticated requests. 

Non-authenticated requests are rejected, and an error is returned. 

The filter `graphql_pre_restrict_endpoint` is also introduced, allowing for the default behavior to be overridden. (for example, WPGraphQL JWT Auth might override this behavior to allow the login mutation to bypass this check), or users _might_ want to allow introspection to be public, but other operations to remain restricted.

Closes #1869


tests/functional/RestrictedQueriesCept.php
```diff
@@ -0,0 +1,83 @@
...
+codecept_debug( $response_array );
+
+/**
+ * Make sure response is properly returning data as expected
+ */
+$I->assertArrayHasKey( 'errors', $response_array, 'The API is restricted and should return an error' );
+
+$I->assertArrayNotHasKey( 'data', $response_array, 'The API is restricted and should not return data' );
+
+$I->wantTo( 'Test restricting the API prevents unauthorized queries' );
```
markkelnar, line 49: This description says it will test preventing unauthorized access. But looks like the test turns off restriction and verifies the query works.  Cool.  Can you add a test for restriction enabled and a request from an authorized user succeeds?

tests/functional/RestrictedQueriesCept.php
```diff
@@ -0,0 +1,83 @@
...
+				node {
+					id
+					title
+				}
+			}
+		}
+	}'
+] ) );
+
+$I->seeResponseCodeIs( 200 );
```
markkelnar, line 35: Return code is 200 when access is restricted?

IRediTOTO:
Hi, can we choose restriced to user by role? Ex: user with role subcriber can only query their info like email, not query post, page,... 
This is important because we dont want someone easy steal own content.

PR #1874, opened 2021-04-29
Feature/#1871 limit number of batch operations

### LINKED ISSUE #1871: Add Setting to restrict number of batch operations in a single request
WPGraphQL supports Batch Queries: https://www.wpgraphql.com/docs/wpgraphql-vs-wp-rest-api/#performance

While this is largely a benefit that can lead to improved performance, it's also possible to use this feature in a way that exhausts the server. 

Limiting the number of operations that can be sent per-batch request can reduce the risk of exhausting the server. 

Server administrators should have the control to set the limit. 

**Action:**
Provide a Setting in the GraphQL Settings to set the max number of operations to allow in Batch requests. 

Batch Requests with more than the allowed max amount should not execute, and the request should respond with an error.

jasonbahl:

**NOTE:** This PR is based off of #1872, so it includes those changes as well. A diff of just these changes can be seen [here](https://github.com/jasonbahl/wp-graphql/compare/feature/%231869-restrict-requests-to-authenticated-users...jasonbahl:feature/%231871-limit-number-of-batch-operations?expand=1))

---- 

This adds a setting to the GraphQL Settings page to disable/enable Batch queries (enabled by default). 

Screen Shot 2021-04-29 at 11 51 44 AM

When batch queries are enabled, clients can send multiple GraphQL operations in a single HTTP request and get an array of responses returned. 

For example (using Postman): 

## Batch Queries Enabled

Screen Shot 2021-04-29 at 11 51 16 AM

## Batch Queries Disabled

With Batch Queries disabled, sending the same request will not execute, but instead return errors and no data:

Screen Shot 2021-04-29 at 11 59 09 AM

Closes #1871


src/WPGraphQL.php
```diff
@@ -125,7 +125,7 @@ private function setup_constants() {
 
 		// Plugin version.
 		if ( ! defined( 'WPGRAPHQL_VERSION' ) ) {
-			define( 'WPGRAPHQL_VERSION', '1.3.4' );
+			define( 'WPGRAPHQL_VERSION', '1.3.5' );
```
markkelnar, line 128: This version number down in this file seems like it's prone to disaster. Maybe another WP way to solve in a different issue ticket.

markkelnar: Oh, I guess this is coming over from the 1.3.5 release branch and master into develop.

src/Server/ValidationRules/RequireAuthentication.php
```diff
@@ -0,0 +1,110 @@
...
+
+		return true;
+	}
+
+	/**
+	 * @param ValidationContext $context
+	 *
+	 * @return callable[]|mixed[]
+	 */
+	public function getVisitor( ValidationContext $context ) {
```
markkelnar, line 66: Where is this getVisitor() used? Am I missing something?

tests/functional/BatchQueriesDisabledCept.php
```diff
@@ -0,0 +1,36 @@
...
+$I->sendPost( 'http://localhost/graphql', json_encode([
+	[
+		'query' => '{posts{nodes{id,title}}}',
+	],
+	[
+		'query' => '{posts{nodes{id,uri}}}'
+	]
+]));
+
+$I->seeResponseCodeIs( 200 );
```
markkelnar, line 25: 200 return code on query error?

src/Request.php
```diff
@@ -643,7 +676,7 @@ private function get_server() {
 			->setSchema( $this->schema )
 			->setContext( $this->app_context )
 			->setValidationRules( $this->validation_rules )
-			->setQueryBatching( true );
+			->setQueryBatching( $this->is_batch_queries_enabled() );
```
markkelnar, line 679: 👍  tested this locally and saw the errors as expected.

src/WPGraphQL.php
jasonbahl: Ah yes. . .I'm not the best at getting develop caught up with master after a release 🤦🏻‍♂️ 

Def something that can be automated with releases I think.

tests/functional/BatchQueriesDisabledCept.php
jasonbahl: Yes. Since GraphQL as a spec isn't tied to any specific transport, HTTP error codes aren't part of the spec and errors are determined _within_ the response, not as part of the HTTP status.

This allows GraphQL to be used beyond HTTP (websockets, for example) because it can behave the same regardless of transport. 

This is a good explanation: https://stackoverflow.com/questions/59729656/http-status-code-handling-in-graphql-apis#59729656

I think there's room for adjustment in some cases, but I think we'd need to really think through the why's of each case and come up with some sort of documented standard if we did want to change this for WPGraphQL specifically.

src/Server/ValidationRules/RequireAuthentication.php
jasonbahl: @markkelnar the `getVisitor` method comes from the ValidationRule class in GraphQL-PHP: https://github.com/webonyx/graphql-php/blob/0d43d7bee3fdcaf99537f6265203b7692f39e6ca/src/Validator/Rules/ValidationRule.php#L33-L43

markkelnar (approved):

PR #1880, opened 2021-04-30
Bugfix: Provide correctly-formatted modified date

### LINKED ISSUE #1696: Post modified date invalid
I'm seeing an issue where the `modified` date for posts isn't in a valid format from what I can tell

This is creating issues when passing into a `Date` function in JavaScript, particularly creating issues in Safari

You can see the difference in the formatting by looking between the different date/modified options:
image

date, dateGmt, and modifiedGmt all seem okay, but not modified

It seems like it may be similar to:
* https://github.com/wp-graphql/wp-graphql/issues/620
* https://github.com/wp-graphql/wp-graphql/pull/821

WPGraphQL Version 1.1.3
WordPress 5.6

Other plugins
image

chriszarate:
Currently, the modified date is passed straight from the model layer (MySQL `DATETIME`). Wrap in `Utils::prepare_date_response` to make it consistent with other date fields.

Fixes #1696.

https://wp-graphql.slack.com/archives/C3NM1M291/p1619813145455500

Any other comments?
-------------------

Where has this been tested?
---------------------------
**Operating System:** Debian

**WordPress Version:** 5.7.1

### DIFF
--- a/src/Model/Post.php
+++ b/src/Model/Post.php
@@ -639,7 +639,7 @@ class Post extends Model {
 					return ! empty( $punged ) ? implode( ',', (array) $punged ) : null;
 				},
 				'modified'                  => function() {
-					return ! empty( $this->data->post_modified ) && '0000-00-00 00:00:00' !== $this->data->post_modified ? $this->data->post_modified : null;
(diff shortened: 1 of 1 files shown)


jasonbahl (changes requested):
@chriszarate looks like we might have some bad assertions as I'm seeing tests failing.

jasonbahl:
@chriszarate I just pushed up a change to the test that was failing to assert that the modified date matches the format of the modified date with your change applied.

PR #1908, opened 2021-05-09
Allowing extensions to support `nodeByUri`

renatonascalves:
This adds two hooks allowing extensions to resolve their nodes when using `nodeByUri`.

no!

Screen Shot 2021-05-09 at 19 58 19

Any other comments?
-------------------
I also did a minor cleanup while I was browsing around the codebase.

Just for reference, here is where I'm using it: https://github.com/wp-graphql/wp-graphql-buddypress/issues/77

Where has this been tested?
---------------------------
**Operating System:** MacOS Big Sur 11.3

**WordPress Version:** Version 5.7.1

### DIFF
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -305,10 +305,10 @@ class NodeResolver {
 		/**
 		 * When this filter return anything other than null, it will be used as a resolved node
 		 * and the execution will be skipped.
-		 * 
-		 * This is to be used in extensions to resolve their own nodes which might not use 
+		 *
+		 * This is to be used in extensions to resolve their own nodes which might not use
 		 * WordPress permalink structure.
-		 * 
+		 *
 		 * @param null $node The node, defaults to nothing.
 		 * @param string $uri The uri being searched.
 		 * @param AppContext $content The app context.
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -302,7 +302,23 @@ class NodeResolver {
 
 		do_action_ref_array( 'parse_request', [ &$this ] );
 
-		$node = null;
+		/**
+		 * When this filter return anything other than null, it will be used as a resolved node
+		 * and the execution will be skipped.
+		 * 
+		 * This is to be used in extensions to resolve their own nodes which might not use 
+		 * WordPress permalink structure.
+		 * 
+		 * @param null $node The node, defaults to nothing.
+		 * @param string $uri The uri being searched.
+		 * @param AppContext $content The app context.
+		 * @param WP $wo WP object.
+		 */
+		$node = apply_filters( 'graphql_pre_resolve_uri', null, $uri, $this->context, $this->wp );
+
+		if ( ! empty( $node ) && ! is_null( $node ) ) {
+			return $node;
+		}
 
 		// If the request is for the homepage, determine
 		if ( '/' === $uri ) {
@@ -418,7 +434,5 @@ class NodeResolver {
 		}
 
 		return $node;
-
 	}
-
 }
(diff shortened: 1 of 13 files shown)


src/Type/InterfaceType/UniformResourceIdentifiable.php
```diff
@@ -57,8 +56,14 @@ public static function register_type( TypeRegistry $type_registry ) {
...
-					return $type;
-
+					/**
+					 * Filter the UniformResourceIdentifiable node type.
+					 *
+					 * @param null|string $type The type definition the node should resolve to.
+					 * @param object $node The $node that is being resolved.
+					 * @param TypeRegistry $type_registry The type registry.
+					 */
+					return apply_filters( 'graphql_resolve_uniform_resource_identifiable_node_type', $type, $node, $type_registry );
```
jasonbahl, line 66: @renatonascalves can you explain this filter a bit more? 

Since we're registering an InterfaceType here, this already gets passed through this filter here: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/WPInterfaceType.php#L58

I'd prefer to keep filters more central like this, vs introducing new filters for every Type that's registered.

src/Data/NodeResolver.php
```diff
@@ -45,6 +45,24 @@ public function __construct( AppContext $context ) {
...
+		 *
+		 * This is to be used in extensions to resolve their own nodes which might not use
+		 * WordPress permalink structure.
+		 *
+		 * @param null $node The node, defaults to nothing.
+		 * @param string $uri The uri being searched.
+		 * @param AppContext $content The app context.
+		 * @param WP $wp WP object.
+		 */
+		$node = apply_filters( 'graphql_pre_resolve_uri', null, $uri, $this->context, $this->wp );
```
jasonbahl, line 60: Nice. I think this is indeed a good addition!

src/Type/Union/PostObjectUnion.php
```diff
@@ -41,7 +40,7 @@ public static function register_type( TypeRegistry $type_registry ) {
 	 *
 	 * @return array
 	 */
-	public static function get_possible_types() {
+	public static function get_possible_types(): array {
```
jasonbahl, line 43: I think it could be good to start adding return types throughout, but let's do that in a separate PR?

src/Data/NodeResolver.php
jasonbahl: Now, the trick is handling cases where the node to return is actually null. 🤔 

Like, what if we had some private data that we might resolve like so (psuedo code): 

```php
add_filter( 'graphql_pre_resolve_uri', function( $node, $uri ) {

   $node_in_db = some_function_to_get_node();
   
   if ( this_node_is_private( $node_in_db ) ) {
     return null;
   } 
   
   return $node;

});
```

In this example, we'd be resolving the data, determining there is no data, or the data isn't safe to show to the requesting user, so we return null. 

Now, the NodeResolver will have executed _that_ and will continue to execute the remaining logic instead of exiting early. 

🤔

src/Type/InterfaceType/UniformResourceIdentifiable.php
renatonascalves: Thanks for the link. Didn't know about that one. This is unnecessary. Will remove it. :)

src/Data/NodeResolver.php
renatonascalves: @jasonbahl 

I put the hook in the top for a few reasons. Maybe very specific to my extension. 

- Currently, external URI is block.
- Homepage is blocked
- URLs return `null` if they are not match to any of the well known WP objects
- My extension URLs are customs, so there was no need to pass by the current `resolve_uri` logic first

In the scenario presented is okay to let the `resolve_uri` execution continues since the query is not type specific. It can match in a type and not in another. If you bail early, you might not get the node you need.

renatonascalves: And the node `Model` handles permissions. So if your node is private, it will return `null` and the execution can continue. Your private node won't show unless you have permissions to see it.

jasonbahl (approved):

Issue #1935, opened 2021-06-01
register_graphql_connection should pass more config data to sub types


jasonbahl:
When registering a connection with `register_graphql_connection()` any additional data in the `$config` array should be passed down to Types/Fields that are registered as a result of the connection. 

For example (pseudo code): 

```php
register_graphql_connection( [ 
  'fromType' => ...
  'toType' => ...
  'auth' => [
    'allowedRoles' => [ 'administrator' ],
  ],
] );
```

should pass the `auth` to `register_graphql_field` that is run during the connection registration. This would allow the connection to set which roles can/cannot resolve the connection. 

Also the `queryClass` could be defined at the connection level and passed down, which would help extensions like WooGraphQL that switch between classes for certain cases.

PR #1938, opened 2021-06-03
Extended connection configurations

kidunot89:
- Add `auth` and `queryClass` support the `register_graphql_connection()` configuration, allowing for auth restricting of connections, as well as querying results with a custom query class.
- `TypesTest` upgraded to a `WPGraphQLTestCase` and new test added.

#1935 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** Mac OS

**WordPress Version:** 5.7.2

### DIFF
--- a/src/Utils/InstrumentSchema.php
+++ b/src/Utils/InstrumentSchema.php
@@ -245,11 +245,6 @@ class InstrumentSchema {
 		$default_auth_error_message = __( 'You do not have permission to view this', 'wp-graphql' );
 
 		/**
-		 * Filter the $auth_error
-		 */
-		$auth_error = apply_filters( 'graphql_field_resolver_auth_error_message', $default_auth_error_message, $field );
-
-		/**
 		 * Check to see if
 		 */
 		if ( $field instanceof FieldDefinition && (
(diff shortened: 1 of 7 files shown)


src/Utils/InstrumentSchema.php
```diff
@@ -258,20 +253,34 @@ public static function check_field_permissions( $source, array $args, AppContext
...
 			/**
-			 * If the schema for the field is configured to "isPrivate" or has "auth" configured,
-			 * make sure the user is authenticated before resolving the field
+			 * Retrieve permissions error message.
 			 */
-			if ( empty( get_current_user_id() ) ) {
-				throw new UserError( $auth_error );
-			}
+			$auth_error = is_array( $field->config['auth'] ) && ! empty( $field->config['auth']['errorMessage'] )
+				? $field->config['auth']['errorMessage']
```
jasonbahl, line 259: I think the auth error message should be a callback, with some context passed to it. 

Because the error message might need some dynamic input to do things like `sprintf( 'You are not allowed to see %s', $field_name )` or whatever. 

Like, this is a super basic example, but the error message will likely be more useful if it can provide some dynamic context about why the error occurred, and passing context to a callback can allow that.

So, if I were configuring this, I'd want: 

```php
register_graphql_field( 'RootQuery', 'testField', [
  'type' => 'String',
  'auth' => [
    'errorMessage' => function( $some_context_passed_in ) {
       return sprintf( 'You are not allowed to see the "%s" field', $field_name_from_the_context_passed_in );
    }
  ]
] );
```

jasonbahl: Actually, similar to how we're doing `call_user_func()` here: https://github.com/wp-graphql/wp-graphql/pull/1938/files#diff-88202b0372659f22fdbdc5fcf28bc69c356ddbadb177abf2e93ee434108051b2R268

I think we can allow String or Callable. 

So you could do: 

```php
register_graphql_field( 'RootQuery', 'testField', [
  'type' => 'String',
  'auth' => [
    'errorMessage' => 'Some static error message',
  ]
] );
```

Or: 

```php
register_graphql_field( 'RootQuery', 'testField', [
  'type' => 'String',
  'auth' => [
    'errorMessage' => function( $field, $field_key, $source, $args, $context, $info, $field_resolver ) {
       return sprintf( 'You are not allowed to see the "%s" field', field_key );
    }
  ]
] );
```

jasonbahl: Meh, this is something we can add support for later

jasonbahl (approved):

PR #1949, opened 2021-06-10
#1751 - Update GraphQL-PHP to v14.8.0

### LINKED ISSUE #1751: Update GraphQL-PHP to v14.5.1
WPGraphQL currently uses GraphQL-PHP v14.4.0, but v14.5+ (released a few weeks ago) has added support for "Interfaces Implementing other Interfaces", which is needed for #1738.

jasonbahl:

- Fixes bug in Term model where `databaseId` wasn't being returned when queried on TermNodes
- Add `WPInterfaceTrait` which provides re-usable helper function for `WPObjectType` and `WPInterfaceType` to share
- Adds functionality to `WPInterfaceType` and `WPObjectType` to support Interfaces implementing Interfaces
- Fix bug where `uri` field is `non_null` String in some Types and `String` in other types. It should be `String` on all Types as there's no guarantee a `ContentNode` or `TermNode` actually will have a `uri` value
- Adds tests for Interfaces implementing Interfaces and Types implementing Interfaces

closes #1751 

Any other comments?
-------------------

With GraphQL adding support for "Interfaces Implementing Interfaces" in 2020 and GraphQL-PHP bringing support in early 2021, WPGraphQL is now adding support as well. 

We can now register an Interface, like so: 

```php
register_graphql_interface_type( 'TestInterfaceOne', [
  'fields' => [
    'one' => [
      'type' => 'String'
    ]
  ]
]);
```

Then, we can register another Interface, and declare that it should implement the first Interface: 

```php
register_graphql_interface_type( 'TestInterfaceTwo', [
  'interfaces' => [ 'TestInterfaceOne' ],
  'fields' => [
    'two' => [
      'type' => 'String',
    ],
  ]
]);
```

And then we can see in the Schema that `TestInterfaceTwo` has fields `one` and `two` and that `TestInterfaceTwo` implements `TestInterfaceOne`.

Screen Shot 2021-06-10 at 1 22 51 PM

We can go a step further and register a third Interface that Implements `TestInterfaceTwo`

```php
register_graphql_interface_type( 'TestInterfaceThree', [
  'interfaces' => [ 'TestInterfaceTwo' ],
  'fields' => [
    'three' => [
      'type' => 'String',
    ],
  ]
]);
```

And we can see that `TestInterfaceThree` now has all three fields, the one declared on itself, the one declared by `TestInterfaceTwo` and the one declared by `TestInterfaceOne`. 

Screen Shot 2021-06-10 at 1 25 36 PM

And then, we can register an ObjectType to the Schema that implements `TestInterfaceThree`: 

```php
register_graphql_object_type( 'TestTypeWithInterfaces', [
  'interfaces' => [ 'TestInterfaceThree' ],
  'fields' => [
    'four' => [
      'type' => 'String',
    ],
  ],
]);
```

And we can see that this Type now has all 4 fields, the one that was registered to the Type, and the 3 from the interfaces, and we can see that the type implements all 3 interfaces. . .the one that was declared explicitly and the ones that one implements: 

Screen Shot 2021-06-10 at 1 27 12 PM

And now we can register a field that returns the Type which implements the interfaces: 

```php
register_graphql_field( 'RootQuery', 'testTypeWithInterfaces', [
  'type' => 'TestTypeWithInterfaces',
  'resolve' => function() {
    // this is mock data we're returning, just for example sake
    return [
      'one' => 'one value',
      'two' => 'two value',
      'three' => 'three value',
      'four' => 'four value',
    ];
  }
] );
```

And now we can use Fragments to query! 

Screen Shot 2021-06-10 at 1 35 34 PM


szepeviktor:
You push tooling into your project until your source code is barely indented!

jasonbahl:
Closing to refactor some of the Interface inheritance. Will re-open as a new PR

PR #1958, opened 2021-06-15
Bug/register interfaces to types regression

### LINKED ISSUE #1957: Regression: register_graphql_interfaces_to_types() not working
With the release of WPGraphQL v1.4.0, the `register_graphql_interfaces_to_types()` method is no longer working. 

This was reported by @chriszarate in Slack. 

It appears to be a regression caused by the `graphql_object_type_interfaces` being removed.

jasonbahl:

- fixes regression to the `register_graphql_interfaces_to_types()` method 
- renames filter `graphql_object_type_interfaces` to `graphql_type_interfaces`. If you were using `graphql_object_type_interfaces` filter in your codebase, you will need to update the filter name as it now applies to both Object and Interface Types.
- applies `graphql_type_interfaces` to both WPInterfaceType and WPObjectType classes
- adds tests for registering interfaces to types and registering interfaces to other interfaces.

closes #1957


chriszarate (approved):
Gave this branch a spin. Filter is called correctly and was I able to add interfaces to existing types via the `graphql_object_type_interfaces` filter. I didn't test the `wp-graphql-gutenberg` plugin but I would expect it to work fine.

Thank you!

shrenikTrukky:
wp-graphql-gutenberg 's query for blocks not working with new version of wp-graphql 1.4.0.

jasonbahl:
@chriszarate @shrenikTrukky thanks for testing and confirming! Apologies for this regression. We have better tests in place now to prevent this in the future.

.github/workflows/testing-integration.yml
```diff
@@ -64,6 +64,7 @@ jobs:
           DEBUG: ${{ matrix.debug }}
           SKIP_TESTS_CLEANUP: ${{ matrix.coverage }}
           SUITES: wpunit
+          COMPOSER_MEMORY_LIMIT: "-1"
```
chriszarate, line 67: ♾️

chriszarate (approved):
Tested again with latest changes, looks good and filter works as expected. Thanks!

.github/workflows/testing-integration.yml
jasonbahl: ya, trying to overcome some memory limit issues, but going to make that a separate issue

chriszarate (approved):

PR #1970, opened 2021-06-22
Change docker image names specific to wp/php build

markkelnar:
When we build the wp-graphql docker images, use a name that better represents the WP and PHP version used for the image. Should allow us to reuse images as well as potentially deploy to docker hub.

Related to https://github.com/wp-graphql/wp-graphql-persisted-queries/issues/1

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …


docker/app.entrypoint.sh
```diff
@@ -39,7 +39,10 @@ if ! $( wp core is-installed --allow-root ); then
 		--allow-root
 fi
 
-# Install and activate WPGatsby
```
jasonbahl, line 42: haha! Good ol copy-pasta 🍝

docker-compose.yml
```diff
@@ -41,12 +41,13 @@ services:
...
       - app_db
-    image: wpgraphql-testing:latest
+    image: wp-graphql-testing:latest-wp${WP_VERSION-5.6}-php${PHP_VERSION-7.4}
     volumes:
       - '.:/var/www/html/wp-content/plugins/wp-graphql'
       - './.log/testing:/var/log/apache2'
       - './codeception.dist.yml:/var/www/html/wp-content/plugins/wp-graphql/codeception.yml'
-    env_file: .env.dist
+    env_file:
+      - .env
```
jasonbahl, line 50: @markkelnar so I do a lot of my testing locally outside of docker and change my .env file locally to credentials for my local MySQL environment, etc but don't touch my .env.dist file as that's used for testing. 

If we make Docker use the `.env` file, then we'll want to track changes to it, and I and others that run tests locally outside of Docker will have to be dilligent not to commit changes to the .env file. 

Why not keep `.env.dist` as the default, but use `.env` as an override locally if/when needed?

markkelnar: Glad this got your attention. I saw your git history changes and had some questions. Let's chat.

bin/run-docker.sh
```diff
@@ -9,39 +9,64 @@ set -eu
...
+    echo "Example use:";
+    echo "  composer build-app";
+    echo "  composer run-app";
+    echo "";
+    echo "  WP_VERSION=5.5.3 PHP_VERSION=7.4 composer build-app";
+    echo "  WP_VERSION=5.5.3 PHP_VERSION=7.4 composer run-app";
+    echo "";
+    echo "  WP_VERSION=5.5.3 PHP_VERSION=7.4  bin/run-docker.sh build -a";
+    echo "  WP_VERSION=5.5.3 PHP_VERSION=7.4  bin/run-docker.sh run -a";
+    exit 1
```
jasonbahl, line 24: 👌🏻

bin/run-docker.sh
```diff
@@ -9,39 +9,64 @@ set -eu
...
+TAG=${TAG-latest}
+WP_VERSION=${WP_VERSION-5.6}
+PHP_VERSION=${PHP_VERSION-7.4}
+
+BUILD_NO_CACHE=${BUILD_NO_CACHE-}
+
+if [[ ! -f ".env" ]]; then
+  echo "No .env file was detected. .env.dist has been copied to .env"
+  echo "Open the .env file and enter values to match your local environment"
+  cp .env.dist .env
```
jasonbahl, line 40: 👌🏻

jasonbahl (approved):

docker/app.entrypoint.sh
```diff
@@ -39,7 +39,10 @@ if ! $( wp core is-installed --allow-root ); then
...
+echo "Running WordPress version: $(wp core version --allow-root)"
+
+# Install and activate wp-graphql
+wp plugin install wp-graphql --allow-root
```
jasonbahl, line 45: Will this install WPGraphQL from the .org repo? 

Don't we want to use the current version of the repo and _not_ install it from .org repo? 🤔

jasonbahl (changes requested):

docker/app.entrypoint.sh
markkelnar: This was to put a version of wp-graphql in the running docker when the plugin is not already mounted in the running docker image.  It'll  use what is mounted locally in your working directory.  So in most cases it'll just say "already installed".  
I might be getting ahead of myself or out of order for the followup work where we use this docker image with another wp-graphql plugin.  In that case, wp-graphql is not found locally and is not installed, so my initial thought was it needs to be installed. I haven't fully flushed that work out.

markkelnar: Maybe we move something like we do in acf when a variable Is specified https://github.com/wp-graphql/wp-graphql-acf/blob/develop/docker/app.entrypoint.sh#L47

markkelnar: Removing that for now until I have a better understanding and use case.

bin/run-docker.sh
```diff
@@ -53,18 +78,18 @@ case "$subcommand" in
...
             case ${opt} in
-				e )
-				env_file=${OPTARG};
-				if [ ! -f $env_file ]; then
-					echo "No file found at $env_file"
-				fi
-				;;
-                a ) docker-compose up --scale testing=0;;
+                a )
+                docker-compose up --scale testing=0
```
markkelnar, line 82: I need a slash here at the end.

jasonbahl (approved):

PR #1973, opened 2021-06-24
Update type of mimeType to list of types

ljanecek:
I think it would be better if there was an option to add more types since WP Query can work with arrays.

No

No

Any other comments?
-------------------
No

Where has this been tested?
---------------------------
**Operating System:** OSX

**WordPress Version:** 5.7.2

### DIFF
--- a/src/Connection/PostObjects.php
+++ b/src/Connection/PostObjects.php
@@ -515,7 +515,9 @@ class PostObjects {
 				'description' => __( 'Filter the connection based on dates', 'wp-graphql' ),
 			],
 			'mimeType'    => [
-				'type'        => 'MimeTypeEnum',
+				'type'        => [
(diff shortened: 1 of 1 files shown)


src/Connection/PostObjects.php
```diff
@@ -515,7 +515,9 @@ public static function get_connection_args( $args = [], $post_type_object = null
 				'description' => __( 'Filter the connection based on dates', 'wp-graphql' ),
 			],
 			'mimeType'    => [
-				'type'        => 'MimeTypeEnum',
+				'type'        => [
```
jasonbahl, line 518: If we make this change, it will be a breaking change to the Schema. 

We could introduce a new `mimeTypes` field which is a `list_of => MimeTypeEnum` and then map that new input in the connection resolver.

PR #1974, opened 2021-06-24
Docker app script for setup without the exec

markkelnar:
The app entrypoint sh script is invoked by the testing script. But also invokes the exec CMD, which starts apache. For the testing image, we don't want to do that, we have a different exec CMD to run.  So it has a 'sed' command to modify the apo entrypoint script.  Kinda gross. We want to invoke the app setup for testing and potentially for other docker image.

Right now, don't really change any behavior. Just make it more useable for the next script that wants to invoke it.

Part of https://github.com/wp-graphql/wp-graphql-persisted-queries/issues/1

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/docker/app.entrypoint.sh
+++ b/docker/app.entrypoint.sh
@@ -2,5 +2,6 @@
 
 # Run app setup script.
 . app-setup.sh
+. app-post-setup.sh
 
 exec "$@"
--- a/docker/app.entrypoint.sh
+++ b/docker/app.entrypoint.sh
@@ -1,52 +1,6 @@
 #!/bin/bash
 
-if [ "$USING_XDEBUG" == "1"  ]; then
-    echo "Enabling XDebug 3"
-    mv /usr/local/etc/php/conf.d/disabled/docker-php-ext-xdebug.ini /usr/local/etc/php/conf.d/
-fi
-
-# Run WordPress docker entrypoint.
-. docker-entrypoint.sh 'apache2'
-
-set +u
-
-# Ensure mysql is loaded
-dockerize -wait tcp://${DB_HOST}:${DB_HOST_PORT:-3306} -timeout 1m
-
-# Config WordPress
-if [ ! -f "${WP_ROOT_FOLDER}/wp-config.php" ]; then
-    wp config create \
-        --path="${WP_ROOT_FOLDER}" \
-        --dbname="${DB_NAME}" \
-        --dbuser="${DB_USER}" \
-        --dbpass="${DB_PASSWORD}" \
-        --dbhost="${DB_HOST}" \
-        --dbprefix="${WP_TABLE_PREFIX}" \
-        --skip-check \
-        --quiet \
-        --allow-root
-fi
-
-# Install WP if not yet installed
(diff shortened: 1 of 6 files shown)


docker/testing.Dockerfile
```diff
@@ -36,9 +36,6 @@ ENV PATH "$PATH:~/.composer/vendor/bin"
 # Configure php
 RUN echo "date.timezone = UTC" >> /usr/local/etc/php/php.ini
 
-# Remove exec statement from base entrypoint script.
-RUN sed -i '$d' /usr/local/bin/app-entrypoint.sh
```
markkelnar, line 40: This tells me we want everything the setup part of the script does to get WP running.

docker/app.entrypoint.sh
```diff
@@ -1,52 +1,6 @@
...
-wp plugin activate wp-graphql --allow-root
-
-# Set pretty permalinks.
-wp rewrite structure '/%year%/%monthnum%/%postname%/' --allow-root
-
-wp db export "${PROJECT_DIR}/tests/_data/dump.sql" --allow-root
+# Run app setup script.
+. app-setup.sh
 
 exec "$@"
```
markkelnar, line 6: docker up/run runs this as the ENTRYPOINT and "$@" is apache-foreground to start apache.

kidunot89 (dismissed):
@markkelnar Great work, I hadn't thought of doing it that way. 👍🏿

markkelnar:
> @markkelnar Great work, I hadn't thought of doing it that way. 👍🏿

Thanks. I might separate this even more as I'm trying to build on it with another repo. Don't know yet.

docker/app.entrypoint.sh
```diff
@@ -1,52 +1,7 @@
...
-# Activate wp-graphql
-wp plugin activate wp-graphql --allow-root
-
-# Set pretty permalinks.
-wp rewrite structure '/%year%/%monthnum%/%postname%/' --allow-root
-
-wp db export "${PROJECT_DIR}/tests/_data/dump.sql" --allow-root
+# Run app setup script.
+. app-setup.sh
+. app-post-setup.sh
```
markkelnar, line 5: My thinking with two scripts is that it would allow an inherited dockerfile entrypoint script to run the setup.sh, do some other stuff, then run the post-setup.sh.  Like install wp-graphl from somewhere if it's not installed, like how wp-graphql-acf needs.

docker/app.setup.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
markkelnar, line 1: Technically we don't have to change the name of this script. It could still be app.entrypoint.sh.

markkelnar: Now that I've slept on this, I kinda like the idea of renaming it back to app.entrypoint. Makes it more obvious.

markkelnar: I take that back I forgot the entrypoint.sh file still exists and is needed cause it invokes the `exec $@`.

jasonbahl (approved):

PR #1977, opened 2021-06-28
Support passing the same string for graphql_single_name and graphql_plural_name

### LINKED ISSUE #1903: No Schema and GraphQL Queries return 500 error when a custom post type word plural singular is same
**Issue:** The GraphQL IDE shows No Schema and all queries return 500 error.

**Reproduce:** Declare a custom post type either manually or via the Custom Posts Type UI Plugin.
and set the following to make it show show_in_graphql, 
```
'show_in_graphql' => true,
'graphql_single_name' => 'Deer',
'graphql_plural_name' => 'Deer',
```

Leaving one route off isn't an option as that also breaks things.

With the two routes not sure how to fix this but this is enough of a problem in ENGLISH, let alone other languages with different pluralization rules.  Internationalization fragility.

https://www.vappingo.com/word-blog/101-words-that-are-both-plural-and-singular/

**Workaround:** For anyone else facing this, i'm working around by adding an "s" to the plural even if that doesn't make linguistic sense.

This also seems very fragile and took me a long time to figure out what was causing the issue with no good feedback and the debug logs were no help.  At the very least checking for the issue and warning would be helpful.  Or maybe creating the plural route automatically so it isn't so fragile?  Or catching the error and just not adding an endpoint for that type until it is fixed.

having all the wpgraphql endpoints break is a pretty huge impact for properly leaving the "s" off of a pluralized word that isn't supposed to have one and could be confusing to other users as well.  especially ones where English isn't their first language.

Thanks!

apmatthews:
For both custom post types and custom taxonomies:
- Revert throwing an exception when `graphql_single_name` and `graphql_plural_name` are the same.
- When `graphql_single_name` and `graphql_plural_name` are the same, prepend the word `all` onto the `graphql_single_name` to create the object connection's `fromFieldName`.

Closes #1903 

Any other comments?
-------------------
I'm brand new into the world of GraphQL, so feedback from a more experienced eye would be greatly appreciated. I'm just jumping in to help this issue along since it's more likely to be a problem as people start using the [Atlas Content Modeler plugin](https://github.com/wpengine/atlas-content-modeler).

- There were a few different solutions discussed in #1903. Prepending `all` made the most sense to me,  but I'm happy to go back and explore the alternatives.
- The plural name is used in many different contexts throughout the code base, so I avoided changing `graphql_plural_name` everywhere. Doing so would probably involve running a filter to change the args on `register_post_type()` or `register_taxonomy()` to check for naming conflicts, which seemed a little heavy handed. It would also result in error messages that read awkwardly: "Sorry, you are not allowed to delete allDeer".
- A couple of notable instances of `graphql_plural_name` have been left unchanged despite them being used in the schema. In these cases, I don't believe there is a risk of singular/plural conflicts like there is at the root level. These instances are:
    - **`PostObjectMutation`**: Still sets object terms using the unaltered `graphql_plural_name`.
    - **`PostObjectCreate` input field config**:  Still registers term inputs under the unaltered `graphql_plural_name`.
- Considering the points above, I have only attempted to resolve conflicts between singular and plural names at the root query level. More specifically, I've targeted conflicts between a `PostObject` or `TermObject` type and the respective `RootQueryToPostObject` or `RootQueryToTermObject` connection.

Open questions:

- I _think_ the tests here are sufficient to validate the changes, but again, I am very new to GraphQL and could be missing something.
- The `PostObjects` connection config defaulted to `'fromFieldName' => lcfirst( $graphql_object->graphql_plural_name )`, whereas the TermObjects connection config did not use `lcfirst()`. Was this intentional or should the `TermObjects` connection config also use `lcfirst()`? I left it alone because I think updating it could be a breaking change.
- Does this sufficiently nullify the need for `graphql_single_name` and `graphql_plural_name` to be unique? If so, should the [docs](https://www.wpgraphql.com/docs/custom-post-types/#using-custom-post-types-with-wpgraphql) be updated with this PR?

Where has this been tested?
---------------------------
**Operating System:** macOS Big Sur (Running the project's docker 'app' environment)

**WordPress Version:** 5.6.2 (In the project's docker 'testing' environment)


src/Connection/PostObjects.php
```diff
@@ -362,12 +362,18 @@ public static function get_connection_config( $graphql_object, $args = [] ) {
 			unset( $connection_args['stati'] );
 		}
 
+		$from_field_name = lcfirst( $graphql_object->graphql_plural_name );
+
+		if ( $graphql_object->graphql_single_name === $graphql_object->graphql_plural_name ) {
+			$from_field_name = 'all' . ucfirst( $graphql_object->graphql_single_name );
```
jasonbahl, line 368: I think I'm mostly rubber-ducking here, not necessarily suggesting a specific change to this PR.

One thing that concerns me with the `all${SingleName}` convention is for `Type->Type` connections. 

Like: 

```graphql
{
  allCustomPostType {
    nodes {
       id
       title
       allCustomTaxonomy {
         nodes {
           id
           name
         }
       }
    }
  }
}
```

The `allCustomTaxonomy` reads weird to me when coming from a connection from another Type, because you're not _really_ asking for _all_ nodes of that Type, you're asking for selected nodes of that Type. . .I think. 🤔  

Naming things is hard.

I think this (`all{SingleName}`) pattern is a good improvement over what we have today though!

src/Connection/PostObjects.php
```diff
@@ -362,12 +362,18 @@ public static function get_connection_config( $graphql_object, $args = [] ) {
 			unset( $connection_args['stati'] );
 		}
 
+		$from_field_name = lcfirst( $graphql_object->graphql_plural_name );
+
+		if ( $graphql_object->graphql_single_name === $graphql_object->graphql_plural_name ) {
```
jasonbahl, line 367: Perhaps we can _also_ support this pattern if `graphql_plural_name` is not set. 

Like if I do: 

```php
register_post_type( 'test', [
  'show_in_graphql' => true,
  'graphql_single_name' => 'test'
]);
```
Perhaps the lack of `graphql_plural_name` should also lead to `test` and `allTest` in the Schema with no errors 🤔

src/Connection/PostObjects.php
apmatthews: Totally agree! We only want `all` prepended on RootQuery connections. Once you get into a Type->Type situation I think you're pretty safe from naming collisions between the single/plural, so we can just stick with the provided plural name as the field name.

Turns out I had achieved that for PostObject->TermObject connections but not the other way around! I took another look and think I've got a more targeted solution now. PR has been updated. The example you provided would now read:

```graphql
{
  allCustomTaxonomy {
    nodes {
      id
      name
      deer {
        nodes {
          id
          title
        }
      } 
    }
  }
  allCustomPostType {
    nodes {
       id
       title
       customTaxonomy {
         nodes {
           id
           name
         }
       }
    }
  }
}
```

src/Connection/PostObjects.php
apmatthews: We still need _something_ set for `graphql_plural_name` for the sake of generating error messages and for simpler Type->Type connections. Checking for an unset or empty plural name in all of these case could get messy and cause a lot of branching. Maybe we could add a filter on `register_post_type()` and `register_taxonomy()` args that would set `$args['graphql_plural_name'] = $args['graphql_single_name']` if plural is not provided? I'll give that a try in a follow up PR.

jasonbahl (approved):

PR #1978, opened 2021-06-30
Use docker compose instead of docker-compose for up

markkelnar:
`composer run-app` results in error. When the WP/PHP environment variables were added recently, need to also switch to `docker compose up`.

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/bin/run-docker.sh
+++ b/bin/run-docker.sh
@@ -79,7 +79,7 @@ case "$subcommand" in
         while getopts "e:at" opt; do
             case ${opt} in
                 a )
-                docker-compose up --scale testing=0 \
+                docker compose up --scale testing=0 \
                     -e WP_VERSION=${WP_VERSION} \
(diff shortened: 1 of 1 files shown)


apmatthews:
This resolves the original issue I was running into where `composer run-app` was just giving me docker usage instructions. I think it's ignoring the `--scale testing=0` flag though. `composer run-app` still tries to boot all three services (`app`, `app_db`, and `testing`) and the `testing` service fails when it can't connect to the DB. It doesn't affect my ability to access `app` at localhost:8091, so maybe that can be handled in a separate PR.

markkelnar:
I was running docker version 20.10.6 and did not see this issue.  Updated to 20.10.7 and now see testing also starting.  Gonna look into that.  :hmm:

jasonbahl (approved):

PR #1984, opened 2021-07-02
PHP 8 support

### LINKED ISSUE #1976: wp-graphql composer install FAILs on php8
with
```
php -v
	PHP 8.0.7 (cli) (built: Jun  1 2021 18:43:05) ( NTS gcc x86_64 )
	Copyright (c) The PHP Group
	Zend Engine v4.0.7, Copyright (c) Zend Technologies
	    with Zend OPcache v8.0.7, Copyright (c), by Zend Technologies
	    with Xdebug v3.0.4, Copyright (c) 2002-2021, by Derick Rethans
```

installing wpgraphql via composer FAILs
```
composer require --with-all-dependencies wp-graphql/wp-graphql
	Using version ^1.4 for wp-graphql/wp-graphql
	./composer.json has been updated
	Running composer update wp-graphql/wp-graphql --with-all-dependencies
	Loading composer repositories with package information
	Updating dependencies
	Your requirements could not be resolved to an installable set of packages.

	  Problem 1
	    - wp-graphql/wp-graphql[v1.4.0, ..., v1.4.3] require ivome/graphql-relay-php 0.5.0 -> satisfiable by ivome/graphql-relay-php[v0.5.0].
	    - ivome/graphql-relay-php v0.5.0 requires php >=5.4,<8.0-DEV -> your php version (8.0.7) does not satisfy that requirement.
	    - Root composer.json requires wp-graphql/wp-graphql ^1.4 -> satisfiable by wp-graphql/wp-graphql[v1.4.0, v1.4.1, v1.4.2, v1.4.3].
```

checking deps
@ https://github.com/wp-graphql/wp-graphql/blob/develop/composer.json
```
"require": {
	"php": ">=7.1.0",
	"webonyx/graphql-php": "14.8.0",
	"ivome/graphql-relay-php": "0.5.0"},
```

and
@ https://github.com/ivome/graphql-relay-php/blob/v0.5.0/composer.json
```
"require": {
	"php": ">=5.4,<8.0-DEV",
	"webonyx/graphql-php": ">=0.7.0"
```

and
@ https://github.com/ivome/graphql-relay-php/blob/master/composer.json
```
"require": {
	"php": "^7.1 || ^8.0",
	"webonyx/graphql-php": "^14.0"
```

jasonbahl:
This updates WPGraphQL to support PHP 8

No functional code has changed, we've just updated the composer.json to declare php 8 support and made some adjustments to the test environment

Closes #1976 

Any other comments?
-------------------
- updates php declaration in composer.json to support php 8.0
- updates `graphql-php` to 14.9.0
- update `graphql-relay-php` to 0.6.0
- update `wp-graphql-testcase` to ^2.1.1
- update test matrix to run tests on php 8
- update TypesTest.php to work with newest version of WPGraphQL Test Case


docker/app.setup.sh
```diff
@@ -8,6 +8,9 @@ fi
 # Run WordPress docker entrypoint.
 . docker-entrypoint.sh 'apache2'
 
+echo "Delete the wp-config.php"
```
markkelnar, line 11: What problem was this solving? Is this wp-config.php from a local WP that's already been running? So you want to remove it each time the app starts?

jasonbahl: When @kidunot89 and I were walking through things it seems like the php8 docker image already had a wp-config.php but we want to use ours that we generate with wp-cli, so this deletes the one from Docker and uses wp-cli to generate one with our creds.

markkelnar: Was it this failure? https://github.com/wp-graphql/wp-graphql/runs/2975991532?check_suite_focus=true#step:8:40

jasonbahl: Ah yes. I think it was actually WP 5.7.2, not PHP 8. long day.

markkelnar: There's a note in the WP docker hub page

> NOTE: Since March 2021, WordPress images use a customized wp-config.php that pulls the values directly from the environment variables defined above (see wp-config-docker.php in docker-library/wordpress#572 and docker-library/wordpress#577). As a result of reading environment variables directly, the cli container also needs the same set of environment variables to properly evaluate wp-config.php.

Looking at these now in case there's something else we'll need to consider.

jasonbahl: ok, so we can change our environment variable names to match and get rid of the delete/regenerate wp-config?

jasonbahl:
@markkelnar thanks for the tweaks!

Issue #1873, opened 2021-04-28
GraphQL WordPress Plugin Performance Issue


gagan0123:
GraphQL Plugin's performance deteriorates significantly with the number of post-types and taxonomies defined in the WordPress.

### Reproduce the bug:
1. Try to add more than 20-30 post types, with around the same number of taxonomies defined for them. 
2. Check out the response timing of GraphQL endpoint.

To further add to this issue, even caching does not resolve this performance problem. I've tried with 
1. Object Caching (Redis)
2. MySQL caching(I know I shouldn't have)
3. Using a [GraphQL caching plugin for WordPress](https://github.com/valu-digital/wp-graphql-cache)

Screencast for the same:

[![](http://img.youtube.com/vi/Et7iFbV-H8A/0.jpg)](http://www.youtube.com/watch?v=Et7iFbV-H8A "GraphQL WP Plugin Performance Issue")

### Query statistics:
| Request Type:                                | WordPress Homepage | GraphQL Request(just endpoint) | WP-JSON Request(just endpoint) | GraphQL Request(Single Post) | WP-JSON Request(Single Post) |
|----------------------------------------------|--------------------|--------------------------------|--------------------------------|------------------------------|------------------------------|
| Timing with default WordPress                | 85 ms              | 220 ms                         | 80 ms                          | 230 ms                       | 80 ms                        |
| Timinig with 10 post types and 10 taxonomies | 90 ms              | 400 ms                         | 110 ms                         | 407 ms                       | 100 ms                       |
| Timing with 50 post types and 50 taxonomies  | 160 ms             | 1650 ms                        | 260 ms                         | 1650 ms                      | 230 ms                       |

Here's a link to XDebug profiling for the above requests(I've skipped the 10 post type in there)
https://drive.google.com/drive/folders/1-U_8Gb4BsL_UEHMyO0NKthRRkXQmq_2i?usp=sharing

#### What version of `WPGraphQL` you're using, and the platform(s) you're running it on
`WPGraphQL` version: 1.3.5

#### What other plugins you're using
No plugin active on the site except `WPGraphQL`

#### The behavior you expect to see, and the actual behavior
The number of post types and taxonomies defined in a WordPress site should have minimal/no impact on the performance of GraphQL queries. 

Any website running WooCommerce, has multiple post types defined for things like:
1. Orders
2. Products
3. Variations
4. Coupons

having custom taxonomies like:
1. Product type
1. Product visibility
1. Product categories
[...]

PR #2000, opened 2021-07-12
Fix lazy-loading types

chriszarate:
This is a ~draft~ PR exploring a fix to #1873.

One specific performance problem we noticed was that [lazy-loading of types](https://webonyx.github.io/graphql-php/schema-definition/#lazy-loading-of-types) was not working. Upon investigation, it seems there are several reasons for this:

1. `TypeRegistry::register_type` instantiates type classes immediately, instead of waiting for the `$schema_config->typeLoader` to be called.
2. When an object or interface type is loaded, all of its fields are immediately defined and their types loaded, instead of providing callables to allow lazy-loading of just the fields that are needed.
3. When connections are registered, the type is being loaded solely to check if it exists.
4. The entire schema is being loaded and instrumented via `$schema->getTypeMap()` in order to wrap field resolvers and provide the `graphql_resolve_field` filter.

The first three problems are addressed in this PR; the final issue needs discussion about the best way to fix it. The current method of wrapping field resolvers is incompatible with lazy type loading. Note that my temporary "fix" simply disables the filter, which also disables field sanitization and breaks some tests. cc: @jasonbahl 

I've done some anecdotal performance analysis but would appreciate some more structured testing from those that have experience performance issues. cc: @gagan0123 @andreasradloff @drtechie

I've added a test to `AssertValidSchemaTest.php` to hopefully prevent future disruption of lazy type loading.


tests/wpunit/AssertValidSchemaTest.php
```diff
@@ -210,4 +210,55 @@ public function testContentTypesOfTaxonomy() {
...
+
+					return [];
+				},
+			]
+		);
+
+		register_graphql_field(
+			'RootQuery',
+			'example',
+			[ 'type' => 'TestLazyType' ]
```
jasonbahl, line 247: I think we'll need to also add a test for cases where fields are of wrapping Types. 

like `'type' => [ 'list_of' => [ 'non_null' => 'TestLazyType' ] ]`

When I was exploring fixing this issue I believe I was running into something similar to: https://github.com/webonyx/graphql-php/issues/779

It could have been caused by something else I hadn't yet addressed in the WPGraphQL codebase, though. 

I'll explore this today.

src/Utils/InstrumentSchema.php
```diff
@@ -105,7 +97,7 @@ protected static function wrap_fields( array $fields, string $type_name ) {
 			 * @throws Exception
 			 * @since 0.0.1
 			 */
-			$field->resolveFn = function ( $source, array $args, AppContext $context, ResolveInfo $info ) use ( $field_resolver, $type_name, $field_key, $field ) {
+			$field->resolveFn = static function ( $source, array $args, AppContext $context, ResolveInfo $info ) use ( $field_resolver, $type_name, $field_key, $field ) {
```
chriszarate, line 100: What does the `static` keyword do on this closure? Seems redundant inside a static method.

src/WPGraphQL.php
```diff
@@ -364,18 +367,17 @@ private function filters() {
 		/**
 		 * Instrument the Schema to provide Resolve Hooks and sanitize Schema output
 		 */
-		add_filter(
-			'DISABLED_graphql_schema',
+		add_filter( 'graphql_get_type',
 			[
-				'\WPGraphQL\Utils\InstrumentSchema',
-				'instrument_schema',
+				InstrumentSchema::class,
```
chriszarate, line 372: TIL

tests/wpunit/AssertValidSchemaTest.php
chriszarate: Added!

src/Utils/InstrumentSchema.php
jasonbahl: There seems to be some performance impact with explicitly declaring static on closures: https://stackoverflow.com/questions/19899468/php-closures-why-the-static-in-the-anonymous-function-declaration-when-bindin#answer-64481894

That said, my real reasoning for the change was that one of the IDE rules I have setup was yelling at me to declare this static. 🤣

tests/wpunit/AssertValidSchemaTest.php
jasonbahl: Amazing! Thanks so much! 🙌🏻

jasonbahl:
@chriszarate thanks so much for working on this! Very much appreciated. 

I've pushed up some changes as well to cover some additional cases where some types might need to be eagerly loaded still. For example the `DefaultTemplate` Type is only exposed as via ResolveType on an interface and GraphQL-PHP needs to know about this kind of Type in advance. (See the notes about passing Types to the schema here: https://webonyx.github.io/graphql-php/schema-definition/)

I've added the ability to pass `'eagerlyLoadType' => true` when registering a Type (see: https://github.com/wp-graphql/wp-graphql/pull/2000/files#diff-fea80346fad1aa40b8c8872a4d3a8f73a343b677fe4e8410e1224fcfb502a1a5R430) and any type that has that will be eagerly loaded and all other Types will be Lazily loaded. 
 
 With these changes in place I've ran some super non-scientific tests. 
 
 With the following code in place to test with many Post Types and Taxonomies (as inspired by @gagan0123): 
 
 ```php
	for ($x = 0; $x <= 20; $x++) {
		register_post_type( 'post_type_'. $x, [
			'show_in_graphql' => true,
			'graphql_single_name' => 'PostType' . $x,
			'graphql_plural_name' => 'allPostType' . $x
		]);
		register_taxonomy( 'taxonomy_' .$x, [ 'post_type_' . $x ], [
			'show_in_graphql' => true,
			'graphql_single_name' => 'Taxonomy' . $x,
			'graphql_plural_name' => 'allTaxonomy' . $x
		]);
	}
} );
 ```
 
 ## Before this PR applied:
 
- Introspection Query: 8.26s
- Posts query: 403ms

## After this PR applied:

- Introspection Query: 1.68s
- Posts query: 154ms

src/Registry/TypeRegistry.php
```diff
@@ -774,7 +821,13 @@ public function get_type( string $type_name ) {
...
 	 */
 	public function get_types() {
-		return $this->types;
+
+		// The full map of types is merged with eager types to support the
+		// rename_graphql_type API.
+		//
+		// All of the types are closures, but eager Types are the full
+		// Type definitions up front
+		return array_merge( $this->types, $this->get_eager_type_map() );
```
chriszarate, line 830: I don't think we need to merge this, since `get_eager_type_map()` calls `get_type()`, which populates `$this->types`.

Seems like `get_eager_type_map` could just be `get_eager_types` and return `void`, since it populates `$this->types`

TylerBarnes:
🔥 🔥 🔥 

Just tried this on gatsbyjs.com and there's a noticeable improvement!

### before

introspection 4.040s
node sourcing 96.539s

### after

introspection 2.816s
node sourcing 87.128s

jasonbahl:
I've run some tests similar to those that @gagan0123 ran in #1873 and this PR is showing some significant improvements!

To get the numbers below, I executed 10 requests of each scenario and then averaged the times. 

### Query Statistics:

| Request Type                             | WP Homepage | REST Endpoint | GraphQL Endpoint | REST Singe Post | GraphQL Single Post | GraphiQL Introspection |
|------------------------------------------|-------------|---------------|------------------|-----------------|---------------------|------------------------|
| WPGraphQL 1.3.5                          | 70 ms       | 77 ms         | 150 ms           | 73 ms           | 160 ms              | 2585 ms                |
| PR #2000                                 | 77 ms       | 72 ms         | 93 ms            | 73 ms           | 105 ms              | 658 ms                 |
| WPGraphQL 1.3.5 + 15 CPTs and Taxonomies | 87 ms       | 78 ms         | 292 ms           | 79 ms           | 284 ms              | 6295 ms                |
| PR #2000 + 15 CPTs and Taxonomies        | 89 ms       | 77 ms         | 150 ms           | 79 ms           | 157 ms              | 1322 ms                |

While this shows that WPGraphQL still has some overhead overhead compared to the WP homepage and REST API requests, this PR shows some SIGNIFICANT improvements! 

I've got some other ideas on how we can try and reduce that additional overhead further and look forward to exploring it even more. 

----

To help visualize this data even more, I created some before/after charts:

## WPGraphQL v1.3.5

Screen Shot 2021-07-23 at 1 18 16 PM

### PR #2000 

Screen Shot 2021-07-23 at 1 18 48 PM

Prior to the PR we can see that GraphQL requests were near and above 150ms and after we can see it closer to 100ms!

----

## WPGraphQL v1.3.5 (+ 15 Custom Post Types / Taxonomies)

Screen Shot 2021-07-23 at 1 20 03 PM

## PR #2000 (+ 15 Custom Post Types / Taxonomies)

Screen Shot 2021-07-23 at 1 20 14 PM

Prior to the PR, the GraphQL requests were near the 300ms and now they're cut down closer to 150ms!

----

src/Registry/TypeRegistry.php
jasonbahl: Without the `array_merge()` everything fails 😬 

Screen Shot 2021-07-23 at 1 23 32 PM

The reason we need this:

`get_eager_type_map()` returns an array of Types that need to be loaded up-front, instead of referencing a callback. GraphQL-PHP needs this for Types that are not referenced directly by a field, but only show as part of a Union or Interface. 

So while `$this->types[]` looks like: 

```php
[
  'RootQuery' => function() {}, // callback
  'RootMutation' => function() {}, // callback 
  'Post' => function() {}, // callback
  ...
]
```

The eager type map looks like: 

```php
[
  'EagerType' => WPObjectType // Full Type Def, not a callback
]
```

So the output of `get_types()` needs to be a mix of the lazy loaded Types that map to a callback (which will load a Type Object) _and_ the eager types that map directly to a Type Object: 

```php
[
  'RootQuery' => function() {},
  'RootMutation' => function() {},
  'Post' => function() {},
  'EagerType' => WPObjectType 
  ...
]
```

chriszarate: Yup yup, got it. I read that method incorrectly. Thanks!

jasonbahl:
Here's one more chart that improves developer experience quite a bit: 

### Introspection in GraphiQL

This is how long GraphiQL in the WP Admin takes to load the schema

Screen Shot 2021-07-23 at 2 33 29 PM

With 15 Custom Post Types and Taxonomies registered it was taking WPGraphQL 6+ seconds to return the Schema. 

Now, it's taking about 1 second! 

## Before

~6 seconds before GraphiQL has the Schema and is ready to use

graphiql-loading-1 3 5

## Now

~1 second and GraphiQL is ready for interaction

graphiql-loading-pr#2000

TylerBarnes:
dang, 6 seconds to 1 second is impressive! :clap: amazing work!

moonmeister:
Someone giv3e @chriszarate an award. This is stellar.

drtechie:
**Before plugin update**

Screenshot 2021-08-14 at 11 50 32 AM

**After plugin update**

Screenshot 2021-08-14 at 11 50 42 AM

The percentage change for requests served per 60 seconds is 49.26% 🔥 🔥 

Thank you @chriszarate @jasonbahl

src/Utils/InstrumentSchema.php
renatonascalves: Nice! Tks for the link.

PR #2001, opened 2021-07-12
Use mariadb instead of mysql for tests

chriszarate:
Use `mariadb` instead `mysql` image in Docker environment. The `mariadb` image provides images for ARM platforms (including M1 / "Apple Silicon" Macs) without the need for additional Docker configuration.

Where has this been tested?
---------------------------
**Operating System:** macOS on M1 chip

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,7 @@ services:
       local:
 
   app_db:
-    image: mysql:5.7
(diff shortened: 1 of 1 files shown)


markkelnar:
This is a great idea.

jasonbahl (approved):

PR #2006, opened 2021-07-13
Run tests from PROJECT_DIR var

markkelnar:
Use the $PROJECT_DIR from the .env file instead of the hardcoded directory in the script. This will allow other docker images to resuse this docker file to base their tests on and run codeception tests from their own PROJECT_DIR.

Related to work for persisted queries plugin tests.

Should not change behavior for this repo.


jasonbahl:
@markkelnar for this PR "I am thankful"

PR #2015, opened 2021-07-15
Debug schema diff

szepeviktor:
Let's see CI output! 👀

### DIFF
--- a/.github/workflows/schema-linter.yml
+++ b/.github/workflows/schema-linter.yml
@@ -78,6 +78,5 @@ jobs:
 
       - name: Run Schema Inspector
         run: |
-          diff -u /tmp/${{ steps.get-latest-tag.outputs.tag }}.graphql /tmp/schema.graphql
           # This schema and previous release schema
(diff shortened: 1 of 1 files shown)


.github/workflows/schema-linter.yml
```diff
@@ -79,4 +79,4 @@ jobs:
       - name: Run Schema Inspector
         run: |
           # This schema and previous release schema
-          node_modules/.bin/graphql-inspector diff /tmp/${{ steps.get-latest-tag.outputs.tag }}.graphql /tmp/schema.graphql
+          npx graphql-inspector diff /tmp/${{ steps.get-latest-tag.outputs.tag }}.graphql /tmp/schema.graphql
```
jasonbahl, line 82: Probably don't need to install this if we're using npx, eh?

https://github.com/wp-graphql/wp-graphql/blob/86ff565b801aed661adec7dc3129b1469a007aa8/.github/workflows/schema-linter.yml#L74-L77

szepeviktor: You're right.

```
npx --yes -p @graphql-inspector/cli -c 'graphql-inspector diff a b'
```

But this is not explicit :(

PR #2019, opened 2021-07-19
Tag and deploy testing image for use from the registry

markkelnar:
Similar to how we push the built docker image for the app, let's also push the testing built image to the github container registry.  This will help with other plugins that want to reuse that testing image/container flow instead of rebuilding all the pieces.  

First use case as part of https://github.com/wp-graphql/wp-graphql-persisted-queries/issues/2.


jasonbahl (approved):

markkelnar:
😱

jasonbahl:
👊🏻

PR #2032, opened 2021-07-27
Release/v1.6.0

jasonbahl:
# Release Notes

### Chores / Bugfixes

- ([#2000](https://github.com/wp-graphql/wp-graphql/pull/2000)): This fixes an issue where all Types of the Schema were loaded for each GraphQL request. Now only the types required to fulfill the request are loaded on each request. Thanks @chriszarate!
- ([#2031](https://github.com/wp-graphql/wp-graphql/pull/2031)): This fixes a performance issue in the WPGraphQL model layer where determining whether a User is a published author was generating expensive MySQL queries on sites with a lot of users and a lot of content. Thanks @chriszarate!

----

# Performance Improvements

This Release has a focus on performance improvements. 

## Lazy Loading Schema

GraphQL-PHP, the GraphQL library used under the hood, supports "Lazy Loading" of GraphQL Types, which allow Types to be registered as callbacks, and the full Type definitions are loaded only when needed. 

For example, the following query: 

```graphql
{
  posts {
    nodes {
      id
      title
    }
  }
}
```

Should only load the following Type Definitions: 

- RootQuery
- RootQueryToPostConnection
- RootQueryToPostConnectionEdge
- Post
- (any Interfaces, such as Node and NodeWithTitle, that are implemented by these Types)

However, WPGraphQL wasn't using this feature of GraphQL-PHP quite right and was instead loading ALL GraphQL Type definitions on _every_ GraphQL request. 

This caused a lot of unnecessary overhead. 

The improvements in this release show significant reduction in response time. 

Below are some details on how things have improved. Results may vary based on your specific server setup, your schema, the queries you run, etc, but in general every user of WPGraphQL should see performance improvements after this release.

### Query Statistics:

| Request Type                             | WP Homepage | REST Endpoint | GraphQL Endpoint | REST Singe Post | GraphQL Single Post | GraphiQL Introspection |
|------------------------------------------|-------------|---------------|------------------|-----------------|---------------------|------------------------|
| WPGraphQL 1.3.5                          | 70 ms       | 77 ms         | 150 ms           | 73 ms           | 160 ms              | 2585 ms                |
| v1.6.0                                | 77 ms       | 72 ms         | 93 ms            | 73 ms           | 105 ms              | 658 ms                 |
| WPGraphQL 1.3.5 + 15 CPTs and Taxonomies | 87 ms       | 78 ms         | 292 ms           | 79 ms           | 284 ms              | 6295 ms                |
| v1.6.0 + 15 CPTs and Taxonomies        | 89 ms       | 77 ms         | 150 ms           | 79 ms           | 157 ms              | 1322 ms                |

To help visualize this data even more, I created some before/after charts:

## WPGraphQL v1.3.5

Screen Shot 2021-07-23 at 1 18 16 PM

### v1.6.0

Screen Shot 2021-07-23 at 1 18 48 PM

Prior to the PR we can see that GraphQL requests were near and above 150ms and after we can see it closer to 100ms!

----

### WPGraphQL v1.3.5 (+ 15 Custom Post Types / Taxonomies)

Screen Shot 2021-07-23 at 1 20 03 PM

### v1.6.0 (+ 15 Custom Post Types / Taxonomies)

Screen Shot 2021-07-23 at 1 20 14 PM

Prior to the PR, the GraphQL requests were near the 300ms and now they're cut down closer to 150ms!

And, all together we can see the different types of requests for v1.3.5 and v1.6.0:

Screen Shot 2021-07-23 at 2 42 37 PM

### Faster Introspection

Introspection Queries (used for tools such as GraphiQL and Gatsby Source WordPress) is also MUCH faster now. 

This is how long GraphiQL in the WP Admin takes to load the schema

Screen Shot 2021-07-23 at 2 33 29 PM

With 15 Custom Post Types and Taxonomies registered it was taking WPGraphQL 6+ seconds to return the Schema. 

Now, it's taking about 1 second! 

**v1.3.5**

~6 seconds before GraphiQL has the Schema and is ready to use

graphiql-loading-1 3 5

**v1.6.0**

~1 second and GraphiQL is ready for interaction

graphiql-loading-pr#2000

----

## Improved User Model Check

The WPGraphQL Model Layer does a series of checks on each node to make sure that the requesting user has permission to see the nodes requested. 

When a list of posts is queried in GraphQL, the Author(s) of the posts must also be loaded behind the scenes to properly determine whether each post in the list can be returned to the requesting user. 

Let's say a list of draft posts were queried. If the request were public, no results would be returned. But, if the request were authenticated, then we'd need to know who the requesting user is, what capabilities they have, and whether they're the owner of the post(s) being returned. 

If the requesting user is the Author of the post, they can see it. If they're not the author, then we need to check their capabilities. If they don't have proper capabilities to read others posts, they shouldn't see the post. 

Long story short, querying a list of posts requires more than just returning raw data from the database, it requires added context of the requesting user, of the application layer (roles and capabilities of each post_type, etc) and ownership of each node being returned.

In the past, one check WPGraphQL performed to ensure User nodes were public, was to [run the](https://github.com/wp-graphql/wp-graphql/compare/master...release/v1.6.0-beta?expand=1#diff-2886a9a9577b1d57d2364a968402b5d7416d8091cbadab8f43c15b6400b671f4L157) `count_user_posts()` function for each author loaded for the request. 

This was expensive and required many SQL queries to fulfill. 

In fact, for a "simple" query such as: 

```graphql
{
  posts( first: 100 ) {
    nodes {
      id
      title
      content
    }
  }
}
```

This would generate 113 MySQL queries to fulfil: 

Screen Shot 2021-07-26 at 10 32 02 PM

But with the changes here, we're down to 13 queries, reducing the total queries by 99 for this request, and cutting the total SQL query time nearly in half!

Screen Shot 2021-07-26 at 10 41 11 PM

Results will vary from site-to-site based on data sets, etc, but this is a significant improvement and should improve performance for most users of WPGraphQL.


chriszarate (approved):
Gave this another test on a site with a large posts table and post_authors table. I still think there are additional improvements we could make to the UserLoader, but this is a huge improvement and good enough to ship—especially since this issue hits so many adopters. I can keep poking around additional improvements.

PR #2051, opened 2021-08-03
Bug/#2047 - Allow Types to share the name of a global PHP function

### LINKED ISSUE #2047: Fatal error in 1.6.1
After updating from wp-graphql 1.5.8. to 1.6.1 my nextjs's queries result in a fatal error with the following stack trace. Downgrading to 1.5.8 fixes the issue.

**Update:** the same error occurs when querying the schema in the wp-graphql IDE.

wp-graphql 1.6.1
wp-graphql-jwt-authentication 0.4.0
wp-graphql-acf 0.4.0

`````
PHP Fatal error:  Uncaught ArgumentCountError: header() expects at least 1 parameter, 0 given in /***/wp-content/plugins/wp-graphql/vendor/webonyx/graphql-php/src/Type/Schema.php:406
Stack trace:
#0 /***/wp-content/plugins/wp-graphql/vendor/webonyx/graphql-php/src/Type/Schema.php(406): header()
#1 /***/wp-content/plugins/wp-graphql/vendor/webonyx/graphql-php/src/Type/Definition/FieldDefinition.php(208): GraphQL\Type\Schema::resolveType('Header')
#2 /***/wp-content/plugins/wp-graphql/vendor/webonyx/graphql-php/src/Utils/TypeInfo.php(287): GraphQL\Type\Definition\FieldDefinition->getType()
#3 /***/wp-content/plugins/wp-graphql/vendor/webonyx/graphql-php/src/Language/Visitor.php(466): GraphQL\Utils\TypeInfo->enter(Object(GraphQL\Language\AST\FieldNode))
#4 /***/wp-content/plugins/wp in /***/wp-content/plugins/wp-graphql/vendor/webonyx/graphql-php/src/Type/Schema.php on line 406
`````

### PR DESCRIPTION
**Pull request by jasonbahl**
--------------------------------------------------

This fixes a bug where GraphQL Types that shared the name of a global PHP function would try and lazy load by calling that PHP function. 

For example, a GraphQL Type named `Header` would try and call `header()` to lazy load, which is not what should happen.

This PR does the following:

- Update the TypeRegistry to prefix types in the registry with `graphql_type_`
- Update the logic around `is_callable()` in the setup_type_modifiers function
- Adds a test to register a type that shares the name of a global PHP function (header`

Closes #2047 

Any other comments?
-------------------

With the following snippet in place:

```php
register_graphql_field( 'RootQuery', 'header', [
	'type' => 'Header',
	'description' => __( 'This field is named after a PHP function to test that it does not cause conflicts', 'wp-graphql' ),
	'resolve' => function() {
		return 'it works!';
	}
]);

register_graphql_object_type( 'Header', [
	'fields' => [
		'test' => [
			'type' => 'String'
		],
	],
]);
```

### Before

The Schema won't load at all 😱 

And trying to query the `header` field of the `Header` type returns a fatal error. 

Screen Shot 2021-08-03 at 9 14 05 AM

### After
The `Header` Type can be seen in the Schema and the `header` field can be queried successfully:

Screen Shot 2021-08-03 at 9 12 30 AM

### DIFF
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -175,11 +175,6 @@ class TypeRegistry {
 	 * @return string
 	 */
 	protected function format_key( string $key ) {
-
-		if ( false === strpos( $key, 'graphql_type_', 0 ) ) {
-			return strtolower( 'graphql_type_' . $key );
-		}
-
 		return strtolower( $key );
 	}
 
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -995,7 +995,7 @@ class TypeRegistry {
 				$field = $this->prepare_field( $field_name, $config, $type_name );
 
 				if ( ! empty( $field ) ) {
-					$fields[ $field_name ] = $this->prepare_field( $field_name, $config, $type_name );
+					$fields[ $field_name ] = $field;
 				}
 
 				return $fields;
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -858,13 +858,6 @@ class TypeRegistry {
 	 */
 	protected function prepare_field( $field_name, $field_config, $type_name ) {
 
-		/**
-		 * If the Field is a Type definition and not a config
-		 */
-		if ( ! is_array( $field_config ) ) {
-			return $field_config;
-		}
-
 		if ( ! isset( $field_config['name'] ) ) {
 			$field_config['name'] = lcfirst( $field_name );
 		}
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -906,6 +906,9 @@ class TypeRegistry {
 	}
 
 	/**
+	 * Processes type modifiers (e.g., "non-null"). Loads types immediately, so do
+	 * not call before types are ready to be loaded.
+	 *
 	 * @param mixed|string|array $type The type definition
 	 *
 	 * @return mixed
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -885,45 +885,26 @@ class TypeRegistry {
 			return null;
 		}
 
+		/**
+		 * If the type is a string, create a callable wrapper to get the type from
+		 * type registry. This preserves lazy-loading and prevents a bug where a type
+		 * has the same name as a function in the global scope (e.g., `header()`) and
+		 * is called since it passes `is_callable`.
+		 */
 		if ( is_string( $field_config['type'] ) ) {
-			$field_config['type'] = $this->format_key( $field_config['type'] );
+			$field_config['type'] = function () use ( $field_config ) {
+				return $this->get_type( $field_config['type'] );
+			};
 		}
 
 		/**
-		 * If type is not already a callable, create one to preserve lazy-loading.
+		 * If the type is an array, it contains type modifiers (e.g., "non_null").
+		 * Create a callable wrapper to preserve lazy-loading.
 		 */
-		if ( ! is_callable( $field_config['type'] ) ) {
-
-			if ( is_string( $field_config['type'] ) ) {
-				if ( false !== strpos( $field_config['type'], 'graphql_type_' ) ) {
-					if ( $this->has_type( $field_config['type'] ) ) {
-
-						$field_config['type'] = function () use ( $field_config ) {
-							if ( is_string( $field_config['type'] ) ) {
-								return $this->get_type( $field_config['type'] );
-							}
-
-							return $this->setup_type_modifiers( $field_config['type'] );
-						};
-					} else {
-						$field_config['type'] = function () use ( $field_config ) {
-							if ( is_string( $field_config['type'] ) ) {
-								return $this->get_type( $field_config['type'] );
-							}
-
-							return $this->setup_type_modifiers( $field_config['type'] );
-						};
-					}
-				}
-			} else {
-				$field_config['type'] = function () use ( $field_config ) {
-					if ( is_string( $field_config['type'] ) ) {
-						return $this->get_type( $field_config['type'] );
-					}
-
-					return $this->setup_type_modifiers( $field_config['type'] );
-				};
-			}
+		if ( is_array( $field_config['type'] ) ) {
+			$field_config['type'] = function () use ( $field_config ) {
+				return $this->setup_type_modifiers( $field_config['type'] );
+			};
 		}
 
 		if ( ! empty( $field_config['args'] ) && is_array( $field_config['args'] ) ) {
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -838,8 +838,6 @@ class TypeRegistry {
 					$prepared_field = $this->prepare_field( $field_name, $field_config, $type_name );
 					if ( ! empty( $prepared_field ) ) {
 						$prepared_fields[ $this->format_key( $field_name ) ] = $prepared_field;
-					} else {
-						unset( $prepared_fields[ $this->format_key( $field_name ) ] );
 					}
 				}
 			}
--- a/tests/wpunit/AssertValidSchemaTest.php
+++ b/tests/wpunit/AssertValidSchemaTest.php
@@ -426,4 +426,56 @@ class AssertValidSchemaTest extends \Codeception\TestCase\WPTestCase {
 		$this->assertArrayNotHasKey( 'errors', $actual );
 
 	}
+
+	/**
+	 * Many moons ago, fields could pass a Type definition instead of an array.
+	 *
+	 * This test ensures older plugins that extend WPGraphQL in this way still work
+	 *
+	 * @throws Exception
+	 */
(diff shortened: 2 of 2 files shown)


jasonbahl:
Note: This doesn't just affect native global functions. 

Any PHP function that can be called, even added by WordPress or WP plugins/themes that share the name of a GraphQL Type will experience this bug. 

For example, if we add a Type named `WP_Send_Json`, this will also break things as `wp_send_json()` is a callable. 

```php
register_graphql_field( 'RootQuery', 'header', [
	'type' => 'WP_Send_Json',
	'description' => __( 'This field is named after a PHP function to test that it does not cause conflicts', 'wp-graphql' ),
	'resolve' => function() {
		return 'it works!';
	}
]);

register_graphql_object_type( 'WP_Send_Json', [
	'fields' => [
		'test' => [
			'type' => 'String'
		],
	],
]);
```

src/Registry/TypeRegistry.php
```diff
@@ -174,7 +174,12 @@ public function __construct() {
...
 	 * @return string
 	 */
-	protected function format_key( $key ) {
+	protected function format_key( string $key ) {
+
+		if ( false === strpos( $key, 'graphql_type_', 0 ) ) {
+			return strtolower( 'graphql_type_' . $key );
+		}
+
 		return strtolower( $key );
```
chriszarate, line 183: We should just always prepend. Otherwise we would create a collision between two user-defined types named `header` and `graphql_type_header`. Maybe unlikely, but not much to gain with this logic on an internal-facing key.
```suggestion
		return strtolower( 'graphql_type_' . $key );
```

src/Registry/TypeRegistry.php
```diff
@@ -805,7 +807,7 @@ public function get_type( string $type_name ) {
 	 * @return bool
 	 */
 	public function has_type( string $type_name ) {
-		return isset( $this->type_loaders[ $type_name ] );
+		return isset( $this->type_loaders[ $this->format_key( $type_name ) ] );
```
chriszarate, line 810: oh oops!

tests/wpunit/AssertValidSchemaTest.php
```diff
@@ -284,7 +284,7 @@ public function testSchemaSupportsLazyLoadingTypes() {
 	}
 
 	public function testSchemaCallsLazyLoadingTypesWhenNeeded() {
-		$filter_calls = [];
+		$filter_calls = [ 'graphql_get_type' ];
```
chriszarate, line 287: What's the reason for this change?

tests/wpunit/AssertValidSchemaTest.php
```diff
@@ -350,6 +350,78 @@ public function testSchemaCallsLazyLoadingTypesWhenNeeded() {
 		$this->assertSame( 'bar', $actual['data']['example']['foo'] );
 
 		$expected_filter_calls = [ 'graphql_wp_object_type_config', 'graphql_object_fields', 'graphql_get_type' ];
-		$this->assertSame( $expected_filter_calls, $filter_calls );
+		foreach ( $expected_filter_calls as $expected_filter_call ) {
+			$this->assertTrue( in_array( $expected_filter_call, $filter_calls, true ) );
+		}
+
+		$this->assertSame( count( $expected_filter_calls ), count( $filter_calls ) );
+	}
```
chriszarate, line 358: How is this different than the `assertSame`?

tests/wpunit/AssertValidSchemaTest.php
```diff
@@ -350,6 +350,78 @@ public function testSchemaCallsLazyLoadingTypesWhenNeeded() {
...
+				],
+			],
+		]);
+
+		$query = '
+		{
+		  header {
+			test  
+		  }
+	    }
```
chriszarate, line 383:
```suggestion
		{
			header {
				test  
			}
		}
```

tests/wpunit/AssertValidSchemaTest.php
```diff
@@ -350,6 +350,78 @@ public function testSchemaCallsLazyLoadingTypesWhenNeeded() {
...
+				],
+			],
+		]);
+
+		$query = '
+		{
+		  wpSendJson {
+			test  
+		  }
+	    }
```
chriszarate, line 417:
```suggestion
		{
			wpSendJson {
				test  
			}
		}
```

tests/wpunit/AssertValidSchemaTest.php
jasonbahl: Tests for me were returning them in different order. Always the same 3, but sometimes in different order. 

Perhaps that's something we should investigate more deeply instead of change this test

tests/wpunit/AssertValidSchemaTest.php
jasonbahl: see below comment: 

>>> Tests for me were returning them in different order. Always the same 3, but sometimes in different order.
>>> Perhaps that's something we should investigate more deeply instead of change this test

src/Registry/TypeRegistry.php
jasonbahl: 😬

src/Registry/TypeRegistry.php
jasonbahl: I think the problem with _always_ prepending is that the format_key gets used on inserting types into the registry _and_ getting them out. 

If we always prepend and try: `get_type( 'graphql_type_post' )` then it will actually try to do `get_type( 'graphql_type_graphql_type_post' )` and find nothing. 

I guess I was trying to guard agains explicitly asking for `graphql_type_$name`, but I think you're right that it should be safe to always prepend 🤔

src/Registry/TypeRegistry.php
```diff
@@ -883,17 +885,45 @@ protected function prepare_field( $field_name, $field_config, $type_name ) {
...
+					if ( $this->has_type( $field_config['type'] ) ) {
+
+						$field_config['type'] = function () use ( $field_config ) {
+							if ( is_string( $field_config['type'] ) ) {
+								return $this->get_type( $field_config['type'] );
+							}
+
+							return $this->setup_type_modifiers( $field_config['type'] );
+						};
+					} else {
```
chriszarate, line 908: Repeating a ton of logic here. I think we could move it into `get_type` and dry this up quite a bit. Happy to take a crack at it if you want.

src/Registry/TypeRegistry.php
chriszarate: I think all this key logic needs to be localized in `has_type` and `get_type` (as mentioned below) and then it should work.

tests/wpunit/AssertValidSchemaTest.php
chriszarate: I see that as a reason for the change below, but not for inserting this into the initial array. This change means that it will always look like `'graphql_get_type'` was called, even if it wasn't.

jasonbahl: oh shoot. Yes, I see 🤦🏻

src/Registry/TypeRegistry.php
jasonbahl: Have at it. I was getting to a point where my brain was fried and I was almost forgetting what I was trying to solve. lol

src/Registry/TypeRegistry.php
```diff
@@ -841,8 +838,6 @@ public function prepare_fields( $fields, $type_name ) {
 					$prepared_field = $this->prepare_field( $field_name, $field_config, $type_name );
 					if ( ! empty( $prepared_field ) ) {
 						$prepared_fields[ $this->format_key( $field_name ) ] = $prepared_field;
-					} else {
-						unset( $prepared_fields[ $this->format_key( $field_name ) ] );
```
chriszarate, line 845: redundant

src/Registry/TypeRegistry.php
```diff
@@ -863,13 +858,6 @@ public function prepare_fields( $fields, $type_name ) {
 	 */
 	protected function prepare_field( $field_name, $field_config, $type_name ) {
 
-		/**
-		 * If the Field is a Type definition and not a config
-		 */
-		if ( ! is_array( $field_config ) ) {
```
chriszarate, line 869: redundant

src/Registry/TypeRegistry.php
```diff
@@ -998,7 +998,7 @@ function ( $fields ) use ( $type_name, $field_name, $config ) {
 				$field = $this->prepare_field( $field_name, $config, $type_name );
 
 				if ( ! empty( $field ) ) {
-					$fields[ $field_name ] = $this->prepare_field( $field_name, $config, $type_name );
```
chriszarate, line 1001: this is called twice unnecessarily, see line 998

tests/wpunit/AssertValidSchemaTest.php
```diff
@@ -349,7 +349,81 @@ public function testSchemaCallsLazyLoadingTypesWhenNeeded() {
 		$this->assertArrayNotHasKey( 'errors', $actual );
 		$this->assertSame( 'bar', $actual['data']['example']['foo'] );
 
-		$expected_filter_calls = [ 'graphql_wp_object_type_config', 'graphql_object_fields', 'graphql_get_type' ];
+		sort( $filter_calls );
```
chriszarate, line 352: sorting seemed a little cleaner than the loop

Issue #1756, opened 2021-03-03
User Error: expected iterable, but did not find one for field Post.pinged


tsdexter:
When loading the graphql endpoint for my site there is an error on the `Post.pinged` field.

Expected behaviour: loading the endpoint has no errors

Actual behaviour: 100 posts are queried and each returns an error: `User Error: expected iterable, but did not find one for field Post.pinged`  causing `gatsby develop` to fail with: 
```
ERROR 

gatsby-source-wordpress  WPGraphQL returned a debug message on startup:

User Error: expected iterable, but did not find one for field Post.pinged.


not finished createSchemaCustomization - 0.707s
not finished  gatsby-source-wordpress  ensuring plugin requirements are met - 0.699s
```

I am using WP Gatsby Version 1.0.1 and WP GraphQL Version 1.2.5

tsdexter:
I've temporarily fixed this with: 

```
function graphql_interface_fields_callback( $fields, $typename ) {
    // Maybe modify $example in some way.
    if ($typename == 'NodeWithTrackbacks') {
      return [
					'toPing'     => [
						'type'        => 'String',
						'description' => __( 'URLs queued to be pinged.', 'wp-graphql' ),
					],
					'pinged'     => [
						'type'        => 'String',
						'description' => __( 'URLs that have been pinged.', 'wp-graphql' ),
					],
					'pingStatus' => [
						'type'        => 'String',
						'description' => __( 'Whether the pings are open or closed for this particular post.', 'wp-graphql' ),
					],
				];
    }
}
add_filter( 'graphql_interface_fields', 'graphql_interface_fields_callback', 10, 2 );
```

But I'm unsure of the underlying issue and how changing that `type` might affect other things

tsdexter:
It looks like this is related to using https://github.com/ProjectNami/projectnami (MS SQL database) which throws errors from some of the (My)SQL in the code. We're moving away from Nami so I won't be needing a fix. 

Should I close the issue or leave it in case you plan to support Nami in the future?

jasonbahl:
@tsdexter I'm not familiar with Project Nami. If it's behaving differently than standard MySQL databases, that seems like possibly a bug with that project and not with WPGraphQL? 

Based on your comment about moving away, I'll close this issue but keep an eye out for others using Project Nami to see if it's something we should more seriously consider supporting.

tsdexter:
[...]

PR #2093, opened 2021-09-11
Fix: Post.pinged not returning iterable.

justlevine:
Post.pinged was returning a `string` instead of `array|null`, throwing`User Error: expected iterable, but did not find one for field Post.pinged`.

Issue was caused because `implode()` was being returned instead of used solely to check if `get_pung()` was empty.

`get_pung()` can return `[ $key => "" ]`, which fails the `empty()` check. Using `implode()` is the correct way to address this, but it needs to be in the `empty()` check, not returned if `! empty()`.

#1756 

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl), Windows (Local by Flywheel).

**WordPress Version:** 5.8.1


dunxtand:
Hi there, I don't want to rush your development process, but do you have any estimate of when this specific fix might be released in an update?

jasonbahl:
@justlevine ok, looks like the latest failure was just a hiccup 

All good now.

PR #2022, opened 2021-07-20
Feature/#1738 introduce common interfaces

jasonbahl:
What does this implement/fix? Explain your changes.
---------------------------------------------------

This pull request introduces a number of new Interfaces and applies them throughout the Schema. 

Interfaces introduced in this PR are: 

- `CommentConnection`
- `CommenterConnection`
- `Connection`
- `ConnectionEdge`
- `ContentNodeConnection`
- `ContentTypeConnection`
- `HierarchicalNode`
- `MenuConnection`
- `MenuItemConnection`
- `MenuItemLinkableConnection`
- `Previewable`
- `TaxonomyConnection`
- `TermNodeConnection`
- `UserConnection`

Does this close any currently open issues?
------------------------------------------
closes #1738 

Any other comments?
-------------------
These connections add a lot of value to Schema organization, code reusability, and introspection. 

One can now use GraphiQL to find all instances of Connections in the Schema like so: 

connection-interface-implementations

One could also look for all Connections that implement a more specific Connection Interface, such as `ContentNodeConnection`. This interface is applied to any node that has a connection to any type of Content Node. So a User to Post connection is a `ContentNodeConnection` and a User to Page connection is also a ContentNodeConnection. 

content-node-connection-implementations

These shared connection interfaces allow client developers to have a better understanding of shared properties of connections and can allow for reduced code when querying similar data from the Graph. 

On the server side, developers can more easily add features to specific connections or general connection interfaces.

PR #2099, opened 2021-09-17
Move `taxonomy` field to the `TermNode` interface

jeanfredrik:
Right now the `taxonomy` field used to get a term’s taxonomy is registered on each individual term type. This PR changes this so that it’s registered on the common `TermNode` interface instead. That means you can now make this query which was not possible before:

```gql
query MyQuery {
  terms {
    nodes {
      taxonomy {
        node {
          name
        }
      }
    }
  }
}
```

Where has this been tested?
---------------------------
**Operating System:** MacOS

**WordPress Version:** 5.5.6

### DIFF
--- a/src/Connection/Taxonomies.php
+++ b/src/Connection/Taxonomies.php
@@ -29,27 +29,6 @@ class Taxonomies {
 
 		$taxonomies = get_taxonomies( [ 'show_in_graphql' => true ], 'objects' );
 
-		if ( is_array( $taxonomies ) && ! empty( $taxonomies ) ) {
-			foreach ( $taxonomies as $taxonomy ) {
-				register_graphql_connection(
-					[
-						'fromType'      => $taxonomy->graphql_single_name,
-						'toType'        => 'Taxonomy',
-						'fromFieldName' => 'taxonomy',
-						'oneToOne'      => true,
-						'resolve'       => function ( Term $source, $args, $context, $info ) {
-							if ( empty( $source->taxonomyName ) ) {
-								return null;
-							}
-							$resolver = new TaxonomyConnectionResolver( $source, $args, $context, $info );
-							$resolver->set_query_arg( 'name', $source->taxonomyName );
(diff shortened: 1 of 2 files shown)


jasonbahl:
@jeanfredrik thanks for the pull request! 

I think there's a lot of value in the simplicity offered here. 

However, it does come at a cost of losing some contextual relational "edge" data. 

For example, today I could do the following: 

```graphql
query MyQuery {
  terms {
    nodes {
      __typename
      ... on Tag {
        taxonomy {
          someTagToTaxonomyField # some edge field only in context of category->taxonomy connections
          node {
            name
          }
        }
      }
      ...on Category {
        taxonomy {
          someCategoryToTaxonomyField # some edge field only in context of category->taxonomy connections
          node {
            name
          }
        }
      }
    }
  }
}
```

With the change you're proposing, the ability to query this "edge" data would go away, because the connections would be vague `TermNode -> Taxonomy` and not specific `Tag->Taxonomy` or `Category->Taxonomy`. 

I believe this might be solve-able by introducing a generic `TaxonomyConnection` Interface, as seen in the PR here: https://github.com/wp-graphql/wp-graphql/pull/2022

Then, we could have a generic: TermNode->Taxonomy connection, and have specific `Tag->Taxonomy` and `Category->Taxonomy` connections that implement the `TermNode->Taxonomy` Interface. 

This would allow for the more simple queries that you want, while still allowing for some more specific relational data to be exposed. 

I'm going to leave this PR open for now and circle back to #2022 and see if we can get both behaviors working well together.

jeanfredrik:
Ok, that makes sense. What I'm primarily after is a way to associate a generic TermNode with its Taxonomy. I don’t see any way to do that right now. I've created [another PR](https://github.com/wp-graphql/wp-graphql/pull/2143) that adds a `taxonomyName` field to `TermNode` which would solve my issue. Maybe that’s a better solution for now?

PR #2132, opened 2021-10-28
Fixes slug queries on custom post types.

### LINKED ISSUE #2082: Custom Post Types query using slug as unique identifier returns no data in WPGraphQL
# Context
I've created a Custom Post Type in WordPress named "projects" and now I'm trying to query a single project using the slug as an identifier. In other words, I want the idType to be the SLUG, not ID. I tried the following query but it does not work.

image

However, this query works (it's a common query that is using the ID) :

image

The query above is not what i'm looking for, but at least i know that my CPT is registered correctly and that my data can be fetched somehow.

# Why i think this is a problem?
Because I used the same type of query for querying WordPress posts (not custom posts types, just posts) and it works fine. I'm being returned the correct data.

image

Any help is appreciated ;) I don't know if i'm missing something with my query, or if custom post types don't work the same way that posts do. 

```
WPGraphQL version: 1.3.10
Other WordPress plugins installed: Advanced Custom Fields, Custom Post Type UI
```

ChrisWiegman:
Fixes slug not available in CPTs per https://github.com/wpengine/atlas-content-modeler/issues/290 and similar issues.

Fixes https://github.com/wpengine/atlas-content-modeler/issues/290
May fix #2082 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** MacOS

**WordPress Version:* 5.8.1

### DIFF
--- a/tests/wpunit/ContentNodeInterfaceTest.php
+++ b/tests/wpunit/ContentNodeInterfaceTest.php
@@ -356,20 +356,54 @@ class ContentNodeInterfaceTest extends \Codeception\TestCase\WPTestCase {
 	 */
 	public function testContentNodeFieldBySlug() {
 
-		$args = array(
-			'public'              => false,
-			'show_in_graphql'     => true,
-			'graphql_single_name' => 'book',
-			'graphql_plural_name' => 'books',
-			'label'               => 'Books',
+		$post_types = array(
+			'book' => array(
+				'public'              => false,
+				'show_in_graphql'     => true,
+				'graphql_single_name' => 'book',
+				'graphql_plural_name' => 'books',
+				'label'               => 'Books',
+			),
+			'test' => array(
+				'public'              => false,
+				'publicly_queryable'  => true,
+				'show_in_graphql'     => true,
+				'graphql_single_name' => 'test',
+				'graphql_plural_name' => 'tests',
+				'label'               => 'Tests',
+			),
+			'cat' => array(
+				'public'              => false,
+				'publicly_queryable'  => false,
+				'show_in_graphql'     => true,
+				'graphql_single_name' => 'cat',
+				'graphql_plural_name' => 'cats',
+				'label'               => 'Cats',
+			),
 		);
 
-		register_post_type( 'book', $args );
+		foreach ( $post_types as $post_type => $args ) {
+			register_post_type( $post_type, $args );
+		}
 
-		$post_id = $this->factory()->post->create( [
+		$post_id_book = $this->factory()->post->create( [
 			'post_type'   => 'book',
 			'post_status' => 'publish',
 			'post_title'  => 'Test Post',
+			'post_author' => $this->admin
+		] );
+
+		$post_id_test = $this->factory()->post->create( [
+			'post_type'   => 'test',
+			'post_status' => 'publish',
+			'post_title'  => 'Test Test',
+			'post_author' => $this->admin
+		] );
+
+		$post_id_cat = $this->factory()->post->create( [
+			'post_type'   => 'cat',
+			'post_status' => 'publish',
+			'post_title'  => 'Test Cat',
 			'post_author' => $this->admin
 		] );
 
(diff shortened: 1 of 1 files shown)


jasonbahl:
@ChrisWiegman @victorcngo one thing to note, there is another bug that when fixed, might make this behavior come back. 

When registering a post type, we can register a post_type with: 

- `public`: true/false
- `publicly_queryable`: true/false

If a post_type is registered with both values as `false`, then content within that post type will be treated as non-public, so a post_type that is registered without declaring `public=>true` would be treated as non-public content, and querying an entity by slug would return null again. 

This would be intentional, though, to respect Access Control privileges of WordPress.

See this issue: https://github.com/wp-graphql/wp-graphql/issues/2134

When registering a post_type, I would recommend explicitly declaring whether you want the post_type to be fully public, meaning each entry has its own URL, such as `site.com/my-custom-type/title-of-post`, or if you want the post_type to be non-public, but publicly queryable. 

The latter would be something like a "Features" post_type that you want to publicly query for display within another page, but you don't want each entry to be accessed via a public URL individually. 

So, if a post_type were registered with the code in issue #2082, the query by slug will return `null` for public requests as the entries would be considered non public. 

To fix this, the registration could be changed: 

**FROM:**

```php
/* Register CPT for projects */
   register_post_type( 'projects', [
      'show_ui' => true,
      'labels'  => [
        //@see https://developer.wordpress.org/themes/functionality/internationalization/
        'menu_name' => __( 'Projects', 'your-textdomain' ),
      ],
      'show_in_graphql' => true,
      'hierarchical' => false,
      'graphql_single_name' => 'project',
      'graphql_plural_name' => 'projects',
      'capability_type' => 'post',
      'supports' => array('title', 'author', 'editor', 'thumbnail','excerpt','revisions','page-attributes'),
      'cptp_permalink_structure' => '%post_name%'
   ] );
} );
```

**TO:**

```php
/* Register CPT for projects */
   register_post_type( 'projects', [
      'show_ui' => true,
      'labels'  => [
        //@see https://developer.wordpress.org/themes/functionality/internationalization/
        'menu_name' => __( 'Projects', 'your-textdomain' ),
      ],
      'show_in_graphql' => true,
      'hierarchical' => false,
      'graphql_single_name' => 'project',
      'graphql_plural_name' => 'projects',
      'capability_type' => 'post',
      'supports' => array('title', 'author', 'editor', 'thumbnail','excerpt','revisions','page-attributes'),
      'cptp_permalink_structure' => '%post_name%'<
      'publicly_queryable' => true # <-- this declares that the entries in this type should be queryable by public users
   ] );
} );
```

jasonbahl:
See the corresponding PR for the issue I mentioned, which will be adjusting permissions for post types to better respect the `public` / `public_queryable` arguments of post_type registration: https://github.com/wp-graphql/wp-graphql/pull/2135

tests/wpunit/ContentNodeInterfaceTest.php
```diff
@@ -351,5 +351,43 @@ public function testContentNodeFieldByQueryArgUri() {
...
 	}
 
+	/**
+	 * @throws Exception
+	 */
+	public function testContentNodeFieldBySlug() {
+
+		$args = array(
+			'public'              => false,
+			'show_in_graphql'     => true,
```
jasonbahl, line 361: @ChrisWiegman let's maybe add 2 variations of this test. 

one where the post type is: 

`'public' => false`

And another where the post_type is: 

```php
...
'public' => false,
'publicly_queryable' => true,
```

If the post_type is set to false, when #2135 is merged, the "correct" behavior will be to not return a post of a `public=>false` post_type to a public user, but if `public=>false, publicly_queryable=>true` then the posts of that post_type can be returned to public users in GraphQL.

ChrisWiegman: I've added cases with the requested configs. For now they don't meet your description above (false to both still returns the post) but I think that's intended until another PR is merged?

jasonbahl: Correct. I'll get both PRs merged and ensure tests are accurate for the new behavior.

PR #2135, opened 2021-10-29
Respect 'publicly_queryable' post type setting.

### LINKED ISSUE #2134: Respect `publicly_queryable` setting for post types
Currently, if a post type in WordPress has `publicly_queryable` set to `false`, its posts are still accessible to unauthenticated users. WPGraphQL should respect the `publicly_queryable` setting.

It's easy to conflate `publicly_queryable` with `public` and/or `show_in_graphql` settings. Here is the breakdown that @jasonbahl and I came up with for remembering what effect each setting has:

- `show_in_graphql` —> add to GraphQL schema
- `public` —> Give this a URI
- `publicly_queryable` —> accessible to all unauthenticated users

I'll submit a PR so that a post is considered private if `publicly_queryable` is set to `true` (whether explicitly, or implicitly by inheriting the `public => true` value for the post type).

## End Result

So the end result of this change should be this behavior:

- If `publicly_queryable` is `true` for a post type, its posts should be accessible to unauthenticated users.
- If `publicly_queryable` is `false` for a post type, its posts should be accessible only to authenticated users who have the required capability.

kellenmace:
Fixes #2134.

Consider a post as private if 'publicly_queryable' is set to false so that only authenticated users with the required capability can access it.

Yes, #2134.

Any other comments?
-------------------
Please test this to double-check that it results in the intended behavior.

I also updated the paragraph at the top of the https://www.wpgraphql.com/docs/custom-post-types/#public-vs-private-data section of the docs to reflect this change (wp-admin edit page is here: https://content.wpgraphql.com/wp-admin/post.php?post=5116&action=edit).

### DIFF
--- a/src/Model/Post.php
+++ b/src/Model/Post.php
@@ -328,7 +328,7 @@ class Post extends Model {
 		/**
 		 * Published content is public, not private
 		 */
-		if ( 'publish' === $this->data->post_status && $this->post_type_object->publicly_queryable ) {
+		if ( 'publish' === $this->data->post_status && $this->post_type_object && $this->post_type_object->publicly_queryable ) {
 			return false;
 		}
 
--- a/src/Model/Post.php
+++ b/src/Model/Post.php
@@ -328,7 +328,7 @@ class Post extends Model {
 		/**
 		 * Published content is public, not private
 		 */
-		if ( 'publish' === $this->data->post_status ) {
+		if ( 'publish' === $this->data->post_status && $this->post_type_object->publicly_queryable ) {
 			return false;
 		}
 
--- a/src/Model/Post.php
+++ b/src/Model/Post.php
@@ -328,7 +328,7 @@ class Post extends Model {
 		/**
 		 * Published content is public, not private
 		 */
-		if ( 'publish' === $this->data->post_status && $this->post_type_object && $this->post_type_object->publicly_queryable ) {
+		if ( 'publish' === $this->data->post_status && $this->post_type_object && ( true === $this->post_type_object->public || true === $this->post_type_object->publicly_queryable ) ) {
 			return false;
 		}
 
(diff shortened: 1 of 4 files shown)


kellenmace:
This line of code could have alternatively been written as:
`if ( 'publish' === $this->data->post_status && ! empty( $this->post_type_object->publicly_queryable ) ) {`
to account for `$this->post_type_object` possibly being set to `null`.

I think doing
`if ( 'publish' === $this->data->post_status && $this->post_type_object && $this->post_type_object->publicly_queryable ) {`
is a bit more readable, though, so I went with that. Matter of preference, though!

src/Model/Post.php
```diff
@@ -328,7 +328,7 @@ public function is_private() {
 		/**
 		 * Published content is public, not private
 		 */
-		if ( 'publish' === $this->data->post_status ) {
+		if ( 'publish' === $this->data->post_status && $this->post_type_object && $this->post_type_object->publicly_queryable ) {
```
jasonbahl, line 331: @kellenmace this is interesting. 

I think if `public` is true, we need to respect it first. 

Because post types _can_ be `public => true, publicly_queryable => false`, such as the `page` post_type: https://github.com/WordPress/WordPress/blob/8447683474ebdee773fc4fca7b3e44dea45060f4/wp-includes/post.php#L45-L68, which appears to be breaking a lot of tests right now 🤔

jasonbahl: @kellenmace I pushed up a change so it's now:

```php
if ( 'publish' === $this->data->post_status && $this->post_type_object && ( true === $this->post_type_object->public || true === $this->post_type_object->publicly_queryable ) ) {
  return false;
}
```

This means that if _either_ `public` _or_ `publicly_queryable` are true, then WPGraphQL should treat the entries as public.

This would allow for "utility" post_types, such as a Slideshow post_type or an FAQ post_type, etc to be registered as `public=>false, publicly_queryable=>true` and the entries would not get a public url, _but_ would be queryable in WPGraphQL, so they could still be used to compose public facing pages, even without having their own public urls.

Issue #2140, opened 2021-11-04
`WPConnectionType` $config cannot be filtered.


justlevine:
# Overview
The [`WPConnectionType` class](https://github.com/wp-graphql/wp-graphql/blob/a40f46b935e0e4d2cd46b92603c900cbb017a29d/src/Type/WPConnectionType.php#L15) does not make use of any wordpress filters. This means that there's no way to overwrite the connection `$config`, which is a major problem when trying to extend the connections generated for by WPGraphQL for custom post types, or changing the behavior of any registered by another plugin. 

Considering all the other `WP{type}Type` classes have filters in their constructor, I think this should be considered a bug, and not a Feature Request.

## Context
I've spent the last week trying to extend a Custom Post Type, to use its own model, dataloader, resolvers, etc. The whole situation is a PITA for what should be a fairly straightforward task, and as more advanced plugins add WPGraphQL support is essential. 
Without a major refactor `PostObjects` or `DataSource` to allow this (going to open a separate issue soon), you're required to `add_filter()` half-a-dozen values. However, without any way to filter the connection `$config`, the alternative is to stop WPGraphQL from registering those connections entirely, and then manually reimplementing all the connection args etc. (See the hundreds of lines of code in WPGraphQL for Woocommerce, that really should be unnecessary and essentially duplicate core WPGraphQL behavior).

## Version Information
WPGraphQL v1.6.7

PR #2141, opened 2021-11-04
fix: Add missing `graphql_wp_connection_type_config` filter.

justlevine:
This PR adds the `graphql_wp_connection_type_config` filter to WPConnectionType, providing basic parity with the other `WP{$}Type` classes, and making it possible to filter existing connections and resolvers.

This is an additive change. There are no unit test regarding other filters, so I'm assuming none are needed here. Also it seems the docs for filters arent in the repo anymore (autogenerated?) but if they do need to go somewhere, let me know.

#2140 

Usage Example
-------------------
```php
add_filter( 'graphql_wp_connection_type_config',
  function( array $config ) : array {
    if ( 'RootQuery' !== $config['fromType'] || 'MyPostType' !== $config['toType'] ) {
      return $config;
    }

    $config['resolve'] = function( $source, array $args, AppContext $context, ResolveInfo $info ) {
      return ( new MyPostTypeConnectionResolver( $source, $args, $context, $info ) )->get_connection();
    }
  },
);

```

Where has this been tested?
---------------------------
**Operating System:** WSL2 (Ubuntu 20.04 )

**WordPress Version:** 5.8.1

### DIFF
--- a/src/Type/WPConnectionType.php
+++ b/src/Type/WPConnectionType.php
@@ -134,7 +134,7 @@ class WPConnectionType {
 		 * Filter the config of WPConnectionType
 		 *
 		 * @param array        $config         Array of configuration options passed to the WPConnectionType when instantiating a new type
-		 * @param WPConnectionType $wp_connection_type The instance of the WPObjectType class
+		 * @param WPConnectionType $wp_connection_type The instance of the WPConnectionType class
 		 */
 		$config = apply_filters( 'graphql_wp_connection_type_config', $config, $this );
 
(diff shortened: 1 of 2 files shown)


jasonbahl:
@justlevine 

> There are no unit test regarding other filters, so I'm assuming none are needed here.

Ideally it would be great to have tests for filters. I believe there's a long standing issue to add more test coverage for filters. 

The more tests we have for things like this that extension plugins rely on, the better. It will help prevent regressions if we have specific tests for APIs like this. 

If you have time and the desire to add a test for this filter, that would be much appreciated!

justlevine:
@jasonbahl  test  has been added 😎

PR #2143, opened 2021-11-05
feat: Add `taxonomyName` field

jeanfredrik:
This PR adds a `taxonomyName` field to the `TermNode` type, since there’s currently no other way to associate a generic term node to its taxonomy. This is also potentially more performant compared to `taxonomy->name` since the taxonomy doesn’t have to be loaded to get the taxonomy name.

### DIFF
--- a/src/Type/InterfaceType/TermNode.php
+++ b/src/Type/InterfaceType/TermNode.php
@@ -97,6 +97,10 @@ class TermNode {
 						'type'        => 'Int',
 						'description' => __( 'The taxonomy ID that the object is associated with', 'wp-graphql' ),
 					],
+					'taxonomyName' => [
+						'type'        => 'String',
+						'description' => __( 'The name of the taxonomy that the object is associated with', 'wp-graphql' ),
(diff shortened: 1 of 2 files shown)


jasonbahl (approved):

jasonbahl:
Thanks @jeanfredrik!

jasonbahl:
@jeanfredrik one thing I just thought of is that you can query `__typename` as well. 

For example: 

```graphql
{
  terms {
    nodes {
      __typename
      taxonomyName
      name
    }
  }
}
```

Screen Shot 2021-12-17 at 3 15 21 PM

The `__typename` field returns the GraphQL Type of the node, where the newly added `taxonomyName` returns the internal name of the type. 🤷🏻‍♂️

jeanfredrik:
Ah I didn’t think of that! I actually could’ve solved my specific problem with that. The actual name is more convenient though when you want to match them with their `Taxonomy` node because `Taxonomy.graphqlSingleName` has the wrong case.

PR #2182, opened 2021-12-16
fix: checking for route instead of domain+route

### LINKED ISSUE #2181: bug: Cannot make graphql request to domain that is not `site_url`
## Original issue: 

We had a customer reporting that they were getting an error when querying graphql that indicated html was being returned instead of JSON. They'd just added a new domain. 

## Diag:
We tracked it back to how WPGraphQL detects if the request being made is a graphql request in [`is_graphql_http_request`](https://github.com/wp-graphql/wp-graphql/blob/9f8ac9291491ca4812a03d177cf3a33668bf561a/src/Router.php#L139-L213). When the `site_url` domain doesn't match the domain used to make the request, this returns false even if the route contains `/graphql(/)`.

### Debugger showing issue: 
image

### URL config during above screenshot (wpenext.local) was used to access.
image

moonmeister:

Ignores the domain and just makes sure the configured graphql path is the root path.

fixes #2181
https://wpengine.atlassian.net/browse/NEB-844

Any other comments?
-------------------
Should write from tests, but I'll leave that to the experts. 

Right now `/graphql/this/that/the/other/`  passes. If we wanted we could test the path to make sure it's only `/graphql` or `/graphql/` but I'm not sure if it is an issue. 

Where has this been tested?
---------------------------
Tested it on my local PC as per below and also tested on WPEngine and validated it works. 

**Operating System:** …
macOS 11.16.1

**WordPress Version:** …
5.8.2

### DIFF
--- a/src/Router.php
+++ b/src/Router.php
@@ -169,14 +169,18 @@ class Router {
 			// Check the server to determine if the GraphQL endpoint is being requested
 			if ( isset( $_SERVER['HTTP_HOST'] ) && isset( $_SERVER['REQUEST_URI'] ) ) {
 
-				// $host = wp_unslash( $_SERVER['HTTP_HOST'] );
+				$host = wp_unslash( $_SERVER['HTTP_HOST'] );
 				$uri  = wp_unslash( $_SERVER['REQUEST_URI'] );
 
-				if ( ! is_string( $uri ) ) {
+				if ( ! is_string( $host ) ) {
+					return false;
+				} elseif ( ! is_string( $uri ) ) {
 					return false;
 				}
 
 				$len                     = strlen( self::$route );
+
+				// the off by one offset validates it's at /graphql or /graphql/ and not /somepath/graphql
 				$is_graphql_http_request = ( substr( $uri, 1, $len ) === self::$route );
 			}
 		}
--- a/src/Router.php
+++ b/src/Router.php
@@ -177,9 +177,12 @@ class Router {
 				} else if ( ! is_string( $uri ) ) {
 					return false;
 				}
- 
-				$graphql_url = wp_unslash( parse_url( site_url( self::$route ), PHP_URL_PATH ) );
-				$request_url = wp_unslash( parse_url( $uri, PHP_URL_PATH ) );
+
+				$parsed_site_url = parse_url( site_url( self::$route ), PHP_URL_PATH );
+				$graphql_url = ! empty( $parsed_site_url ) ? wp_unslash( $parsed_site_url ) : self::$route;
+
+				$parsed_request_url = parse_url( $uri, PHP_URL_PATH );
+				$request_url = ! empty( $parsed_request_url ) ? wp_unslash( $parsed_request_url ) : null;
 
 				// Determine if the route is indeed a graphql request
 				$is_graphql_http_request = str_replace( '/', '', $request_url ) === str_replace( '/', '', $graphql_url );
--- a/src/Router.php
+++ b/src/Router.php
@@ -178,6 +178,12 @@ class Router {
 					return false;
 				}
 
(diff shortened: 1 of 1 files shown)


src/Router.php
```diff
@@ -178,21 +178,10 @@ public static function is_graphql_http_request() {
...
-				// Strip protocol.
-				$replaced_path = preg_replace( '#^(http(s)?://)#', '', $full_path );
-				if ( ! empty( $replaced_path ) ) {
-					$full_path = $replaced_path;
-				}
-				$replaced_url = preg_replace( '#^(http(s)?://)#', '', $site_url );
-				if ( ! empty( $replaced_url ) ) {
-					$site_url = $replaced_url;
-				}
+				$len                     = strlen( self::$route );
```
moonmeister, line 181: Thinking this might be a better one. and eliminates routes like `/graphql/test` triggering as valide. not sure if there's a performance difference

```suggestion
								$is_graphql_http_request = boolval( preg_match('/^\/graphql\/?$/', $uri) );
```

jasonbahl:
@moonmeister it looks like this PR is failing tests right now. Looks like it's causing the functional tests to receive an HTML payload instead of a GraphQL payload, the opposite of the problem you were trying to fix?

jasonbahl:
I've got some ideas, though. I'll pull this down locally and see what I can work up.

jasonbahl:
@moonmeister I pushed some code and tests are now passing. 

However, I'm not sure how to verify this fixes the client's issue, as I still don't know the steps to reproduce. 

Can you share what steps I can take to get a WordPress site into the same state as the user that reported the issue?

I'm not quite sure I fully understand how they got in this position 🤔

jasonbahl:
@moonmeister are they using something like a domain mapping plugin? That might be the context I'm missing?

jasonbahl:
In any case, it would be nice to be able to reproduce this, ideally in the context of a test, to ensure it continues to work

moonmeister:
The site_url must match the domain being used to request data previously. So using local I changed the site_url in the db and accessed using the locals configured domain.

jasonbahl:
@moonmeister if I change the domain using Local, it also changes the value in the database. 

If the value wasn't changed in the database, how would one interact with the site at all? 🤔

jasonbahl:
If my domain in local doesn't match my site address in WordPress settings, I can't access the site at all until I update the database again directly

jasonbahl:
In any case, can you confirm that the updated code works? Maybe we can do a screenshare so you can show me what steps to take to reproduce?

moonmeister:
That's why you edit the db dicectly

jasonbahl:
If you edit the db directly nothing on the site works, including the admin. That's what I don't understand. 

I wouldn't expect the GraphQL endpoint to work if nothing else on the site works 🤔

moonmeister (commented):
Good call out. If you using domain based I think this would be fine. If using path based it may be an dissue cause you'd be comparing `/mysite/graphql` vs `/graphql`

jasonbahl:
@moonmeister this is working with our latest changes: 

Screen Shot 2021-12-20 at 11 07 54 AM


### test case

I'm having trouble getting a proper test written for this case as it requires some specific Docker setup that I'm not super familiar with yet. 

I'm hoping @markkelnar might be able to help me get the Docker test environment setup right to test this from a different domain than what is set in the siteurl option in the database. 

Here's a quick video walking through the situation and the fix.

https://www.youtube.com/watch?v=qK32G_QV-pw

I'll open up another issue to work on getting a proper test in place for this.

jasonbahl:
New issue for testing is here: https://github.com/wp-graphql/wp-graphql/blob/develop/docker/app.Dockerfile#L57

PR #2195, opened 2021-12-29
Fix type of parent ID when creating a response to a parent comment

mikelei1997:
I was trying to create a comment under another comment (i.e. reply to a comment), but it looks like setting `parent` in `createComment`'s input field does not work at all, the comment is posted directly under the post, not a response to another comment.
Changed the type of `parent` to `Int` and the issue was gone.
According to https://developer.wordpress.org/reference/functions/wp_new_comment/, when creating a new comment with parent ID, `comment_parent` field should be integer, not ID/string. Currently we're taking `parent` argument as ID.

No

There's no error message for this bug.
The following command would post the comment directly under post 3538 (`parent` does not work):
```
mutation CREATE_COMMENT {
  createComment(
    input: {commentOn: 3538 (Post ID), content: "This is a test comment, yo", author: "Jason", authorEmail: "xyz@gmail.com", parent: "Y29tbWVudDoxMDQ1"  (Parent Comment ID)}
  ) {
    success
  }
}
```
The following command would post a response under comment "Y29tbWVudDoxMDQ1" / 1045 after I changed type of `parent` to `Int`:
```
mutation CREATE_COMMENT {
  createComment(
    input: {commentOn: 3538 (Post ID), content: "This is a test comment, yo", author: "Jason", authorEmail: "xyz@gmail.com", parent: 1045  (Parent Comment ID)}
  ) {
    success
  }
}
```

Any other comments?
-------------------
First time creating a PR to an open source project, please let me know if I need to change anything, thanks!

Where has this been tested?
---------------------------
**Operating System:** Windows 11

**WordPress Version:** 5.8.2

### DIFF
--- a/src/Mutation/CommentCreate.php
+++ b/src/Mutation/CommentCreate.php
@@ -59,7 +59,7 @@ class CommentCreate {
 				'description' => __( 'Type of comment.', 'wp-graphql' ),
 			],
 			'parent'      => [
-				'type'        => 'ID',
+				'type'        => 'Int',
 				'description' => __( 'Parent comment of current comment.', 'wp-graphql' ),
 			],
 			'date'        => [


spencersmb:
+1 Having an issue with trying to post a comment as a reply to a parent comment.

justlevine:
This is working locally for me, but we should probably add a wpunit test for creating child comments.

justlevine:
On further review (https://github.com/wp-graphql/wp-graphql/issues/997), this seems to be the wrong approach. 
The type _should_ be `ID`, and you should be feeding it the WP database ID (as either a `String` or an `Int`).

Rather, we need to (plugin wide) either indicate that most where args expect a `databaseId` and not a relay ID or ideally create a helper function that accepts either a global id or db id and resolve it to the `int` db id.

jasonbahl:
closing in favor of #2328

PR #2196, opened 2021-12-30
Bug/#2166 content node preview for drafts is null

### LINKED ISSUE #2166: Retrieving a `contentNode` as preview for draft posts results in `null`
When retrieving a `contentNode` with `asPreview` set to `true` for a draft post, the response will be `null`.

## Versions:
- WP GraphQL: 1.6.7
- WordPress: 5.8.2

## Other Plugins Installed:

None

## Current Behavior

The response is `null`

## Expected Behavior

The response should include the proper content node

## How to reproduce

1. Create a draft page
2. Create a query with the following content:
   ```graphql
    query MyQuery {
      contentNode(id: "xxxx", idType: DATABASE_ID, asPreview: true) {
        id
      }
    }
   ```
3. Replace "xxxx" with the draft post ID
4. Notice the response:
     ```json
     {
       "data": {
         "contentNode": null
       }
     }
    ```

## Screenshots
Draft Page:

draft-page

`contentNode` with `asPreview`:

with-as-preview

`contentNode` without `asPreview`:

without-as-preview

apmatthews:
Supersedes #2189.

Other than rebasing, cleaning up the diff, and updating the tests, this includes [one functional change](65c2794b633dee2ab5d851dbb6caff05c0934912) from what was done in #2189. Instead of returning the draft ID as the `previewRevisionDatabaseID`, we return `null` when the post has no revisions. Returning the draft ID seemed misleading, since it's not actually a revision ID. This is covered in the updated tests.

_Description from #2189:_

This fixes some bugs with querying content nodes asPreview and getting null responses when the content node is in a draft state.

Closes #2166

Any other comments?
-------------------
Given the following post_type registration:

```php
	register_post_type( 'previewTest', [
		'graphql_single_name' => 'PreviewTest',
		'graphql_plural_name' => 'Preview',
		'show_in_graphql' => true,
		'supports' => [ 'title', 'editor' ],
		'show_ui' => true,
		'public' => true,
		'label' => 'Preview Tests'
	]);
});
```

We could then create a post of the previewTest post_type, and without publishing it, click preview.

Then query said post like so:

```gql
{
  notPreview: previewTest(id: 1473, idType: DATABASE_ID, asPreview: false) {
    ...PreviewTestFields
  }
  preview: previewTest(id: 1473, idType: DATABASE_ID, asPreview: true) {
    ...PreviewTestFields
  }
}

fragment PreviewTestFields on PreviewTest {
  id
  databaseId
  isPreview
  uri
  status
  content
  previewRevisionDatabaseId
}
```

## BEFORE

This would return the following results:

Screen Shot 2021-12-20 at 2 35 58 PM

even though we could see the preview of the draft post in WordPress itself.

Screen Shot 2021-12-20 at 2 36 10 PM

## AFTER
After this PR, we can now use the same ID to query, but we get the preview (the draft post) returned as expected:

Screen Shot 2021-12-20 at 2 38 12 PM


src/Type/ObjectType/RootQuery.php
```diff
@@ -197,10 +197,21 @@ public static function register_type() {
 										'check_enabled'  => false,
 									]
 								);
-								$post_id   = ! empty( $revisions ) ? array_values( $revisions )[0] : null;
+								$post_id   = ! empty( $revisions ) ? array_values( $revisions )[0] : $post_id;
 							}
 
-							return ! empty( $post_id ) ? $context->get_loader( 'post' )->load_deferred( $post_id ) : null;
+							$allowed_post_types   = get_post_types( [ 'show_in_graphql' => true ] );
```
apmatthews, line 203: @jasonbahl I wasn't sure whether the next several lines here were actually part of this bug fix or if they were part of your other feature branch. They seem to make sense in this context so I left them in place, but we could probably revert back to the one-liner return statement in this PR if the `allowed_post_types` check is coming with your `contentTypeName` feature.

jasonbahl (approved):

PR #2203, opened 2022-01-05
Add filter for post insert type during create

markkelnar:
A filter on the post input args for specific post type when post is created. Allows for args to be filtered, validated, renamed, etc.

Used for [this persisted queries work ](https://github.com/wp-graphql/wp-graphql-persisted-queries/pull/40) to filter input on mutation create/update. 

re: https://github.com/wp-graphql/wp-graphql-persisted-queries/issues/38

### DIFF
--- a/src/Mutation/PostObjectCreate.php
+++ b/src/Mutation/PostObjectCreate.php
@@ -296,17 +296,6 @@ class PostObjectCreate {
 			}
 
 			/**
-			 * Fires after a single term is created or updated via a GraphQL mutation
-			 *
-			 * The dynamic portion of the hook name, `$taxonomy->name` refers to the taxonomy of the term being mutated
-			 *
-			 * @param int    $post_id       Inserted post ID
-			 * @param array  $args          The args used to insert the term
-			 * @param string $mutation_name The name of the mutation being performed
-			 */
-			do_action( "graphql_insert_{$post_type_object->name}", $post_id, $input, $mutation_name );
-
-			/**
 			 * This updates additional data not part of the posts table (postmeta, terms, other relations, etc)
 			 *
 			 * The input for the postObjectMutation will be passed, along with the $new_post_id for the
--- a/src/Mutation/PostObjectCreate.php
+++ b/src/Mutation/PostObjectCreate.php
@@ -296,6 +296,17 @@ class PostObjectCreate {
 			}
 
 			/**
+			 * Fires after a single term is created or updated via a GraphQL mutation
+			 *
+			 * The dynamic portion of the hook name, `$taxonomy->name` refers to the taxonomy of the term being mutated
+			 *
+			 * @param int    $post_id       Inserted post ID
+			 * @param array  $args          The args used to insert the term
+			 * @param string $mutation_name The name of the mutation being performed
+			 */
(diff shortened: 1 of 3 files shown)


src/Mutation/PostObjectCreate.php
```diff
@@ -295,6 +295,17 @@ public static function mutate_and_get_payload( $post_type_object, $mutation_name
...
+			/**
+			 * Fires after a single term is created or updated via a GraphQL mutation
+			 *
+			 * The dynamic portion of the hook name, `$taxonomy->name` refers to the taxonomy of the term being mutated
+			 *
+			 * @param int    $post_id       Inserted post ID
+			 * @param array  $args          The args used to insert the term
+			 * @param string $mutation_name The name of the mutation being performed
+			 */
+			do_action( "graphql_insert_{$post_type_object->name}", $post_id, $input, $mutation_name );
```
jasonbahl, line 307: @markkelnar I think it might be better to centralize an action like this for all mutations to use. 

I'm thinking we could maybe put this action here: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Registry/TypeRegistry.php#L1160

That way all mutations are hookable, including mutations created by 3rd party plugins, etc.

I'm thinking it might be good to have a _before_ and _after_ action, like: 

```php

$user_input = $args['input'];

$filtered_input = apply_filters( 'graphql_mutation_input', $user_input, $context, $info, $whatever_else_seems_relevant )

$response = call_user_func( $mutateAndGetPayload, $filtered_input, $context, $info );

$filtered_response = apply_filters( 'graphql_mutation_resonse', $response, $filtered_input, $user_input, $context, $info );

// then return the filtered response. 
```

this would allow plugins to hook in and modify filters _before_ and _after_ the execution of _any_ mutation's resolver.

markkelnar: Interesting.  My change was to mimic an [action that already exists](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Mutation/PostObjectUpdate.php#L175) in the post object update.

jasonbahl: Ya, that was one of the first mutations in the codebase, and the filter was added for some specific needs. Now that some patterns have emerged, it makes more sense to centralize filters like this that can be used across the board. 

I'm not 💯 opposed to adding hooks/filters for specific mutations when it makes sense, but ideally I'd like to centralize things that can be centralized, and I _think_ this is one of those cases?

src/Registry/TypeRegistry.php
```diff
@@ -1157,7 +1157,12 @@ public function register_mutation( $mutation_name, $config ) {
 						// Translators: The placeholder is the name of the mutation
 						throw new Exception( sprintf( __( 'The resolver for the mutation %s is not callable', 'wp-graphql' ), $mutation_name ) );
 					}
-					$payload = call_user_func( $mutateAndGetPayload, $args['input'], $context, $info );
+
```
jasonbahl, line 1160: @markkelnar nice! Are you able to get your code in the persisted queries plugin working with this change?

jasonbahl (approved):

PR #2228, opened 2022-01-27
Make some user fields optional

victormattosvm:
When I try to update some user fields like last name, website, etc, with empty values, they're not being updated because of a condition that checks if they're empty and so ignores them. I should be able to update with empty values.

What I've done:
1. I've changed the condition (!empty) to (isset) for optional fields. I haven't changed the required ones.
2. The code was reordered into two groups: Required and optional fields

mutation: updateUser

...

...

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** 5.8.3
**GraphQL Version: 1.6.12

### DIFF
--- a/src/Data/UserMutation.php
+++ b/src/Data/UserMutation.php
@@ -121,6 +121,59 @@ class UserMutation {
 	public static function prepare_user_object( $input, $mutation_name ) {
 
 		$insert_user_args = [];
+		
+		/**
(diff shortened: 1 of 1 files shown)


jasonbahl (approved):
Thanks for this @victormattosvm!

Issue #1850, opened 2021-04-09
Receiving wrong contentType when idType is set to URI


JariSavolainen:
I have registered a custom post type. So when I query a page that has same slug/uri as the custom post type, I receive the custom post type instead. I tested this with contentNode as well. I have WP GraphQL Version 1.3.5 installed (WP 5.7).

**So this ...**
```
query getPageByURI {
  page(id: "example", idType: URI) {
    pageId
    title
    contentType {
      node {
        label
      }
    }
  }
  contentNode(id: "example", idType: URI, contentType: PAGE) {
    contentType {
      node {
        label
      }
    }
    ... on Page {
      pageId
    }
    ... on CustomPostType {
      customposttypeId
    }
  }
}
```

**... returns this ...**
```
{
  "data": {
    "page": {
      "pageId": 12345,
      "title": "Example",
      "contentType": {
        "node": {
          "label": "CustomPostTypes"
        }
      }
    },
    "contentNode": {
      "contentType": {
        "node": {
          "label": "CustomPostTypes"
        }
      },
      "customposttypeId": 12345
    }
  }
}
```

jasonbahl:
@JariSavolainen thanks for reporting. This definitely looks like a bug. 

I believe the code path for this is in `/src/Types/RootQuery.php` and then in `/src/Data/NodeResolver.php` if you were interested in trying to chase it down before I can get to it.

CptCompiler:
this does not only happen with custom post types.

I have a post with the uri `test` and no page with this uri.

if I do: 
```
page(id: "/test/", idType: URI)
```

I get the post. Should be null I think

jasonbahl:
@CptCompiler can you share a bit more info?

What happens if you visit `yoursite.com/test` in a browser? What loads? 

Then when you make the GraphQL query for a page by `/test/` uri, what's the response in GraphiQL?

What version of WPGraphQL are you using? The latest release (released yesterday: https://github.com/wp-graphql/wp-graphql/releases/tag/v1.5.2) did include a bugfix for another similar issue (https://github.com/wp-graphql/wp-graphql/pull/1994).

CptCompiler:
Sorry, I updated to the latest version (1.5.2) without checking the old version i had. But it's working now.

bhardie:
Based on the comments above it looks like this was resolved after 1.5.2.  If this is still an issue, please open a new issue referencing this issue.

Issue #1910, opened 2021-05-10
Can't query `/shop/` page by contentNode with uri argument when woocommerce is enabled (no wp-graphql-woocommerce required)


ardiewen:
Really strange and specific issue here... thought it might be downstream to `wp-graphql-woocommerce` and [wp-grapql/wp-graphql-woocommerce#491](https://github.com/wp-graphql/wp-graphql-woocommerce/issues/491), but it's present even without the downstream plugin enabled.

Steps to replicate:

1. Fresh wordpress 5.7.1 and woocommerce 5.2.2 install with wp-graphql 1.3.8, this triggers some pages to be automatically created, such as `/shop`, `/cart`, `/checkout` etc...

2. Try to query all content nodes, confirming that the shop page with id of `cG9zdDo0NA==` and URI of `/shop/` shows up.:
<img width="2028" alt="Screen Shot 2021-05-10 at 6 45 39 PM" src="https://user-images.githubusercontent.com/39966887/117733642-ed1e7c00-b1bf-11eb-8a88-77b8b74e8458.png">

3. Try to query `contentNode` by URI of `/shop/` (no wp-graphql-woocommerce enabled):
<img width="2028" alt="Screen Shot 2021-05-10 at 6 47 08 PM" src="https://user-images.githubusercontent.com/39966887/117733779-21923800-b1c0-11eb-8bc5-54aad9c58cf1.png">

4. Try to query `contentNode` by ID of `` (no wp-graphql-woocommerce enabled):
<img width="2028" alt="Screen Shot 2021-05-10 at 6 50 13 PM" src="https://user-images.githubusercontent.com/39966887/117734013-8f3e6400-b1c0-11eb-9a60-f88fafbfe4a7.png">

5. Turn woocommerce off, forcing all the default included woocommerce pages to disappear, recreating a "shop" page, and then querying it doesn't seem to do anything. Seems that something has been altered in the database maybe? The query looks identical to step 3, where the SQL trace doesn't seem to even be looking for a content node.

### Environment

wp-graphql 1.3.8
wordpress 5.7.1
woocommerce 5.2.2

blauwhelm:
I don't know if this is related, but I'm having a similar problem with fetching a custom post type.

So this gives null:

```graphql
{
  contentNode(id: "/locaties/gregorius", idType: URI) {
    slug
    ... on Location {
      title
    }
  }
}
```

And this gives a valid response:

```graphql
{
[...]

PR #2235, opened 2022-02-12
fix: resolve pages that double as post type archives

justlevine:
Certain custom post types (e.g. Woocommerce) use a WP `Page` that doubles as a post type archive. Currently querying for these by `page( idType: URI )` returns a `WP_Post_Type` instead of the associated page. 

This PR checks if the provided URI is a post type archive, and if so resolves it to the page, before falling back to returning a `WP_Post_Type`.

#1910 , #1850 (I think)

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (WSL2 + devilbox)

**WordPress Version:** 5.9


justlevine:
I dont think those test fails are coming from me...

jasonbahl:
@justlevine I'd love to see some good tests around this to make sure the intended functionality works properly now and to prevent regressions down the road. 

Would you be interested in writing some tests?

jasonbahl:
@justlevine I pulled the code down and tried to run through the WooCommerce "shop" page situation, and either I'm missing a step or this still doesn't address that situation. 

I set up WooCommerce following their wizard, so I have a page named "shop" that can be visited on my WordPress site at `/shop`.

But when I query `nodeByUri( uri: "/shop" ) { __typename }` I get a null response. No node can be found.

jasonbahl:
Possibly the WooCommerce situation wasn't what you optimized for though?

What are the steps to reproduce the other situation? Looking at the issue, it seems as if at least one commenter suggested the issue was resolved: https://github.com/wp-graphql/wp-graphql/issues/1850#issuecomment-880655490

justlevine:
To test:
1. Autogenerate the shop page by installing activating Woocommerce.
2. Disable Woocommerce and rebuild the permalinks.
3. Run the following query:
```graphql
{
  page(id:"/shop", idType:URI) {
    __typename
  }
  nodeByUri(uri:"/shop"){
    __typename
  }
}
```
Both will resolve to page.
4. Enable Woocommerce and expose the `Product` cpt to the schema (add [this gist](https://gist.github.com/justlevine/ffecb991060758a7f5bfdfdfcbbfdf6a) to `mu-plugins`
5. Rerun the query.
`page.__typename` will resolve to `Page`.
`nodeByUri.__typename` will resolve to `ContentType`. (Couldnt think of another way to expose both types for the same identifier, but I'm open to suggestions).

**However**:
6. [🚨]If the `product` post type isnt exposed to WPGraphQL (comment out the code from # 4), both `page` and `nodeByUri` will resolve to `null`. 
7. [⚠] If you set Settings > Reading > Posts page to a static page and then query `nodeByUri(uri:"/my-posts-page")`, `__typename` resolves to `Page`. This is inconsistent with the new behavior.

## tl;dr
Needs some more work and thorough unit testing.

jasonbahl:
@justlevine, your last comment was:

> Needs some more work and thorough unit testing.

Is this still the case, or is this in a state to be reviewed again?

justlevine:
I think anything I tried to do here is a waste of time until we can cleanup/refactor `NodeResolver` (#2342).

jasonbahl:
Ok cool. We'll circle back then

PR #2236, opened 2022-02-12
feat: Add filter `graphql_model_prepare_fields`

### LINKED ISSUE #2198: Need more intuitive method to extend Model `$fields`
# Tl;dr
Currently, adding a field to an existing GraphQL model is bad DX, requiring two WP Filters: first to set the field to null, and then to resolve the value. 

The fix is to pass the model $data to `graphql_return_model_data`.

# Problem

There's currently no intuitive method for adding a field to an existing WPGraphQL model.

[`graphql_return_model_data`](https://www.wpgraphql.com/filters/graphql_return_modeled_data/) in its current form is only really suitable for removing fields. Since `Model::$data` isnt passed to the filter, there is no way to resolve the new field to anything meaningful. 

Meanwhile, [`graphql_pre_return_field_from_model`](https://www.wpgraphql.com/filters/graphql_pre_return_field_from_model/) only handles fields that have already been registered.

As such, the only way to currently handle adding fields to a Model is to use _both_ filters, first setting the field to `null` in `graphql_return_model_data, and _then_ resolving it in `graphql_pre_return_field_from_model`:

```php
add_filter(
  'graphql_return_modeled_data',
  function( array $fields, string $model_name ) {
    if ( 'myModel' === $model_name ) {
      $fields['myCustomField'] = null;
    }
    return $fields;
  },
  10,
  2
);

add_filter(
  'graphql_pre_return_field_from_model'
  function( $return_field, string $key, string $model_name, $data ) {
    if ( 'myModel' === $model_name && 'myCustomField' === $key ) {
      $return_field = get_my_custom_field_value( $data );
    }
    return $return_field;
  },
  10,
  4
);
```

# Recommended Solution

The ideal solution (imo) would be to simply add the model `$data` to [`graphql_return_modeled_data`.

This is easy enough to do without breaking back-compat, by just appending `$this->data` as the last arg [in the filter](https://github.com/wp-graphql/wp-graphql/blob/194e89306ce0ebd158887f92b3b035dfc5a14972/src/Model/Model.php#L486).

__However__, from a DX standpoint, $data should really be earlier in the list of args:

```php
$this->fields = apply_filters( 'graphql_return_modeled_data', $this->fields, $this->get_model_name(), $this->data, $this->visibility, $this->owner, $this->current_user );
```

That would however constitute a breaking change.

justlevine:
Adds `graphql_model_prepare_fields` filter, and deprecates `graphql_model_prepare_fields`.

**Notes:** 
1. Didnt think it made sense to add a new `deprecated.php` file for just one deprecation, so i'm using `apply_filters_deprecated()` inline.
2. opted against passing the class instance `$this` as a parameter since a) it's more in line with the other class filters, b) the only property missing is `$restricted_cap` which doesn't seem necessary at this point in the lifecycle, and c) it means we dont need to audit which class properties/getters should be added/elevated.
3. the `$version` param on `apply_filters_deprecated()` is currently set to version `1.7.0` because of semver. It should be changed to whatever version this pr is released on.

fixes #2198 

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (WSL2 + devilbox)

**WordPress Version:** 5.9.0


jasonbahl (approved):

jasonbahl:
@justlevine I updated the docblock. Thanks for this PR!

PR #2269, opened 2022-03-04
Bug/#2268 graphiql changes overwhelms browser history

### LINKED ISSUE #2268: [v1.7.0] New WPGraphiQL pollutes browser history with url param changes
The new version of WPGraphiQL bundled with v1.7.0 adds every single state change to the browser history. In past versions, while changes updated the browser string, it wasn't part of the browser history.

Not sure if this change was intentional, as a way to undo changes by going 'back' in the browser (I'll leave it to you to decide if the browser history is the best way to do that vs storing the history in a separate WPGraphQL undo/redo feature), but even if it was, the back button still only changes the url in the address bar, and you would need to manually reload the page to get the old query information.

Also, other state changes pollute the Browser history that definitely shouldnt, such as `isQueryComposerOpen`.

image

## Expected behavior

_Either_ don't store changes to the url's query parameters in the browser history _or_ reload the GraphiQL state when going back/forward in the browser.

## Actual behavior

Every update to the url's query parameter is stored in the browser history, but navigating back/forward in the browser doesnt change the code to reflect the changed parameters.

### When reporting a _bug_, please be sure to include the following:
- [x] A descriptive title
- [x] An *isolated* way to reproduce the behavior (example: GitHub repository with code isolated to the issue that anyone can clone to observe the problem)
- [x] What version of `WPGraphQL` you're using, and the platform(s) you're running it on
- [x] What other plugins you're using
- [x] The behavior you expect to see, and the actual behavior

jasonbahl:
This provides a custom history object to the QueryParamProvider which configures the history to be replaced, instead of pushed. 

This allows queries to be typed / changed in GraphiQL without adding a history entry in the browser for each individual change.

closes #2268

### DIFF
--- a/packages/wpgraphiql/components/App/App.js
+++ b/packages/wpgraphiql/components/App/App.js
@@ -72,17 +72,8 @@ export const AppWithContext = () => {
       <QueryParams config={filteredQueryParamsConfig}>
         {(renderProps) => {
           const { query, setQuery } = renderProps;
-
-          const updateQuery = (query) => {
-            console.log( {
-              current: history.state,
-              updateQuery: query
-            })
-            setQuery(query);
-          }
-
           return (
-            <AppContextProvider queryParams={query} setQueryParams={updateQuery}>
+            <AppContextProvider queryParams={query} setQueryParams={setQuery}>
               <ApolloProvider client={client(getEndpoint())}>
                 <FilteredApp />
               </ApolloProvider>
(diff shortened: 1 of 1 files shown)


justlevine (approved):
confirmed working in the following browsers on Windows 10:
- Chrome 99.0.4844.51
- Firefox 97.0.1
- Edge 99.0.1150.30

PR #2278, opened 2022-03-07
Add composer post-install script

### LINKED ISSUE #2273: Add composer postinstall script to install npm dependencies so developers only have to run one command
As a contributor to WPGraphQL, I would like to only run one command `composer install` to have the plugin ready for local development. 

As of the 1.7.0 release, I also need to run `npm install && npm run build` in order to use GraphiQL

jasonbahl:
This PR updates the composer scripts to include a post-install script which installs npm dependencies and builds the graphiql app. 

This:

- update .gitattributes, to **_not_** ignore the packages directory, as it includes packages needed to build the JS files
- update `composer.json` to exclude the packages directory for plugin builds. The final result, after building the plugin, does not need the packages directory

closes #2273 

Any other comments?
-------------------
We should be able to also remove the `npm install` / `npm run build` scripts from workflows that already run composer install, as this will be redundant now.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -30,5 +30,4 @@
 /phpstan.neon.dist export-ignore
 /schema.graphql export-ignore
 /node_modules export-ignore
-/packages export-ignore
 /artifacts export-ignore
--- a/composer.json
+++ b/composer.json
@@ -86,7 +86,10 @@
     "fix-cs": [
       "php ./vendor/bin/phpcbf"
     ],
-    "phpstan": ["phpstan analyze --ansi --memory-limit=1G"]
+    "phpstan": ["phpstan analyze --ansi --memory-limit=1G"],
+    "post-install-cmd": [
+      "npm ci && npm run build"
+    ]
   },
   "archive": {
     "exclude": [
@@ -98,7 +101,8 @@
       "!.wordpress-org/",
       "docs/",
       "wp-graphql.zip",
-      "!build"
+      "!build",
+      "packages"
     ]
   }
 }


.gitattributes
```diff
@@ -30,5 +30,4 @@
 /phpstan.neon.dist export-ignore
 /schema.graphql export-ignore
 /node_modules export-ignore
-/packages export-ignore
```
jasonbahl, line 33: Instead of ignoring the `/packages` directory when downloading the plugin from github, this includes it, as it's needed for running `npm install && npm run build`

composer.json
```diff
@@ -86,7 +86,10 @@
     "fix-cs": [
       "php ./vendor/bin/phpcbf"
     ],
-    "phpstan": ["phpstan analyze --ansi --memory-limit=1G"]
+    "phpstan": ["phpstan analyze --ansi --memory-limit=1G"],
+    "post-install-cmd": [
```
jasonbahl, line 90: This runs `npm ci && npm run build` after composer deps are installed, so that the GraphiQL App is built and ready to use.

composer.json
```diff
@@ -98,7 +101,8 @@
       "!.wordpress-org/",
       "docs/",
       "wp-graphql.zip",
-      "!build"
+      "!build",
+      "packages"
```
jasonbahl, line 105: this excludes the `packages` directory when the plugin is archived, to be ready to be deployed to WordPress.org

The directory is no longer ignored when the plugin is cloned/downloaded, so the JS app can be built, but we don't want to include the raw JS in the distributed plugin.

Issue #1738, opened 2021-02-26
Common Interfaces that other Interfaces can Implement


jasonbahl:
The GraphQL Spec recently added support for Interfaces implementing other Interfaces, and the GraphQL PHP implementation even more recently added support for this. 

WPGraphQL should provide some common interfaces that can be implemented by other Interfaces. 

## Example

All connections in the graph currently return edges and nodes. The Type of Edge and Type of Node vary, but this is common to all connections, so we can provide a common `Connection` Interface.

```graphql
Interface Connection {
  edges: [ ConnectionEdge ]
  nodes: [ node ]
}
```

Then, we can add slightly more specific Interfaces. For example, all connections that connect _to_ Post objects (of any post type) are connections to `ContentNodes`, so we can have an Interface like so: 

```graphql
Interface ContentNodeConnection implements Connection {
  edges: [ ContentNodeConnectionEdge ]
  nodes: [ ContentNode ]
}
```

Then, connections to a specific Post Type could implement a more specific Connection: 

```graphql
Interface PostConnection implements ContentNodeConnection {
  edges: [ PostConnectionEdge ]
  nodes: [ Post ]
}
```

Then, we can have additional general interfaces such as: 

```graphql
Interface TermObjectToContentNodeConnection implements ContentNodeConnection {
  edges: [ ContentNodeConnectionEdge ]
  nodes: [ ContentNode ]
}
```

And then a specific implementation: 

```graphql
Type TagToPostConnectionEdge implements TermObjectToContentNodeConnection {
  edges: [ TagToPostConnectionEdge ]
  nodes: [ Post ]
}
```

## Added Value

This adds value to both client and server developers by providing more clarity via Introspection and more specific ways to customize the Schema on the server. 

### Introspection

Currently it's difficult to determine all the connections in the graph, but if all connections implemented a common `Connection` interface, Introspection could enable consumers to easily know about _all_ Connections by querying for all Types in the Schema that implement the Connection interface. 

Currently, trying to find all connections via introspection is, well, interesting: 

![Screen Shot 2021-02-26 at 11 52 52 AM](https://user-images.githubusercontent.com/1260765/109342497-2aae7280-7829-11eb-9422-5e40119b4443.png)

### Server Customization

Having Broad and Specific Interfaces and Types will allow server developers extending the Schema and underlying behavior more precise control over how to hook into the Graph. 

[...]

PR #2290, opened 2022-03-10
Register comments connection on NodeWithComments

chriszarate:
Register comments connection on `NodeWithComments` interface instead of one individual post types.

n/a

n/a

Any other comments?
-------------------
I like interfaces. :)

Where has this been tested?
---------------------------
**Operating System:* macOS

**WordPress Version:** 5.9

### DIFF
--- a/src/Connection/Comments.php
+++ b/src/Connection/Comments.php
@@ -75,42 +75,29 @@ class Comments {
 		);
 
 		/**
-		 * Register Connections from all existing PostObject Types to Comments
+		 * Register connection on NodeWithComments
 		 */
-		$allowed_post_types = \WPGraphQL::get_allowed_post_types();
-		if ( ! empty( $allowed_post_types ) && is_array( $allowed_post_types ) ) {
-			foreach ( $allowed_post_types as $post_type ) {
-				$post_type_object = get_post_type_object( $post_type );
+		register_graphql_connection(
+			self::get_connection_config(
+				[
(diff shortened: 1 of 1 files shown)


jasonbahl:
@chriszarate thanks for the PR! 

I ultimately want to get here! 

What I think needs to happen, still, is that Connections from/to Interfaces need to be able to also register edge Types for the specific relationships. 

For example, if we were to have a generic `ContentNode -> Term` connection, then the following Types should also be generated in the Schema:

### Types for the generic connection
- ContentNodeToTermConnection
- ContentNodeToTermConnectionEdge

### Specific Types for the specific connections
- PostToTagConnection implements ContentNodeToTermConnection
- PostToTermConnectionEdge implements ContentNodeToTermConnectionEdge
- PostToCategoryConnection implements ContentNodeToTermConnection
- PostToCategoryConnectionEdge implements ContentNodeToTermConnectionEdge

----

This allows for Specific relationships to have specific edge data exposed in the Schema. 

For example, if we had some relational data _only_ between Pages and Comments but not other post types, with the current PR we would _have_ to expose the edge data to all `ContentNode -> Comments` connections, which would be inaccurate. 

I believe #1738 starts laying some of this ground work, but there's definitely more work to do in the Connections realm to get to this final destination. 

Happy to chat more about it and see what we can do to get there. 

Would be 🔥  to have the `register_graphql_connection` API do most of this work with little-to-no change to the registration of connections.

PR #2293, opened 2022-03-11
Fix: Connections missing nodes when `before` or `after` are empty.

justlevine:
This PR fixes certain connections missing nodes when the `before` or `after` where args are empty.

Issue stems from the fact that `get_nodes()` in several connection resolvers simply checked for `isset( $this->args['{before|after'}])` before slicing the array.

Along those lines, I also got rid of the unnecessary `isset()` checks in a couple of places.

## to test:
Run the following query and confirm `withOffset` and `withoutOffset` are the same:

```graphql
{
  withOffset: taxonomies(first:2, before:""){
    nodes{
      name
    }
  }
  withoutOffset: taxonomies(first:2){
    nodes{
      name
    }
  }
}
```

`taxonomies` should be replaced with the other relevant connection types: `contentTypes`, `registeredScripts`, `registeredStylesheets`, `plugins`, `taxonomies`, `themes`, and `roles`.

Without PR (v1.7.2)
image

With PR:
image

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2) + Devilbox  + PHP v8.0.15

**WordPress Version:** v5.9.1

### DIFF
--- a/src/Data/Connection/UserConnectionResolver.php
+++ b/src/Data/Connection/UserConnectionResolver.php
@@ -133,7 +133,7 @@ class UserConnectionResolver extends AbstractConnectionResolver {
 				} elseif ( ! empty( $orderby_input['field'] ) ) {
 
 					$order = $orderby_input['order'];
-					if ( isset( $this->args['last'] ) && ! empty( $this->args['last'] ) ) {
+					if ( ! empty( $this->args['last'] ) ) {
 						if ( 'ASC' === $order ) {
 							$order = 'DESC';
 						} else {
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -470,7 +470,7 @@ abstract class AbstractConnectionResolver {
 	 * @return int|null
 	 */
 	public function get_after_offset(): ?int {
-		if ( isset( $this->args['after'] ) && ! empty( $this->args['after'] ) ) {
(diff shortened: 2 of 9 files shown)


justlevine:
Added tests.

Note:
- I was sloppy with the commit history.  I refactored some tests to use `WPGraphQLTestCase`, fixed `UserConnectionPaginationTest::testPaginateForwardAndBackward()`, and added barebones missing pagination tests for `contentTypes`, `registeredScripts`, `registeredStylesheets`, and `taxonomies`, in addition to testing for an empty `before`/`after` string.
- There's some commented out code regarding `last` args, as those tests will only work once #2294 is committed.

justlevine:
Worthwhile to note that I didnt try 'backfilling' missing connection tests in the new test files I created, only added those relevant to the changes in this PR.

src/Data/Connection/UserConnectionResolver.php
```diff
@@ -133,7 +133,7 @@ public function get_query_args() {
 				} elseif ( ! empty( $orderby_input['field'] ) ) {
 
 					$order = $orderby_input['order'];
-					if ( ! empty( $this->args['last'] ) ) {
+					if ( isset( $this->args['last'] ) && ! empty( $this->args['last'] ) ) {
```
justlevine, line 136: I'm pretty sure the `isset()` is redundant, and included in `! empty()`.

jasonbahl (approved):

Issue #2087, opened 2021-09-09
[Bug] `pageInfo` cursors for `plugins` and `themes` are not correctly returning `arrayconnection:{id}`


justlevine:
It seems the `endCursor` and `startCursor` on `plugins` and `themes` is incorrectly returning the bas64encode value of `arrayconnection:{path_to/plugin.php}` instead of `arrayconnection:{#}` (which is what's required by the `before`/`after` params).

## To replicate:

```graphql
plugins(first:5){
  pageInfo{
    endCursor # arrayconnection:{path_to/plugin.php}
    startCursor 
  }
  notWorking: plugins(first:5, after: $endCursor ){ # expects arrayconnection:{#}
  pageInfo{
    endCursor
    startCursor # arrayconnection:{path_to/plugin.php}
  }
  nodes { # returns []
    name
  }
}
```
WPGraphQL v1.6.4
WordPress: v5.8.1

PR #2294, opened 2022-03-11
fix: refactor AbstractConnectionResolver::get_nodes() to prevent children double slicing

justlevine:
This PR refactors `AbstractConnectionResolver::get_nodes()` to use a separate function (`AbstractionConnectonResolver::get_ids_for_nodes()` for slicing the query ids by the `query_amount` and `offset`.

This serves three purposes:
1. Improves the performance of child `get_nodes()` methods, since the models are only fetched for the actual nodes to return.
2. Makes extending `AbstractConnectionResolver` more DRY, since for most use cases in core, `get_nodes()` is really only used to change how the offsets are handled.
3. Fixes issues like #2087 , since currently `parent::get_nodes()` slices by `query_amount` _before_ slicing by `offset`, which is the entire reason `get_nodes()` needs to be overwritten to begin with.

#2087

Any other comments?
-------------------
 - The underlying issue is that some queries require the ids to be sliced by `query_amount` _before_ slicing by `offest`, while others need it to happen after. I see no solution to this that isn't a breaking change (if it is possible at all).
 - While this PR is _not_ a breaking change, plugins that overwrite `AbstractConnectionResolver::get_nodes()` by relying on `parent::get_nodes()` will still be broken as the latter is still slicing by `query_amount` before `offset.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (WSL + Devilbox) + PHP 8.0.15

**WordPress Version:** 5.9.1

### DIFF
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -592,27 +592,27 @@ abstract class AbstractConnectionResolver {
 	}
 
 	/**
-	 * Get_nodes
+	 * Get_ids_for_nodes
 	 *
-	 * Get the nodes from the query.
+	 * Gets the IDs from the query.
 	 *
 	 * We slice the array to match the amount of items that was asked for, as we over-fetched
 	 * by 1 item to calculate pageInfo.
 	 *
 	 * For backward pagination, we reverse the order of nodes.
 	 *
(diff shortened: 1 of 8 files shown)


justlevine:
While the code changes are independent, adding tests require #2293 (to avoid dealing with merge conflicts). Marking this as blocked in the interim.

jasonbahl (approved):

Issue #1845, opened 2021-04-06
Page/Post Templates should be registered to the schema using get_page_templates()


jasonbahl:
Currently WPGraphQL makes use of the global `wp_get_theme()->get_post_templates()` to find all templates. 

This method is not filterable. 

We should switch to use `get_page_templates()` to allow the templates to be filterable. 

We will also need to specify each post_type in order for the filters to be properly applied. 

See: https://core.trac.wordpress.org/browser/tags/5.7/src/wp-includes/class-wp-theme.php#L1265
See: https://core.trac.wordpress.org/browser/tags/5.7/src/wp-includes/class-wp-theme.php#L1281

CalebBarnes:
I was quickly trying to make a WP plugin to allow devs to add templates without creating PHP files in the theme but ran into this issue that my templates are not appearing in`wp_get_theme()->get_post_templates()`

Would love to see this update! Would be a nice boost in productivity since most headless sites don't even need a theme. 👍

I am using the wp filter `theme_page_templates` to add templates from an acf options page. 

Example here if you are interested: https://github.com/CalebBarnes/custom-templates-creator/blob/main/custom-templates-ui.php#L48-L65

jasonbahl:
Closing as this was fixed by @justlevine and released in v1.8.0 (see: https://github.com/wp-graphql/wp-graphql/releases/tag/v1.8.0)

PR #2296, opened 2022-03-11
Register post/page templates with get_page_templates()

justlevine:
This PR replaces the use of `wp_get_theme()->get_post_templates()` with the filterable `wp_get_theme()->get_page_templates()`.

For sanity, the `register_graphql_object_type()` loop has been moved from class `TypeRegistry` to class `Type\InterfaceType\ContentTemplate`, following the pattern established in `RootQuery::register_post_object_fields()`, and `RootQuery::register_term_object_fields()`.

#1845 

Any other comments?
-------------------
- I would recommend the pattern used here be adopted for the rest of the registration loops in TypeRegistry. Happy to work on a PR.
- From my understanding of the previous code (unconfirmed), _all_ templates were registered to the schema, not just those used by recognized Post Types. As they could only be queried via the `templates` field on a recognized post type, I don't think we should consider this a breaking change.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox), PHP 8.0.15

**WordPress Version:** 5.9.2

### DIFF
--- a/src/Model/Post.php
+++ b/src/Model/Post.php
@@ -535,7 +535,7 @@ class Post extends Model {
 				},
 				'template'                  => function () {
 
-					$registered_templates = wp_get_theme()->get_post_templates();
+					$registered_templates = wp_get_theme()->get_page_templates( null, $this->data->post_type );
 
 					$template = [
 						'__typename'   => 'DefaultTemplate',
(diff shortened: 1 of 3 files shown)


jasonbahl (approved):
@justlevine this is great!

PR #2299, opened 2022-03-15
fix: MediaItemUpdate mutation comparing string author ID with number returned from get_current_user_id

abaicus:

When running a `updateMediaItem` mutation as a user who can't edit another users' posts, a `UserError` gets thrown, even if this should go through for users who can update their post and are trying to do that. 

The `$author_id` variable was always a string. Both in the initial assignment when pulling it out of the media item post [here](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Mutation/MediaItemUpdate.php#L85), and for the case where a different user ID is provided [here](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Mutation/MediaItemUpdate.php#L109).

This PR should cast the variable to int, and make sure that the comparison with the `get_current_user_id` function call is valid [here](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Mutation/MediaItemUpdate.php#L116).

I have searched around, but it seems there isn't an open issue regarding this.

Where has this been tested?
---------------------------
**Operating System:** macOS Monterey 12.2.1 (21D62)

**WordPress Version:** 5.9.2

### DIFF
--- a/src/Mutation/MediaItemUpdate.php
+++ b/src/Mutation/MediaItemUpdate.php
@@ -82,7 +82,7 @@ class MediaItemUpdate {
 			if ( null === $existing_media_item ) {
 				throw new UserError( __( 'No mediaItem with that ID could be found to update', 'wp-graphql' ) );
 			} else {
-				$author_id = $existing_media_item->post_author;
+				$author_id = absint( $existing_media_item->post_author );
 			}
 
 			/**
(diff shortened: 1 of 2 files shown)


jasonbahl (approved):

jasonbahl:
@abaicus thanks for the PR! 

I added a test case for this to ensure we don't see regressions to this behavior in the future. 🙏🏻

PR #2319, opened 2022-04-02
fix: pass `WP_Comment` object to `comment_text` filter

### LINKED ISSUE #2309: The Filter hook comment_text should contain 3 params such as apply_filters( 'comment_text', $comment_text, $comment, $args )
In 168 line of /src/Model/Comment.php, 

return $this->html_entity_decode( apply_filters( 'comment_text', $content ), 'contentRendered', false );

This comment_text hook generates an error because it has only one param.

In 1028 line wp-includes/comment-template.php, you can find the following code.
       /**
	 * Filters the text of a comment to be displayed.
	 *
	 * @since 1.2.0
	 *
	 * @see Walker_Comment::comment()
	 *
	 * @param string          $comment_text Text of the current comment.
	 * @param WP_Comment|null $comment      The comment object. Null if not found.
	 * @param array           $args         An array of arguments.
	 */
      echo apply_filters( 'comment_text', $comment_text, $comment, $args );

justlevine:
This PR fixes an issue with the Comment model's `contentRendered` field, where the `WP_Comment` object was not passed to the `comment_text` wp filter.

closes #2309 

Any other comments?
-------------------
There are no tests to write. The behavior of the plugin hasnt changed, we're just fixing our use of WP's api.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + php8)

**WordPress Version:** 5.9.2

### DIFF
--- a/src/Model/Comment.php
+++ b/src/Model/Comment.php
@@ -165,7 +165,7 @@ class Comment extends Model {
 				'contentRendered'    => function () {
 					$content = ! empty( $this->data->comment_content ) ? $this->data->comment_content : null;
 
-					return $this->html_entity_decode( apply_filters( 'comment_text', $content ), 'contentRendered', false );
+					return $this->html_entity_decode( apply_filters( 'comment_text', $content, $this->data ), 'contentRendered', false );
 				},
 				'karma'              => function () {
 					return ! empty( $this->data->comment_karma ) ? $this->data->comment_karma : null;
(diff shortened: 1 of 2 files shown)


jasonbahl (dismissed):

jasonbahl:
Thanks @justlevine! 🙌🏻 

I added a test to ensure this filter can be called without issue and that the filtered value is properly passed to the end result.

Issue #2308, opened 2022-03-25
The filter commentType for commets doesn't work 


Talented-Business:
query MyQuery {
  comments(
    where: {commentType: "custom_comment", contentId: "67"}
  ) {
    edges {
      node {
        content
        type
        databaseId
      }
    }
  }
}
When usign above query, it doesn't filter the comments, it returns all comments.
Please fix it.

jasonbahl:
This functionality appears to be working. 

I inserted 3 comments into the database related to post id `1074`

2 are of the comment_type: `comment` and one is `pingback`

![Image](https://user-images.githubusercontent.com/1260765/182487797-6d4f54ac-24a9-4bca-9523-36eefa4527fc.png)

I can filter by `commentType: "comment"` and get just the comments:

![Image](https://user-images.githubusercontent.com/1260765/182487865-28b5bde8-2734-487a-ae4f-2ee8331e54cd.png)

Or I can filter by `commentType: "pingback"` and get just the pingback:

![Image](https://user-images.githubusercontent.com/1260765/182487942-0e31eacc-c491-4f0b-8bae-eb880a83c7f0.png)

PR #2320, opened 2022-04-02
Fix: comment connection filtering by `commentType`

justlevine:
This PR fixes filtering comment connection queries by `commentType`, `commentTypeIn`, and `commentTypeNotIn`.

~~It also changes the GraphQL type of `commentTypeNotIn` from `String` to `'list_of' => 'String'`, in line with the [`WP_Comment_Query` api](https://developer.wordpress.org/reference/classes/WP_Comment_Query/__construct/).~~

#2308 

Any other comments?
-------------------
1. I added tests to ensure `commentType` works, but I didn't backfill tests for other comment connection where args.
2. While the type of `commentTypeNotIn` changed, I not sure if `string`=> `string[]` is considered breaking since you can still pass a `String` value.  If so, I can move that to a separate PR. **Edit**: it _is_ breaking.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + php8 )

**WordPress Version:** 5.9.2


jasonbahl:
Here's a snippet users can use, should they need to make this field a String again, instead of a listOf strings

```php
add_filter( 'graphql_input_fields', function( $fields, $type_name, $config, $type_registry ) {

	if ( ! array_key_exists( 'commentTypeNotIn', $fields ) ) {
		return $fields;
	}

	$fields['commentTypeNotIn']['type'] = 'String';

	return $fields;

}, 10, 4 );
```

justlevine:
> I believe the only case this would be breaking is if users have explicitly used a variable in the query for the commentTypeNotIn field.

My concern is WP auto-updates. 
Since you can still pass a string _as a string_ to a `listOf`, I'm less worried about a user needing to revert the field to be a single value, and more about WP auto-updates breaking someone's build before they have a chance to read the changelog.

jasonbahl (approved):

Issue #2131, opened 2021-04-05
`createComment` mutation doesn't set correctly the author


aresrioja10:
**Describe the bug**
I'm trying to use the writeReview from GraphQL IDE and I've detected that it doesn't set correctly the author. It always set the logged user as the author.

**To Reproduce**
1. Use this example (changing the commentOn field if neccesary):
```
mutation writeReview {
  writeReview(input: {rating: 5, author: "John Doe", authorEmail: "john.doe@email.com", commentOn: 3468, content: "Lorem ipsum", date: "01/03/2021"}) {
    clientMutationId
    rating
    review {
      author {
        node {
          name
          email
        }
      }
    }
  }
}
```

2. Although we've set that the author is John Doe, the return data says that is admin@email.com (the logged user):

```
"writeReview": {
      "clientMutationId": null,
      "rating": 5,
      "review": {
        "author": {
          "node": {
            "name": "developer",
            "email": "admin@email.com"
          }
        }
      }
    }
```

**Expected behavior**
It should set the user defined in the mutation params.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Chrome

**Plugin Versions**
- **WooGraphQL Version:** 0.8.1
- **WPGraphQL Version:** 1.3.3
- **WordPress Version:** 5.7
- **WooCommerce Version:** 5.1.0

justlevine:
I've reproduced this in WPGraphQL v1.7.2

```graphql
mutation {
  createComment(
    input: {
      author: "John Doe",
      authorEmail: "john.doe@email.com",
      commentOn: 56,
      content: "Lorem ipsum",
      date: "01/03/2021"
    }
  ) {
    comment {
      author {
        node {
          email
          name
        }
      }
    }
  }
}
```
![image](https://user-images.githubusercontent.com/29322304/161253751-1224c5f5-eb25-495d-bda9-cdc519679864.png)

jasonbahl:
Closed by https://github.com/wp-graphql/wp-graphql/pull/2323

Released in v1.8.0: https://github.com/wp-graphql/wp-graphql/releases/tag/v1.8.0

PR #2323, opened 2022-04-03
Fix: Use comment author when admin are creating/updating a different users comment.

justlevine:
This PR fixes `CommentMutation::prepare_comment_object()` to correctly set a comment's author when it differs from the currently logged-in user.

#2131 

Any other comments?
-------------------
- Note the two commented-out asserts in `CommentMutationsTest::testCreateComment()`. For some reason, a subscriber is unable to view their own comment. The resolver creates the model, and the model's visibility is visible, but for some reason `DataSource::resolve_comment()` is coming back null. 
I reverted the changes and tried the asserts, but  `comments` still returns null in the tests (didnt check manually in WPGraphiQL), so I believe it's a pre-existing bug. If so a separate issue should be created.

Where has this been tested?
---------------------------

**WordPress Version:** 5.9.2

### DIFF
--- a/tests/wpunit/CommentMutationsTest.php
+++ b/tests/wpunit/CommentMutationsTest.php
@@ -227,7 +227,7 @@ class CommentMutationsTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase {
 
 		$this->assertArrayNotHasKey( 'errors', $actual );
 		$this->assertTrue( $actual['data']['createComment']['success'] );
-		// $this->assertEquals( $this->subscriber->ID, $actual['data']['createComment']['comment']['author']['node']['databaseId'] );
+		$this->assertEquals( $this->subscriber, $actual['data']['createComment']['comment']['author']['node']['databaseId'] );
 
 		// Test logged in user different than author.
 		wp_set_current_user( $this->admin );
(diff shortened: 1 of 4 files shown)


jasonbahl:
@justlevine ah yes. This one is tricky. 

It looks like the comment is "pending" and subscribers don't have permission to view comments that aren't approved yet. 

However, an authenticated user, including a subscriber, _should_ be able to view their own comment. 

I'll look into this 👀

justlevine:
I've had similar trouble in general when using the data loader to resolve mutations outputs, but here it _does_ work for admin. I still don't fully understand how `Deferred`s work...

justlevine:
Yeah it was redundant. Unless I overlooked something theyre now covered by  `testCreateComment`

tests/wpunit/CommentMutationsTest.php
```diff
@@ -168,57 +171,87 @@ public function testCreateComment() {
...
+			'author'    => null,
+			'email'     => null,
 		];
+		wp_set_current_user( $this->admin );
 
-		/**
-		 * use --debug flag to view
-		 */
-		codecept_debug( $actual );
+		$actual = $this->graphql( compact( 'query', 'variables' ) );
```
justlevine, line 214: replaces L83

tests/wpunit/CommentMutationsTest.php
```diff
@@ -518,37 +520,32 @@ public function testCanCreateCommentNotLoggedIn() {
...
+			'author'    => 'Comment Author',
+			'email'     => 'subscriber@example.com',
+		];
 
 		wp_set_current_user( 0 );
 
-		$actual = do_graphql_request( $mutation, 'createCommentTest', $variables );
-
-		codecept_debug( $actual );
+		$actual = $this->graphql( compact( 'query', 'variables' ) );
```
justlevine, line 548: duplicates assert on L103

jasonbahl (approved):

Issue #2201, opened 2022-01-04
Feature Request: And/Or Operators


jadamconnor:
Other GraphQL libraries feature And/Or [operators](https://dgraph.io/docs/graphql/queries/and-or-not/). An example:

```
posts(where: {_and: [{categoryName: "foo"}, {categoryName: bar}]}) {
    edges {
        node {
            id
        }
    }
}
```

Is there some specific limitation preventing this from being implemented with WPGraphQL?

jasonbahl:
@jadamconnor the only limitation so far has been coming up with a good plan around it, and spending the time to execute the plan!

It's been discussed a bit here already: https://github.com/wp-graphql/wp-graphql/issues/1385

I think this is the way to go, for sure. 

The early design around the `where` args was, I think, a mistake, as it was an attempt to treat GraphQL simply as a mirror of internal WordPress mechanisms and not a Graph. 

I think formalizing a "spec" of sorts around the and/or operators, how plugins can hook in to add their own, how the input should affect underlying queries, etc would be great.

jasonbahl:
Thanks again for this request, it remains one of the most asked-for improvements to querying with WPGraphQL. As part of an audit of open issues, a heads up on where this is going: we plan to introduce a new comprehensive RFC for newly designed Connection `filter` and `sort` args that would replace the current connection `where` args, designed as a proper spec rather than a mirror of WP_Query, including support for and/or style compound filtering like what was requested here. I'll link the RFC here and in #1385 when it's published so the use cases from this thread can be discussed against a concrete proposal.

jasonbahl:
Closing this for now. Will keep #1385 open for now until we have the RFC I mentioned above prepared.

PR #2325, opened 2022-04-04
-  changes `commentTypeNotIn` field to a listOf String instead of String

### LINKED ISSUE #2321: `commentTypeNotIn` input arg should be type `String[]`
Per the [`WP_Comment_Query` api](https://developer.wordpress.org/reference/classes/WP_Comment_Query/__construct/), `commentTypeNotIn` should be changed [here](https://github.com/wp-graphql/wp-graphql/blob/58ecca1c263ca377787d863b731c6d342d3c017a/src/Connection/Comments.php#L284) from `'type' => 'String'` to `'type' => [ 'list_of' => 'String' ]`.

This is a breaking schema change.

jasonbahl:
This changes the `commentTypeNotIn` field on Comment connections to be a `list_of => string` type instead of a `string` type. 

closes #2321 

Any other comments?
-------------------

This is a breaking change, as queries that were written like so:

```graphql
query getComments( $notIn: String ) {
  comments( where: { commentTypeNotIn: $notIn ) {
    ...
  }
}
```

would need to be changed like so: 

```graphql
query getComments( $notIn: [String] ) { # <-- note the change to the variable declaration
  comments( where: { commentTypeNotIn: $notIn ) {
    ...
  }
}
```

## Breaking Change

This includes a minor breaking change to the Schema, as seen here: https://github.com/wp-graphql/wp-graphql/runs/5823051931?check_suite_focus=true

Should you need to keep backward compatibility with the previous shape of the `commentTypeNotIn` field, this snippet will change the field back to a string instead of a `listOf( string )`

```php
add_filter( 'graphql_input_fields', function( $fields, $type_name, $config, $type_registry ) {

	if ( ! array_key_exists( 'commentTypeNotIn', $fields ) ) {
		return $fields;
	}

	$fields['commentTypeNotIn']['type'] = 'String';

	return $fields;

}, 10, 4 );
```


jasonbahl:
Closing this out. The linked issue (#2321) was already closed as a duplicate, and rather than making one-off breaking changes to individual `where` args, we're planning a broader refactor of connection filtering and sorting that will address the input types for comment queries (and other connections) holistically. See #2201 for related discussion.

Issue #532, opened 2018-10-01
New comment mutations needs to use the global relay id


kidunot89:
Comment mutations relies on the post-object's `postId` field to identify the parent object that is being add to. It should probably use the relay global id of the post-type object or comment object it being added on. As it is know you currently can't add comments on comments either.

jasonbahl:
@kidunot89 ya, I was exploring having mutations accept _either_ a Global Relay ID _or_ the Database ID. 

The `ID` Type in GraphQL will accept input as a String or an Integer, so we _should_ be able to declare any ID inputs as the `ID` Type, and then input could look accept either type of ID and we would just validate one way or the other. . .something to this tune (total pseudo code):

```
if ( ! empty( $args['someId`] ) ) {
   if ( absint( $args['someId'] ) {
     $id = absint( $args['someId'] );
   } else {
     $id = relay_function_that_I_cant_remember_to_get_the_id( $args['someId'] );
   }
   return ! empty( $id ) ? $id : $id;
}
```

kidunot89:
That's what I was thinking. I think its bad practice to cart the `postId` on all that post's comment components just to create sub-comments, which what we have to do now, unless you want to keep it in a store which I also think is bad practice.

jasonbahl:
Related to #109

jasonbahl:
related #2322

justlevine:
Pretty sure this can be closed, as per the tracking in #998.

PR #2328, opened 2022-04-05
feat: allow global/db ids in comment mutation ID inputs

justlevine:
This PR enables users to supply _either_ a database ID or a global ID to comment mutation inputs of `type` ID. Part of #998 .
The description of `commentOn` was changed to specify it only takes a database ID  (until #532 can be handled).
Also a new test was added for creating child comments.

Invalidates #2195 

Any other comments?
-------------------
- I'm intentionally handling the different parts of #998 in separate PRs to make merging easier. 
- I removed the empty input `UserError` from `CommentUpdate`, since `id` is non-null and its exclusion will already throw a GraphQL error.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (Wsl2 + devilbox + PHP 8.0.15)

**WordPress Version:** 5.9.2

### DIFF
--- a/src/Data/CommentMutation.php
+++ b/src/Data/CommentMutation.php
@@ -6,6 +6,7 @@ use Exception;
 use GraphQL\Error\UserError;
 use GraphQL\Type\Definition\ResolveInfo;
 use WPGraphQL\AppContext;
+use WPGraphQL\Utils\Utils;
 
 /**
  * Class CommentMutation
@@ -85,7 +86,7 @@ class CommentMutation {
 		}
 
 		if ( ! empty( $input['parent'] ) ) {
-			$output_args['comment_parent'] = $input['parent'];
+			$output_args['comment_parent'] = Utils::get_database_id_from_id( $input['parent'] );
 		}
 
 		if ( ! empty( $input['type'] ) ) {
--- a/src/Mutation/CommentCreate.php
+++ b/src/Mutation/CommentCreate.php
@@ -36,7 +36,7 @@ class CommentCreate {
 		return [
 			'commentOn'   => [
 				'type'        => 'Int',
-				'description' => __( 'The ID of the post object the comment belongs to.', 'wp-graphql' ),
+				'description' => __( 'The database ID of the post object the comment belongs to.', 'wp-graphql' ),
 			],
 			'author'      => [
 				'type'        => 'String',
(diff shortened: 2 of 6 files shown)


justlevine:
@jasonbahl this is proof of concept Once I get feedback/approval on this, I'll start work on the other mutations, then connections.

jasonbahl (dismissed):
@justlevine this is wonderful! 👏🏻 

I like the iterative approach you're taking as well, tackling small bits at a time. 

Have you done a pretty thorough audit and have a list of all the fields that need to be addressed? 

I wonder if it might make sense to create a feature branch for this, which is opened as a PR to develop, then we can make small iterative PRs to that feature branch, and work toward a larger merge with several of these PRs accumulated?

Maybe use one Issue and some checkboxes as well to track what inputs need to be changed, etc?

Could help us with this current phase, but also help us set up for when we're ready to make the breaking changes (changing `Int` to `ID` in various places).

justlevine:
Sorry, accidentally pushed media items to this PR, just reverted.

>Have you done a pretty thorough audit and have a list of all the fields that need to be addressed?

I was going to, but then I realized its just a simple search for `=> 'ID'`. 
image

I can take a beat and write them out if you want.

> I wonder if it might make sense to create a feature branch for this, which is opened as a PR to develop, then we can make small iterative PRs to that feature branch, and work toward a larger merge with several of these PRs accumulated?

Your call. I personally try to avoid feature branches whenever for the same reason I'm submitting these PRs piecemeal - merge conflicts are a PITA.

If you're looking to prep for a near-term release and are worried about confusing users by enhancing the mutations/connection args incrementally, then feature branch is probably the way to go. But I can also make the argument that since the accepted `ID` inputs are confusing users _in their current state_, so even if they were added incrementally over a few releases, it'd still lessen the confusion. 🤷‍♂️

Without being too optimistic, updating mutation inputs is pretty straightforward. The only thing that might slow me down is backfilling missing tests.

Connections are going require a larger discussion over in #998 - the question of where in the lifecycle to make the change needs to be thought out foundationally (e.g. is the db or global id saved to `$query_args['graphql_args']`), and on an individual basis (even with AbstractConnectionResolver, each connection have their own order for calling parent methods / processing the args).

jasonbahl (approved):

PR #2332, opened 2022-04-06
Bug/#2282 version graphiql with releases

### LINKED ISSUE #2282: Installing WPGraphQL with Composer doesn't include the built GraphiQL app
When installing WPGraphQL from composer, using `composer require wp-graphql/wp-graphql` starting with v1.7.0, the GraphiQL App is not included anymore. 

In previous versions, I was building the GraphiQL app locally and committing / versioning the built app. The built files were included in the distribution to Packagist and therefore included in the installed files from composer. 

Now, the built JavaScript app is not versioned in Github, but instead it is built for deploying to WordPress.org. 

However, the JavaScript app is _not_ built before deploying to Packagist.org. 

I'm not fully sure if there's a way to do this, as Packagist.org picks up the package when a release is tagged. If theres some sort of "middleware" I can hook into between release on Github and Packagist getting the files, that would be one option to ensure the JavaScript app gets built and included in the packagist archive. 

Another option, would be to update the documentation to guide users that install with composer, to install from wpackagist.org, which gets the built plugin from the WordPress.org repo.

jasonbahl:
This PR does the following: 

- adds the `build` directory & built assets for GraphiQL back to the repo 
- removes `post-install-cmd` from composer.json so that running `composer install` does not also run `npm ci && npm run build`
- Adds a github workflow that checks to see if any given Pull Request or push to develop/master includes changes to `/packages` directory, and if so, run the `npm ci && npm run build` command, then commit the changes from that command back to the codebase
- leave the `build` directory in `.gitignore` so that contributors aren't including changes to these files in their PRs directly

closes #2282 

Any other comments?
-------------------
This should allow users that install WPGraphQL from Composer to keep installing from Composer, as Composer will have the built JS app versioned. 

Contributors that are working on the GraphiQL portion of the plugin should also be able to contribute without interruption. Make changes, build the app locally, push changes, etc and the Workflow should handle building the app consistently and committing the build assets


kidunot89 (dismissed):
@jasonbahl Looks good 👍🏿

.github/workflows/build-graphiql.yml
```diff
@@ -0,0 +1,40 @@
...
+          npm ci && npm run build
+
+      # Commit the built assets for GraphiQL back to the repo
+      - name: Commit built GraphiQL
+        id: commit-changes
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: Apply GraphiQL build changes
+          # Since the build directory is .gitignored, we force the changes
+          push_options: --force
```
esamattis, line 40: This feels wrong since it is under `push_options`.  I think you want to add the `--force` flag to `git add`.

Also not sure why the `stefanzweifel/git-auto-commit-action` is needed. You could just call the git cli directly:

```
git add --force builddir
git commit -m "Commit build"
git push origin HEAD:${{ github.head_ref }}
```

jasonbahl:
> This might not work 🤔

Ah yes, you're correct. It doesn't look like it works as I hoped 🤔 

So ya, I think we version the build directory (and remove it from .gitignore), then anytime a push is made to master/develop, if there were changes to the `packages` or `build` directory, we run the build and commit and push the results of the build.

PR #2337, opened 2022-04-09
Feat/user object mutation id inputs

justlevine:
This PR enables users to supply either a database ID or a global ID to user mutation inputs of type ID. Part of #998 

#2194 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + php8.0.15)

**WordPress Version:** 5.9.2

### DIFF
--- a/src/Mutation/UserDelete.php
+++ b/src/Mutation/UserDelete.php
@@ -5,6 +5,7 @@ namespace WPGraphQL\Mutation;
 use GraphQL\Error\UserError;
 use GraphQLRelay\Relay;
 use WPGraphQL\Model\User;
+use WPGraphQL\Utils\Utils;
 
 /**
  * Class UserDelete
@@ -80,19 +81,21 @@ class UserDelete {
 	 */
 	public static function mutate_and_get_payload() {
 		return static function ( $input ) {
-			/**
-			 * Get the ID from the global ID
-			 */
-			$id_parts = Relay::fromGlobalId( $input['id'] );
+			// Get the user ID.
+			$user_id = Utils::get_database_id_from_id( $input['id'] );
 
-			if ( ! current_user_can( 'delete_users', absint( $id_parts['id'] ) ) ) {
+			if ( empty( $user_id ) ) {
+				throw new UserError( 'The user ID passed is invalid', 'wp-graphql' );
+			}
+
+			if ( ! current_user_can( 'delete_users', $user_id ) ) {
 				throw new UserError( __( 'Sorry, you are not allowed to delete users.', 'wp-graphql' ) );
 			}
 
 			/**
 			 * Retrieve the user object before it's deleted
 			 */
-			$user_before_delete = get_user_by( 'id', absint( $id_parts['id'] ) );
+			$user_before_delete = get_user_by( 'id', $user_id );
 
 			/**
 			 * Throw an error if the user we are trying to delete doesn't exist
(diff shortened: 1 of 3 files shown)


src/Mutation/UserDelete.php
```diff
@@ -80,19 +81,21 @@ public static function get_output_fields() {
...
-			/**
-			 * Get the ID from the global ID
-			 */
-			$id_parts = Relay::fromGlobalId( $input['id'] );
+			// Get the user ID.
+			$user_id = Utils::get_database_id_from_id( $input['id'] );
 
-			if ( ! current_user_can( 'delete_users', absint( $id_parts['id'] ) ) ) {
+			if ( empty( $user_id ) ) {
+				throw new UserError( 'The user ID passed is invalid', 'wp-graphql' );
```
jasonbahl: should be: 

```
throw new UserError( __( 'The user ID passed is invalid', 'wp-graphql' ) );
```

jasonbahl: Github won't let me make a suggestion 🤔

justlevine: Gone now I believe

jasonbahl: Confirmed.

justlevine:
This also closes #2194

jasonbahl (dismissed):

jasonbahl (approved):

jasonbahl:
@justlevine so you can change your env variable to make multisite tests run or not

`MULTISTE=0` means the tests will run in non-multisite
`MULTISITE=1` means they will run in multisite

I tend to run tests locally outside of docker (faster for me) so I can run them like so: `MULTISITE=true vendor/bin/codecept run wpunit` or `MULTISITE=false vendor/bin/codecept run wpunit`

justlevine:
> From what I can tell, they're running as expected.

I'm not by my computer, but that's the only reason I can think that the reassign test is failing now with a bad reassign, on _all_ the tests (and not just 5.9 MU). Can be confirmed if you a put a `codecept_debug` in the is_multisite() conditional in that test. Not sure _why_ that would be, though

justlevine:
https://github.com/wp-graphql/wp-graphql/blob/33255ada6524870245a9367be6420e6a37149553/src/Mutation/UserDelete.php#L140

justlevine (commented):
Should this be in the `is_multisite()` check, or am I guilty of premature optimization?

src/Mutation/UserDelete.php
```diff
@@ -133,12 +133,22 @@ public static function mutate_and_get_payload() {
 			if ( ! function_exists( 'wpmu_delete_user' ) ) {
 				require_once ABSPATH . 'wp-admin/includes/ms.php';
 			}
+			if ( ! function_exists( 'remove_user_from_blog' ) ) {
```
justlevine, line 136: Should this be in the is_multisite() check, or am I guilty of premature optimization?

jasonbahl:
@justlevine I mistyped. instead of `MULTISITE=false vendor/bin/codecept run wpunit` it needs to be `MULTISITE=0 vendor/bin/codecept run wpunit`

I can verify now by placing this inside the `mutate_and_get_payload` function in the `UserDelete.php` file and running the test command. 

` MULTISITE=0 vendor/bin/codecept run tests/wpunit/UserObjectMutationsTest.php:testDeleteUserWithReassign --debug`

CleanShot 2022-05-16 at 16 56 23

` MULTISITE=1 vendor/bin/codecept run tests/wpunit/UserObjectMutationsTest.php:testDeleteUserWithReassign --debug`

CleanShot 2022-05-16 at 16 57 16

src/Mutation/UserDelete.php
jasonbahl: nope, I was moving this already! The test actually didn't like it outside the `is_multisite()` call! So good suggestion 😄

jasonbahl: just pushed this change 😄

PR #2341, opened 2022-04-11
fix: Wrong term URIs for subsites

jeanfredrik:
```php
// src/Type/ObjectType/TermObject.php

$url = $term->link;
if ( ! empty( $url ) ) {
  $parsed = wp_parse_url( $url );
  if ( isset( $parsed ) ) {
    $path  = isset( $parsed['path'] ) ? $parsed['path'] : '';
    $query = isset( $parsed['query'] ) ? ( '?' . $parsed['query'] ) : '';
    return trim( $path . $query );
  }
}
return '';
```

This differs from how `uri` is fetched on `PostObject` and doesn’t return the correct uri for subsites on netowrk installs that use subdirs. This PR changes this so it uses the same method as `PostObject->uri`:

```php
// src/Model/Term.php

$link = $this->link;
return ! empty( $link ) ? str_ireplace( home_url(), '', $link ) : null;
```

This also moves the logic from `WPGraphQL\Type\ObjectType\TermObject` to `WPGraphQL\Model\Term` which seems not to have been used before 🤔


src/Model/Term.php
```diff
@@ -194,15 +194,9 @@ protected function init() {
...
+					$link = $this->link;
 
-					if ( is_wp_error( $link ) ) {
-						return null;
-					}
-
-					$stripped_link = str_ireplace( home_url(), '', $link );
-
-					return trailingslashit( $stripped_link );
+					return ! empty( $link ) ? str_ireplace( home_url(), '', $link ) : null;
```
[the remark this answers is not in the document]
jeanfredrik: There’s only one `return` 🤔

justlevine:
@jeanfredrik that test is correct. In WP native, term links are filtered before being output, and not an identical copy of what's saved in the DB.

justlevine:
From my understanding, the changing from the filtered `get_term_link()` to the unfiltered `$term->link` shouldn't change behavior for multisite URIs on its own.
Can you share the values you're getting pre/post this PR?

jeanfredrik:
But why the use of `parse_url` though? Wouldn’t removing home_url from `term_link` be enough?

jeanfredrik:
Before the PR the subsite’s subdir is included in the uri, e.g. `/mysubsite/posts/categories/mycategory/` instead of `/posts/categories/mycategory/`. This is because `parse_url` is used instead of replacing the `home_url` with an empty string as is done with posts.

jeanfredrik:
`$term->link` is also filtered though, here: https://github.com/whitespace-se/wp-graphql/blob/c213b76f449fae5e08c445d157758204f6d3baae/src/Model/Term.php#L166-L169

justlevine:
And if you replace TermNodeTest#L699-702 with your `str_ireplace()` logic, does the test pass.
If so, and we confirm that 

```php
add_filter( 'term_link', function ( $term_link ) {
	$frontend_uri = 'http://localhost:3000/';
	$site_url     = trailingslashit( site_url() );
	return str_replace( $site_url, $frontend_uri, $term_link );
});
``` still includes the port in the returned URI, then the test can just be updated and we're good to go no breaking changes.

(Sorry, im not by my comp til later this evening)

**Review by justlevine — COMMENTED** [review]
looks good to me :-)

**Comment by jasonbahl** [comment]
@theodesp Would you mind reviewing this PR? Since you worked on this a few months ago and are familiar with this space and have integrations that use that filter, I'd like to see if this change causes any regressions with your integrations. 🙏🏻

**Comment by theodesp** [comment]
@jasonbahl There is definitely some regression changes here when resolving the term uri:

**Current**
Screenshot 2022-04-19 at 17 00 12

**PR Update**
Screenshot 2022-04-19 at 16 59 46

**Comment by justlevine** [comment]
@theodesp anything special about your env other than ACM? Also how are you setting the site uri to `localhost:3000`?

Its working fine for me on a fixed domain, so I'm betting it has something to do with the fact that this solely relies on `home_url()` in the `Model`, instead of using `wp_parse_url()` in the `TermObject.uri` resolver to essentially ignore the domain name.

image

Not sure if ignoring the domain name is a bug (external redirects?) or a feature, but if it's the latter or we just dont want to wait for a breaking release, its an easy fix to reimplement `wp_parse_url()` on the model.

**Comment by theodesp** [comment]
@justlevine I'm using the [Faust.js](https://faustjs.org/) headless plugin and I set the FrontEnd URI to point to the Headless site at localhost:3000

**Comment by justlevine** [comment]
Indeed, the issue stems from the fact that FaustWP does [its own `str_replace()` for `home_url()`](https://github.com/wpengine/faustjs/blob/1762188abddeef809916b31d415851da3dfcb826/plugins/faustwp/includes/replacement/functions.php#L55) on [`term_link`](https://github.com/wpengine/faustjs/blob/canary/plugins/faustwp/includes/replacement/callbacks.php#L198), so by the time [it gets to WPGraphQL](https://github.com/wp-graphql/wp-graphql/blob/b4990207b2e82dfcd1515f96c458516a5c7bc84b/src/Model/Term.php#L167), its recognized as an _external_ url and therefore isnt made relative.

I _think_ the solution is for FaustJS to use [the `home_url` filter](https://developer.wordpress.org/reference/hooks/home_url/) instead of their current approach, but I say that without knowing the reasons that led to it.

As far as this PR goes, not sure if it makes sense to wait for the change before committing, or to go back to ignoring the base url via `wp_parse_url()` without differentiating between an actual site link and an external one.

**Comment by justlevine** [comment]
From @apmatthews over in the [FaustJS Discord](https://discord.com/channels/836253505944813629/877940107016019998/976203803625914378)

>We use a custom setting for the frontend URL in FaustWP because home_url is often used in contexts that assume home_url points to a WordPress frontend. For example, the Gutenberg editor uses home_url to locate the REST API. If we were to use (or filter, as you suggest) the home_url() setting to point at the headless frontend, Gutenberg (and potentially other things) would break. There are workarounds in some cases (i.e. filtering the rest_url to make an exception to the home_url replacement), but the decision was ultimately to play it safe -- use a custom frontend URL setting and only replace where necessary rather than modify the home_url() globally.

As such, short term I recommend we go back to using `wp_parse_url()` instead of `home_url()`, (and if nobody beats me too it, ill update this PR when I get a chance), so we're not actively causing them headaches

Long-term and from a broader codebase perspective, I'm a bit more conflicted. I still think being able to differentiate between local/external links before making a `uri` relative is important, and definitely think switching out [other `home_url()` calls in the codebase](https://github.com/wp-graphql/wp-graphql/search?home_url%28%29) is the wrong move.

We could possibly work around this and make the `uri` results consistent and significantly simpler for _everyone_ (including FaustJS and WPGraphQL extensions for 3rd party plugins) with a helper method for getting the `home_url()` + a wpgraphql-specific filter. Gonna open up a separate issue to explore that.

File: src/Model/Term.php [hunk 2]
```diff
@@ -194,15 +194,9 @@ protected function init() {
 					return $queue;
 				},
 				'uri'                      => function () {
```
justlevine, line 196: Replacing the callback with the following _should_ do the trick (I'm not 100% sure the slash situation with `WP_Site->$path`):

```php
'uri'  => function () {
	$link = $this->link;

	if ( empty( $link ) ) {
		return null;
	}

	$parsed = wp_parse_url( $link );

	if ( empty( $parsed ) ) {
		return null;
	}

	// @todo We should probably check the the URL is for the current site before making it relative.

	$path  = $parsed['path'] ?? '';
	$query = isset( $parsed['query'] ) ? ( '?' . $parsed['query'] ) : '';

	// Strip out multisite directory.
	if ( ! empty( $path ) && is_multisite() ) {
		$site = get_site();

		if ( $site instanceof \WP_Site ) {
			$path = str_ireplace( ltrim( untrailingslashit( $site->path ), '/' ), '', $path );
		}
	}

	return  trim( $path . $query );
},
```

justlevine (commented):
These changes should fix the underlying issue while preserving Faust's use of `term_link`. (I dont have permissions to update the PR branch directly).

src/Model/Term.php
```diff
@@ -192,15 +192,9 @@ protected function init() {
...
```
jasonbahl, line 194: @jeanfredrik for whatever reason, Github won't let me provide this suggestion inline, but after reviewing and hashing this out for a while, @josephfusco and I believe this should be updated to the following:

```php
'uri'                      => function () {
	$link = $this->link;

        // parse the link to ensure it's a valid url. 
	$maybe_url = wp_parse_url( $link );

	// If $maybe_url returns false, it means it isn't a valid URL, so we can assume
        // the link has been filtered upstream and we can return the link value.
	if ( false === $maybe_url ) {
		return $link;
	}

	// Replace the home_url in the link in order to return a relative uri.
	// For subdirectory multisites, this replaces the home_url which includes the subdirectory.
	return ! empty( $link ) ? str_ireplace( home_url(), '', $link ) : null;
},
```

jasonbahl (changes requested):

Issue #1679, opened 2021-01-12
RootQuery to ContentNode connection returns posts it shouldn't


jasonbahl:
The `RootQueryToContentNode` connection is returning posts of `post_types` that are not public. 

For example, with the [WPGatsby](https://github.com/gatsbyjs/wp-gatsby/) plugin active, the following query from a public user: 

```
{
  contentNodes {
    nodes {
      __typename
    }
  }
}
```

Is returning nodes of the `ActionMonitor` Type, a non-public post_type. 

The Model Layer for SURE needs to be fixed to ensure non-public posts are not being returned, even if an underlying query asks for them. 

We should also optimize the connection queries to only ask for things we know the user _can_ ask for. 

Some context that _might_ be helpful to resolving this: 

- The post types that are queried for are centrally checked in the Connection Resolver here: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Connection/PostObjectConnectionResolver.php#L56-L57. Limiting this centrally to public post types could help. 

- Here (https://github.com/wp-graphql/wp-graphql/blob/develop/src/Model/Post.php#L329-L331) the Post model is allowing posts of any post type to be public, as long as they're published

justlevine:
The temporary workaround is to `graphql_post_object_connection_query_args` to unset unwanted post types. E.g for `action_monitor`

```php
/**
 * Remove unwanted posts types from `contentNodes`.
 **/
add_filter(
  'graphql_post_object_connection_query_args', 
    function( array $query_args, $source, array $args, $context, $info) {
      if( 'contentNodes' === $info->fieldName ) ) {
        // You can make this more generic by using the WP_Post_Type's params to decide whether to unset()
        unset( $query_args['post_type']['action_monitor'] );
      }
      return $query_args;
  }, 
  10,
  5
);
```
![image](https://user-images.githubusercontent.com/29322304/165675551-25e5ef3c-969f-4b34-b9fa-fa3f01b46048.png)


> The Model Layer for SURE needs to be fixed to ensure non-public posts are not being returned, even if an underlying query asks for them.

For an actual fix in core, I think we need to decide what exactly determines if a post is `non_public` ( e.g. `exclude_from_search` ), which is likely depends on the context connection. Eg. `action_monitors` still needs to be available on `query { actionMonitors { ... } }`

jasonbahl:
@justlevine ya, I think my original intent of the `ContentNode` Interface was to have it apply to any `public` post type. 

[...]

PR #2353, opened 2022-04-26
Feat: Refactor get_allowed_post_types() and get_allowed_taxonomies()

justlevine:
What does this implement/fix? Explain your changes

This PR refactors `WPGraphQL::get_allowed_post_types()` and `WPGraphQL::get_allowed_taxonomies()` to correctly store data in the `WPGraphQL::$allowed_post_types` and `WPGraphQL::$allowed_taxonomies` class properties.

Additionally, it stores the `WP_Post_Type` and `WP_Taxonomy` objects, allowing the user to retrieve _either_.

This will ensure that the list of possible post types / taxonomies is the same throughout the schema, and will also let us significantly clean up the codebase by removing now-extraneous rehydration calls ( e.g. `get_post_type( \WPGraphQL::allowed_post_type( )[ $n ] )` ) in a future PR.

### Details

* The array of `WP_Post_Type`s and `WP_Taxonomies` are stored to their respective parameters _after_ any `...entities_allowed_{type}` filters have been applied, but before any args passed to the methods.
* `WPGraphQL::get_allowed_post_types()` now takes an `$output` argument _as the first parameter_ that lets a user decide if they need the `names` or the `objects, and relocates the`$args` to the second parameter. This is _backwards compatible_ and if an array is passed as the first arg, it will automatically reassign it.
* `WPGraphQL::get_allowed_taxonomies()` accepts those same two parameters. There's no need to worry about back-compat, since it currently doesnt accept any.
* Numerous calls to `get_post_types()` and `get_taxonomies()` have been replaced with the respective `get_allowed...()` methods. This fixes a handful of bugs where the list of "allowed" types was different throughout the plugin, which would break the schema (e.g. querying `contentTypes` after deregistering a core type ).

* The `WPGraphQL::$allowed_post_types` and `WPGraphQL::$allowed_taxonomies` properties are now `protected`, to enforce the static getter pattern.

Any other comments?
-------------------

* As mentioned, this PR intentionally doesnt refactor the _usage_ of these methods (e.g rehydrating the post object by first getting the names and then looping through them and calling `get_post_type()` ). That's a separate improvement deserving of its own PR.
* I also didnt change many of the calls in `NodeByUri`, since that's already a finicky piece of code, and ideally we'd first get #2342 and better test how using a cached list of post/tax types might affect filters on the stored query vars.
* Relatedly: the deprecation notice is currently commented out until we're no longer `_doing_it_wrong()` ourselves.
* Storing an array of _objects_ in `WPGraphQL::$allowed_post_types` and `WPGraphQL::$allowed_taxonomies` is _non-breaking_ since theyre still array types, and they weren't actually holding anything beforehand. That said, while changing those properties to `protected` is ideal and wouldn't affect any users, ~since it's technically a breaking change, I'll leave it to @jasonbahl to decide.~ (edit: added by Jason).
* The rationale behind the change in argument order for `WPGraphQL::get_allowed_post_types()` is to conform with PHP best practices, since it's significantly more likely to toggle between `names`/`objects` outputs than it is to pass args. Technically we don't even need to deprecate passing _just_ the `$args` as shorthand for `WPGraphQL::get_allowed_post_types( 'names', $args )` , and can just remove the commented out `_doing_it_wrong`. Again, I'll let @jasonbahl  decide 😎.

Where has this been tested?
---------------------------

**WordPress Version:** 5.9.3

### DIFF
--- a/src/Connection/Taxonomies.php
+++ b/src/Connection/Taxonomies.php
@@ -27,7 +27,7 @@ class Taxonomies {
 			]
 		);
 
-		$taxonomies = get_taxonomies( [ 'show_in_graphql' => true ], 'objects' );
+		$taxonomies = \WPGraphQL::get_allowed_taxonomies( 'objects' );
 
 		if ( is_array( $taxonomies ) && ! empty( $taxonomies ) ) {
 			foreach ( $taxonomies as $taxonomy ) {
--- a/src/Connection/TermObjects.php
+++ b/src/Connection/TermObjects.php
@@ -210,7 +210,7 @@ class TermObjects {
 						]
 					),
 					'resolve'        => function ( Post $post, $args, AppContext $context, ResolveInfo $info ) {
-						$taxonomies = get_taxonomies( [ 'show_in_graphql' => true ] );
+						$taxonomies = \WPGraphQL::get_allowed_taxonomies();
 						$terms      = wp_get_post_terms( $post->ID, $taxonomies, [ 'fields' => 'ids' ] );
 						if ( empty( $terms ) || is_wp_error( $terms ) ) {
 							return null;
--- a/src/Data/Connection/ContentTypeConnectionResolver.php
+++ b/src/Data/Connection/ContentTypeConnectionResolver.php
@@ -91,7 +91,7 @@ class ContentTypeConnectionResolver extends AbstractConnectionResolver {
 		}
 
 		$query_args = $this->get_query_args();
-		return array_values( get_post_types( $query_args ) );
+		return array_values( \WPGraphQL::get_allowed_post_types( 'names', $query_args ) );
 	}
 
 	/**
--- a/src/Data/Connection/PostObjectConnectionResolver.php
+++ b/src/Data/Connection/PostObjectConnectionResolver.php
@@ -53,7 +53,7 @@ class PostObjectConnectionResolver extends AbstractConnectionResolver {
 		if ( 'revision' === $post_type || 'attachment' === $post_type ) {
 			$this->post_type = $post_type;
 		} elseif ( 'any' === $post_type ) {
-			$post_types      = get_post_types( [ 'show_in_graphql' => true ] );
+			$post_types      = \WPGraphQL::get_allowed_post_types();
 			$this->post_type = ! empty( $post_types ) ? array_values( $post_types ) : [];
 		} else {
 			$post_type = is_array( $post_type ) ? $post_type : [ $post_type ];
--- a/src/Data/Connection/TermObjectConnectionResolver.php
+++ b/src/Data/Connection/TermObjectConnectionResolver.php
@@ -48,7 +48,7 @@ class TermObjectConnectionResolver extends AbstractConnectionResolver {
 		/**
 		 * Set the taxonomy for the $args
 		 */
-		$all_taxonomies = get_taxonomies( [ 'show_in_graphql' => true ] );
+		$all_taxonomies = \WPGraphQL::get_allowed_taxonomies();
 		$query_args     = [
 			'taxonomy' => ! empty( $this->taxonomy ) ? $this->taxonomy : $all_taxonomies,
 		];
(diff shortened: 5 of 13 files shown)


jasonbahl:
> significantly reducing the number of database calls from repeatedly calling get_post_types() and get_taxonomies() respectively

@justlevine where were you seeing significant calls to the database from these functions? Maybe when using something like Custom Post Type UI or something like that? Curious what was causing the extraneous calls to the DB

justlevine:
>>significantly reducing the number of database calls from repeatedly calling get_post_types() and get_taxonomies() respectively
>
>@justlevine where were you seeing significant calls to the database from these functions? Maybe when using something like Custom Post Type UI or something like that? Curious what was causing the extraneous calls to the DB.

Honestly not sure. Now that you mention it I'm remembering that the wp functions are supposed to hit the `globals` first, but they were still leaking through on complex queries. Maybe it's an issue with my local object caching...

The original reason I wrote this was to eventually be able to clean up all the messy rehydration loops caused by only storing the names, and then I noticed both the drop in hits and how enforcing the pattern preemptively fixes bugs like with `contentTypes` queries or null type errors when deregistering core types. Might have jumped to conclusions prematurely.

I'm way out of my league on writing performance benchmarks, but maybe that makes sense before replacing some of existing `get_post_types` and `get_taxonomies` calls. Is the script you used [to benchmark 1.6.0](https://github.com/wp-graphql/wp-graphql/releases/tag/v1.6.0) public?

jasonbahl:
Post Types and Taxonomies in core are registered in code, so there's no database queries at all for them. 

I think the database queries must be from plugins that register Custom Post Types / Taxonomies via a GUI, like CPT UI. . .which is very common, so I'm happy to support those cases. It seems like those plugins should also be more responsible with how they filter `get_post_types`, etc, if they are indeed causing new queries to occur on each request.

I was curious to see if we could "prove" the impact this has. I think this is a beneficial refactor, regardless though. 

> Is the script you used [to benchmark 1.6.0](https://github.com/wp-graphql/wp-graphql/releases/tag/v1.6.0) public?

You can find it here: 

https://github.com/wp-graphql/wp-graphql/issues/1873#issue-869378973

justlevine:
> I think the database queries must be from plugins that register Custom Post Types / Taxonomies via a GUI, like CPT UI. . .which is very common, so I'm happy to support those cases. It seems like those plugins should also be more responsible with how they filter get_post_types, etc, if they are indeed causing new queries to occur on each request.

Rank Math was the culprit. I was using a `class_exists()` instead of `is_plugin_active()`, so even though it was disabled  their code was still being loaded. Fixed that and those calls went away 😊. For safety I ran `install-test-env` on a new docker instance, and didnt notice any notable performance differences with/without the pr.

> ...ok, now that I think about it more, I don't think the benchmarks were a formal script. I believe it was me loading the Schema and recording the time in a spreadsheet 🤔 😆

Haha nvm then. I'ma see if theres anything I can do about those codeclimate smells, and let you know when this is ready for review.

justlevine:
I've got nothing. post types and taxonomies are _so similar_, but any abstractions I tried made it even harder to read (`$objects, $allowlist, `graphql_{$entity_type}_entities_allowed_{$wp_type}`.

I'm open to suggestions or just leaving it as-is.

jasonbahl:
@justlevine I updated the properties to be protected. I'm cool with that change. 

> closes https://github.com/wp-graphql/wp-graphql/issues/1679 possibly others.

Can you share how this closes that issue? 

I just ran the query from that issue, and I'm seeing the same results described in the issue:

CleanShot 2022-04-27 at 16 09 40

justlevine:
It's working fine here, lemme investigate.

image

justlevine:
Oops. Forgot to disable my other plugins just now, and like with the db calls it was leaking through from my dev env. Will abstract and post the temporary workaround there.

Since clearly I overpromised in the description, I'm going to go back and review everything from this pr.

justlevine:
Okay the PR description has been updated to describe what this _actually_ does and why 😅

Issue #2191, opened 2021-12-22
Date Archives don't resolve with `nodeByUri` queries


jasonbahl:
In WordPress, if you have dates setup in your permalinks, you can visit a url such as `/2021/12` and get an archive of posts from December 2021. 

Or you could do `/2021` and get an archive of posts from the year 2021. 

Or you could do `/2021/12/22` and get an archive of posts from December 22, 2021. 

In WPGraphQL, these paths passed to `nodeByUri` will return null. 

**WordPress core behavior:**

I can visit the url `/2013/03/15/` and see an archive of posts for March 15, 2013. 

Inspecting the DOM, I can see the body class outputting "archive date"

![Screen Shot 2021-12-22 at 11 14 58 AM](https://user-images.githubusercontent.com/1260765/147137214-4ba0847b-d932-4501-be01-030e5d8990c8.png)

**WPGraphQL Behavior:**

If I enter the same uri in WPGraphQL, I get null:

![Screen Shot 2021-12-22 at 11 16 08 AM](https://user-images.githubusercontent.com/1260765/147137329-4a38635b-79be-4075-9c10-956355471afa.png)

I would expect to get a `ContentType` node in response

justlevine:
Related: #2366

justlevine:
Related: #2491

Issue #2190, opened 2021-12-21
Post Format archives don't resolve with nodeByUri query


jasonbahl:
After enabling post formats using the snippet below:

```php
add_theme_support( 'post-formats', array( 'aside', 'gallery', 'link', 'image', 'quote', 'status', 'video', 'audio', 'chat' ) );
```

And creating a post assigned to the `video` format, like so: 

![Screen Shot 2021-12-21 at 2 05 41 PM](https://user-images.githubusercontent.com/1260765/146997633-26cfb9c3-93fe-42cb-8643-6a852c06e583.png)

I can visit the url `/type/video` and get an archive of posts of said post-format. 

![Screen Shot 2021-12-21 at 2 03 42 PM](https://user-images.githubusercontent.com/1260765/146997712-fda7ea42-b879-41db-8875-74ec4c1d2964.png)

We can see in the "body class" that WordPress is resolving this to the archive of the post format. 

However, if we take the uri and use it in a `nodeByUri` query, we get a `null` response. 

**query**:

```graphql
query GetNodeByUri($uri: String!) {
  node: nodeByUri(uri: $uri) {
    ...NodeByUri
  }
}

fragment NodeByUri on UniformResourceIdentifiable {
  __typename
  uri
  id
  ...DatabaseIdentifier
  ...ContentType
  ...User
  ...TermNode
  ...ContentNode
  ...MediaItem
}

fragment DatabaseIdentifier on DatabaseIdentifier {
  databaseId
}

fragment MediaItem on MediaItem {
  id
  mimeType
}

fragment ContentType on ContentType {
  name
  isFrontPage
  isPostsPage
}

fragment TermNode on TermNode {
  isTermNode
  slug
  taxonomyName
}

fragment ContentNode on ContentNode {
  isContentNode
  slug
  contentType {
    node {
      name
    }
  }
  template {
    templateName
  }
}

fragment User on User {
  nicename
  databaseId
}
```

**variables**: 

```
{
  "uri": "/type/video"
}
```

## Result

The result is null. WPGraphQL doesn't properly return the PostFormat node as expected. 

![Screen Shot 2021-12-21 at 2 03 22 PM](https://user-images.githubusercontent.com/1260765/146998124-0802795b-bd0d-44e1-9df0-0f2034389483.png)

## Expectation

I expect to get a node of the `PostFormat` type in response to the query. 

Similar to if I queried a category by URI, like so:

[...]

PR #2355, opened 2022-04-28
Bugfix/#1679 ensure content nodes are public post types

### LINKED ISSUE #1679: RootQuery to ContentNode connection returns posts it shouldn't

jasonbahl:
This updates the ContentNode type to represent public post types, instead of _all_ post types. 

The ContentNode Interface was intended to be shared by public post types. The `ContentNode` Interface implements the `UniformResourceIdentifiable` Interface, which is an Interface for public facing nodes that can be identified by a URI. 

Non Public post types should not be considered Content Nodes as they don't have a public URI. 

The following are examples of Post Types in WordPress core that are not public, don't have a URI, and should _not_ be considered a `ContentNode` in WPGraphQL:

- [custom_css](https://github.com/WordPress/WordPress/blob/c6dbcf85766063b20aeb23ca207cc476818cdda2/wp-includes/post.php#L174) 
- [customize_changeset](https://github.com/WordPress/WordPress/blob/c6dbcf85766063b20aeb23ca207cc476818cdda2/wp-includes/post.php#L215)
- [oembed_cache](https://github.com/WordPress/WordPress/blob/c6dbcf85766063b20aeb23ca207cc476818cdda2/wp-includes/post.php#L252)
- [user_request](https://github.com/WordPress/WordPress/blob/c6dbcf85766063b20aeb23ca207cc476818cdda2/wp-includes/post.php#L264)

- closes #1679 
- closes https://github.com/wp-graphql/wp-graphql-woocommerce/issues/362#issuecomment-733103202

Any other comments?
-------------------

## Before

Querying a list of content nodes exposed "utility" post type content, such as the `ActionMonitorAction` type in the WPGatsby plugin. 

The action monitor type is intended to be a Publicly Queryable type, but is not publicly identifiable by URI, and therefore should not be considered a `ContentNode` 

CleanShot 2022-04-28 at 12 08 42

## After

After this PR, the `ActionMonitorAction` post type is not returned when querying for `contentNodes`

CleanShot 2022-04-28 at 12 10 35

## BREAKING CHANGE?

This can be a breaking change for some users. It's not a technical breaking change to the Schema, but it could impact users that were querying non-public post types and getting that data back in the context of `ContentNode` queries. So, while it doesn't. 

It's possible, via filters, to add the functionality back (allow other Types to be treated as Content Nodes) should that be very important for specific users.


justlevine:
Filter for reverting. 

Needs testing that it catches all use cases, since we're we cant directly overwrite the constructor param passed to the resolver, only overwrite the `post_type` query arg that the `PostObjectConnectionResolver` set. (Are there other where args besides `contentTypes` that influence which `post_type` is set?)

```php
/**
 * Include non-public post types in `contentNode
 **/
add_filter(
  'graphql_post_object_connection_query_args', 
  function( array $query_args, $source, array $args, $context, $info) {
    if( 'contentNodes' === $info->fieldName ) && empty( $args['where']['contentTypes'] ) {
      $query_args['post_type'] = \WPGraphQL::get_allowed_post_types();
    }

    return $query_args;
  }, 
  10,
  5
);

**Comment by justlevine** [comment]
Should this PR do the same for termNodes? Currently they all return terms from _all_ taxonomies, and not just from public terms.

**Comment by jasonbahl** [comment]
> Should this PR do the same for termNodes? Currently they return terms from all taxonomies, and not just from public terms.

Yes, I think now would be a good time to address both. Good call out. I'll follow up, and also add some tests for filtering back to old behavior, that way we can document to folks how they can get the old behavior if they feel like it was a feature, not a bug.

**Comment by jasonbahl** [comment]
@justlevine I think we'll need to come up with a name for these other types of Post Types and Taxonomies. . .like core WP has private taxonomies such as:

- wp_theme: https://github.com/WordPress/WordPress/blob/74dcee85d065fe9f700f3459c44b845aba98f01a/wp-includes/taxonomy.php#L192
- wp_template_part_area: https://github.com/WordPress/WordPress/blob/74dcee85d065fe9f700f3459c44b845aba98f01a/wp-includes/taxonomy.php#L211
- nav_menu: https://github.com/WordPress/WordPress/blob/74dcee85d065fe9f700f3459c44b845aba98f01a/wp-includes/taxonomy.php#L113

For menus (and menu_items), we leave them out and provide a custom way to interact with menus in the Schema, independent from the mapping of post types/taxonomies to the Schema. 

For these other ones, right now WPGraphQL doesn't set these to `show_in_graphql => true`, but of course users could filter them to `show_in_graphql => true`, and we need ensure they're treated properly in that case. 

I do think that if they are private, they shouldn't be exposed in most of the connections we have today, but I do think we should consider how they should be exposed, as the data is likely important, even if only to authenticated parties. 

I'd like to find a name that is declarative of what these things are. In my mind I consider them `Utility Post Types` and `Utility Taxonomies` but I don't think those, (specifically "Post Types") translate well to a GraphQL Schema. Taxonomies and Terms are more widely used words in technology, but "Post Types" don't really describe what these things are, unless you're a WordPress developer, but a lot of folks interacting with a GraphQL Schema are not WordPress developers and I've tried (very imperfectly) to reduce the amount of WordPress-isms in the Schema. (i.e. Post Type => ContentNode, post_tag => Tag, etc). 

Anyway, what I think we need to figure out, is GraphQL Type and Field names for:

- Public Post Types: ContentType / ContentNode
- Private Post Types: ? / ?

- Public Taxonomies: Taxonomy / TermNode
- Private Taxonomies: PrivateTaxonomy? UtilityTaxonomy? / PrivateTermNode? UtilityTermNode?

----

In my experience, Taxonomies and Terms are used almost the same whether they're private or public, the difference usually being public ones have archive pages and private ones don't, but they're used to group things together. 

Post Types have wider range of utility, in my experience. 

Public Post Types are typically for public facing content, but private post types are all over the place. Contact Forms, Redirects, Reusable Gutenberg Blocks, CSS for Customizer, Changesets for Customizer, Revisions (of any other post type),  Oembed Cache, User Request (for GDPR, I believe?), etc. 

The use of Post Types is soooo wide. 

---

Perhaps, we simply don't expose private post types and taxonomies at all, even if they are set to `show_in_graphql` and even if there are use cases where admin users would want to query the data? 

The assumption being that a post type or taxonomy marked private has a "different" use than standard public post types, and they should be mapped to the Schema in a more specific way (like we do with Nav Menus / Nav Menu Items) 🤔

**Comment by jasonbahl** [comment]
> I'm still playing catch-up with GraphQL in non WordPress contexts, but so far my understanding is interfaces are best used to describe the shape of a type, not the context it's used in. A ContentNode isn't just a Node thats public, it's a generic name for types that are shaped like WP_Post_Type

@justlevine most of the utility post types I described store data the way post types store data, but are used and shaped very differently. 

For example most of them don't have a uri, which is a field declared by the ContentNode interface. And many are intended to be used in vastly different ways (like redirects, oembed cache, etc). Their shape and utility are very different from other Post Types that are used for "content"/

Also, the Schema will show all Types that implement that interface in the docs, which is a mismatch of behavior. 

Seeing that you could query `{ contentNodes { nodes { ...someFields } } }` and the Schema showing that you _might_ get a `CssChangeset` or an `OembedCache` is misleading to the client developer working with the Schema. 

If the fields returning `ContentNode` are returning only public post types, then the user looking at the Schema should be presented with an Interface that only Public Types implement. 

I think there's benefit to having _another_ Interface (not named ContentNode) that would spread across _all_ PostTypes, public or private. I don't know what to name this thing, and I don't like `PostType` or `PostNode` as a name. 😆 

I do think there are use cases where folks want to be able to query across literally any post type, much like you might do:

```php
new WP_Query( [ 
  'post_type' => 'any' 
]);
```

But querying for `contentNodes` the expectation is that you're doing: 

```
new WP_Query( [ 
  'post_type' => 'public_post_types' 
]);
```

The Schema should reflect that. The `ContentNode` interface should only represent `Public` post types that users can expect as possible return Types of a `contentNodes` query

jasonbahl:
@justlevine I updated the PR to only apply the `TermNode` Interface to public taxonomy types. 

ex: 

`type Tag implements TermNode`

but, if the taxonomy `wp_template_part_area` (see: https://github.com/WordPress/WordPress/blob/74dcee85d065fe9f700f3459c44b845aba98f01a/wp-includes/taxonomy.php#L208) were filtered to `show_in_graphql` it would _not_ implement the `TermNode` interface. 

The downside to this, now, is that if someone were to filter the post type `wp_template_part` and `wp_template_part_area` to both `show_in_graphql` there would be no clear way to query the terms associated with the template part 🤔 

For example:

```php
add_filter( 'register_post_type_args', function( $args, $post_type ) {

	if ( 'wp_template_part' === $post_type ) {
		$args['show_in_graphql'] = true;
		$args['graphql_single_name'] = 'TemplatePart';
		$args['graphql_plural_name'] = 'TemplateParts';
	}

	return $args;

}, 10, 2 );

add_filter( 'register_taxonomy_args', function( $args, $taxonomy ) {

	if ( 'wp_template_part_area' === $taxonomy ) {
		$args['show_in_graphql'] = true;
		$args['graphql_single_name'] = 'TemplatePartArea';
		$args['graphql_plural_name'] = 'TemplatePartAreas';
	}

	return $args;

}, 10, 2 );
```

This adds the private post type `template_part` and private taxonomy `template_part_area` to the Schema. 

But now the `TemplatePartArea` doesn't implement the `TermNode` Interface, so this Type is missing fields like `description` that are added by that interface. 

On the upside, terms like `Category` have connections to public things like `enqueuedStylesheets`, but the private taxonomy `TemplatePartArea` does not have connections to `enqueuedStylesheets` (because that connection is a `TermNode->EnqueuedStylesheet` connection and now a private Taxonomy doesn't implement the TermNode interface) 

I'm confusing myself right now and going to take a break from this and come back to it after the weekend. 

Ultimately, I think we need to resolve what I was saying above: 


What I think we need to figure out, is GraphQL Type and Field names for:

- **Public Post Types**: ContentType / ContentNode
- **Private Post Types**: ? / ?
- **All Post Types**: ? / ?

- **Public Taxonomies**: PublicTaxonomy? / PublicTermNode?
- **Private Taxonomies**: PrivateTaxonomy? UtilityTaxonomy? / PrivateTermNode? UtilityTermNode?
- **All Taxonomies**: Taxonomy / TermNode

justlevine:
[Written before your last commit/comment. Reviewing that now]
@jasonbahl thanks for the clarification

> Seeing that you could query { contentNodes { nodes { ...someFields } } } and the Schema showing that you might get a CssChangeset or an OembedCache is misleading to the client developer working with the Schema.

Those types aren't publicly queryable either. We can both agree that at a bare miniminum the client developer is only going to see items that [are supposed to be viewable](https://github.com/WordPress/wordpress-develop/blob/1cffa3f82b2de2df13e89767a4714de6fbe3de78/src/wp-includes/post.php#L2253)

> If the fields returning ContentNode are returning only public post types, then the user looking at the Schema should be presented with an Interface that only Public Types implement.
>
> I think there's benefit to having another Interface (not named ContentNode) that would spread across all PostTypes, public or private. I don't know what to name this thing, and I don't like PostType or PostNode as a name. 😆

I guess I'm wondering is whether the expectation from the `ContentNode` _interface_ is/should be "Public post type objects", and not as a GraphQL equivalent to `new WP_Query( $args_from_where_args )`, (still only including items where `is_post_type_viewable()` is true). 

This rises from the parallel to `RootQueryToTermNodeConnection` which is "clearly" (I say from a users perspective) intended as a counterpart to WP_Term_Query.

There's reason to assume `ContentNode` is the same as `TermNode` but with a less confusing name than `PostObjectNode`. This is strengthened by the fact that `ContentType` _is_ the equivalent of `WP_Post_Type`. 

If  that's not/shouldnt be the expectation, then yeah 100% a new interface/ rootQuery connection is necessary to handle all post type objects types like `ActionMonitorAction` that are defined to be publicly queryable, but arent strictly "Content". However, that means we'd need to rename `ContentType`.

E.g. (names for illustration only. We both know how bad I am at naming things 😎):

```graphql
 query RootQuery {
   taxonomies { # all publicly_queryable taxonomies. Works like get_taxonomies().
     __typename # Taxonomy
   }
  terms{ # the terms for all publicly_queryable taxonomies. Works like new WP_Term_Query().
     __typename # Term
   }
   postTypes{ # all publicly_queryable post types. Works like get_post_types()
     __typename # PostType
   }
   postObjectNodes { # the post objects for all publicly_queryable post types. Works like new WP_Query()
     __typename # PostObjectNode
   }
   contentNodes { # the post objects for all public post types. Essentially a filtered connection to postObjectNodes
     __typename # Maybe it needs a new interface, maybe its just a PostObjectNode.
   }
 }
```
> In my experience, Taxonomies and Terms are used almost the same whether they're private or public, the difference usually being public ones have archive pages and private ones don't, but they're used to group things together.

For this and the above, it definitely feels that that `RootQueryToTermNodeConnection` definitely **should not** only be limited public terms, but if we want to set good dx for most use cases, then we can set the default included taxonomies to only those that are public..

jasonbahl:
(Brain dumping from my phone, so formatting is going to be weird)

Here's an idea:


public post type: ContentType / ContentNode

private post type: PrivateModel / PrivateModelType

all post types: Model / ModelType


🤔 

interface Model implements Node {
  id: ID!
}

interface ContentNode implements Model, UniformReaourceIdentifiable { 
  id: ID!
  uri: String
}

type Post implements ContentNode { ... }

type Redirect implements Model { ... }

Then a query for contentNodes returns content of public post types, and the schema self documents public post types as the possible types of the ContentNode interface. 

A query for "models" would return nodes of _any_ post type 



A

justlevine:
Agree with the issues you raise, and that this is essentially a naming things issue more than anything else. 

>What I think we need to figure out, is GraphQL Type and Field names for:
>
>Public Post Types: ContentType / ContentNode
>
>Private Post Types: ? / ?
>
>All Post Types: ? / ?
>
>Public Taxonomies: PublicTaxonomy? / PublicTermNode?
>
>Private Taxonomies: PrivateTaxonomy? UtilityTaxonomy? / PrivateTermNode? UtilityTermNode?
>
>All Taxonomies: Taxonomy / TermNode

I'm still not 💯convinced that that the interfaces should be based on an attribute _value_ instead of the possible existence of that attribute.

In that model, we still have `Taxonomy` / `TermNode`, but public and private (and for good measure graphql equivalents to `show_ui` , `exclude_from_search` etc) terms are retrieved with where args. 

Separately, A `NodeWithArchive` interface could be created that represents 'public' taxonomies (in this context) i.e. those that have an `archivePath` and maybe even a connection to the object at the end of that path (would be useful for avoiding all the headaches with trying to show archives on the frontend by hardcoding routes / using `nodeByUri`  e.g. #2191 , #2190, #1910 ).

`ContentNode` still exists, but its treated like `NodeWithAuthor` or `UniformResourceIdentifiable`, i.e. as a WP_Post that has Content related fields. **[Currently]( https://github.com/wp-graphql/wp-graphql/blob/369b9f0725b29d070db80206951fd1a480ade2b0/src/Type/ObjectType/PostObject.php#L28) ,all `PostObject` types are a `ContentNode` and a `NodeWithTemplate`, but that just isnt true to real-life**, or rather it isn't true if you're defining a `ContentNode` as a public post typ.e

I dont see a shape-based reason for `ContentType` to exist separately from whatever the schema's equivalent to `WP_Post_Type` is called.

(Even with this approach, we can still have `RootQuery.publicTerms` or `RootQuery.publicPostObjects`, or whatever else we want to represent data contexts over data shapes.)

justlevine:
> (Brain dumping from my phone, so formatting is going to be weird)
> 
> Here's an idea:
> 
> public post type: ContentType / ContentNode
> 
> private post type: PrivateModel / PrivateModelType
> 
> all post types: Model / ModelType
> 
> 🤔

I think that obfuscates things even more than just just calling the all types `PostObjectType` and `PostObjectNode`, even with the existence of the `Post` object type. Why would `Model` be associated specifically with post objects, instead of anything else that can be modeled. We then also need a clear way to differentiate between the abstract php `Model` and the `Post extends Model` that is reflected in the the schema as just `Model`.

justlevine:
Regarding using `ContentType` or `ContentNode` as either public-facing post types _or_ post types with a specific subset of fields, it seems WordPress core disagrees

> There are many different types of content in WordPress. These content types are normally described as Post Types, which may be a little confusing since it refers to all different types of content in WordPress. For example, a post is a specific Post Type, and so is a page.

[source](https://developer.wordpress.org/themes/basics/post-types/)

Probably (along with the `PostObject` usage of `ContentNode` as an interface) where my confusion stems from.

jasonbahl:
Ya. Perhaps the answer is just use WordPress terminology, even if it's not ideal. 43% of the web has been built on it already! If you don't know what a PostType or PostObject is, and your first time using WordPress data is via a GraphQL Schema, perhaps you will learn what it is! 😆 

Might be easier to just call it something familiar even if not "perfect"

jasonbahl:
Ok so thinking more about this and I might be redundant but here goes:

Public Post types have a uri and should be ContentNodes. The ContentNode interface describes nodes that have a Uri. They're nodes of published content. 

While they're often different, it's common to need to fetch them together. 

For example search queries, nav menus, etc. 

We need to be able to get `contentNodes` of various types in a single request and can use `...on Post`, `...on Page` etc to differentiate the fields we need for each. 

Private Post Types are very wide in their utility. And, in my experience, it's rare that different private types are needed at once. For example I've never personally needed to query for both oembed_cache, redirects and customizer changesets at one time. 

So, for public post types, I think it's important to have connection queries that can return ContentNode interface (any public post type)

For private post types, it's still important to expose them in the Schema, but not as part of the ContentNode interface. 

Redirects, oembed_cache, etc can be queried individually or in a connection of their respective type, and they can be marked publicly queryable as well, but they'll be treated more so like independent Types as a pose to a shared grouping of Types. 

There should still be an interface to describe these private types. Perhaps `UtilityNode` / `UtilityType`, but probably not a beee to expose a root connection for this interface. 

All Post Types could be `DataNode` / `DataType`

So, we could have:

```
# A data entry node (a node stored in the posts table. i.e. get_post() )

interface DataNode implements Node {
  id: ID!
}
```

```
# The type definition for a set of data ( i.e. get_post_type_object() )

interface DataType implements Node {
  id: ID!
}
```

```
# The type definition of a public facing data type (get_post_type_object() for public post type)

interface ContentType implements DataType, UniformResourceIdentifiable {
  id: ID!
  uri: String!
}
```

```
# Public facing Data node. i.e. get_post() of a public post type

interface ContentNode implements DataNode, UniformResourceIdentifiable {
  id:ID!
  uri: String!
}
```

```
# Node with specific utility (get_post() of a private post_type()). Can be publicly queryable, just not identifiable by a public URI (i.e. ActionMonitor)

interface UtilityNode implements DataNode {
  id: ID!
}
```

```
# The type definition of a non-public Data Type (get_post_type_object() for a private post type). Can be publicly queryable, just not identifiable by a public URI 

interface UtilityType implements DataType {
  id: ID!
}
```


Then, we can query:

```
dataTypes {
  nodes {
    __typename
  }
}
```

And get: 

- OembedCache
- UserRequest
- CustomizerChangeset
- Post
- Page
(any other post type, public or not)

etc


```
contentTypes {
  nodes {
    __typename
  }
}
```

would just be:

- Page
- Post
- House
- Car
- Product
- etc
(public post types) 

```
utilityTypes {
  nodes {
    __typename
  }
}
```

would be:

- ActionMonitor
- OembedCache
- CustomizerChangeset
- etc
(non-public post types)

justlevine:
> Public Post types have a uri and should be ContentNodes.
> ...
> All Post Types could be DataNode / DataType

The more we discuss this and I think about it the more I disagree ([most of twitter too](https://twitter.com/jasonbahl/status/1520185663429087233?s=21&t=zUb1AkrQAaZdZfGQY4KyLQ)  ). I think Public posts types should be ContentNodes _that_ have a URI, and Private post types to be ContentNodes without one.  At this point, I'm deferring to you on the topic 😎.

That said, if we do need a generic name to describe a `WP_Post` (instead of sticking with `ContentNode` and having a `PublicContentNodes` for those with a URI ), it needs to be specific enough to differentiate itself between other modeled data types (something no one on twitter mentioned). I definitely don't think `DataNode` cuts it... sounds more like a union of _all_ data models, and there's nothing that inherently indicates that a `DataNode` is only the `WP_Post` subset of possible data nodes.

PS: this is the first link when I google [WordPress data types](https://code.tutsplus.com/tutorials/understanding-and-working-with-data-in-wordpress--cms-20567). I think you're right with [your previous comment](https://github.com/wp-graphql/wp-graphql/pull/2355#issuecomment-1113900042), and if we're going to go against the grain and use `ContentType`/`ContentNode` for only public-facing types, we should stick with possibly-foreign-but-easy-to-decipher `PostObjectType`/`PostObjectNode`.

> Private Post Types are very wide in their utility. And, in my experience, it's rare that different private types are needed at once. For example I've never personally needed to query for both oembed_cache, redirects and customizer changesets at one time.

This is a very good point. 
Abstracting this away from any specific implementation of `ContentNode` it that for means for good DX:
* We should have a connection from `RootQuery` to a list of all public post objects.
* We should have a connection from `RootQuery` to lists of the individual post objects by type, for both public and private objects.
* We should have a connection from `RootQuery` to a list of all post types, with a way to filter them to only get the private ones. It seems unnecessary to to have a separate connection specifically to private post types, since it's both a rare usage and there is nothing unique about the object shape (explained below).
* _Maybe_ we should have a connection from `RootQuery` to a list of all public post types, but due to the shape (explained below), we might be equally served by having the `RootQuery` => all post types, default filtered to `public`, no different than how connections to individual lists of post objects default to `published`.

> There should still be an interface to describe these private types. Perhaps UtilityNode / UtilityType, but probably not a beee to expose a root connection for this interface.
> ...
> ```
># Node with specific utility (get_post() of a private post_type()). Can be publicly queryable, just not identifiable by a public URI (i.e. ActionMonitor)
>
>interface UtilityNode implements DataNode {
>  id: ID!
>}
> ```

I disagree that a separate interface type should be created and the reason is shown from your definition. There's no difference between a `UtilityNode` and a `DataNode`, so why clutter up the schema with an alias. 
Similarly for `interface ContentType implements DataType` - currently there is no difference, but if/when we want to to better support archives, itl probably be better served by a `NodeWithArchive` interface that can be applied across types the same way `UniformResourceIdentifiable` codes.

----
A query following the above would look like (ignoring the names used. Aliases can theoretically be their own field on `RootQuery`, im just illustrating the underlying logic from an interface POV ):

```graphql
query {
  postObjectTypes {
    nodes {
      __typename # all \WP_Post_Types
      ... on NodeWithArchive { # theoretical inteface
        archiveSlug
      }
    }
  }

  contentTypes: postObjectTypes( where: { public: true } ) {
     nodes {
      __typename # all public \WP_Post_Types
     # interface fragment only if `ContentType` isnt its own interface.
     }
  }

  privatePostObjectTypes: postObjectTypes( where: { public: false } ) {
    nodes {
      __typename # all private \WP_Post_Types
    }
  }

  # same pattern for taxonomies.
  
  postObjectNodes {
    nodes {
      __typename # all \WP_Posts
      ... on UniformResourceIdentifiable { # or on ContentNode. This is the only time we need an interface to describe the object shape.
        __typename # public \WP_Posts
        uri
      }
    }
  }
  
  contentNodes: postObjectNodes( where: { public: true } ) {
    nodes {
      __typename # public \WP_Posts
      # interface fragment only if `ContentNode` isnt its own interface.
    }
  }
  
  privatePostObjectNodes: postObjectNodes( where: {public: false} ) {
    nodes {
      __typename # private \WP_Posts
    }
  }
  
  # Same pattern works for taxonomies
}
```

jasonbahl:
Ok, so I might have been too brief in my pseudo type definitions and explanations on why I think there needs to be a difference between Public, Private and All post types. 

In WordPress, post types that are public have a uri. Post types that are not public, do not. 

The `ContentNode` Interface, which implements the `UniformResourceIdentifiable` Interface, was intended to be implemented on public post_types, and therefore the query: 

```
{
  contentNodes {
    nodes {
       id
       uri
    }
  }
}
```

Would return a list of nodes of any public facing post type. 

This query should _not_ return Types that do not implement the `UniformResourceIdentifiable` Interface (non-public post types). 

The `ContentNode` Interface should only show public facing post types as Possible Types when doing an Introspection query, (and when looking at docs in GraphiQL, etc) because public post types are the only Types a client application should expect from doing a `contentNodes` or `contentNode(id: $id)` query. 

It would be misleading and inaccurate for Schema Documentation to show "Private Post Type A" and "Private Post Type B" as Possible Types for `contentNodes` queries if the intent is to only ever return `public` post types from this query. 

For example, on the current `develop` branch, and with WPGatsby active, I can see the `ActionMonitorAction` Type as a `PossibleType` of the `ContentNode` Interface:

CleanShot 2022-05-02 at 10 10 32

I can then query for a list of `ContentNode` nodes like so:

```graphql
{
  contentNodes {
    nodes {
      __typename
      id
      uri
    }
  }
}
```

And, in response, I get a payload that includes action monitor nodes. 

```json
{
  "data": {
    "contentNodes": {
      "nodes": [
        {
          "__typename": "ActionMonitorAction",
          "id": "cG9zdDo2MDI=",
          "uri": "/action_monitor/jasonbahl-1651266947/"
        },
        {
          "__typename": "ActionMonitorAction",
          "id": "cG9zdDo2MDA=",
          "uri": "/action_monitor/update-setting-active_plugins-1651252079/"
        },
        {
          "__typename": "ActionMonitorAction",
          "id": "cG9zdDo1OTg=",
          "uri": "/action_monitor/hello-world-1651097319/"
        },
        {
          "__typename": "TestModel",
          "id": "cG9zdDo1NDI=",
          "uri": "/testmodel/542/"
        },
        {
          "__typename": "Page",
          "id": "cG9zdDo1MjI=",
          "uri": "/my-account/"
        },
        {
          "__typename": "Page",
          "id": "cG9zdDo1MjE=",
          "uri": "/checkout/"
        },
```

We can see that we're even getting a URI for the action monitor nodes, but, we really shouldn't because WordPress itself doesn't respect URI for private post types. 

If I visit the uri of one of the action montitor nodes, such as: `http://pq.local/action_monitor/jasonbahl-1651266947/`

I get a 404. 

CleanShot 2022-05-02 at 10 14 35

This is a bug, and why this PR (and issue #1679) exist. 

The `ContentNode` interface implements the `UniformResourceIdentifiable` Interface because the original intent was to allow client applications to query across public post types.

I believe to fix the bug, we need to move the application of the `ContentNode` Interface to only be applied to GraphQL Types that represent public post types, like we do with the `UniformResourceIdentifiable` Interface here: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/ObjectType/PostObject.php#L28-L32 

If we do that, then the `ContentNode` type will appropriately only show public post types as possible types: 

CleanShot 2022-05-02 at 10 22 16

With the suggested change in place (applying the `ContentNode` interface only to public post types) I now only see public post types as the Possible Types of the ContentNode.

And now, if I run the same query to get content nodes, I appropriately only get public facing nodes that have a URI. No more nodes that will give a uri that will 404. 

CleanShot 2022-05-02 at 10 25 11

So now, as a client developer looking at the Schema, I have accurate Schema documentation. I can accurately see that the possible Types of a ContentNode and know that when I execute a `contentNodes` query, it will return me nodes of public post types. I know that querying for `contentNodes` won't return me other Types that it shouldn't, such as `ActionMonitorAction` or `CustomizerChangeset`, etc. 

To expand further, let's take one of the WordPress core Post Types and filter it to show in GraphQL. 

```php
add_filter( 'register_post_type_args', function( $args, $post_type ) {
	
	if ( 'custom_css' === $post_type ) {
		$args['show_in_graphql'] = true;
		$args['graphql_single_name'] = 'CustomCssNode';
		$args['graphql_plural_name'] = 'CustomCssNodes';
	}

	return $args;

}, 10, 2 );
```

This takes the core WordPress `custom_css` post type and exposes it to WPGraphQL. 

This post type is private (no uri). It stores data for use in WordPress but is not intended to be queried the same way other "content" is. 

While it's technically a post_type, it's use is very different than public post types that are intended to be visited via URI, mapped to a template, etc. 

This type of data is `utility` data, or `config` data, perhaps. It's important to be able to be queried via WPGraphQL. Developers will want access to this data (just like they want access to action monitor actions, etc), but it shouldn't be confused as being the same thing as other public post types. 

With this snippet in place, though, if I go make some CSS changes using the customizer

CleanShot 2022-05-02 at 10 37 23

Then all of a sudden I'm getting custom CSS nodes in my `contentNodes` queries 😱 

CleanShot 2022-05-02 at 10 43 29

This isn't right! 

Should the Custom CSS be queryable publicly? Sure! It's needed to change the background color (at least in my case). But, it shouldn't be considered a "ContentNode" because it doesn't _actually_ have a public facing uri. 

We can see a uri returned in the response, but if I visit that URI, we see that WordPress doesn't actually respect it, it 404s (but you can see that my CSS change was respected. . .so the data was used publicly, just not as a "content node". 

So, again this Type of "post_type" should be publicly queryable, but NOT treated as other public post types intended for delivering content via URI.

I think we need a way to name things that are public post types (current naming convention is `ContentNode`). 

If we fix this bug, as this PR does (or close to it), then we'd be able to query public post types using the `contentNodes` and `contentNode` query. But, we wouldn't have a way to query across non-public types, or all post_types (public or private). 

I think for public applications, the most common entry point for things like this would be `contentNodes` entry point. 

Like, if I wanted a search form to search across my post types, I could do something like: 

```graphql
query SearchContentNodes( $keyword: String ) { 
  contentNodes( where: { search: $keyword } ) { 
     __typename
     id
     uri
     ...on NodeWithTitle {
        title
     }
     ...on NodeWithAuthor {
       author {
         node {
            id
            name
            avatar {
               url
            }
         }
       }
     }
  } 
}
```

I would expect to only get public facing content (I don't want action monitor and Custom CSS in my results). 

But!

If I was building an admin application, I might want a way to make similar queries to this, but query across private types, or All types. 

Much like if I were doing `new WP_Query( [ 'post_type' => 'all' ]);`

For this to work, we need an interface that can be applied to `all` post types, not just public post types. 

We could name this something WordPress-y like `PostObjectNode` or something to that tune, but I feel like that requires some domain knowledge of the client developer. When I browse the Github GraphQL Schema, I don't feel like I'm required to know what system it's built on. The Types are descriptive of what the things actually are, not the system it's built on. I don't feel that `post_type` or `post_type_object` describe what content stored in the `*_posts` table actually is. 

This is what I'm trying to figure out naming for.

What do we call these nodes of data that exist in the posts table, but aren't used for publishing content. They're used for other purposes, but sometimes, as a developer, it's still nice to be able to access them together. 

This is why my mind goes to something like `UtilityNode`. 

```
Utility

adjective
1.
useful, especially through being able to perform several functions.
"a utility truck"

2.
functional rather than attractive.
"utility clothing"
```

`UtilityNode`: Functional, rather than being attractive? 

These nodes provide utility, but they're not content made to be viewed directly by users, like other "content nodes" are. 

Another option could be to let `ContentNode` be the name for _all_ post types. We could argue that Custom CSS, redirects, etc are all content of some form, even if not the type of content you think you would render as "pages" on a website. 

The snippet of CSS I saved is a type of content, still.

So to make the distinction between _all_ post types (contentNodes) and public/private ones, perhaps we simply name them:

- **ContentNode** = posts of _any_ post type (this would no longer implement the `UniformResourceIdentifiable` Interface if we go this route, and would be applied to _all_ post types regardless of public/private status)
- **PublicContentNode** = posts of any public post type (this would now implement the `UniformResourceIdentifiable` Interface, and this interface would be applied to any post type that is public)
- **PrivateContentNode** = posts of any private post type (this would not implement the `UniformResourceIdentifiable` interface, and this interface would be applied to any post type that is NOT public)

Then, users could query for `{contentNodes{ ... }}` and get posts of any type, including Action Monitor Actions and Custom CSS. 

But, users that actually want _only_ public content, could query for: `{ publicContentNodes { ... } }` and get _just_ posts of public post types. 

And users that want to query _only_ private nodes could query for: `{ privateContentNodes { ... }}` and get posts of private post types, such as Custom CSS, action monitor actions, etc.

jasonbahl:
One reason I don't like the `PublicContentNode` vs `PrivateContentNode` naming convention, is that it feels misleading. 

You can have a private post type that's actually publicly queryable, but not intended to be represented as "content". 

Like the `CustomCSS` example. It's publicly queryable, as the client application needs to get it and use it, but it's not considered "content" in the same way a post or page is. 

If it's publicly queryable, but called "PrivateContentNode" that's super weird to me. 

Something like `UtilityNode` doesn't describe if it's public/private, just describes that the thing is `Functional, rather than being attractive`. 

There can be public or private utility nodes, still. Some will require auth, like the `user_request` post_type is both private and not publicly queryable. But Custom CSS is private, but publicly queryable, as is the WPGatsby Action Monitor (private, but publicly queryable).

jasonbahl:
I think this is where my mind is at currently:

`ContentNode`: public facing published content (`posts`, `pages`, `cars`, `houses`, etc) (posts of public post types)

`UtilityNode`: nodes of data stored in the posts table, but not intended to be public facing "content" (post types like `custom_css`, `oembed_cache` , etc) (posts of private post types)

`DataNode`: anything stored in the wp_posts table, public or private. So a `Post`, `Page`, `Car`, `House`, `CustomCss`, `oEmbedCache`, etc would all be considered a DataNode

All ContentNodes are DataNodes, but not all DataNodes are Content Nodes.

All Utility Nodes are DataNodes, but not all DataNodes are Utility Nodes.

Utility Nodes are not Content Nodes, and Utility Nodes _do not_ have a URI (WordPress 404s if you visit their permalink, for example. . .WPGraphQL shouldn't even have a uri field for these nodes)

Content Nodes are not Utility Nodes, and Content Nodes _do_ have a URI that users can visit in a browser to see the content.

justlevine:
Thanks so much for taking the time to go into all that detail! We may have started with different premises, but we agree on the conclusion. 

I promised I wouldnt vote for using `ContentNode` as the base type anymore 🤫, so I'll just warn about the dangers of using our own custom naming conventions.

> We could name this something WordPress-y like PostObjectNode or something to that tune, but I feel like that requires some domain knowledge of the client developer. When I browse the Github GraphQL Schema, I don't feel like I'm required to know what system it's built on. The Types are descriptive of what the things actually are, not the system it's built on. I don't feel that post_type or post_type_object describe what content stored in the *_posts table actually is.

💯. 
We can't account for the myriad of ways people extend WP, and in general the further we depart from the understood usage (either in _completely generic terms_ or based on assumed domain knowledge of the client developer), the harder we it is to describe them accurately. Using a functional naming convention is impossible if it cant actually describe what things actually are.

E.g. a [WooCommerce shop order](https://github.com/woocommerce/woocommerce/blob/efdaf77d806b46aea9655f64958450fc77f3032f/plugins/woocommerce/includes/class-wc-post-types.php#L409) is not a public post type, but neither an unsullied developer nor one with WP knowledge is going to logically connect shop orders to utility types, and a search to define either term or both together won't clarify things either.

TylerBarnes:
@jasonbahl I agree with that but I would simplify it to just `DataNode` and `ContentNode`. I think `DataNode` covers the `UtilityNode` use case 🤔

justlevine:
> 

> I think this is where my mind is at currently:
> 
> * `ContentNode`: public facing published content (`posts`, `pages`, `cars`, `houses`, etc) (posts of public post types)
> * `UtilityNode`: nodes of data stored in the posts table, but not intended to be public facing "content" (post types like `custom_css`, `oembed_cache` , etc) (posts of private post types)
> * `DataNode`: anything stored in the wp_posts table, public or private. So a `Post`, `Page`, `Car`, `House`, `CustomCss`, `oEmbedCache`, etc would all be considered a DataNode
> 
> All ContentNodes are DataNodes, but not all DataNodes are Content Nodes.
> 
> All Utility Nodes are DataNodes, but not all DataNodes are Utility Nodes.
> 
> Utility Nodes are not Content Nodes, and Utility Nodes _do not_ have a URI (WordPress 404s if you visit their permalink, for example. . .WPGraphQL shouldn't even have a uri field for these nodes)
> 
> Content Nodes are not Utility Nodes, and Content Nodes _do_ have a URI that users can visit in a browser to see the content.

Where do taxonomies and terms fit into this naming convention?

jasonbahl:
> @jasonbahl I agree with that but I would simplify it to just DataNode and ContentNode. I think DataNode covers the UtilityNode use case 🤔

Users can query public content via `{contentNodes { ... }}` and _all_ content of _all_ post types via `{ dataNodes { ... } }`

Perhaps we have just the 2. 

That said, I do think it could be helpful to have an interface that is shared across private Post Types. 

For Introspection sake, it could be nice for developers to be able to identify the difference between a Node, a ContentNode, a DataNode, a Utility Node, a TermNode (a public term node? a private term node), etc. 

They're all "nodes" but their "possible types" are different in each case. This can be helpful for building various admin tools, etc. 

So ya, I think it might be possible to just have 2 Interfaces (ContentNode = public), (DataNode = all), but I do think there's utility (pun intended) to having the `UtilityNode` interface (or equivalent if this name doesn't feel right still). 🤔

jasonbahl:
> Edit to @TylerBarnes 's edit 😎: A departure from wp is good - so long as its generic while staying descriptive. Im definitely good with content node. I worry data node becomes confusing and less descriptive when we recognize that other types of 'data' exist in the schema and theres nothing generic to explain why those exist separately.

Maybe instead of `DataNode` and `UtilityNode` we steal a play out of the Atlas Content Modeler playbook and use the `Model` terminology. 

In ACM, post types are "Models" and Taxonomies / Terms are still Taxonomies / Terms (these aren't WordPress-y lingo). 

So we could have: 

- Model (all post types)
- DataModel / UtilityModel (private post types)
- ContentModel (public post types)

So then you have the concept of:

- Models
  - ContentModel: (public post types) Model intended for publishing content 
    - Post
    - Page
    - Car
    - House
  - DataModel / UtilityModel: (private post types) Model with some utility other than published content with URI
    - Custom CSS
    - Oembed Cache
    - redirects

Taxonomy       
Term
Users


----

> Where do taxonomies and terms fit into this naming convention? (Regardless, I agree with @TylerBarnes that a separate utility isnt necessary ).

😭 

I don't know! haha. 

Similar issue, though. 

Public Taxonomies have archives and a uri you can visit, and private ones do not. 

I was thinking maybe something like: 

- **TermNode**: terms of any taxonomy
- **ContentTermNode**: terms of public taxonomies
- **UtilityTermNode**: terms of private taxonomies

But, what's weird is you can have private taxonomies associated with public post types, and you can have private post types associated with public taxonomies, so things get weird.

Not sure the term `Content` translates as well to taxonomies. 

You can also have (with some elbow grease) Taxonomies associated with users (@kellenmace has done this for some headless apps)

justlevine:
Im not sure using `Node > Model (post types) > ContentModel (public post types )` is more clear than `Node > ContentNode (post types ) > PublicContentNode`, and I think it becomes even more confusing than your worry about `public` vs `publicly_queryable` when you :
1. try to rename the `WP_Post_Type` and `WP_Taxonomy` interfaces to fit the new conventions.
2. try to understand why other `Node` child interfaces (e.g. comments, terms, taxonomies.. menus are a post type in FSE so not an issue long term ) arent also a `Model` or `DataModel`, or `DataType` or what have you.

And with that said, I now defer to you both :-)

TylerBarnes:
I'm not sure public vs private is the right line to draw for what's "content" and what's "data". What about private content sites where you need to log in? IMO it makes more sense to draw the line at "is it a reasonable assumption that a page would normally be created from this node where this node is the primary content? if so it's content, if not it's data".

jasonbahl:
> I'm not sure public vs private is the right line to draw for what's "content" and what's "data". What about private content sites where you need to log in? IMO it makes more sense to draw the line at "is it a reasonable assumption that a page would normally be created from this node where this node is the primary content? if so it's content, if not it's data".

@TylerBarnes agreed! 

The term `private` is used multiple ways in WordPress, which confuses things further. 

Like, you can have a "post" in a public post type, that does have a uri, but you can set the status to private for that individual post, which means it requires auth to resolve. 

Then, you can register a post_type as `private => false`, but also set it to be `publicly_queryable => true` which means you can have content in a private type but the content is actually public 😱 

This stuff gets so confusing. lol. 

I have a feeling, no matter how good we get at naming these Interfaces, it will remain confusing. 🤣

jasonbahl:
Ok, I'm back with more thoughts. 

## Public Post Types

I think we all largely agree that `ContentNode` should represent CMS content intended to be published and viewed at a specific url (hence implementing the `UniformResourceIdentifiable` Interface). 

So, we can expect that any post type registered as `public => true` should have the `ContentNode` interface applied, and the `uri` field for the node should properly resolve. 

I should then also be able to query across **public** post types with the `contentNodes` query (what this PR originally set out to achieve): 

```graphql
{
  contentNodes {
    nodes { ... }
  }
}
```

This query should translate, _roughly_, to: 

```php
$public_post_types = get_post_types( 'names', [ 'show_in_graphql' => true, 'public' => true ] );
new WP_Query( [ 'post_type' => $public_post_types ]);
```

Of course, you could use `$where` args to filter further. 

The `ContentTypeEnum` should also be made up of only public post types. 

So, I could narrow this query to `where: { contentTypes: [ Post, Page ]}` for example, but, I **_would not_** be able to filter by a private post type, like: `where: { contentTypes: [ OembedCache ]}` as `OembedCache` is a private post type and not a `ContentNode` and should not show up in the `ContentTypeEnum`. 

## Private Post Types

If a post type is registered as `'public' => false` then it implements the `Node` interface, and other applicable interfaces, such as `NodeWithTitle`, etc, but it would NOT implement the `ContentNode` interface. 

It would also not implement another Interface (such as `UtilityNode`, as I previously proposed) which would group it with other Post Types. 

A private Post Type would be largely treated as its own Type of Node. 

For example, if I registered the following post type:

```php
register_post_type( 'redirect', [ 
  'public' => false,
  'show_in_graphql' => true,
  'graphql_single_name' => 'Redirect',
  'graphql_plural_name' => 'Redirects',
  'supports' => [ 'title' ],
] );
```

This would (roughly) end up generating the following in the Schema: 

```graphql
type Redirect implements Node & NodeWithTitle {
  id: ID!
  title: String
}

extend type RootQuery {
  redirect(id: ID!): Redirect
  redirects( $connectionArgs ): RootQueryToRedirectConnection
}
```

jasonbahl:
I want to add more thoughts as well. 

If we were building an administrative interface for editing content in WordPress, similar to the WP Admin, we would need to be able to get a list of all post types so we could add specific UIs to interact with them. 

If `ContentNode / ContentType` becomes synonomous with `WP_Query( [ 'post_type' = $public_post_types ])` (types with a URI)

Then we would still need a way to query for _all_ post types in the admin and get a list of _all_ Types. 

This is where an Interface (or Union) would come in handy for representing _all_ Post Types. 

```graphql
{
  dataTypes { # essentially get_post_type_objects()
    node { # essentialy WP_Post_Type
       id
       name
       dataNodes { # essentially new WP_Query( [ post_type => $data_type->name ] )
          nodes {
            __typename
            id
            ...on ContentNode {
               uri
            }
          }
       }
    }
  }
}
```

If we had something generic like a generic `DataType` / `DataNode` interface (or possibly concede and do `PostObjectType / PostObjectNode`) then we could query across _all_ post types to do things like build admin interfaces that span across types. 

But the `contentNodes` query would still be the common entry point for querying across _public_ types (with a uri). 

🤔 

So, similar to @CodeProKid thoughts, we could have:

```graphql
"""
A node is an object that can be identified by ID
"""
interface Node {
  id: ID!
}

"""
DataType Enum represents _all_ post types, public or private
"""
enum DataTypeEnum {
  Post
  Page
  Redirect
  OembedCache
}

"""
ContentTypeEnum represents public post types (nodes that have a uri)
"""
enum ContentTypeEnum {
  Post
  Page
}

"""
A Data Node is a node stored in the *_posts table. Their uses can be varied, but they have some things in common, like updated date, etc.
"""
interface DataNode implements Node {
  id: ID!
  date: DateTime!
  modified: DateTime!
  dataType: DataTypeEnum!
}

""" 
ContentNode is a DataNode (node in the *_post table) that is public (has a uri)
"""
interface ContentNode implements Node & DataNode {
  id: ID!
  date: DateTime!
  updated: DateTime!
  contentType: ContentTypeEnum!
  uri: String
}

"""
oembed_cache post_type is 'public => false'. Public users don't interact with this data set directly. This data is used internally when resolving post_content, etc.
"""
type OembedCache implements DataNode {
  id: ID!
}


"""
Redirect would be a post_type registered as 'public => false, publicly_queryable => true' as these nodes aren't directly queryable by URI, but they're publicly needed to generate lists of redirects. They're a DataNode, but not a ContentNode (no uri). 
"""
type Redirect implements DataNode {
  id: ID!
  date: DateTime!
  updated: DateTime!
  postType: PostTypeEnum!
  from: URL
  to: URL
}

""" 
A Post is a public facing DataNode that can be viewed at a uri. It implements other interfaces (based on post_type_supports() such as `NodeWithTitle` and `NodeWithAuthor`, etc
"""
type Post implements DataNode & ContentNode & NodeWithTitle & NodeWithAuthor {
  id: ID!
  date: DateTime!
  updated: DateTime!
  postType: PostTypeEnum!
  uri: String
  categories: [Category!]
  tags: [Tag!]
}
```

So now, we could do the following: 

```
{
  contentTypes {
     node {
        __typename
        name
     }
  }
}
```

And get a list of nodes back like so: 

```json
[
  {
    "__typename": "ContentType",
    "name": "Post"
  },
  {
    "__typename": "ContentType",
    "name": "Page"
  }
]
```

But a query like: 

```graphql
{
  dataTypes {
     node {
        __typename
        name
     }
  }
}
```

might give us a list of nodes like so: 

```json
[
  {
    "__typename": "ContentType",
    "name": "Page"
  },
  {
    "__typename": "ContentType",
    "name": "Post"
  },
  {
    "__typename": "DataType",
    "name": "CustomCSS"
  },
  {
    "__typename": "DataType",
    "name": "Redirect"
  },
  {
    "__typename": "DataType",
    "name": "OEmbedCache"
  },
]
```

We could do public facing search queries like so: 

```graphql
{
  contentNodes( where: { search: $keyword } ) { ... }
}
```

And we could do admin searches (across all post types) like so:

```graphql
{
  dataNodes( where: { search: $keyword } ) { ... }
}
```

jasonbahl:
Some naming I'm considering 🤔 : 

```
DataNode = object stored in *_posts table
DataType = WP_Post_Object (get_post_type_object())
```
or 
```
EntryNode = object stored in *_posts table
EntryType = WP_Post_Object (get_post_type_object())
```
or
```
Model = object stored in *_posts table
ModelType = WP_Post_Object (get_post_type_object())
```

jasonbahl:
As far as terms/taxonomies go, I'm thinking we have a similar problem here: 

Just like Post Types, taxonomies can be: 

- `'public' => true/false`
- `'publicly_queryable' => true/false`

So, possibly we go with: 

```
TermNode = terms of _any_ taxonomy (get_term() / WP_Term)
ContentTermNode = terms of _public_ taxonomy (get_term() / WP_Term)

Taxonomy = any taxonomy (get_taxonomy() / WP_Taxonomy)
TaxonomyEnum = names of any taxonomy
ContentTaxonomy = any public taxonomy (get_taxonomy / WP_Taxonomy)
ContentTaxonomyEnum = names of any public taxonomy
```

mindctrl:
I'm still catching up on this thread 😅 but just wanted to share a few thoughts as I read.

> We can see that we're even getting a URI for the action monitor nodes, but, we really shouldn't because WordPress itself doesn't respect URI for private post types.

Action monitor post types are queryable on the front end. There's an archive page and single post views work too. Maybe you need to flush permalinks?
image


The reason `action_monitor` is viewable is because `publicly_queryable` is set to `true`. [Reference](https://github.com/gatsbyjs/wp-gatsby/blob/209880d4be7f8518b656a11f43d1aa009856b277/src/ActionMonitor/ActionMonitor.php#L187).

WordPress has this to say about `publicly_queryable`:
```
 *     @type bool         $publicly_queryable    Whether queries can be performed on the front end for the post type
 *                                               as part of parse_request(). Endpoints would include:
 *                                               * ?post_type={post_type_key}
 *                                               * ?{post_type_key}={single_post_slug}
 *                                               * ?{post_type_query_var}={single_post_slug}
 *                                               If not set, the default is inherited from $public.
 ```

This is backed by the way `is_post_publicly_viewable()` and `is_post_type_viewable()` work. 
 
 IMO, if the intent of the Gatsby plugin is to not expose those post types to the public, they should register them as `publicly_queryable => false`, though I know that's not really the focus of this conversation. Just wanted to mention this as a clarifying point to the whole "public" vs "private" thing. I think it matters because you were saying "But the contentNodes query would still be the common entry point for querying across public types (with a uri).", and in the example of action_monitor types they do have a URI.

> Then, you can register a post_type as `private => false`, but also set it to be `publicly_queryable => true` which means you can have content in a private type but the content is actually public

`register_post_type()` doesn't support a `private` argument as far as I see.

**Comment by justlevine** [comment]
@mindctrl 
> Then, you can register a post_type as private => false

I believe that is a mistype. The intention here is that (obstensibly) when `public` is true there is a public facing uri, and when it's false there isn't . A "private" post type / taxonomy (i.e. `'public' === false )` can still be `publicly_queryable` (determines the model permissions for unverified users), and a CPT/tax that is neither public nor publicly queryable can still be queried in WPGraphQL by users with the correct permissions if `show_in_graphql` is set.

> There's an archive page and single post views work too.

Can you confirm whether they have an actual URI or are just accessible with query vars (like in your screenshot). In general I worry that there's too many ways the registration args are used in the wild for us to be able to accurately give various interfaces self-documenting names that aren't tied to WP terminology, but I was under the impression that WPGatsby's action monitor _does_ meet those assumptions.

**Comment by mindctrl** [comment]
@justlevine 

> Can you confirm whether they have an actual URI or are just accessible with query vars (like in your screenshot).

Yes, they have a URI. Good idea to check that! Screenshot (logged out just to check lack of auth):

image

> The intention here is that (obstensibly) when public is true there is a public facing uri, and when it's false there isn't .

It's possible for this to be not true, assuming we're talking about the same thing when we say "public" (confusing all around!). e.g. this is public and has a public facing URI:
`register_post_type( 'moo', [ 'publicly_queryable' => true, 'public' => false ] );`

> In general I worry that there's too many ways the registration args are used in the wild for us to be able to accurately give various interfaces self-documenting names that aren't tied to WP terminology

I think you might be right. In addition to the all the ways people do it, I think there's a very good chance people misunderstand how `public` and `publicly_queryable` work (maybe including me 😆).

**Comment by mindctrl** [comment]
Edited my previous message to fix an incorrect statement I made. 🤦

**Comment by jasonbahl** [comment]
> register_post_type() doesn't support a private argument as far as I see.

@justlevine @mindctrl correct, this was a typo. `public => true/false` is what I meant.

----

@mindctrl ah yes, I've confirmed that `public => false, publicly_queryable => true` does indeed generate a proper permalink! And the content is accessible at that permalink 🤔 

I must not have been flushing my permalinks during my other tests here.

But! It still leaves these posts out of most other public interfaces, such as sitemaps, search, etc. 

So, the idea is that if you know how to get to the thing, you can get to it, publicly, but it's not going to be grouped together with other content that's primarily intended to make up the pages of the site. 

## public => false

I registered the following post type (and flushed my permalinks):

```php
register_post_type( 'publicly_queryable', [
	'public' => false,
	'label' => 'Publicly Queryable',
	'show_ui' => true,
	'publicly_queryable' => true,
	'show_in_graphql' => true,
	'show_in_rest' => true,
	'graphql_single_name' => 'PubliclyQueryable',
	'graphql_plural_name' => 'PubliclyQueryables'
]);
```

With `public => false` the WordPress editor doesn't even give me the "Permalink" setting to view the post. 

CleanShot 2022-05-04 at 09 22 09

And you can see this content is left out of the sitemap. 

CleanShot 2022-05-04 at 09 16 02

And the content is not included in search results:

CleanShot 2022-05-04 at 09 32 41

I can, however, access the page by it's permalink (I was wrong about that above!)

CleanShot 2022-05-04 at 09 25 05

But, it seems that WordPress is trying hard to prevent this thing from being grouped together with other post types intended to be used as "content".

Hence my theory that there's a difference between "ContentNodes" (public post types) and "Other, not public, post types"

## public => true

If I change my post type registration to `public => true` (and flush my permalinks)

```php
register_post_type( 'publicly_queryable', [
	'public' => true,
	'label' => 'Publicly Queryable',
	'show_ui' => true,
	'publicly_queryable' => true,
	'show_in_graphql' => true,
	'show_in_rest' => true,
	'graphql_single_name' => 'PubliclyQueryable',
	'graphql_plural_name' => 'PubliclyQueryables'
]);
```

Now the editor gives me a Permalink UI to view the content:

CleanShot 2022-05-04 at 09 45 59

And the content is included in sitemaps:

CleanShot 2022-05-04 at 09 47 23

WordPress is treating this as "content" that makes up the site, where `public => false` post types were treated much differently. You could access the data if you knew how, but in general, the data of those post types aren't treated as "content"

And thus, here we are. 

How do we define / name these differences in a way that makes sense. 

What are the use cases folks need. 

When I originally implemented the `ContentNode` Interface and `RootQuery->ContentNode` connection, the goal was to query across nodes that are intended to be used as content (public => true) post types. 

This PR is aiming to get that working as originally intended. 

But, comes at a loss of "how can we interact with these other nodes that are not considered `ContentNode`"?

What are they called? How do we query them? etc. 



> I think you might be right. In addition to the all the ways people do it, I think there's a very good chance people misunderstand how public and publicly_queryable work (maybe including me 😆).

Me too! lol. It's evolved over so many years by different people with different context. Even core seems to treat things slightly differently based on how you access data. 

For example, the content in the "custom_css" post type is clearly publicly accessible content, as we can see the output of it as a public user, but the post_type is set as private, and _not_ set to publicly_queryable. We should assume that this is fully private, but yet it's output as publicly visible data. So, WordPress registers it as private, but outputs the content as public 🤦🏻‍♂️ . Confusing indeed. 

A lot of this is because WordPress was (and still is) built with the assumption that the CMS is _also_ the rendering layer, and with headless we've come to muck things up. 

So we (WPGraphQL specifically) rely on these registry rules perhaps more than core WordPress itself, because in monolith WordPress, even if something is `public => false, publicly_queryable => false`, you can still write a `WP_Query` or direct `sql` statement to "break the rules" and get the data anyway, which is the case with the `custom_css`. It's registered as non-public, non-publicly_queryable content, but core WordPress "breaks the rules" and gets the content and outputs it to the public anyway.

TylerBarnes:
> IMO, if the intent of the Gatsby plugin is to not expose those post types to the public, they should register them as `publicly_queryable => false`, though I know that's not really the focus of this conversation. Just wanted to mention this as a clarifying point to the whole "public" vs "private" thing. I think it matters because you were saying "But the contentNodes query would still be the common entry point for querying across public types (with a uri).", and in the example of action_monitor types they do have a URI.

@mindctrl that's required by WPGraphQL as of a recent version for a post type to be public in the api. I made that change because action monitor posts should be publicly queryable by Gatsby sites without auth.

TylerBarnes:
I think since this release https://github.com/wp-graphql/wp-graphql/releases/tag/v1.6.7

justlevine:
[It seems](https://github.com/WordPress/wordpress-develop/blob/e1b417f5dc17dcdf45ac56ecf1a9e5cdd26039aa/src/wp-includes/class-wp-post-type.php#L611) that `$args['rewrite']` is responsible for whether a `publicly_queryable` type gets a permalink.

Assuming we accept the definition of an interface as a unique grouping of shared fields (https://www.apollographql.com/docs/apollo-server/schema/unions-interfaces/#interface-type), maybe it makes sense to revisit this as part of #1738? _What_ we call the interface is only important after we have a clear definition of what fields the interface groups, and under what conditions they're applied to a given GraphQL type ...

In the interim, we can temp-fix #1679 by adding a `isPublic: bool|undefined` where arg to `contentNodes`, and either leave it undefined for a non-breaking workaround, or setting the default value to `true` for a breaking change that mirror's the original (and possibly future) intent of the connection.

jasonbahl:
> Some individual action monitor posts are private (anything related to authenticated previews).

@TylerBarnes ya, some posts / pages are private too. 

I think what we're trying to ultimately figure out, is how can users interact with: 

- Public data that's intended to be represented by a URI, and is stored by WordPress in the *_posts table
- Public data that should be queryable without authentication, but isn't intended to be included in sitemaps, and the primary use case isn't visiting the thing at a uri (action montitor actions, custom_css, etc)
- Private data that should only be visible to authenticated users with proper capabilities (contact form submissions, the core "user_request" post_type, etc).


I think we're all largely on the same page that fully public post_types should be queryable together, and the `ContentNode` interface and `{ contentNodes { ... } }` query solves for that. 

For these other things, what do we do with them? How do we expose them to the Schema? What do we call them? What Interfaces are needed to express how they're similar but different than other things? etc?

It's clear that some naming in WordPress is already super confusing. ex: `public` / `publicly_queryable`. . .some of us have a decade+ of WordPress experience and are still confused by this. 

How can we express in the WPGraphQL Schema what these things are by naming them well and providing good descriptions in the Schema for them? lol.

jasonbahl:
> Assuming we accept the definition of an interface as a unique grouping of shared fields (https://www.apollographql.com/docs/apollo-server/schema/unions-interfaces/#interface-type)

@justlevine ya, so Interfaces ensure Types that implement it all have those fields, but not all Types that have those fields need to implement said interface. 

For example, if we had an interface such as `DataNode` to represent posts of any post type, and we decided the only field that we felt was common across _all_ post types, was an ID, then we could have an interface like so: 

```
interface DataNode implements Node {
  id: ID!
}
```

Then, we could have our Post Types implement that interface, but not our term types (even though terms also have an ID)

```
type Post implements DataNode  & Node{
  id: ID!
  title: String
  ...otherFields
}

type Category implements TermNode & Node {
  id: ID!
  name: String
  ...otherFields
}
```

Even though a Category has an ID field, and _could_ technically implement the DataNode interface, we don't want `Category` to be referenced as a `possibleType` of the `DataNode` interface as we know that querying for a `DataNode` will never return a Category. 

We know that querying for a DataNode, anywhere in the graph, should be limited only to posts (of any post type). 

Much like we know that a `ContentNode`, anywhere in the graph, should be limited to only posts of public post types. 

I do agree that there's much overlap with #1738! And, because I'm an over-analyzer (as you all know from this discussion) that PR has been hanging for quite a while. Perhaps now is the time to get that one finalized as well, as they're solving very similar problems.

jasonbahl:
> Agenda item for the [upcoming meeting](https://wp-graphql.slack.com/archives/CDE8Y3MN3/p1651170633748049) perhaps?

yes

chriszarate:
Another late arriver to a great conversation. There are two things that strike me upon reading the points here:

1. There are a lot of dichotomies proposed here—Public vs. Private, Content vs. Data, Function vs. Utility—but all of them turn out to be incredibly subjective. What value do these classifications bring, especially if they will be confusing or controversial?

2. On top of this, we've established that the implementation details of how WordPress (or a third-party plugin) chooses to register a post type can be confusing even to experienced WordPress devs. Those details (e.g., `"rewrite" => true`) often have little bearing on how we think about the underlying data—or on how a WPGraphQL consumer might want to query for it. Why automatically perpetuate decisions that a headless implementor might want to revisit?

> I think we're all largely on the same page that fully public post_types should be queryable together, and the ContentNode interface and { contentNodes { ... } } query solves for that.

I agree with this. (Although I would probably advocate for an even narrower definition with better alignment. `ContentNodes` should have `post_content`. However, I am happy to abandon this if only for backwards compatibility.)

Beyond `ContentNode`, I think interfaces should be purely aligned with the data model and I would advocate against further "convenience" interfaces that will probably end up subverting how the implementor thinks about their data. If we make it easy for implementors to specify the post types they want to query against a generic interface, I think that's enough.

I will sit with my reading a bit more, though.

PR #2356, opened 2022-04-29
chore: refactor codebase using `WPGraphQL::get_allowed_{type}( 'objects ')`

justlevine:
This PR builds on #2353, by replacing the previous hydration pattern that used allowed post type / taxonomy names to get their objects, with the updated `WPGraphQL::get_allowed_{type}( 'object' ) method. 

e.g  this:
```php
$allowed_post_types = \WPGraphQL::get_allowed_post_types();

if ( ! empty( $allowed_posts_types ) && is_array( $allowed_post_types ) {
  foreach( $allowed_post_types as $post_type ) {
    $post_type_object = get_post_type_object( $post_type );

    if ( $post_type_object instanceof \WP_Post ) {
      // do stuff with $post_type_object.
    }
  }
}
```
becomes this :

```php
$allowed_post_types = \WPGraphQL::get_allowed_post_types( 'objects' );

foreach( $allowed_post_types as $post_type_object ) {
  // do stuff with $post_type_object
}

This pattern is more readable, easier to test and maintain, and will probably improve some codeclimate scores too.

```
no

Any other comments?
-------------------
* For consistency (and to better help code scanning), I also renamed taxonomy object variables to `$tax_object` in the few places they were using a different variation. 

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + PHP 8.0.15)

**WordPress Version:** 5.9.3

### DIFF
--- a/src/Connection/TermObjects.php
+++ b/src/Connection/TermObjects.php
@@ -25,9 +25,6 @@ class TermObjects {
 	 */
 	public static function register_connections() {
 
-		$allowed_taxonomies = \WPGraphQL::get_allowed_taxonomies();
-		$allowed_post_types = \WPGraphQL::get_allowed_post_types();
-
 		register_graphql_connection(
 			[
 				'fromType'       => 'RootQuery',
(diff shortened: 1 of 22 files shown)


justlevine:
All those codeclimate smells are on preexisting logic thats unchanged in this PR. Marking them as resolved.

src/Connection/TermObjects.php
```diff
@@ -52,179 +49,162 @@ public static function register_connections() {
...
+						$resolver->set_query_arg( 'include', $ancestor_ids );
+
+						return $resolver->get_connection();
 					},
 				] );
 			}
 		}
 
+		// Register a connection from each post type that
+		foreach ( $allowed_post_types as $post_type_object ) {
```
justlevine, line 176: I'm unclear why this foreach loop is separate from earlier one on L85.

jasonbahl: ah yes, so the one on line 85 is within the `foreach ( $allowed_taxonomies as $tax_object ) {` loop. 

So, this is getting the intersection of allowed post types for each allowed taxonomy so we can create connections:

- from TaxonomyA -> PostTypeA, etc

This foreach, on line 176 is top level post types, not nested within another foreach. 

I'm open to re-thinking how this was done if you have ideas to dry it up or whatever.

PR #2360, opened 2022-05-04
Chore: PHPStan 1.6 fixes (first pass)

justlevine:
What does this implement/fix? Explain your changes.
---------------------------------------------------
This PR fixes ~63~ 66 stan errors when using PHPStan v1.6, as well as a pesky formatting issue that was causing WPCS to fatal error on PHP8.

**Note**: no changes to actual code or the projects dependencies are in this PR. That will happen in a separate PR to fix the remaining ~73~ 70 errors (which do require changes to the code to resolve).

### To Test
1. On `develop`, run `composer require --dev szepeviktor/phpstan-wordpress phpstan/phpstan`, followed by `composer phpstan`.
2. Checkout this PR, and then run  `composer require --dev szepeviktor/phpstan-wordpress phpstan/phpstan`, followed by `composer phpstan`.
3. Confirm the number of PHPStan errors dropped from 136 to 73

Any relevant logs, error output, GraphiQL screenshots, etc?
-------------------------------------
**Before::**
image

**After**
image

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 ( WSL2 + devilbox + PHP8.0.15)

**WordPress Version:** 5.9.3

PR #2361, opened 2022-05-04
Chore: PHPStan 1.6 fixes (secondpass)

justlevine:
This PR builds off of #2360 to fix another 51 errors in PHPStan 1.6.x. Unlike that PR, this one includes changes to the actual code.

To make merging easier #2360 has been squashed into this PR.

**Note**: the dependencies themselves are still not updated in this PR, to allow for merging before the remaining errors are resolved.

### Details
- stan: Remove "impossible" unnecessary conditionals (e..g `isset` in a property that _must_ exist ), and add some missing type hints.
- fix: `Model\MenuItem::init()` now correctly checks for `$this->fields` before initializing.
- fix: The specific `WP_Error` messages are now thrown during a failed `MediaItemCreate` and `MediaItemUpdate` mutation.

### To Test
- On develop, `run composer require --dev szepeviktor/phpstan-wordpress phpstan/phpstan`, followed by `composer phpstan`.
- Checkout this PR, and then run `composer require --dev szepeviktor/phpstan-wordpress phpstan/phpstan`, followed by `composer phpstan`.
- Confirm the number of PHPStan errors dropped from 136 to 20

image

Any other comments?
-------------------

Where has this been tested?
---------------------------

**WordPress Version:** 5.9.3


justlevine:
I need help with the remaining 20 errors (either in this PR if they're low hanging fruit, or in a third-pass PR). Specifically:

* `src/Data/Connection/PostObjectConnectionResolver.php:234`
`$query_args['post_parent']` won't ever exist. Is the intent `$args['where']['post_parent']` or something else?

* `src/Data/DataSource.php:350` and `src/Type/ObjectType/CommentAuthor.php:81`
`get_avatar_data()` always returns an `array` with `found_avatar` as a `bool` indicating if the avatar exists. We need to decide if we always want an `Avatar` model (current behavior), or to resolve null if `found_avatar` is false.

* `src/Data/Loader/PostObjectLoader.php:144`
Is the correct behavior here to set `$loaded_posts[ $key ]` to `null` on a non-existent WP_Post, or to leave it out of the returned array altogether?

* `src/Data/NodeResolver.php:360`
My favorite file 🙃.  Want a second set of eyes to confirm there isn't a missing case when `'page' === $show_on_front` but a valid `$page_id` isnt set.

* `src/Model/MenuItem.php:160`
Is this supposed to be a different conditional, or is it really a duplicate.

* `src/Model/Post.php:190`, `src/Model/Post.php:353`, `src/Model/Term.php:83`, `src/Model/User.php:107`.
Is there a possibility that `$this->data` is _not_ going to be a `WP_Post`, `WP_Term`, or `WP_User` object? And is it possible those will be an _invalid_ object (i.e. where the database ID is falsy)?

* `src/Model/Post.php:394`
Same question as above applies to `$parent_post_type_object`.

* `src/Server/ValidationRules/QueryDepth.php:107`
What does an  `InlineFragmentNode` without valid children actually look like?

* `src/Type/Enum/MediaItemSizeEnum.php:27`
Is there a reason we're providing our own default instead of relying on [`get_intermediate_images_sizes()`](https://developer.wordpress.org/reference/functions/get_intermediate_image_sizes/)?

* `src/Type/Enum/MediaItemStatusEnum.php:24`
Should possible attachment `$post_stati` be filterable or allow for custom-registered stati (like the broader [`get_post_stati()`](https://developer.wordpress.org/reference/functions/get_post_stati/) does... I cant seem to find a way to get post-type-specific stati only...)?

* `src/Type/ObjectType/RootQuery.php:227, 696, 783`.
I dont understand how `Deffered` classes work well enough to resolve this.

justlevine:
> [Coverage Status](https://coveralls.io/builds/48825591)
> 
> Coverage decreased (-0.04%) to 79.288% when pulling **[be48f48](https://github.com/wp-graphql/wp-graphql/commit/be48f48077113c081204112ec7a1fe827139b728) on justlevine:chore/phpstan-1.6-second-pass** into **[c1c2255](https://github.com/wp-graphql/wp-graphql/commit/c1c2255762394b5ec56ce6b8f6e56cbb212433b7) on wp-graphql:develop**.

These are all preexisting coverage issues, just a few more lines since the thrown `UserErrors` now try to use the `WP_Error` message.

justlevine:
> Code Climate has analyzed commit [be48f48](https://github.com/wp-graphql/wp-graphql/commit/be48f48077113c081204112ec7a1fe827139b728) and detected **15 issues** on this pull request.
> 
> Here's the issue category breakdown:
> 
> Category	Count
> Complexity	9
> Duplication	6
> View more on [Code Climate](https://codeclimate.com/github/wp-graphql/wp-graphql/pull/2361).

Same here. Other than the duplications, its just resurfacing previous issues.

src/Type/ObjectType/RootQuery.php
```diff
@@ -610,7 +610,7 @@ function ( $post ) use ( $allowed_post_types ) {
 						'type'        => 'User',
 						'description' => __( 'Returns the current user', 'wp-graphql' ),
 						'resolve'     => function ( $source, array $args, AppContext $context ) {
-							return isset( $context->viewer->ID ) && ! empty( $context->viewer->ID ) ? DataSource::resolve_user( $context->viewer->ID, $context ) : null;
+							return ! empty( $context->viewer->ID ) ? DataSource::resolve_user( $context->viewer->ID, $context ) : null;
```
jasonbahl, line 613: I don't care _that much_ but I've been getting rid of more and more of the references to DataSource methods and calling the loaders directly, like so: 

```suggestion
							return ! empty( $context->viewer->ID ) ? $context->get_loader( 'user' )->load_deferred( $context->viewer->ID ) : null;
```

jasonbahl (dismissed):

justlevine:
@jasonbahl Here's the fix for the broken test. I dont want to create a merge conflict if you're in middle of something 😇

https://github.com/justlevine/wp-graphql/blob/10db8f8fc48a1439ab5fd36495ca2edfc3ae68e7/src/Type/ObjectType/EnqueuedStylesheet.php#L32

image

jasonbahl:
I need help with the remaining ~20~ 19 errors (either in this PR if they're low hanging fruit, or in a third-pass PR). Specifically:

> * [src/Data/Connection/PostObjectConnectionResolver.php:234](https://github.com/justlevine/wp-graphql/blob/057363dba45fbca3ce4a73d628e89a669f820258/src/Data/Connection/PostObjectConnectionResolver.php#L234)
`$query_args['post_parent']` won't ever exist. Is the intent `$args['where']['post_parent']` or something else?

Ya, we map the where args via `$input_fields = $this->sanitize_input_fields( $this->args['where'] );`

So ya, this is supposed to unset `$input_fields['parent']` before the array_merge happens below. 

That's where we map the input args with the query_args, then we unset the post_parent argument after that merge

Pushed a commit with the update.

> * [src/Data/DataSource.php:350](https://github.com/justlevine/wp-graphql/blob/c3b61f0eb8a8f9537771f7ea3badc757a38e3ce9/src/Data/DataSource.php#L350) and [src/Type/ObjectType/CommentAuthor.php:81](https://github.com/justlevine/wp-graphql/blob/3963a844b1f1b07742a20e3cb8b3cdbf1cd51971/src/Type/ObjectType/CommentAuthor.php#L81)
`get_avatar_data()` always returns an `array` with `found_avatar` as a `bool` indicating if the avatar exists. We need to decide if we always want an `Avatar` model (current behavior), or to resolve null if `found_avatar` is false.

This is where WordPress and Static Analysis don't always line up. 

The `get_avatar_data()` function says it returns an array, but what it _really_ returns is the results of `apply_filters( 'get_avatar_data', $args, $id_or_email );`

And a plugin could easily do:

```php
add_filter( 'get_avatar_data', function() { return 'potatoes'; } );
```

or 

```php
add_filter( 'get_avatar_data', function() { return null; } );
```
So, while we believe that it will most likely be an array, we don't have a guarantee, because of the filtered return.

Part of me wants to add a @phpstan-ignore-next-line here explaining why we're checking if it is a non-empty array, but part of me wants to let it break if someone is indeed being a turd and returning something weird from that filter. 

I do like the idea of bailing early if `false === $avatar['not_found']` but that looks like it will be a breaking change to do it (although most likely a good one, as who wants to get an avatar object with no values returned? 🤔 )

...

Ok, so after looking at it, get_avatar_data will return a fallback url from gravatar even if an avatar is not found. 

So, we don't want to bail early if found_avatar is false, we want to bail early if 'url' is empty. 

Pushed an update to compensate for this. Fixes the phpstan warning and keeps tests passing.

> * [src/Data/Loader/PostObjectLoader.php:144](https://github.com/justlevine/wp-graphql/blob/d0dbfd2a7690e5a57396e2f85283793ae154eeea/src/Data/Loader/PostObjectLoader.php#L144)
Is the correct behavior here to set `$loaded_posts[ $key ]` to `null` on a non-existent WP_Post, or to leave it out of the returned array altogether?

Ah! So here we're already setting `$loaded_posts = []` so we don't need to `return ! empty( $loaded_posts ) ? $loaded_posts : [];`

Pushed a commit.

* [src/Data/NodeResolver.php:360](https://github.com/justlevine/wp-graphql/blob/d0dbfd2a7690e5a57396e2f85283793ae154eeea/src/Data/NodeResolver.php#L360)
My favorite file 🙃.  Want a second set of eyes to confirm there isn't a missing case when `'page' === $show_on_front` but a valid `$page_id` isnt set.

* [src/Model/MenuItem.php:165](https://github.com/justlevine/wp-graphql/blob/29d4130001d52d8df0e2fbd3bd0e4e9a46a5364c/src/Model/MenuItem.php#L165)
Is this supposed to be a different conditional, or is it really a duplicate.

* [src/Model/Post.php:190](https://github.com/justlevine/wp-graphql/blob/88b569b1a51884feb18d6ad22ac3c001c87fa6aa/src/Model/Post.php#L190), [src/Model/Post.php:353](https://github.com/justlevine/wp-graphql/blob/88b569b1a51884feb18d6ad22ac3c001c87fa6aa/src/Model/Post.php#L353), [src/Model/Term.php:83](https://github.com/justlevine/wp-graphql/blob/365b021efeddfc99df1fc7898c071294cac741b6/src/Model/Term.php#L83), [src/Model/User.php:107](https://github.com/justlevine/wp-graphql/blob/057363dba45fbca3ce4a73d628e89a669f820258/src/Model/User.php#L107).
Is there a possibility that `$this->data` is _not_ going to be a `WP_Post`, `WP_Term`, or `WP_User` object? And is it possible those will be an _invalid_ object (i.e. where the database ID is falsy)?

* [src/Model/Post.php:394](https://github.com/justlevine/wp-graphql/blob/88b569b1a51884feb18d6ad22ac3c001c87fa6aa/src/Model/Post.php#L394)
Same question as above applies to `$parent_post_type_object`.

* [src/Server/ValidationRules/QueryDepth.php:107](https://github.com/justlevine/wp-graphql/blob/9a43fb06fa140567f1fe29b577cb5ae89549c7aa/src/Server/ValidationRules/QueryDepth.php#L107)
What does an  `InlineFragmentNode` without valid children actually look like?

* [src/Type/Enum/MediaItemSizeEnum.php:19](https://github.com/justlevine/wp-graphql/blob/9b43b680b2d26d5779a29825c371fa09557418d1/src/Type/Enum/MediaItemSizeEnum.php#L19)
Is there a reason we're providing our own default instead of relying on [`get_intermediate_images_sizes()`](https://developer.wordpress.org/reference/functions/get_intermediate_image_sizes/)?

* [src/Type/Enum/MediaItemStatusEnum.php:24](https://github.com/justlevine/wp-graphql/blob/9b43b680b2d26d5779a29825c371fa09557418d1/src/Type/Enum/MediaItemStatusEnum.php#L24)
Should possible attachment `$post_stati` be filterable or allow for custom-registered stati (like the broader [`get_post_stati()`](https://developer.wordpress.org/reference/functions/get_post_stati/) does... I cant seem to find a way to get post-type-specific stati only...)?

> * [src/Type/ObjectType/RootQuery.php:227](https://github.com/justlevine/wp-graphql/blob/d0dbfd2a7690e5a57396e2f85283793ae154eeea/src/Type/ObjectType/RootQuery.php#L227), [696](https://github.com/justlevine/wp-graphql/blob/d0dbfd2a7690e5a57396e2f85283793ae154eeea/src/Type/ObjectType/RootQuery.php#L696), [783](https://github.com/justlevine/wp-graphql/blob/d0dbfd2a7690e5a57396e2f85283793ae154eeea/src/Type/ObjectType/RootQuery.php#L783).
I dont understand how `Deferred` classes work well enough to resolve this. 

Ah! Looks like we just need to bail early if we're not getting what we expect from the deferred. 

Pushed a commit.

src/Model/Comment.php
```diff
@@ -93,18 +93,17 @@ protected function is_private() {
...
 			return true;
 		}
 
 		// A comment is considered private if it is attached to a private post.
-		if ( empty( $commented_on ) || true === ( new Post( $commented_on ) )->is_private() ) {
+		if ( true === ( new Post( $commented_on ) )->is_private() ) {
 			return true;
 		}
 
 		// NOTE: Do a non-strict check here, as the return is a `1` or `0`.
```
justlevine, line 105: comment can be dropped.

src/Mutation/MediaItemCreate.php
```diff
@@ -253,10 +253,15 @@ public static function mutate_and_get_payload() {
 			 * post_status (inherit if not entered)
 			 * post_mime_type (pulled from the file if not entered in the mutation)
 			 */
-			$attachment_id = wp_insert_attachment( $media_item_args, $file['file'], $attachment_parent_id );
+			$attachment_id = wp_insert_attachment( $media_item_args, $file['file'], $attachment_parent_id, true );
 
-			if ( 0 === $attachment_id || is_wp_error( $attachment_id ) ) {
-				throw new UserError( __( 'The Media Item failed to create', 'wp-graphql' ) );
+			if ( is_wp_error( $attachment_id ) ) {
```
justlevine, line 258: the `0 === $attachment_id` should be added back: https://developer.wordpress.org/reference/functions/wp_insert_attachment/

src/Mutation/MediaItemUpdate.php
```diff
@@ -145,8 +145,13 @@ public static function mutate_and_get_payload() {
 			 */
 			$post_id = wp_update_post( $clean_args, true );
 
-			if ( 0 === $post_id || is_wp_error( $post_id ) ) {
-				throw new UserError( __( 'The media item failed to update', 'wp-graphql' ) );
+			if ( is_wp_error( $post_id ) ) {
```
justlevine, line 148: `0 === $post_id` should be added back: https://developer.wordpress.org/reference/functions/wp_update_post/

src/Mutation/PostObjectCreate.php
```diff
@@ -296,13 +296,6 @@ public static function mutate_and_get_payload( $post_type_object, $mutation_name
 				}
 			}
 
-			/**
-			 * If the $post_id is empty, we should throw an exception
-			 */
-			if ( empty( $post_id ) ) {
```
justlevine, line 302: should be added back (or refactored to match the other tests for a return value of `0`): https://developer.wordpress.org/reference/functions/wp_insert_post/

src/Mutation/PostObjectUpdate.php
```diff
@@ -160,7 +160,7 @@ public static function mutate_and_get_payload( $post_type_object, $mutation_name
 			/**
 			 * Throw an exception if the post failed to update
 			 */
-			if ( 0 === $post_id || is_wp_error( $updated_post_id ) ) {
+			if ( ! $post_id || is_wp_error( $updated_post_id ) ) {
```
justlevine, line 163: it should be `0 === $updated_post_id`. https://developer.wordpress.org/reference/functions/wp_update_post/

src/Type/ObjectType/MediaSize.php
```diff
@@ -64,7 +64,7 @@ public static function register_type() {
 
 							if ( ! empty( $image['ID'] ) ) {
 								$src = wp_get_attachment_image_src( absint( $image['ID'] ), $image['name'] );
-								if ( is_array( $src ) && isset( $src[0] ) && ! empty( $src[0] ) ) {
+								if ( is_array( $src ) && ! empty( $src[0] ) ) {
```
justlevine, line 67: This can be simplified ([as done in the post model](https://github.com/wp-graphql/wp-graphql/blob/14de401d563f3dec99ba40c680d0f7e73b004746/src/Model/Post.php#L821)) to `! empty( $src )`

src/Type/ObjectType/RootQuery.php
```diff
@@ -220,6 +221,12 @@ public static function register_type() {
 
 							return absint( $post_id ) ? $context->get_loader( 'post' )->load_deferred( $post_id )->then(
 								function ( $post ) use ( $allowed_post_types ) {
+
+									// if the post isn't an instance of a WP_Post or Post model, return
+									if ( ! $post instanceof \WP_Post && ! $post instanceof Post ) {
```
justlevine, line 226: Should we be converting a `WP_Post` to a `Post` model before returning on L234?
(Might be beyond the scope of this PR, but if we're explicitly allowing `\WP_Post` and not just its `Model` on the loader, it might make sense to address it now).

src/Type/ObjectType/RootQuery.php
```diff
@@ -707,6 +714,12 @@ public static function register_post_object_fields() {
 
 						return absint( $post_id ) ? $context->get_loader( 'post' )->load_deferred( $post_id )->then(
 							function ( $post ) use ( $post_type_object ) {
+
+								// if the post isn't an instance of a WP_Post or Post model, return
+								if ( ! $post instanceof \WP_Post && ! $post instanceof Post ) {
```
justlevine, line 719: see above re. the return type, if were accepting both a `WP_Post` and a `Post` model

src/Type/ObjectType/RootQuery.php
```diff
@@ -790,10 +803,16 @@ function ( $post ) use ( $post_type_object ) {
 						return $context->get_loader( 'post' )->load_deferred( $post_id )->then(
 							function ( $post ) use ( $post_type_object ) {
 
+								// if the post type object isn't an instance of WP_Post_Type, return
 								if ( ! $post_type_object instanceof \WP_Post_Type ) {
 									return null;
 								}
 
+								// if the post isn't an instance of a WP_Post or Post model, return
+								if ( ! $post instanceof \WP_Post && ! $post instanceof Post ) {
```
justlevine, line 812: See above re the return type, if were accepting both a `WP_Post` and a `Post` model.

src/Mutation/PostObjectUpdate.php
```diff
@@ -160,8 +160,13 @@ public static function mutate_and_get_payload( $post_type_object, $mutation_name
 			/**
 			 * Throw an exception if the post failed to update
 			 */
-			if ( 0 === $post_id || is_wp_error( $updated_post_id ) ) {
```
justlevine, line 163: `0 === $post_id` is already ruled out on [line 80](https://github.com/wp-graphql/wp-graphql/blob/d5a2c6c1b77e2aff401f971ad3e9fe0863f850d1/src/Mutation/PostObjectUpdate.php#L80), so we can use the same logic /throws we do elsewhere.

src/Type/ObjectType/RootQuery.php
jasonbahl: ah! Yes, this is actually already a Post model, not a WP_Post. I thought I had changed it before I committed last night, but I apparently did not. Good catch! 🙌🏻

.github/workflows/code-quality.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           restore-keys: php-${{ matrix.php }}N-${{ runner.os }}-composer
 
       - name: Install dependencies
-        run: composer install --no-progress
+        run: composer update --no-progress && composer require --dev szepeviktor/phpstan-wordpress phpstan/phpstan
```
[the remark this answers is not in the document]
jasonbahl: It probably is enough.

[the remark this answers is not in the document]
jasonbahl: confirmed

Issue #2362, opened 2022-05-04
MenuItems should have a relative `uri`.


josephfusco:
As a headless WordPress user, I'd like menu paths to be relative to the current WordPress site in order to manage multiple Faust frontends from a multisite using subdirectories.

### Current
<img width="1385" alt="GraphQL response showing multisite subdirectory prepended to path" src="https://user-images.githubusercontent.com/6676674/166810643-64b21e8f-b7d7-4f92-a8ab-158f9b17bc20.png">

### Expected
<img width="1385" alt="GraphQL response showing menu paths without multisite subdirectory" src="https://user-images.githubusercontent.com/6676674/166810714-51da5929-844b-48f1-8ef2-1931b88ea721.png">

## Environment

- **wordpress**: `5.9.3`
- **wp-graphql**: `1.8.0`
- **faustwp**: `0.7.7`

## Related Issues
- https://github.com/wp-graphql/wp-graphql/issues/259
- https://github.com/wp-graphql/wp-graphql/issues/1001

justlevine:
Hey @josephfusco,
If I understand correctly, the issue isn't that the `path` has the subdirectory in it (it's supposed to be the directory path), but rather that `menuItems` don't have a `uri` field (i.e. a relative permalink).
Please let me know if I'm mistaken.

PS: I don't know your actual site config, but I'm assuming the `url` shouldnt be `localhost:3000/my-slug`, but `localhost:3000/my-subdir/my-slug`. If so that's definitely a bug. Can you confirm your subsite's home url (and test with faust disabled to see if the problem is with the plugin or with faust)?

josephfusco:
@justlevine Correct, my semantics may be off with where the relative permalink should live, but the goal would be to have a a multisite subdirectory site be the primary backend for Faust/Next application without the subdirectory in the Next route.

I would agree with your observation that `localhost:3000/my-slug` should be `localhost:3000/my-subdir/my-slug` as the subdirectory exists in the URL with faustwp deactivated.

<img width="1192" alt="Screen Shot 2022-05-05 at 8 54 56 AM" src="https://user-images.githubusercontent.com/6676674/166927277-15b89bd4-43ef-4052-bb1c-4922b223dcc2.png">

josephfusco:
See comment https://github.com/wp-graphql/wp-graphql/pull/2363#issuecomment-1119710067

Issue #2275, opened 2022-03-07
Setup tests to run in multisite


jasonbahl:
As the maintainer of WPGraphQL, I would like to run tests in a multisite environment, to ensure functionality works as expected on multisite setups. 

Currently all tests are running in single site WordPress installs. 

We had a regression to a multisite feature #2274 and we should be able to prevent this by having test suite(s) run in multisite. 

I would like to test #2274 as the first test in the suite(s) that run in multisite.

jasonbahl:
I have this working locally! 👏🏻 

With an environment variable we can now run tests in multisite or non-multisite

Running in multisite has exposed some existing bugs with our plugin resolvers, so I'm investigating that so I can get the initial tests passing when I open my PR

PR #2363, opened 2022-05-04
Add relative `uri` to MenuItem

josephfusco:
This fixes an issue with using wp-graphql in a WordPress multisite using subdirectories where the individual site subdirectory is included with the menu path.

https://github.com/wp-graphql/wp-graphql/issues/2362

**Before**
Screen Shot 2022-05-04 at 3 24 03 PM

**After**
Screen Shot 2022-05-04 at 3 23 49 PM

Where has this been tested?
---------------------------
**Operating System:** macOS 12.3.1
**WordPress Version:** 5.9.2

### DIFF
--- a/src/Type/ObjectType/MenuItem.php
+++ b/src/Type/ObjectType/MenuItem.php
@@ -118,6 +118,12 @@ class MenuItem {
 						'type'        => 'String',
 						'description' => __( 'URL or destination of the menu item.', 'wp-graphql' ),
 					],
+					// Note: this field is added to the MenuItem type instead of applied by the "UniformResourceIdentifiable" interface
+					// because a MenuItem is not identifiable by a uri, the connected resource is identifiable by the uri.
+					'uri'              => [
+						'type'        => 'String',
+						'description' => __( 'The uri of the resource the menu item links to', 'wp-graphql' ),
+					],
 					'path'             => [
 						'type'        => 'String',
 						'description' => __( 'Path for the resource. Relative path for internal resources. Absolute path for external resources.', 'wp-graphql' ),
--- a/src/Type/ObjectType/MenuItem.php
+++ b/src/Type/ObjectType/MenuItem.php
@@ -21,7 +21,7 @@ class MenuItem {
 			'MenuItem',
 			[
 				'description' => __( 'Navigation menu items are the individual items assigned to a menu. These are rendered as the links in a navigation menu.', 'wp-graphql' ),
-				'interfaces'  => [ 'Node', 'DatabaseIdentifier' ],
+				'interfaces'  => [ 'Node', 'DatabaseIdentifier', 'UniformResourceIdentifiable' ],
 				'connections' => [
 					'connectedNode' => [
 						'toType'               => 'MenuItemLinkable',
@@ -113,10 +113,6 @@ class MenuItem {
 					'title'            => [
 						'type'        => 'String',
 						'description' => __( 'Title attribute for the menu item link', 'wp-graphql' ),
-					],
-					'uri'              => [
-						'type'        => 'String',
-						'description' => __( 'The relative path to the current site.', 'wp-graphql' ),
 					],
 					'url'              => [
 						'type'        => 'String',
--- a/src/Type/ObjectType/MenuItem.php
+++ b/src/Type/ObjectType/MenuItem.php
@@ -21,7 +21,7 @@ class MenuItem {
 			'MenuItem',
 			[
 				'description' => __( 'Navigation menu items are the individual items assigned to a menu. These are rendered as the links in a navigation menu.', 'wp-graphql' ),
-				'interfaces'  => [ 'Node', 'DatabaseIdentifier', 'UniformResourceIdentifiable' ],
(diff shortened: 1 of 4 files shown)


justlevine:
I think a better approach to this would be to add a `uri` field in addition to the existing `url`, instead of changing the expectation (and breaking behavior) of either `url` (a fully qualified url) and `path` (the relative path on the server).

This is also how `User` objects do it:
image

josephfusco:
@justlevine I added a new `uri` field rather than modifying the pre-existing `path` as suggested.

Here's how it looks now

Screen Shot 2022-05-05 at 1 08 47 PM

justlevine:
Hey @josephfusco , I tweaked your code a bit to follow the patterns used for other `uri` fields. Can you confirm it still works on multisite? (It should be fine, but multisite support is still a bit spotty ).

It seems that `MenuItems` aren't really tested, so I'm gonna take a minute to backfill those before marking this ready.

josephfusco:
Hey @justlevine I greatly appreciate your help with this!

After giving this more thought, I'm going to close this out add to fix to the faustwp WordPress plugin. It feels like the more appropriate place to put this seeing that faustwp is the one that is replacing the subdirectory.

justlevine:
@josephfusco I believe this should stay open. While its true that Faust is doing a bad replace on the url, there are a lot of use cases for a relative uri, and it makes the object behave more uniformly with other `Nodes` that can have an associated link. 

If you prefer, I can create a separate PR - lemme know!

josephfusco:
@justlevine Works for me! Feel free to keep it on this PR :)

justlevine:
@jasonbahl not sure why the schema linter is failing 🤔

jasonbahl:
@justlevine looks probably related with this latest release of the schema linter: https://github.com/cjoudrey/graphql-schema-linter/releases/tag/v3.0.0

jasonbahl:
looks like we'll need to explicitly install graphql in the workflow now

jasonbahl:
I just fixed the Schema linter in #2368

justlevine:
@jasonbahl  And it's passing again 🚀. Ready for your review :)

jasonbahl:
Ah issue for multisite tests here: https://github.com/wp-graphql/wp-graphql/issues/2275

jasonbahl:
Ok, so I got #2375 merged, so now we have tests running in multisite! 

I'll work up a test for this so we can get this merged.

justlevine:
> I'll work up a test for this so we can get this merged.

Only thing that should be missing is an `assertEquals()` for `$actual['data']['menuItem']['path']`;

src/Type/ObjectType/MenuItem.php
```diff
@@ -21,7 +21,7 @@ public static function register_type() {
 			'MenuItem',
 			[
 				'description' => __( 'Navigation menu items are the individual items assigned to a menu. These are rendered as the links in a navigation menu.', 'wp-graphql' ),
-				'interfaces'  => [ 'Node', 'DatabaseIdentifier' ],
+				'interfaces'  => [ 'Node', 'DatabaseIdentifier', 'UniformResourceIdentifiable' ],
```
jasonbahl, line 24: Ah, so even though the `MenuItem` type has a `uri` field, Menu Items are not actually uniquely identifiable by a URI. 

If you take the `MenuItem.uri` field and do a `nodeByUri( $uri: MenuItem.uri ) { ... }` query, you will get another node (Post, Page, etc)

The `MenuItem` Type should not implement the `UniformResourceIdentifiable` interface, as a MenuItem is not identifiable by a URI. It links to other nodes that are identifiable by uri, but it by itself is not.

jasonbahl: If we look in the Schema now for the `nodeByUri` field, we see that it returns `UniformResourceIdentifiable` Interface, which now has `MenuItem` as a Possible Type, which is misleading, as `nodeByUri` will never return an object with type `MenuItem`

CleanShot 2022-05-13 at 16 32 49

jasonbahl:
I left some notes about the `uri` field / `UniformResourceIdentifiable` Interface. 

I think I'm ok with having the `uri` field as a non-breaking change here, but I think we should remove the `UniformResourceIdentifiable` interface, as this Type is not actually identifiable by a uri. The nodes that MenuItem links to are identifiable by the uri, but not the MenuItem itself.

jasonbahl:
I pushed up a commit to remove `UniformResourceIdentifiable` from the `MenuItem` type

jasonbahl:
Pushed a follow-up commit adding the `uri` field back to the `MenuItem` Type since it was removed when removing the `UniformResourceIdentifiable` Interface from the Type. 

Now, we can query the `uri` field of a MenuItem, but `MenuItem` will not be shown as a `possibleType` in `nodByUri` queries.

jasonbahl:
I think we still might need to have a specific test where we do something like the following:

- create a new blog in the multisite
- switch to blog
- create a menu with some menu items 
- assign the menu to a location (so it's public)
- query the menu
- assert that the blog directory is _not_ in the uri

🤔

PR #2376, opened 2022-05-13
Bug/#2371 auth callback should work for mutations

### LINKED ISSUE #2371: Add hooks and filters for mutations
### Feature request
The ability to alter the capabilities required to perform mutations on a given custom post type.

(Previously discussed in #2365.)

### Current situation
Changing capabilities to create or update a post is tricky today because permission checks are not easy to adjust — they're baked into the `mutate_and_get_payload` anonymous function in `wp-graphql/src/Mutation/PostObjectCreate.php`:

https://github.com/wp-graphql/wp-graphql/blob/c1c2255762394b5ec56ce6b8f6e56cbb212433b7/src/Mutation/PostObjectCreate.php#L200-L206

### Possible solutions
- Extract permissions checks for all mutations into a `permissionCallback` that runs at the beginning of the mutation logic (after type/input checking, but before mutations occur), and returns a filterable boolean to allow/deny the mutation.
- Wrap existing capabilities checks with filters.

Both options would need to pass in some contextual info (the mutation/post type they're acting on).

jasonbahl:
This updates `register_graphql_mutation` to work with the current `auth` and `isPrivate` methods for restricting resolvers at the Schema level. 

related to: #2371 

Given the following snippet: 

```php
register_graphql_mutation( 'isPrivateMutation', [
	'inputFields' => [],
	'outputFields' => [
		'test' => [
			'type' => 'String'
		]
	],
	'isPrivate' => true,
	'mutateAndGetPayload' => function() use ( $expected ) {
		return [
			'test' => 'test_value';
		];
	}
]);
```

The following query can be executed: 

```php
mutation {
  isPrivateMutation(input:{ clientMutationId: "test" }) {
    test
  }
}
```

### BEFORE

As a public user, I'm able to execute the mutation and get the data back

CleanShot 2022-05-13 at 15 54 56

### AFTER

Now, we can see the mutation is properly restricted, because of the `isPrivate` config is set to true on the mutation

CleanShot 2022-05-13 at 15 54 11

Any other comments?
-------------------
This doesn't solve/close #2371. It just brings an existing layer of auth restrictions at the Schema level to be supported by mutations, and adds some better tests around this functionality.

### DIFF
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -1070,7 +1070,7 @@ class TypeRegistry {
 		$this->register_field(
 			'rootMutation',
 			$mutation_name,
-			[
+			array_merge( $config, [
 				'description' => sprintf( __( 'The payload for the %s mutation', 'wp-graphql' ), $mutation_name ),
 				'args'        => [
 					'input' => [
@@ -1099,7 +1099,7 @@ class TypeRegistry {
 
 					return $payload;
 				},
-			]
+			])
 		);
 
 	}
--- a/src/Utils/InstrumentSchema.php
+++ b/src/Utils/InstrumentSchema.php
@@ -216,7 +216,7 @@ class InstrumentSchema {
 			return true;
 		}
 
-		if ( ! isset( $field->config['auth'] ) || ! is_array( $field->config['auth'] ) ) {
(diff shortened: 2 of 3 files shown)


justlevine (approved):
Looks good and working locally on my end.

Issue #2384, opened 2022-05-17
Add helper method for retrieving the Frontend base URL


justlevine:
## What

We should add a filterable helper method to WPGraphQL for retrieving the site's frontend URL (and possibly the relative URI).

## Why
The `home_url()` method is used in many different contexts, both by WP core and 3rd-party plugins. That makes things like url replacements (a common need in headless environments where the actual frontend and the "wp frontend" aren't the same) a major pain, and can cause inconsistencies and edge cases that often require additional code to handle.

Creating a helper method would both make WPGraphQL DRY-er and more testable, and make it easier for devs trying to support 3rd-party plugins by providing a single consistent source of truth to alter.

Two recent examples of this issue (I'm sure there's more if I did a deep dive):
- #2324
- #2341

## How
Im thinking something along the lines of
```php

public static function get_home_url() : string {
   $home_url = home_url();
   
   return apply_filters( 'graphql_home_url', $home_url );
}
```

We could even DRY out the `str_replace()` used [throughout the plugin](https://github.com/wp-graphql/wp-graphql/search?home_url) with the following:

```php

public static function get_relative_uri( string $url ) {
  $home_url = WPGraphQL::get_home_url();
  
  return str_ireplace( $home_url, '', $url );
}
```

Extension authors (and core) could then use the above methods without needing to worry about possible side effects from `home_url()`, and users/plugin authors who want to reference a headless frontend instead of WP's can just use the filter once, instead of having to handle the various schema use cases separately. 

E.g. for FaustJS - a significant amount of [their replacements code](https://github.com/wpengine/faustjs/tree/1762188abddeef809916b31d415851da3dfcb826/plugins/faustwp/includes/replacement) (which currently doesn't  account for all possible edge cases) could be reduced to a single filter call:

```php

add_filter( 'graphql_home_url', 
  function( string $url ) {
    return  faustwp_get_setting( 'frontend_uri' );
  }
);
```

<!--
### When you open an _issue_ for a feature request, please add as much detail as possible:
- [x] A descriptive title
- [x] A description of the problem you're trying to solve, including *why* you think this is a problem
- [x] An overview of the suggested solution
- [x] If the feature changes current behavior, reasons why your solution is better
--!>

renatonascalves:
I'd argue that `home_url` is that source of truth. It has a function and a hook: https://developer.wordpress.org/reference/hooks/home_url/

Why introduce another layer of functions/methods and hooks?

[...]

PR #2386, opened 2022-05-18
Feat/add get home url helper

justlevine:
Adds the `Utils::get_home_url()` and `Utils::get_relative_uri()` helper functions, along with the `graphql_home_url` filter, to allow for DRY and consistent `uri`s regardless of site setup (regular, subsite, subdir, host replacement, etc).

#2384 

Any other comments?
-------------------
- This PR **does not** apply the `graphql_home_url` filter to fully-qualified links, such as those returned from `get_permalink()`. To do so would require `str_replace()` on all those fields, which _to me_ feels like unnecessary overhead. 
IMO devs should ideally be using WP's `home_url`, and I can't think of a use case where this filter would be required over something along the lines of :
```php
add_filter( 'home_url', function( $url, $path, $scheme ) {
  if( is_graphql_request() || 'graphql' === $scheme ) {
    $url = 'my-frontend.com' . $path;
  } 
  return $url;
}, 10, 3 );
```
Would be helpful if someone less biased than myself would review and decide if filtering fully-qualified links is necessary.

- The new methods *are not* currently exposed as access functions. I'm unclear of the criteria that makes a method 'worthy' of an access function over a direct static method call.

Where has this been tested?
---------------------------

**WordPress Version:** 5.9.3

### DIFF
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -7,6 +7,7 @@ use WP;
 use WP_Post;
 use WPGraphQL\AppContext;
 use WPGraphQL\Model\Post;
+use WPGraphQL\Utils\Utils;
 
 class NodeResolver {
 
@@ -102,6 +103,7 @@ class NodeResolver {
 				[
 					wp_parse_url( site_url() )['host'],
 					wp_parse_url( home_url() )['host'],
+					wp_parse_url( Utils::get_home_url() )['host'],
 				],
 				true
 			) ) {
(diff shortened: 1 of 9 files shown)


jasonbahl (approved):
Good stuff! Thanks for this ❤️

justlevine:
(@jasonbahl Im actually a bit surprised this got merged without more discussion on #2384 and #2341. deferring to you, but you might want to follow up in those threads).

jasonbahl:
@justlevine maybe I did jump the gun. I'll follow up on those threads and see if I have more thoughts

Issue #2297, opened 2022-03-14
Version missing from EnqueuedAsset types


benknight:
Script and stylesheet versions don't resolve correctly because they reference an undefined filed named `$version` on the dependency objects. Desired behavior for the resolver would be to return `$ver` if it's set, and fall back to `$default_version` so that consumers can use query string `?v=...` to manage caching the same way native themes do.

https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/InterfaceType/EnqueuedAsset.php#L53

See:

https://developer.wordpress.org/reference/classes/wp_dependencies/
https://developer.wordpress.org/reference/classes/_wp_dependency/
https://developer.wordpress.org/reference/classes/wp_scripts/

justlevine:
Hey @benknight thanks for reporting the issue, and the possible fix!

Few notes after investigation:
- Many scripts and stylesheets dont have a `$ver` specified in `global $wp_{scripts|styles}`, which as far as I can tell is on only programmatic way to get the list of assets. Please correct me if I'm wrong.
- The `$default_version` doesnt exist on the individual `_WP_Dependency`, but instead is a property of WP_{Scripts|Styles}. We can get it by getting both globals and then using the `$type` to get the property dynamically.

> consumers can use query string `?v=...` to manage caching the same way native themes do.

Not 100% I understand what this would look like in headless. My experience with managing scripts/styles in WP is mainly configuring caching plugins (in native), or stripping them out entirely (aka why I ♥ decoupled ). So while this seems like an easy PR, it's probably best someone who understands the nuances writes it.

### Code sample
(Assuming there arent any special cases to consider)
```php
// EnqueuedAsset.php::L53
'version' => [
  'type'        => 'String',
  'description' => __( 'The version of the enqueued asset', 'wp-graphql' ),
  'resolve' => static function( $asset ) {
     global $wp_styles, $wp_scripts;
   
     $version = ! empty( $asset->ver ) ? $asset->ver : ${'wp_' . $asset->type . 's'}->default_version;
     return $version ?: null;
  },
]
```

benknight:
> Not 100% I understand what this would look like in headless. My experience with managing scripts/styles in WP is mainly configuring caching plugins (in native), or stripping them out entirely (aka why I ♥ decoupled ). So while this seems like an easy PR, it's probably best someone who understands the nuances writes it.

[...]

PR #2387, opened 2022-05-18
fix: resolve asset versions

justlevine:
This PR fixes issues with `EnqueuedAsset` types not always having a set version, providing the [default version](https://developer.wordpress.org/reference/classes/wp_dependencies/) when none is explicitly set.

#2297 

Before: 
image

After:
image

Any other comments?
-------------------
Also backfilled tests for _some_ of the other missing `EnqueuedAsset` fields: `extra`, `handle`, `src`. Unsure how to test `dependencies`, since we can't query for a specific asset, nor predict what array position in `nodes` will contain an asset with a set dependency.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (Wsl2 + devilbox + php 8.0.15)

**WordPress Version:** 5.9.3

### DIFF
--- a/src/Type/InterfaceType/EnqueuedAsset.php
+++ b/src/Type/InterfaceType/EnqueuedAsset.php
@@ -52,7 +52,7 @@ class EnqueuedAsset {
 				],
 				'version'      => [
 					'type'        => 'String',
-					'description' => __( 'The version of the enqueued asset', 'wp-graphql' ),
+					'description' => __( 'The version of the enqueued asset. Defaults to the current version of WordPress.', 'wp-graphql' ),
 				],
 				'src'          => [
 					'type'        => 'String',
--- a/src/Type/ObjectType/EnqueuedScript.php
+++ b/src/Type/ObjectType/EnqueuedScript.php
@@ -21,7 +21,7 @@ class EnqueuedScript {
 			'description' => __( 'Script enqueued by the CMS', 'wp-graphql' ),
 			'interfaces'  => [ 'Node', 'EnqueuedAsset' ],
 			'fields'      => [
-				'id'  => [
+				'id'      => [
 					'type'    => [
 						'non_null' => 'ID',
 					],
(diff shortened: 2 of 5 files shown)


src/Type/InterfaceType/EnqueuedAsset.php
```diff
@@ -52,7 +52,7 @@ public static function register_type( TypeRegistry $type_registry ) {
 				],
 				'version'      => [
 					'type'        => 'String',
-					'description' => __( 'The version of the enqueued asset', 'wp-graphql' ),
+					'description' => __( 'The version of the enqueued asset. Defaults to the current version of WordPress.', 'wp-graphql' ),
```
jasonbahl, line 55: I wonder if we should not disclose this as it could leave sites vulnerable. 

If a security report were announced about a specific version of WordPress and folks were able to identify that any given site was using that version from this query, it would help bad actors identify which sites are vulnerable. 🤔

src/Type/ObjectType/EnqueuedScript.php
```diff
@@ -21,19 +21,26 @@ public static function register_type() {
...
 					'resolve' => function ( \_WP_Dependency $script ) {
 						return isset( $script->src ) && is_string( $script->src ) ? $script->src : null;
 					},
 				],
+				'version' => [
+					'resolve' => function ( \_WP_Dependency $script ) {
+						global $wp_scripts;
+
+						return isset( $script->ver ) && is_string( $script->ver ) ? (string) $script->ver : $wp_scripts->default_version;
+					},
```
jasonbahl, line 42: along the lines of my other comment, when we default to the current WordPress version, perhaps we should hash it to make it an arbitrary value? maybe something like:

```php
$version = \WPGraphQL\Router::$route . $wp_scripts->default_version;
$hashed_version = md5( $version );
```

jasonbahl (changes requested):
I left some comments about the version and one idea for it.

justlevine:
@jasonbahl If we're really worried about security, the version number for the asset could also tell malicious actors about vulnerabilities.

That said, both a 'naked' version number, and using WP's version as a fallback are [default WP behavior](https://github.com/WordPress/WordPress/blob/3fec80568e955b822c39df4075d9e97acc1e4ef2/wp-includes/class.wp-dependencies.php#L246), so I'll leave the question to you as to how far we want to stray from core WP?

Two possible alternatives:

- A recipe that replaces version numbers with hashes, for those who want to security harden their installs without a plugin like AIO Security or WordFence that obsfucates things like this.
- Take a page from `RAW`/`RENDERED`, and add a query arg that allows _authenticated_ users to get the version but defaults to the hash. Not sure if this is technically breaking from a GraphQL perspective, but it would break FaustJS (as GQTY converts fields with args to functions).

jasonbahl:
Perhaps we leave it as is for now, but leave the comment out of the description. 🤔

justlevine:
okay I fixed the description

jasonbahl (approved):

PR #2396, opened 2022-05-28
chore: merge master branch to develop

justlevine:
What does this implement/fix? Explain your changes.
---------------------------------------------------
[v1.8.2](https://github.com/wp-graphql/wp-graphql/releases/tag/v1.8.2) was merged directly with the `master` branch, instead of the usual `release/branch` -> `develop` -> `master`.

This PR pulls those changes back onto the `develop` branch.

Does this close any currently open issues?
------------------------------------------
…

Any relevant logs, error output, GraphiQL screenshots, etc?
-------------------------------------
(If it’s long, please paste to https://ghostbin.com/ and insert the link here.)

Any other comments?
-------------------
Might make sense to set some [branch protection rules]( https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule) to enforce the following rules:
1. Only PRs from `develop` can be merged with master.
2. All other PRs must be merged to `develop`.

Ref: 
- https://carldesouza.com/prevent-direct-merges-to-master-with-branch-protection-in-github/
- https://github.community/t/improvement-suggestion-settings-branch-pattern-that-can-only-merge-to-target-branch-pattern/171870/2

Where has this been tested?
---------------------------
**Operating System:** N/A

**WordPress Version:** N/A

Issue #2394, opened 2022-05-26
NodeByUri will not query WooCommerce products since v1.8.1


jmotes-lapero:
v1.8.2:

<img width="946" alt="image" src="https://user-images.githubusercontent.com/94563084/170400530-596198d8-d881-4933-8b3a-cb29e6cab311.png">

v1.8.0:

<img width="1072" alt="image" src="https://user-images.githubusercontent.com/94563084/170400578-c2634191-6dd8-4d84-a1b1-11275c2660cb.png">

I'm using WooGraphQL v0.11.0 but experienced the issue with v0.10.5 as well.

jasonbahl:
@jmotes-lapero Can you provide a bit more info to reproduce this? 

Like, are you able to reproduce this on a minimal install of WordPress with just WPGraphQL, WooCommerce and WooGraphQL and nothing else? 

If so, what data set is needed to reproduce this?

Is the product published? etc

jasonbahl:
@jmotes-lapero does it work in v1.8.1 as well? Or is that version broken too? Would be good to narrow down the culprit to 1.8.1 or 1.8.2 if possible.

jmotes-lapero:
@jasonbahl it breaks starting with 1.8.1 in our project. I'll work on a minimal reproduction today. I assumed it would be easy to reproduce since 1.8.0 works fine for us, but sounds like I was wrong. Sorry about that!

jmotes-lapero:
Rolling back to 1.8.0:

<img width="1306" alt="image" src="https://user-images.githubusercontent.com/94563084/170510127-11ae3083-30ec-4bdf-b7b1-0e343dd1531f.png">

And on 1.8.1:

<img width="1108" alt="image" src="https://user-images.githubusercontent.com/94563084/170510337-68f5dd1f-2833-42e9-8c8f-88f4a4390778.png">

justlevine:
@jmotes-lapero Are you sure the only thing that changed is the WPGraphQL version and not maybe the WC version? Because I *cant* get it to work in v1.8.0, v1.7.2 or v1.6.12.

**v1.8.0**
![image](https://user-images.githubusercontent.com/29322304/170587567-5189e7cd-5a7d-42d4-9844-a1f5ad83cbb3.png)

**v1.6.12**
![image](https://user-images.githubusercontent.com/29322304/170588413-3fdc6941-a06a-4fb6-83eb-1ba87566571f.png)

justlevine:
[...]

PR #2399, opened 2022-05-28
feat: schema customization with post type args

justlevine:
Adds additional WPGraphQL-specific `$args` to `WP_Post_Type` and `WP_Taxonomy` for granular registration/modification of PostObject and TermObject GraphQL types.

Builds off of #1486, with feedback from #2394.

### What's changed
* dev: Use `register_post_type_args` and `register_taxonomy_args` filters for setting WPGraphQL default args (instead of directly on the globals ), so they can be overruled by the user.
* feat: adds the following WPGraphQL properties to `WP_Post_Type` and `WP_Taxonomy`:
```php
// The "kind" of GraphQL type to register. Can be `interface`, `object`, or `union`.
'graphql_kind'                     => 'object',
// The callback used to resolve the type. Only used if `graphql_kind` is an `interface` or `union`.
'graphql_resolve_type'             => null,
// An array of custom interfaces the type should implement.
'graphql_interfaces'               => [],
// An array of default interfaces the type should exclude.
'graphql_exclude_interfaces'       => [],
// An array of custom connections the type should implement.
'graphql_connections'              => [],
// An array of default connection **field names** the type should exclude.
'graphql_exclude_connections'      => [],
// An array of possible type the union can resolve to. Only used if `graphql_kind` is a `union`.
'graphql_union_types'              => [],
// Whether to register default fieldto the schema.
'graphql_register_root_field'      => true,
// Whether to register default connections to the schema.
'graphql_register_root_connection' => true,
// Whether to register the "create" mutation for this post type. Default true.
'graphql_register_create_mutation' => true,
// Whether to register the "update" mutation for this post type. Default true.
'graphql_register_update_mutation' => true,
// Whether to register the "delete" mutation for this post type. Default true.
'graphql_register_delete_mutation' => true,
```
* chore: consolidate `PostObject` and `TermObject` registration logic into `WPGraphQL\Registry\Utils\PostObject` and `WPGraphQL\Registry\Utils\TermObject`. 

### To do:
- [ ] ~Allow user to set custom DataLoader / Model.~ (punted to #2466)
- [x] Add tests for _custom_ post types and taxonomies.

#1486

Any other comments?
-------------------
- ~Had to cherry-pick from #2396 to get this to work with `wp-graphql-testcase`.~ (Rebased on `develop`). 
-  Deprecated `WPGraphQL\Type\ObjectType\PostObject` and `WPGraphQL\Type\ObjectType\TermObject`, since those namespaces dont make sense when a Post / Term object can be an `interface | object | union` type.
- The `todo` tags need to be replaced with the correct version number. 
- ~I might need help with overloading the dataloader/model. I'm not 100% percent clear where in the lifecycle these need to be replaced, and how to do this in a way that doesn't cripple the connection resolvers.~ (punted to #2466)
- This PR **does not** include the following granular settings from #1486
  - [`connect_to_comments`](https://github.com/wp-graphql/wp-graphql/pull/1486/files#r499819148),  [`connect_to_taxonomies`](https://github.com/wp-graphql/wp-graphql/pull/1486/files#r499824486), [`supports_content_node`](https://github.com/wp-graphql/wp-graphql/pull/1486/files#r499823901) - can be handled with `graphql_exclude_{interfaces|connections}`.
  - [`resolve_from_uri`](https://github.com/wp-graphql/wp-graphql/pull/1486/files#r499822189)
  - [`register_schema_type`](https://github.com/wp-graphql/wp-graphql/pull/1486/files#r499825867)
  - [`supports_menu_item_object_union`](https://github.com/wp-graphql/wp-graphql/pull/1486/files#r499822945)
  - [`supports_post_object_union`](https://github.com/wp-graphql/wp-graphql/pull/1486/files#r499823651)
Some of these are unnecessary, whereas others seem to require more discussion / a clear use case.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 ( WSL2 + devilbox + PHP 8.0.15 )

**WordPress Version:** 6.0


src/WPGraphQL.php
```diff
@@ -419,50 +421,112 @@ public function init_admin() {
 	 * @return void
 	 */
 	public static function show_in_graphql() {
```
justlevine, line 423: This whole file is getting pretty large. Might be time to break the post type / taxonomy registration stuff to a separate class.

If so what should we name it? (Extensions use `WPGraphQL/CoreSchemaFilters`, but that doesn't really mesh with our file structure.

src/Connection/TermObjects.php
```diff
@@ -52,15 +52,15 @@ public static function register_connections() {
...
 		/**
 		 * Loop through the allowed_taxonomies to register appropriate connections
 		 */
 		foreach ( $allowed_taxonomies as $tax_object ) {
+			if ( ! $tax_object->graphql_register_root_connection ) {
+				continue;
+			}
 
-			$root_query_from_field_name = $tax_object->graphql_plural_name;
+			$root_query_from_field_name = lcfirst( $tax_object->graphql_plural_name );
```
jasonbahl, line 63: good catch. 

should this instead use `\WPGraphQL\Utils\Utils::format_field_name( $tax_object->graphql_plural_name )`?

src/Data/PostObjectType.php
```diff
@@ -0,0 +1,517 @@
+<?php
```
jasonbahl, line 1: I feel like this file belongs under Registry vs Data, since the goal here is to add things to the registry. 

I know there are some things in the Data directory already doing similar things (re: settings), and I think that should get moved out too. (tbh, settings need a bit of an overhaul in general, but different discussion)

jasonbahl (commented):
did a quick scan and left some comments. will come back and look in more depth shortly

src/Data/PostObjectType.php
justlevine: Im down for this change.

PS: I was modeling it after the `Data/{Type}Mutation classes (not settings), and breaking away from that pattern could allow us to relocate some of the extra code in `WPGraphQL.php` while still maintaining logical consistency in the class. 

Any specific [ suggestions on the naming](https://github.com/wp-graphql/wp-graphql/pull/2399#discussion_r884188865)?

justlevine:
Apologies for the status spam. commit https://github.com/justlevine/wp-graphql/commit/2d38d1f3bdc29914300b4705076192af645c0183 broke my git history.
Had to reset and reapply the commits manually. 

On the bright side, its now rebased on the latest `develop`

justlevine:
@jasonbahl I dont see a clear path towards model/dataloader support as those are usually hard-coded and called devoid of the `WP_Post_Type` / `WP_Taxonomy` object. 

I'm open to suggestions, but it might make sense to polish this off with the remaining tests, and leave that part for a future PR.

jasonbahl:
>  @jasonbahl I dont see a clear path towards model/dataloader support as those are usually hard-coded and called devoid of the WP_Post_Type / WP_Taxonomy object.

> I'm open to suggestions, but it might make sense to polish this off with the remaining tests, and leave that part for a future PR.

@justlevine ya, let's hold off on that. I have _some_ ideas for sure, but we can tackle it separately.

jasonbahl:
Ooops. Accidentally clicked "Close with comment" instead of "Comment" 🤦🏻‍♂️

src/WPGraphQL.php
```diff
@@ -422,50 +424,112 @@ public function init_admin() {
...
 
+	/**
+	 * Sets up the default post types to show_in_graphql.
+	 *
+	 * @param array  $args      Array of arguments for registering a post type.
+	 * @param string $post_type Post type key.
+	 *
+	 * @return array
+	 */
+	public static function setup_default_post_types( $args, $post_type ) {
```
justlevine, line 443: @jasonbahl I wanted to move this and `setup_default_taxonomies()` into `Registry\Utils\{Post|Term}Object`, but not sure if that still makes sense given the `Utils` namespace.  

Thoughts?

src/WPGraphQL.php
```diff
@@ -422,50 +424,115 @@ public function init_admin() {
...
+			// The callback used to resolve the type. Only used if `graphql_kind` is an `interface` or `union`.
+			'graphql_resolve_type'             => null,
+			// An array of custom interfaces the type should implement.
+			'graphql_interfaces'               => [],
+			// An array of default interfaces the type should exclude.
+			'graphql_exclude_interfaces'       => [],
+			// An array of custom connections the type should implement.
+			'graphql_connections'              => [],
+			// An array of default connection field names the type should exclude.
+			'graphql_exclude_connections'      => [],
```
justlevine, line 516: @jasonbahl should this be renamed to `graphql_exclude_connection_field_names`, to prevent users from thinking they can pass the "to" GraphQL type name (e.g. 'Comment' instead of 'comments'), or is this obvious?

src/WPGraphQL.php
```diff
@@ -422,50 +424,115 @@ public function init_admin() {
...
+			// Whether to register default connections to the schema.
+			'graphql_register_root_field'      => true,
+			'graphql_register_root_connection' => true,
+		];
+
+		$args = wp_parse_args( $args, $defaults );
+
```
justlevine, line 523: @jasonbahl I think this would be the perfect place for an `apply_filters()`, so users can edit the post type / taxonomy args without worrying about the priority of their `register_{post_type|taxonomy}_args` hooks, but i'm not sure what to call it.

We only have the cpt/tax _name_, not whether the item is a cpt/tax, and `grapqhl_register_{my_custom_wp_type}_args` seems ambiguous to what `args` are actually being filtered.

We could wrap the whole `set_wp_type_args()` in a callable that passes a `string $wp_type post_type|taxonomy` depending on which WP hook is called, but that feels messy to me.

justlevine:
Update:
- added `graphql_exclude_interfaces`, `graphql_exclude_connections`, and `graphql_union_types` properties.
- added test parity to `CustomTaxonomyTest.php`.
- *scaffolded* tests for registering types as unions/interfaces.

*Note*: the interface/union functionality doesnt work yet, because the post type / taxonomy connections in `Registry\Utils` are being registered as duplicate fields. I'm not 100% sure why, or what the ideal solution is (I'm assuming we need to add `$args[graphql_kind]` checks to the connection conditionals, but not the specific logic to use). Could use some 👀.

That's the last blocker before cleanup and ready-to-merge.

kidunot89 (dismissed):
@justlevine Looks good code-wise.

justlevine:
Just realized that when debug is disabled, tests with `markTestIncomplete()` don't come back in the report. Addressing this roundaboutly with #2503 , but to see the interface/union tests actually fail locally, uncomment [L1042-1043](https://github.com/justlevine/wp-graphql/blob/52a1130490b5c935d2d07c1e8148e54965ee6596/tests/wpunit/CustomPostTypeTest.php#L1042-1043) and [L1112-1113](https://github.com/justlevine/wp-graphql/blob/52a1130490b5c935d2d07c1e8148e54965ee6596/tests/wpunit/CustomPostTypeTest.php#L1112-1113) lines in `CustomPostTest.php`, and [L663-664](https://github.com/justlevine/wp-graphql/blob/52a1130490b5c935d2d07c1e8148e54965ee6596/tests/wpunit/CustomTaxonomyTest.php#L663-L664) and [L730-731](https://github.com/justlevine/wp-graphql/blob/52a1130490b5c935d2d07c1e8148e54965ee6596/tests/wpunit/CustomTaxonomyTest.php#L730-L731) lines in `CustomTaxonomyTest.php`

src/WPGraphQL.php
jasonbahl: > but i'm not sure what to call it.

I'm thinking we can change the filters to:

```php
add_filter( 'register_post_type_args', [ __CLASS__, 'register_graphql_post_type_args' ], 99, 2 );
add_filter( 'register_taxonomy_args', [ __CLASS__, 'register_graphql_taxonomy_args' ], 99, 2 );
```

Then, have 2 functions like so: 

```php
	/**
	 * Set the GraphQL Post Type Args and pass them through a filter.
	 *
	 * @param array  $args           The graphql specific args for the post type
	 * @param string $post_type_name The name of the post type being registered
	 *
	 * @return array
	 * @throws Exception
	 */
	public static function register_graphql_post_type_args( array $args, string $post_type_name ) {

		$args = self::set_wp_type_args( $args, $post_type_name );

		/**
		 * Filters the graphql args set on a post type
		 *
		 * @param array  $args           The graphql specific args for the post type
		 * @param string $post_type_name The name of the post type being registered
		 */
		return apply_filters( 'register_graphql_post_type_args', $args, $post_type_name );

	}

	/**
	 * Set the GraphQL Taxonomy Args and pass them through a filter.
	 *
	 * @param array  $args          The graphql specific args for the taxonomy
	 * @param string $taxonomy_name The name of the taxonomy being registered
	 *
	 * @return array
	 * @throws Exception
	 */
	public static function register_graphql_taxonomy_args( array $args, string $taxonomy_name ) {

		$args = self::set_wp_type_args( $args, $taxonomy_name );

		/**
		 * Filters the graphql args set on a taxonomy
		 *
		 * @param array  $args          The graphql specific args for the taxonomy
		 * @param string $taxonomy_name The name of the taxonomy being registered
		 */
		return apply_filters( 'register_graphql_taxonomy_args', $args, $taxonomy_name );

	}
```

Now, similar to how there is `register_taxonomy_args` and `register_post_type_args` in core, there can be `register_graphql_taxonomy_args` and `register_graphql_post_type_args`

src/WPGraphQL.php
```diff
@@ -422,50 +424,127 @@ public function init_admin() {
...
 		}
 
-		// Adds GraphQL support for post formats.
-		if ( isset( $wp_taxonomies['post_format'] ) ) {
-			$wp_taxonomies['post_format']->show_in_graphql     = true;
-			$wp_taxonomies['post_format']->graphql_single_name = 'postFormat';
-			$wp_taxonomies['post_format']->graphql_plural_name = 'postFormats';
+		// Ensure union types have a valid list of possible types.
+		if ( empty( $args['graphql_union_types'] ) && 'union' === $args['graphql_kind'] ) {
+			throw new Exception(
```
jasonbahl, line 539: I think this might be too aggressive. I'm seeing fatal errors when I register a post type like so, then load the admin page: 

```php

	register_post_type( 'testType', [
		'show_in_graphql'     => true,
		'graphql_single_name' => 'TestType',
		'graphql_plural_name' => 'TestTypes',
		'graphql_kind'        => 'union',
	] );

} );
```

I think we should be more graceful with this, and either show an error in GraphQL requests, or skip the registration of the invalid type and show a Debug message. 

In any case, I think it should be limited to GraphQL environment rather than throwing an Exception on every WordPress load.

jasonbahl, line 539: I think this might be too aggressive. I'm seeing fatal errors when I register a post type like so, then load the admin page: 

```php

	register_post_type( 'testType', [
		'show_in_graphql'     => true,
		'graphql_single_name' => 'TestType',
		'graphql_plural_name' => 'TestTypes',
		'graphql_kind'        => 'union',
	] );

} );
```

I think we should be more graceful with this, and either show an error in GraphQL requests, or skip the registration of the invalid type and show a Debug message. 

In any case, I think it should be limited to GraphQL environment rather than throwing an Exception on every WordPress load. 

CleanShot 2022-09-08 at 13 05 32

jasonbahl: If we move the logic to check if both `graphql_kind === 'union' && empty( $post_type_object->graphql_union_types )` to PostObject.php where we call `register_graphql_union_type()` I think we can enable a more graceful experience, which I think will be important for when plugins (Custom Post Type UI, etc) register post types and might allow users to leave fields unpopulated. 

I'd hate to see a full on fatal error for the whole site when it's just a mistake affecting WPGraphQL. 

I think we could do something to the tune of: 

```php
if ( ! empty( $post_type_object->graphql_union_types ) ) {
    
    // Set the possible types for the union.
    $config['typeNames'] = $post_type_object->graphql_union_types;

    register_graphql_union_type( $single_name, $config );

} else {

    graphql_debug( 'Registering a post type with "graphql_kind" => "union" requires "graphql_union_types" to also be set', [ 'registered_post_type_object' => $post_type_object ] );

}
```

And that would allow an existing query to succeed, but get a helpful debug message: 

CleanShot 2022-09-08 at 13 13 44

justlevine: > I'd hate to see a full on fatal error for the whole site when it's just a mistake affecting WPGraphQL.


My rationale for the fatal error is that since the schema API is solely for use by PHP developers, it'd be helpful to ensure mistakes get caught immediately instead of accidently slipping upstream. But i'm cool with the graceful approach too.

Do you also prefer`graphql_debug` over a throwing _GraphQL_ `Error`?

jasonbahl: It's hard to say. I think before plugins like Custom Post Type UI, Atlas Content Modeler, etc started allowing folks to do this stuff in a UI and not just in code, it's made it more of a gray area. 

If someone makes a mistake, I want to make sure it's easy to identify, but want to make sure it doesn't tank their existing client applications.

jasonbahl: I think more often, I'd lean on `graphql_debug` as it helps surface issues, but also allows folks to do their thing. 

Like here, we might end up not adding the new Union type to the schema, but we can still let existing queries execute. 

If they go looking for their new Type in the Schema docs or make a query for it, they'll know it's not registered properly and will have a good debug message for it, but exiting schema will remain unaffected.

jasonbahl: Probably case-by-case though, but feels like a generally good experience. Not so harsh that you can't complete your work, but also not so hidden that you can't troubleshoot easily.

src/WPGraphQL.php
jasonbahl: My snippets might not be fully comprehensive, but probably close-ish.

tests/wpunit/CustomPostTypeTest.php
```diff
@@ -509,4 +511,629 @@ public function testQueryCustomPostTypeWithSameValueForGraphqlSingleNameAndGraph
...
+			'graphql_interfaces'  => [ 'WithInterfaceKind' ],
+		]);
+
+		register_post_type( 'child_type_two', [
+			'public'              => true,
+			'show_in_graphql'     => true,
+			'graphql_single_name' => 'ChildTypeTwo',
+			'graphql_plural_name' => 'ChildTypeTwo',
+			'graphql_interfaces'  => [ 'WithInterfaceKind' ],
+		]);
```
jasonbahl, line 1008: This is actually a pretty confusing experience. 

Here we've registered 3 post types, and will get 3 post types in the admin. 

But, we've set the `graphql_kind` for `WithInterfaceKind` to be an `interface`, and the `child_type_one` and `child_type_two` to implement that interface. 

When we create posts of the `WithInterfaceKind` in the wp-admin, we have no way to query for them, because the `{ withInterfaceKinds { nodes { ... } } }` query tries to resolve posts of the `ChildTypeOne` and `ChildTypeTwo` types 🤔 


How are we supposed to use data from the `WithInterfaceKind` post type? 🤔

jasonbahl: Actually, it's trying to resolve posts from the database of the `with_interface_kind` post_type, but then trying to resolve the type as `ChildTypeOne` or `ChildTypeTwo`, but these types won't ever be returned from the database because the resolver is executing this query: 

```
SELECT wp_posts.ID
FROM wp_posts
WHERE 1=1  AND wp_posts.post_type = 'with_interface_kind' AND ((wp_posts.post_status = 'publish')) 
ORDER BY wp_posts.post_date DESC, wp_posts.ID DESC
LIMIT 0, 11
```

So, we're always getting a list of `with_interface_kind` posts, so the `resolveType` function will try and return the `WithInterfaceKind` Type, but the allowed return types are `ChildTypeOne` and `ChildTypeTwo` which will never be returned by the underlying query 🤔 


So we'll always end up with errors like the following:

CleanShot 2022-09-09 at 14 14 18

justlevine: The original plan was to handle this case (vs a more innocuous one where the the same post type can have different GraphQL types - e.g a Post Format, or TEC Event, or WC_Product type) with #2466 . 
Now that it's a separate issue, I need to rewrite this test to check the above simpler use case.

jasonbahl: I think in the case of WooCommerce, the "variations" etc are still of the same "product" post_type, but in this example we're trying to use 3 different post types and the resolver logic doesn't make that connection. 

I think for WooCommerce, we're trying to get to a point where you can do something like: 

```graphql
{
  products { # query from the "product" post type
     nodes {
        __typename
        id
        ...on VariableProduct { someField } # "differentiate type based on some value of the product"
        ...on ProductBundle { someOtherField } # "differentiate type based on some value of the product"
     }
  }
}
```

Where in the test example, we're actually registering 3 different post types, but the resolver is still trying to resolve `WithInterfaceKind` from the `with_interface_kind` post_type. 

I think we might want to test with a scenario like so: 

```php

	register_post_type( 'with_interface_kind', [
		'public'               => true,
		'show_in_graphql'      => true,
		'supports' => [ 'title', 'editor', 'custom-fields', 'author' ],
		'label'                => 'WithInterfaceKind',
		'graphql_single_name'  => 'WithInterfaceKind',
		'graphql_plural_name'  => 'WithInterfaceKinds',
		'graphql_kind'         => 'interface',
		'graphql_resolve_type' => function ( $post ) {

			if ( 'true' === get_post_meta( $post->databaseId, 'is_child_one', true ) ) {
				$kind = 'ChildTypeOne';
			} else {
				$kind = 'ChildTypeTwo';
			}

			return $kind;
		},
	]);

	register_graphql_object_type( 'ChildTypeOne', [
		'eagerlyLoadType' => true,
		'interfaces' => [ 'WithInterfaceKind' ],
		'fields' => [
			'childTypeOneField' => [
				'type' => 'String',
				'resolve' => function( $post ) {
					return get_post_meta( $post->databaseId, 'child_one_field', true );
				}
			]
		]
	]);

	register_graphql_object_type( 'ChildTypeTwo', [
		'eagerlyLoadType' => true,
		'interfaces' => [ 'WithInterfaceKind' ],
		'fields' => [
			'ChildTypeTwoField' => [
				'type' => 'String',
				'resolve' => function( $post ) {
					return get_post_meta( $post->databaseId, 'child_two_field', true );
				}
			]
		]
	]);

} );
```

Here, we register the Post Type as `'graphql_kind' => "interface"`, but instead of registering new post types that implement that interface, we register new Object Types that will implement that interface. 

The idea with the Post Type being an interface is that there can be variables within posts of that post_type that can make them different types, but ultimately they're all still "WithInterfaceKind" types.

The `resolveType` function is now not determining Type by the node's `post_type`, but another differentiating factor. 

We should be confident that registering a post_type will always return a node of that post_type, but the final Type could be something else based off a meta value, a status, a taxonomy relationship, etc. . .but it will always be of the same Post Type (the Interface being the common bond that the object types implement)

jasonbahl: That would allow for this behavior: 

CleanShot 2022-09-09 at 14 39 28

jasonbahl: It might be good to figure out a way to guard against one post_type implementing an `Interface` that represents _another_ post_type. 

Unless there's a strong case for it, I don't see this as something we want to support. 

Post Types are, in my mind, different objects altogether. 

`PostTypeA` should never implement an interface for `PostTypeB`. 

I'm open for examples where that does sound like a good idea though. I could be thinking too narrowly. 

What might be good is to output a `graphql_debug()` message if that happens, letting the user know that the post type attempted to implement an interface represented by another post type, and that's not the intended use. 🤷🏻‍♂️

justlevine: > It might be good to figure out a way to guard against one post_type implementing an Interface that represents another post_type.
>
>Unless there's a strong case for it, I don't see this as something we want to support.

This is a good point. I think I was focused more on how we _could_  (via #2466 ) instead of whether we should.

Even if we do think of a valid use case, since itl require #2466 to implement, we can always revisit the decision then and worse case roll-back the protection.

jasonbahl: Ya, we'd need to go even deeper than #2466 I think. We'd need a way for the entire Connection Resolver to be overridden. 

Like 

```php
register_post_type( 'interface_type', [
  'show_in_graphql' => true,
  'graphql_single_name' => 'IntefaceType',
  'graphql_plural_name' => 'InterfaceTypes',
  'graphql_connection_args' => function( $args ) { 
     # we'd need a way to override the connection args to query across different post_types
     # this post_type should implement the Interface defined by a different post_type
     $args['post_type'] = [ 'interface_type', 'post_type_a', 'post_type_b' ];
     return $args;
  }
  # we'd also need to customize the loader
  ...
  # and the model
  ...
  # and probably some other things
] );

register_post_type( 'post_type_a', [
  'graphql_interfaces' => [ 'InterfaceType' ], # this post_type should implement the Interface defined by a different post_type
  'show_in_graphql' => true,
  'graphql_single_name' => 'PostTypeA',
  'graphql_plural_name' => 'PostTypeAs',

] );

register_post_type( 'post_type_b', [
  'graphql_interfaces' => [ 'InterfaceType' ], # this post_type should implement the Interface defined by a different post_type
  'show_in_graphql' => true,
  'graphql_single_name' => 'PostTypeB',
  'graphql_plural_name' => 'PostTypeBs',
] );
```

src/WPGraphQL.php
justlevine: merged https://github.com/wp-graphql/wp-graphql/pull/2399/commits/263be0cc1aa5fc2296b99dce373d5924cb94ad9f

justlevine: merged: https://github.com/wp-graphql/wp-graphql/pull/2399/commits/263be0cc1aa5fc2296b99dce373d5924cb94ad9f

justlevine: @jasonbahl I think I'm missing something with the above approach.

`graphql_debug()` doesnt output anything on the query (I'm guessing since its happening _before_ the `register_graphql_*()` call. Neither does throwing a GraphQL error at that point in the lifecycle either.

Instead, the RootQuery field is simply not registered, and other queries dont show an issue either.

image

I'm also not entirely sure how I could `graphql_debug()` within `TypeRegisty::register_type()` without losing the context (i.e. the post_type/tax that has a bad `graphql_` property).

> I think before plugins like Custom Post Type UI, Atlas Content Modeler, etc started allowing folks to do this stuff in a UI and not just in code, I used to prefer killing things like this and letting folks know right away. But now it's more of a gray area.
>
> If someone makes a mistake in ACF of CPT UI, this would fatal their site and they couldn't correct the mistake, which is harsh. I want to make sure it's easy to identify, but want to make sure it doesn't tank their existing client applications.

Perhaps the onus needs to be on those 3rd-party plugins that want to support this to make sure their UI requires and validates all the necessary fields?

jasonbahl: Well, the test here is trying to query something that is not in the schema (because we didn't add it to the schema) so it's getting a GraphQL error _before_ execution. 

If we did something like an introspection query the debug message would show.

jasonbahl: I definitely think we need to limit the exception/errors to a more narrow context than _any_
WordPress page load. No matter how much we recommend plugin developers follow some rule here, it will reflect real bad on WPGraphQL if one missing argument of the post type fatals your site. In code or via a GUI plugin. 

Id prefer a more graceful failure than a full on fatal. 

Possibly graphql_debug isn't the right call, but maybe as admin notice or something in addition to graphql_debug?  

Valid queries will output the debug message for developers, (we can at some point, surface the debug messages in GraphiQL better too) and the admin can notify folks of the error too, without killing the whole site.

jasonbahl: I'm open to other options for sure, but I'm opposed to fataling the whole site if a conditional array key/value is missing.

jasonbahl: We might be able to hook into one of the "validation" steps of GraphQL as well to throw a more clear error as to _why_ the field doesn't exist. 

In your test you register a post type, then make a query for it, even though we know it doesn't exist. 

If that query were made, perhaps at run time we can identify the field/type being queried for and if it matches a specific field/type that we have a specific error for we can surface it there, then for all other (valid) queries a debug message is surfaced

jasonbahl: Possibly something like one of these Validation Rules might work? 

src/Server/ValidationRules/RequireAuthentication.php

tests/wpunit/CustomPostTypeTest.php
justlevine: tests fixed in https://github.com/wp-graphql/wp-graphql/pull/2399/commits/9a0c7668dba53b10b2f1a8730ee891b98eb9e670m

src/WPGraphQL.php
justlevine: > If we did something like an introspection query the debug message would show.

What I'm confused about is why it the debug message is showing [when you query `posts`](https://github.com/wp-graphql/wp-graphql/pull/2399#discussion_r966333256), but [not when I do](https://github.com/wp-graphql/wp-graphql/pull/2399#discussion_r967692003).

justlevine:
Code-wise this PR is ready for review.
All thats left is https://github.com/wp-graphql/wp-graphql/pull/2399#discussion_r884188865 which isnt necessary for this to merge.

jasonbahl:
@justlevine ok, so I'm running through some scenarios and finding some issues. 

I'll document. 

We'll start with this basic post_type registration:

```php

  $args = [
    'label' => 'Test Type',
    'public' => true,
    'supports' => ['comments', 'title', 'author', 'editor', 'thumbnail', 'custom-fields' ],
    'show_in_graphql' => true,
    'graphql_single_name' => 'TestType',
    'graphql_plural_name' => 'TestTypes',
  ];

  register_post_type( 'test_type', $args );

} );
```

## graphql_exclude_connection

Because the post type has support for "comments" and "author", this gives us connections: `TestType -> Author` and `TestType -> Comments`

I can query like so now: 

```graphql
{
  testTypes {
    nodes {
      title
      author {
        node {
          name
        }
      }
      comments {
        nodes {
          content
        }
      }
    }
  }
}
```

Let's say I want to "exclude" the Author and Comments connection. 

According to this PR I should be able to add: `'graphql_exclude_connections' => [ 'author', 'comments' ]`

If I do that, the `comments` connection is properly not exposed to the Schema, _but_ the author connection is still present.

We can see that in the screenshot below, that the `TestType` does not have the `comments` field any longer, but does still have the `author` field:

CleanShot 2022-09-20 at 14 00 59

I've not investigated the code, but I suspect this has something to do with one-to-one connections vs. plural connections?

## graphql_exclude_interfaces

Using the registration above for the `test_type` post type, the `TestType` will apply the `ContentNode` interface (as expected). 

CleanShot 2022-09-20 at 14 19 41

If we want this type to _not_ apply the `ContentNode` interface, we should be able to add: 

```php
'graphql_exclude_interfaces' => [ 'ContentNode' ],
```

If we do this, the `TestType` still implements the `ContentNode` interface:

CleanShot 2022-09-20 at 14 32 16


(full registration here): 

```php
$args = [
	'label' => 'Test Type',
	'public' => true,
	'show_in_graphql' => true,
	'graphql_single_name' => 'TestType',
	'graphql_plural_name' => 'TestTypes',
	'supports' => ['comments', 'title', 'author', 'editor', 'thumbnail', 'custom-fields' ],
	'graphql_exclude_interfaces' => [ 'ContentNode' ],
];

register_post_type( 'test-type', $args );
```

If I remove the `supports` line altogether, then the `ContentNode` Interface is not applied.

jasonbahl (changes requested):
I discovered some issues while testing: https://github.com/wp-graphql/wp-graphql/pull/2399#issuecomment-1252882671

I haven't looked yet at the code causing these issues.

justlevine:
> If I do that, the comments connection is properly not exposed to the Schema, but the author connection is still present.

That's because the author isn't registered on the type, but on the `NodeWithAuthor` interface. If [you exclude that interface](https://github.com/wp-graphql/wp-graphql/blob/9cc7fed20c18151bae79dc8e051355e079f22dd1/tests/wpunit/CustomPostTypeTest.php#L734) it should work.

> If we want this type to not apply the ContentNode interface, we should be able to add:
> `'graphql_exclude_interfaces' => [ 'ContentNode' ],`
> If we do this, the TestType still implements the ContentNode interface:

[This should work](https://github.com/wp-graphql/wp-graphql/blob/9cc7fed20c18151bae79dc8e051355e079f22dd1/src/Registry/Utils/PostObject.php#L318), will investigate 🧐

jasonbahl:
> That's because the author isn't registered on the type, but on the NodeWithAuthor interface. If [you exclude that interface]

Gah!

Yes, this is a documentation thing, not a code thing. 🤦🏻‍♂️ 

Thanks for reminding me 😄

justlevine:
> > If we want this type to not apply the ContentNode interface, we should be able to add:
> > `'graphql_exclude_interfaces' => [ 'ContentNode' ],`
> > If we do this, the TestType still implements the ContentNode interface:
> 
> [This should work](https://github.com/wp-graphql/wp-graphql/blob/9cc7fed20c18151bae79dc8e051355e079f22dd1/src/Registry/Utils/PostObject.php#L318), will investigate 🧐

Not by my computer to test, but pretty sure the issue is that [NodeWiithFeaturedImage](https://github.com/wp-graphql/wp-graphql/blob/8c0fb97ea21155b7effd7714c63c37a9dc532e85/src/Type/InterfaceType/NodeWithFeaturedImage.php#L27) includes `ContentNode` (and some others) as inherited interfaces.

Not sure why this inherits those interfaces when other `NodeWith*` types don't. An oversight, maybe?

jasonbahl:
> Not by my computer to test, but pretty sure the issue is that [NodeWiithFeaturedImage](https://github.com/wp-graphql/wp-graphql/blob/8c0fb97ea21155b7effd7714c63c37a9dc532e85/src/Type/InterfaceType/NodeWithFeaturedImage.php#L27) includes ContentNode (and some others) as inherited interfaces.

> Not sure why this inherits those interfaces when other NodeWith* types don't. An oversight, maybe?


Ah! Yes! 💪🏻 

This seems right. 

And this is likely due to my identity crisis of "what is a content node?" (i.e. https://github.com/wp-graphql/wp-graphql/pull/2355)

I think sometimes in my mind `ContentNode` is synonomous with "Post of any post type" and then other times "ContentNode" is synonomous with "Posts of public post types" 🤔 

I think we need to still solve that problem. 🤔

That said, I think this could be a documentation thing, then, too. 

If you `exclude_interfaces`, but another interface that is not excluded includes it, it will still be included. As is the case here. 

I excluded `ContentNode`, but because I did not exclude `NodeWithFeaturedImage` it was still included because `NodeWithFeaturedImage` included it.

Now. . .the question is, do we remove `ContentNode` from `NodeWithFeaturedImage`? My instinct is yes, we should remove that declaration 🤔

justlevine:
> Now. . .the question is, do we remove ContentNode from NodeWithFeaturedImage? My instinct is yes, we should remove that declaration 🤔 (although technically a breaking change at this point if we did, so perhaps leave it and cover in documentation?)

I think I agree with your instincts. 

- In WPGraphQL's codebase, `NodeWithFeaturedImage` is only used on PostObject types, all of which are required by default to have the 'Node', 'ContentNode', 'DatabaseIdentifier', and 'NodeWithTemplate' interfaces by default. So the GraphQL object will still have the same fields if we remove any of those from `NodeWithFeaturedImage`.
- Because `PostObject` GraphQL types are usually autogenerated, its unlikely someone is manually applying this interface to a GraphQL type with the _intention_ to inherit those particular interfaces. It's possibly theyre applying `NodeWithFeaturedImage` to a `TermNode`, but if thats they case, they'd need to be filtering off `ContentNode` anyway, and the other extraneous interfaces are similarly included de facto in `TermObject` types.
- As I mentioned, none of the of the other `NodeWith*` interfaces inherit these additional types. Not just does this imply that the inclusion on `NodeWithFeaturedImage` is a bug, but it also means that even those schemas that are moving to a `nodeByUri` pattern of building up generic queries with interfaces are likely not including the fields in their `... on NodeWithFeaturedImage` fragments. Definitely fewer than e.g. those that could have been using broken pagination as a feature (although to be fair, the necessity of _this_ change is less severe).

True, we could address the extra interfaces on `NodeWithFeaturedImage` solely in the docs, as part of whatever disclaimer that says the API can only remove direct interfaces/connections/fields, but as with `deregister_graphql_field()` doesnt affect inherited types, but IMO educating users about the differences between the two is hard enough without throwing curveballs in the form of unnecessary inheritance.

src/Registry/Utils/TermObject.php
```diff
@@ -0,0 +1,322 @@
...
+			]);
+		}
+
+		// Merge with connections set in register_taxonomy.
+		if ( ! empty( $tax_object->graphql_connections ) ) {
+			$connections = array_merge( $connections, $tax_object->graphql_connections );
+		}
+
+		// Remove excluded connections.
+		foreach ( $tax_object->graphql_exclude_connections as $connection_name ) {
```
jasonbahl, line 245: Does it make sense to support `graphql_exclude_connections` at a lower level? 

For example, in the WPObjectType and WPInterfaceType?

For example, we could then do something like this:

```php
apply_filters( 'graphql_wp_object_type_config', function( $config ) { 
  $config['graphql_exclude_connections'] = [ 'someConnectionToExclude'];
  return $config;
});
```

And this functionality could then work for all types, not just types that were added to the graph by `register_post_type` and `register_taxonomy`?

Like, I know some folks using WPGraphQL to get data from external data sources, like shopify, etc. 

If they had some connections registered, and a plugin author wanted to quickly exclude them, they could use the filter above, and it would work for any type, not just Post/Term types. 🤔

jasonbahl: I suppose even in the situation I brought up, we could just use `apply_filters( 'graphql_wp_object_type_config', ... );` and unset the connections manually as well. 

Maybe not a big deal 🤔

justlevine: I think a whole slew of lower-level filters, along with some additional `(de)register_graphql_*` access methods would be great.

This PR though is about making it easier to for devs to handle auto registering CPTs/Terms _without_ the use of multiple actions/filters.

We can always adjust the underlying logic used to handle the WP_Type class properties in a future PR, and I think that by handling it separately/incrementally we can make a more informed design decision about where these new filters should go, instead of implementing them now and possibly needing to adjust/deprecate them once we have a concrete list of additional use cases/considerations.

src/Type/InterfaceType/NodeWithFeaturedImage.php
```diff
@@ -24,7 +24,7 @@ public static function register_type( TypeRegistry $type_registry ) {
 			'NodeWithFeaturedImage',
 			[
 				'description' => __( 'A node that can have a featured image set', 'wp-graphql' ),
-				'interfaces'  => [ 'Node', 'DatabaseIdentifier' ],
+				'interfaces'  => [ 'Node' ],
```
justlevine, line 27: Still think we should remove `Node` as well for parity with the other `NodeWith` interfaces, and an `exclude` API with the fewest possible quirks. (Why should removing `NodeWithFeaturedImage` be required to remove `Node` and not either remove_all the_ `NodeWith*` interfaces, or without removing _any_ other interfaces. )

If this is an intentional decision, and not something that fell through the cracks of slack vs github, feel free to ignore - though we might consider opening an issue to standardize this behavior and _add_ `Node` to the other `NodeWith*` interfaces.

jasonbahl: Actually, I think we should go the opposite way. 

Any of the `NodeWith*` Interfaces _should_ be guaranteed to be a Node, so they should _all_ implement the `Node` interface. 

I think the oversight was the other way. 

All `NodeWith *` are nodes, and the Schema _should_ express that. 

Documentation should be clear that excluding connections will only exclude connections implemented directly on that type. If the connections are inherited from an Interface, they would need to be excluded/filtered out from that interface (thus affecting all other Types that implement that interface as well).

jasonbahl: Like what I was running into the other day was just me _assuming_ this would work differently. The system was doing what it should be doing, preventing me from doing something silly.

justlevine: Issue created: https://github.com/wp-graphql/wp-graphql/issues/2536

jasonbahl (approved):

jasonbahl:
UPGRADE NOTICE: 

This PR removes the `ContentNode` interface from the `NodeWithFeaturedImage` interface. 

This is considered a breaking change for any client who might have a `...on NodeWithFeaturedImage` fragment referencing fields added by the `ContentNode` interface. 

The following filter can be used to add back the `ContentNode` interface to the `NodeWithFeaturedImage` interface:

```php
add_filter( 'graphql_wp_interface_type_config', function( $config ) {
	if ( $config['name'] === 'NodeWithFeaturedImage' ) {
		$config['interfaces'][] = 'ContentNode';
	}
	return $config;

}, 10, 1 );
```

Issue #2343, opened 2022-04-13
v1.8.0 causing gatsby-source-wordpress error


pauljavascripting:
v1.80 was preventing my headless Wordpress site from building successfully with the following error:

gatsby-source-wordpress  Found a duplicate ID in WordPress - this means you will have fewer nodes in Gatsby than in WordPress. This will need to be resolved in WP by identifying and fixing the underlying bug with your WP plugins or custom code.

I reverted to v1.72 and the error disappeared.

Hope this helps someone out!

justlevine:
@pauljavascripting there's not really a lot to go on here. There's been some reports that v1.8.0 is [catching invalid GraphQL names that were sneaking by undetected in previous versions](https://github.com/wp-graphql/wp-graphql/issues/2313#issuecomment-1089729802).

To properly debug:
1. Deactivate all plugins and custom code except for WPGraphQL and WPGatsby, and confirm gatsby-source-wordpress can introspect.
2. Enable your plugins / custom WPGraphQL functions one-by-one, until you get the error back.
2. En

jasonbahl:
Hello, `wpgraphql.com` is running WPGraphQL v1.8.0 and WPGatsby v2.3.2 and I'm able to successfully build my site on Gatsby. 

I'd like to understand more about this regression, as it seems like it's affecting some other folks in the WPGraphQL Slack channel as well (see: https://wp-graphql.slack.com/archives/C3NM1M291/p1650019084899219)

I'm not currently able to reproduce this based on my use of WPGraphQL + Gatsby. 

If someone could provide some steps to reproduce, that would be helpful. 

A Gatsby Repo + access to a staging site for the WordPress install that's experiencing the issue would be great. 

I have a hunch that it might be related to either certain data sets, or a certain combination of plugins (like WPGraphQL for ACF, or WPGraphQL for WooCommerce, or something to that tune). 

Would be good to be able to reproduce so we can troubleshoot further.

tsdexter:
I am also having similar issues: 

```
warn  gatsby-source-wordpress  Found a duplicate ID in WordPress - this means
 you will have fewer nodes in Gatsby than in WordPress. This will need to be
 resolved in WP by identifying and fixing the underlying bug with your WP plugins
  or custom code.
 
 info  gatsby-source-wordpress  #524 (/tag/centre-for-social-entrepreneurship/)
 is a duplicate of 524 (/tag/centre-for-social-entrepreneurship/)
 
[...]

PR #2406, opened 2022-06-16
chore: update PHPUnit, WPBrowser, and WPGraphQL Test Case

justlevine:
This PR updates PHPUnit to `^9.5.`, WPBrowser to `^3.1.6` and `WPGraphQL Test Case to `~2.3`.

As a result, several failing tests had to be updated.

To address #2343, we need to backfill tests handling connection pagination, and the best way to do that is with the improvements in [WPGraphQL Test Case v2.3.0](https://github.com/wp-graphql/wp-graphql-testcase/releases/tag/v2.3.0).

This already would have happened automatically, but [we rolled back to `v2.1`](https://github.com/wp-graphql/wp-graphql/pull/2396#issuecomment-1140146822)

…

Any other comments?
-------------------
- ~`composer.lock` needs to be regenerated. No idea what PHP version it's built off of, but I doubt it's mine.~
- Note the `@todos` in [RootQueryConnectionTest](https://github.com/justlevine/wp-graphql/blob/85185ac0045457d4ebca00d9361c1a3c041feb5f/tests/wpunit/RootQueryConnectionsTest.php#L81). Making those fields (correctly) return null I believe is a breaking change.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + PHP 8.0.15)

**WordPress Version:** 6.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
 		"appsero/client": "^1.2"
 	},
 	"require-dev": {
-		"lucatume/wp-browser": "^3.1.6",
+		"lucatume/wp-browser": "3.1.0",
 		"codeception/module-asserts": "^1.0",
 		"codeception/module-phpbrowser": "^1.0",
 		"codeception/module-webdriver": "^1.0",
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
 		"appsero/client": "^1.2"
 	},
 	"require-dev": {
-		"lucatume/wp-browser": "3.1.0",
+		"lucatume/wp-browser": "^3.1.6",
 		"codeception/module-asserts": "^1.0",
 		"codeception/module-phpbrowser": "^1.0",
 		"codeception/module-webdriver": "^1.0",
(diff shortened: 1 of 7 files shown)


justlevine:
Code Quality is failing because it needs a new `composer.lock`

Integration tests: 
- `tests/wpunit/PluginCompatibilityTest.php:testAmpWpCompatibility` is failing, either because pretty permalinks arent enabled or because whatever is being tested is actually broken. I'm not sure what this is testing, so I'm not sure how to fix.
image
- `tests/wpunit/UserObjectMutationsTest.php:testSendPasswordResetEmailSentWithEmail` and ``tests/wpunit/UserObjectMutationsTest.php:testSendPasswordResetEmailSentWithUsername` are both passing by me, so I'm unsure what the deal is with them.
image

If these are issues with the functionality and not the tests, I suggest we `$this->markTestIncomplete()`, since the new version of WPGraphQL Test Case is _exposing_ preexisting issues, not introducing them.

PR #2408, opened 2022-06-18
fix: Connection pagination refactor

justlevine:
This PR aims to handle a slew of issues with paginating on GraphQL connections. 
Originally surfaced by #2343, a review showed not just regressions caused by #2294, but others dating back to `v1.6.x` as well. 

- So far, connection pagination tests have been refactored and fleshed out to expose the bugs. 
- Instead of letting the tests fail, a `markTestIncomplete()` notes where the first bug is surfaced. As those are addressed, it's possible the remaining test code will need to be tweaked.

**Next step** is using those tests to fix the exposed issues. Work will continue on this draft PR.

Contributions are welcome, either as comments on this PR, or by your own commits - just remember to `git rebase justlevine/fix/connection-pagination`, resolve any conflicts, and `git push justlevine/fix/connection-pagination --force` when you're done to keep the commit history clean. 

Will be filled out when the PR is read for merge.

Any other comments?
-------------------
...

Where has this been tested?
---------------------------

**WordPress Version:** 6.0


justlevine:
## First bug: 
Fetching multiple `comments` in one query (or in one Codeception test instance), recycles the original `ids` instead of fetching new ones.

## To replicate:
1. Create and approve 3 comments.
2. Run the following query
```graphql
{
  comments(first:2) {
    pageInfo {
      startCursor
      endCursor
    }
    edges {
      cursor
      node {
        databaseId
      }
    }
  }
  after:comments(first:1, after:"YXJyYXljb25uZWN0aW9uOjM="){ # the second cursor.
    pageInfo {
      startCursor
      endCursor
    }
    edges {
      cursor
      node {
        databaseId
      }
    }
  }
}
```
All results should be shown correctly. The reason for this is that the first comments query overfetches by one.

3. Change the `comments(first:2) ` to `comments(first:1)` and rerun the query.
You should see `after` returns null. This is because the comment query overfetch only includes the first two.

## Screenshots
- `first:2`
image
- `first:1`
image

tests/wpunit/CommentConnectionQueriesTest.php
```diff
@@ -88,184 +89,272 @@ public function create_comments() {
...
-				'number'         => 1,
-				'offset'         => 1,
-				'order'          => 'ASC',
-				'orderby'        => 'comment_date',
-				'comment_parent' => 0,
-			]
-		);
-		$second_to_last_comment = $comments[0];
+		codecept_debug( $expected );
+		$this->markTestIncomplete( 'works until here' );
```
jasonbahl, line 328: @justlevine For this to work, we'll need to add functionality similar to what's been done for Post and Term connections: 

See: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Config.php#L187-L216

jasonbahl (commented):
@justlevine I like what you're doing here for sure. 

There's a lot of rough patches with pagination and it would be lovely to get them all shored up. 

I anticipate this taking some time to do comprehensively, so I propose we at least get a fix for the taxonomy pagination regression first, then continue working through other pagination issues as a focus of another release. 

I have a branch locally where I have the Taxonomy pagination working that I can PR up with a few tests, so that we can unblock the Gatsby users that are suffering from the regression, then we can keep working on this to get things in a better spot.

justlevine:
> Were these removed because they're duplicate tests and have coverage elsewhere? or was there another reason for removing them?

Mostly duplicates. What wasnt has been relocated to the {ConnectionType}QueriesTest/{ConnectionType}PaginationTest to keep a standard pattern and ensure tests arent duped/overlooked in the future.

justlevine:
Those cests should not be failing 🤔

justlevine:
Turns out _most_ tests don't clean up after themselves, and some use duplicate names for objects causing tests to pass that shouldnt.

Tried to catch/fix as many as I could. I think all thats left are `WP_Post` and `WP_User` objects.

justlevine:
Work moved to #2457

PR #2410, opened 2022-06-20
Bugfix/taxonomy pagination

### LINKED ISSUE #2343: v1.8.0 causing gatsby-source-wordpress error

jasonbahl:
This fixes a bug with taxonomy pagination where the end cursor wasn't properly being respected for subsequent paginated requests. 

closes #2343 

Any other comments?
-------------------

With the following snippet I've registered a taxonomy for every letter of the alphabet. 

So we end up with the following list of taxonomies:

```json
{
  "category": "category",
  "post_tag": "post_tag",
  "post_format": "post_format",
  "A": "A",
  "B": "B",
  "C": "C",
  "D": "D",
  "E": "E",
  "F": "F",
  "G": "G",
  "H": "H",
  "I": "I",
  "J": "J",
  "K": "K",
  "L": "L",
  "M": "M",
  "N": "N",
  "O": "O",
  "P": "P",
  "Q": "Q",
  "R": "R",
  "S": "S",
  "T": "T",
  "U": "U",
  "V": "V",
  "W": "W",
  "X": "X",
  "Y": "Y",
  "Z": "Z"
}
```

I can then make a query for a set of these taxonomies like so:

```graphql
query testTaxonomies($first: Int, $after: String, $last: Int, $before: String ) {
  taxonomies(first: $first, last: $last, before: $before, after: $after) {
    pageInfo {
      endCursor
      hasNextPage
      hasPreviousPage
      startCursor
    }
    nodes {
      id
      name
    }
  }
}
```

with variables: 

```json
{
  "first": 5
}
```

and we should expect to get the first 5 taxonomies in response:

```json
  "category": "category",
  "post_tag": "post_tag",
  "post_format": "post_format",
  "A": "A",
  "B": "B",
```

We can see this working as expected.

CleanShot 2022-06-20 at 11 14 53@2x

Then, we should be able to take the `pageInfo.endCursor` value and use it as the input value for the `after` variable to query the next page. 

Like so:

```json
{
  "first": 5,
  "after": "YXJyYXljb25uZWN0aW9uOkI="
}
```

This is where the bug was happening. 

We should expect nodes (the _first_ 5 nodes _after_ node "B"):

```json
  "C": "C",
  "D": "D",
  "E": "E",
  "F": "F",
  "G": "G",
```

**BEFORE**

Before this PR, we would get the following output:

```json
{
  "data": {
    "taxonomies": {
      "pageInfo": {
        "endCursor": "YXJyYXljb25uZWN0aW9uOkM=",
        "hasNextPage": true,
        "hasPreviousPage": true,
        "startCursor": "YXJyYXljb25uZWN0aW9uOnBvc3RfdGFn"
      },
      "nodes": [
        {
          "id": "dGF4b25vbXk6cG9zdF90YWc=",
          "name": "post_tag"
        },
        {
          "id": "dGF4b25vbXk6cG9zdF9mb3JtYXQ=",
          "name": "post_format"
        },
        {
          "id": "dGF4b25vbXk6QQ==",
          "name": "A"
        },
        {
          "id": "dGF4b25vbXk6Qg==",
          "name": "B"
        },
        {
          "id": "dGF4b25vbXk6Qw==",
          "name": "C"
        }
      ]
    }
  },
}
```

This is not what we want!

**AFTER**

After this PR, the same query returns the following:

```json
{
  "data": {
    "taxonomies": {
      "pageInfo": {
        "endCursor": "YXJyYXljb25uZWN0aW9uOkc=",
        "hasNextPage": true,
        "hasPreviousPage": false,
        "startCursor": "YXJyYXljb25uZWN0aW9uOkM="
      },
      "nodes": [
        {
          "id": "dGF4b25vbXk6Qw==",
          "name": "C"
        },
        {
          "id": "dGF4b25vbXk6RA==",
          "name": "D"
        },
        {
          "id": "dGF4b25vbXk6RQ==",
          "name": "E"
        },
        {
          "id": "dGF4b25vbXk6Rg==",
          "name": "F"
        },
        {
          "id": "dGF4b25vbXk6Rw==",
          "name": "G"
        }
      ]
    }
  },
}
```

This is what we were expecting! 🥳 🎉 

We can take the end cursor (for node "G") and do another request, expecting nodes "H" through "L"

And we see we get the expected results:

```json
{
  "data": {
    "taxonomies": {
      "pageInfo": {
        "endCursor": "YXJyYXljb25uZWN0aW9uOkw=",
        "hasNextPage": true,
        "hasPreviousPage": false,
        "startCursor": "YXJyYXljb25uZWN0aW9uOkg="
      },
      "nodes": [
        {
          "id": "dGF4b25vbXk6SA==",
          "name": "H"
        },
        {
          "id": "dGF4b25vbXk6SQ==",
          "name": "I"
        },
        {
          "id": "dGF4b25vbXk6Sg==",
          "name": "J"
        },
        {
          "id": "dGF4b25vbXk6Sw==",
          "name": "K"
        },
        {
          "id": "dGF4b25vbXk6TA==",
          "name": "L"
        }
      ]
    }
  },
}
```

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3119,16 +3119,16 @@
         },
         {
             "name": "phpstan/phpstan",
-            "version": "1.7.14",
+            "version": "1.7.15",
             "source": {
                 "type": "git",
                 "url": "https://github.com/phpstan/phpstan.git",
-                "reference": "e6f145f196a59c7ca91ea926c87ef3d936c4305f"
+                "reference": "cd0202ea1b1fc6d1bbe156c6e2e18a03e0ff160a"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/phpstan/phpstan/zipball/e6f145f196a59c7ca91ea926c87ef3d936c4305f",
-                "reference": "e6f145f196a59c7ca91ea926c87ef3d936c4305f",
+                "url": "https://api.github.com/repos/phpstan/phpstan/zipball/cd0202ea1b1fc6d1bbe156c6e2e18a03e0ff160a",
+                "reference": "cd0202ea1b1fc6d1bbe156c6e2e18a03e0ff160a",
                 "shasum": ""
             },
             "require": {
@@ -3154,7 +3154,7 @@
             "description": "PHPStan - PHP Static Analysis Tool",
             "support": {
                 "issues": "https://github.com/phpstan/phpstan/issues",
-                "source": "https://github.com/phpstan/phpstan/tree/1.7.14"
+                "source": "https://github.com/phpstan/phpstan/tree/1.7.15"
             },
             "funding": [
                 {
@@ -3174,7 +3174,7 @@
                     "type": "tidelift"
                 }
             ],
-            "time": "2022-06-14T13:09:35+00:00"
+            "time": "2022-06-20T08:29:01+00:00"
         },
         {
             "name": "phpunit/php-code-coverage",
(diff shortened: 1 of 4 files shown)


justlevine (commented):
Looks good over here 🔥

(For posterity: this works by overloading `AbstractConnectionResolver`'s methods to fix regressions with `TaxonomyConnectionResolver` _only_. Once the actual causes are addressed in the parent class (as these pagination issues affect other connections), much of this code can probably be removed.)

PR #2435, opened 2022-07-06
Add filter in execute() for query string

markkelnar:
For when using the graphql() function directly to invoke a request, ie, during unit testing, queryId is not honored/processed.  The query string is empty and the parser throws error on the query string.

This filter allows the graphql() invocation, which uses the execute() function to return a query string. The smart cache plugin can use this callback filter to look for queryId when the query string is not present in the params and look up a saved query string.

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -611,9 +611,19 @@ class Request {
 			$response = apply_filters( 'pre_graphql_execute_request', null, $this );
 
 			if ( null === $response ) {
+
+				/**
+				 * Allow the query string to be determined by a filter. Ex, when params->queryId is present, query can be retrieved.
+				 */
+				$query = apply_filters(
+					'graphql_execute_query_params',
+					isset( $this->params->query ) ? $this->params->query : '',
+					$this->params
+				);
+
 				$result = \GraphQL\GraphQL::executeQuery(
 					$this->schema,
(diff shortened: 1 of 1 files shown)


src/Request.php
```diff
@@ -611,9 +611,19 @@ public function execute() {
 			$response = apply_filters( 'pre_graphql_execute_request', null, $this );
 
 			if ( null === $response ) {
+
+				/**
+				 * Allow the query string to be determined by a filter. Ex, when params->queryId is present, query can be retrieved.
+				 */
+				$query = apply_filters(
+					'pre_graphql_execute_query_params',
```
jasonbahl, line 619: @markkelnar so the `pre` naming convention indicates filters that run _before_ something and if a response is returned other than null, don't proceed with the normal execution. 

This appears to be just a "normal" filter and not a "pre" filter. 

I think we should consider changing this filter name to `graphql_execute_query_params`

composer.lock
```diff
@@ -386,16 +386,16 @@
         },
         {
             "name": "codeception/codeception",
-            "version": "4.2.0",
+            "version": "4.2.1",
```
jasonbahl, line 389: I don't think you need to push up the composer.lock with this PR?

Issue #1139, opened 2020-01-31
Not possible to query a user from a comment?


jkhaui:
Hi,
Maybe I'm missing something, but is it really not possible to traverse a `user` node from a `comment` node? I.e. when retrieving data for a post, I thought it'd be logical to also retrieve all comments for that post, and for each comment, if the comment's author is a registered user, at minimum you'd naturally want to at least display their username too.

As far as I can see one can only query whether the comment's author was of type User or not a registered user. I love everything else about the plugin but find this quite strange. The most obvious use-case (IMO) for headless WP is to create apps with highly relational data between user interactions - social network style - and associating a comment with a user is an essential feature in achieving this

jasonbahl:
@jkhaui Thanks for using WPGraphQL! 🙌 

You can query posts, the post's comments and the author of the comment, and then keep traversing connections the author has with other content, such as Posts, etc. 

Here's an example query:

```
{
	post(id:"cG9zdDoxMjg4") {
    id
    title
    comments {
      nodes {
        id
        content
        author {
          __typename
          ...on CommentAuthor {
            name
          }
          ...on User {
            firstName
            lastName
          }
        }
      }
    }
  }
}
```

And here's a screenshot of it in action with the results:

![Screen Shot 2020-01-31 at 8 45 21 AM](https://user-images.githubusercontent.com/1260765/73552760-0bff6180-4406-11ea-917b-40dbcbec4758.png)

My guess is that when you got to the `comment.author` field, since it returns a `CommentAuthorUnion`, perhaps that tripped you up? 

The reason it returns a Union, is because some authors are guest authors and some are users. The shape of data you can ask for from a Guest Author is different than what you can ask for if the author is a user. 

This is where this portion of the query kicks in: 

```
 ...on CommentAuthor {
  name
}
...on User {
  firstName
  lastName
}
```

This syntax in GraphQL allows us to query fields of Unions. This is telling GraphQL: "Hey, I don't know what type of user the comment author is, but if it turns out it's a User, I want the firstName and lastName, but if it turns out it's a CommentAuthor (not a user), I want just the name"

This feature of GraphQL allows our applications to have predictable results, even in situations that are commonly unpredictable. 

I hope this answers your question and gets you headed in a direction that allows you to build what you're looking to build.

jkhaui:
[...]

PR #2452, opened 2022-07-22
fix: #2448 - Fixes Comment.author connection resolver to properly load both User and CommentAuthor types

### LINKED ISSUE #2448: Author information is not retrieved when pulling post comments.
Hi, 

first of all thanks for your plugin, it's a great plugin.

https://github.com/wp-graphql/wp-graphql/issues/1139 I saw the error here, the issue was closed, but it was closed without being resolved. Looking at the last comment https://github.com/wp-graphql/wp-graphql/issues/1139#issuecomment-580963408 I am having the same error here.

I have a few custom fields for my User type. But **... on User** doesn't work at all. Only .**..on CommentAuthor** data is coming, I have never been able to solve this problem.

I am sharing a sample screenshot of my application below. I made all users public from the "graphql_object_visibility" section. It just gives me the information on **... on CommentAuhor.** I commented and tested as **2 visitors** and **2 registered members**. However, the result is negative.

Wpgraphql-comment-user-connection

Can you please help with this?

jasonbahl:
This PR fixes the resolver for the `Comment.author` connection. 

The current code will only load User objects if the user is logged in. This isn't expected behavior. Instead, the connection should try to load a User and return it if it's considered a public node, otherwise fallback to a CommentAuhor.

closes #2448
related: #1139

BEFORE:

For non-authenticated requests, the comment author was _always_ loading a `CommentAuthor` type

CleanShot 2022-07-22 at 15 02 35

AFTER:

Now a non-authenticated request will return a mix of `User` and `CommentAuthor` depending on if the author of a Comment is indeed a user, and the user node is safe to be returned publicly

CleanShot 2022-07-22 at 15 01 51

### DIFF
--- a/tests/wpunit/CommentObjectQueriesTest.php
+++ b/tests/wpunit/CommentObjectQueriesTest.php
@@ -721,6 +721,10 @@ class CommentObjectQueriesTest extends \Codeception\TestCase\WPTestCase {
 
 		$this->assertSame( 'User', $typename );
 		$this->assertNotEmpty( $actual['data']['comment']['author']['node']['avatar']['url'] );
+
+		wp_delete_comment( $comment_by_user_id );
+		wp_delete_comment( $comment_by_comment_author_id );
+
 	}
 
 
--- a/tests/wpunit/CommentObjectQueriesTest.php
+++ b/tests/wpunit/CommentObjectQueriesTest.php
@@ -666,7 +666,11 @@ class CommentObjectQueriesTest extends \Codeception\TestCase\WPTestCase {
 	public function testQueryingAvatarOnUserAuthorsIsValidForPublicAndAuthenticatedRequests() {
 
 		// create a comment with a guest author as the author
-		$comment_id = $this->createCommentObject();
+		$comment_by_comment_author_id = $this->createCommentObject([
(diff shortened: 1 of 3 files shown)


tests/wpunit/CommentObjectQueriesTest.php
```diff
@@ -666,7 +666,11 @@ public function dataProviderSwitchUser() {
 	public function testQueryingAvatarOnUserAuthorsIsValidForPublicAndAuthenticatedRequests() {
 
 		// create a comment with a guest author as the author
-		$comment_id = $this->createCommentObject();
+		$comment_by_comment_author_id = $this->createCommentObject([
+			'user_id' => 0
+		]);
+
+		$comment_by_user_id = $this->createCommentObject();
```
justlevine, line 673: Might be beyond the scope of this PR, but maybe we should clean up these comments at the end of the test?
Worried it might affect #2408, although if it does, i can always handle it there...

jasonbahl: ya, good idea. I'm bad at cleaning up after my tests 🤦🏻‍♂️ 

I'll update.

PR #2455, opened 2022-07-24
test: lint and fix wpunit tests

justlevine:
What does this implement/fix? Explain your changes.
---------------------------------------------------
This PR lints and applies some basic fixes/refactors to existing WPUnit tests.
Work on #2408 exposed many cases of data being leaked between tests, however fixing the issues there makes PR review nearly unusable due to both the scope, and unrelated diffs from linting. It also makes merging/rebasing a PITA.

**There are also numerous cases where tests are _passing_ even though they are written incorrectly or there is a bug in the plugin, due to leaky data.**

#2408 is not the first PR where I've encountered this, and I believe this will make future work significantly more efficient. While this may cause some additional merge conflicts on already-open PRs, many of those already have merge conflicts that must be dealt with anyway.

### To test:
Add the following method to the bottom of `WPHelperTest.php` (so it runs last), and run `vendor/bin/codecept run wpunit -vvv`.
```php
public function testEnsureTypeCleanup() {
  $post_types = get_post_types();
  codecept_debug( $post_types );
  foreach ( $post_types as $post ) {
    codecept_debug( $post );
    codecept_debug( get_posts( [ 'post_type' => $post ] ) );
  }

  $taxonomies = get_taxonomies();
  codecept_debug( $taxonomies );
  foreach ( $taxonomies as $taxonomy ) {
    codecept_debug( $taxonomy );
    codecept_debug( get_terms( [ 'taxonomy' => $taxonomy ] ) );
  }

  codecept_debug( get_users() );
}
```
Only default WP post types and taxonomies should appear, the admin user, and _no_ posts or terms.

Does this close any currently open issues?
------------------------------------------
…

Any relevant logs, error output, GraphiQL screenshots, etc?
-------------------------------------
(If it’s long, please paste to https://ghostbin.com/ and insert the link here.)

Any other comments?
-------------------
- While this touches many `wpunit` tests, it does not touch all of them, only those affected by bugs/leaky types discovered in #2408. This is intentional, in an attempt to limit as many merge conflicts as possible.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + PHP 8.0.19)

**WordPress Version:** 6.0.1

Issue #2419, opened 2022-06-24
searchColumns parameter in users Query doesn't work.


developerfan1219:
I would like to query users by email and nickname.
query will be like this:

```
Where
   email like '%test%' OR
   nickname like '%test%'
```

**To Test**

```
query NewQuery {
  users(where: {search: "test", searchColumns: [NICENAME, EMAIL]}) {
    edges {
      node {
        firstName
        lastName
        name
        nickname
        email
      }
    }
  }
}

```
In get_query of UserConnectionResolver, `$this->query_args` before running query is below:

`{ ["count_total"]=> bool(false) ["graphql_args"]=> array(1) { ["where"]=> array(2) { ["search"]=> string(4) "test" ["searchColumns"]=> array(1) { [0]=> string(8) "nicename" } } } ["graphql_cursor_offset"]=> int(0) ["graphql_cursor_compare"]=> string(1) "<" ["graphql_after_cursor"]=> NULL ["graphql_before_cursor"]=> NULL ["number"]=> int(11) ["search"]=> string(6) "*test*" ["search_columns"]=> array(1) { [0]=> string(8) "nicename" } ["fields"]=> string(2) "ID" ["has_published_posts"]=> bool(true) ["orderby"]=> string(10) "user_login" ["order"]=> string(3) "ASC" }`

the query string after running query:

`SELECT wp_users.ID FROM wp_users WHERE 1=1 AND wp_users.ID IN ( SELECT DISTINCT wp_posts.post_author FROM wp_posts WHERE wp_posts.post_status = 'publish' AND wp_posts.post_type IN ( 'post', 'page', 'attachment' ) ) AND (user_login LIKE '%test%' OR user_url LIKE '%test%' OR user_email LIKE '%test%' OR user_nicename LIKE '%test%' OR display_name LIKE '%test%') ORDER BY user_login ASC, wp_users.ID DESC LIMIT 0, 11`

![image](https://user-images.githubusercontent.com/54286458/175447418-98b679b6-c121-4200-81e9-491113f912b3.png)

I would like to get the query string like this.

[...]

PR #2457, opened 2022-07-25
fix: refactor resolvers to better follow Relay spec

justlevine:
This PR addresses numerous underlying issue with paginating GraphQL connections, by refactoring `AbstractConnectionResolver` to better follow the [Relay Connection spec](https://relay.dev/graphql/connections.htm#sec-Pagination-algorithm), while using smaller methods to allow easier extension by child classes.

Originally surfaced by https://github.com/wp-graphql/wp-graphql/issues/2343, a review showed not just regressions caused by https://github.com/wp-graphql/wp-graphql/pull/2294, but others dating back to v1.6.x as well.

Initial work was done in #2408 , which became unwieldy due to leaky tests hiding false positives (addressed in  #2455).

### Changelog
- feat: Add support for paginating a connection by both `before` _and_ `after` args.
- fix: Refactor `AbstractConnectionResolver` to support all parts of Relay Spec, including more robust over-fetching and slicing.
- fix: Implement missing logic for some `{Type}ConnectionResolver::isValidOffset()` methods, and existing ones to be (slightly) more performant.
- fix: Ensure the order of connections paginated by `first` and `last` are the same. [Possibly breaking, see comment below).
- fix: `{Type}ConnectionResolver::get_query_args()` now always returns an array.
- dev: all `{Type}ConnectionResolver::$query` properties have been given a type in the docblock.
- dev: `AbstractConnectionResolver::get_offset()` has been deprecated in favor of `AbstractConnectionResolver::get_offset_for_cursor()`.
- dev: Refactor Cursor Object classes to extend `AbstractCursor`.
- dev: Passing the `$query` to the `PostCursor` and `UserCursor` constructor has been deprecated in favor of passing the `$query->query_vars` instead.
- dev: `PostCursor::get_cursor_post()` `TermCursor::get_cursor_term()` and `UserCursor::get_cursor_user()` have been deprecated in favor of `AbstractCursor::get_cursor_node()`.
- test: Extend Pagination tests to uniformly cover forward/backward pagination.
- remove  `graphql_cursor_offset` and `graphql_cursor_compare` in favor of `graphql_before_cursor` and `graphql_after_cursor`.

TBD.

Any other comments?
-------------------
- ~Several existing tests were removed as they are covered more comprehensively by newer ones.~
- ~This PR does not (yet) include fixes to the `{type}ObjectCursor` classes, which are the source for incomplete tests for `Comment` and `MenuItem` types. These classes still rely on the logic from the now-deprecated `AbstractConnectionResolver::get_offset()`~
- ~There is also an issue when paginating PostObject connections filtered by `search`, which may or may not be connected.~
- ~A separate issue has been surfaced by these tests where `draft` Posts are not available on `menuItems.nodes.connectedNode`, but they _are_ available on `connectedObject`. 🤔~
- I don't know whether the fixes to `last` ordering should be considered a breaking change. *I want to say it isn't*, and definitely feel strongly about that for any regressions caused by #2294 , but if the `order` was backwards before 1.6.x, the question is whether this discrepancy between the plugin and Relay is a bug or expected  behavior users have built around. 

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.01 (Wsl2 + devilbox + php8.0.19)

**WordPress Version:** 6.0.1


jasonbahl:
@justlevine 

> There is also an issue when paginating PostObject connections filtered by search, which may or may not be connected.

Do you know what issue this is? Is it an open Github issue or just a bug you noticed but doesn't have a formal issue opened?

tests/wpunit/PostConnectionPaginationTest.php
```diff
@@ -82,706 +87,587 @@ public function createPostObject( $args ) {
...
+		$actual   = $this->graphql( compact( 'query', 'variables' ) );
 
-		// Assert the first item is the 3rd most recent post
-		$this->assertSame( $this->created_post_ids[2], $actual['data']['posts']['nodes'][0]['databaseId'] );
+		$this->assertValidPagination( $expected, $actual );
+		$this->assertEquals( true, $actual['data']['posts']['pageInfo']['hasPreviousPage'] );
 
-		// Assert the 2nd item is the 4th most recent post
-		$this->assertSame( $this->created_post_ids[3], $actual['data']['posts']['nodes'][1]['databaseId'] );
+		if ( isset( $graphql_args['where']['search'] ) ) {
```
justlevine, line 220: @jasonbahl ☝️

justlevine:
Yeah no prior GH issue, something unearthed during tests. Though come to think of it _maybe_ #2419 is also related?

jasonbahl:
> Yeah no prior GH issue, something unearthed during tests. Though come to think of it maybe https://github.com/wp-graphql/wp-graphql/issues/2419 is also related?

@justlevine seems loosely related, but I don't think directly. 🤔 I think that issue is maybe an issue with the argument getting mapped from the graphql input to the query_args passed to WP_Query? could be wrong tho

tests/wpunit/PostConnectionPaginationTest.php
jasonbahl: @justlevine looks like this was a false bug. 

I tried reproducing and couldn't but then realized that the test was creating 7 posts with `uniqueString` and then asserting that page 3 of 2 posts per page should have `hasNextPage` as `false`. But since there were 7 posts, `hasNextPage` was properly resolving as `true` as page 3 returns nodes 5 & 6, and page 4 would return the 7th node.

I updated the test to only create 6 posts and it's good now. 🙌🏻

jasonbahl:
@justlevine I pushed up a few commits. 

All tests are now passing. No more are marked incomplete 👏🏻 

I'm not incredibly pleased with [the solution I came up with](https://github.com/wp-graphql/wp-graphql/pull/2457/files#diff-ffd0e341c15c329e46bb25d9d9caa195975fb8477af13a986d2e4a5021aa572bR178-R188) for the the pagination issues related to when posts are sorted by "menu_order". I think it needs more thought, but I don't have brilliant ideas at the moment. I got it to work, but I'm not sure I love the implementation. Open to iteration either now, or sometime in the future. 

Anything else I did is of course also open to iteration, cleanup, DRYing up, etc. 

I feel like there's a lot we can do in the cursor classes to DRY things up, but again, my focus was on getting things working first, before making the code DRY and abstracting patterns.

justlevine:
@jasonbahl 

> I'm not incredibly pleased with [the solution I came up with](https://github.com/wp-graphql/wp-graphql/pull/2457/files#diff-ffd0e341c15c329e46bb25d9d9caa195975fb8477af13a986d2e4a5021aa572bR178-R188) for the the pagination issues related to when posts are sorted by "menu_order".

The important thing is that it's working and I dont think it's worth delaying this PR over (I cant think of any alternatives either 😅).

I did a basic refactor of the CursorObject classes so they extend [`AbstractCursor`](https://github.com/wp-graphql/wp-graphql/pull/2457/files#diff-1219a5bac23dd517b3bf72c58cde884f2031d8963ef57a577e6420c9f20f5299). There might me some more room for abstraction in the individual `compare_{...}()` methods, but unless something jumps out at you then I don't think it's worth delaying the PR for either.

One thing I'm stuck on is what to do with the old `graphql_cursor_offset` and `graphql_cursor_compare` query vars.

The only place we're actually using `graphql_cursor_compare` is in [`Config::graphql_wp_query_cursor_pagination_stability()`](https://github.com/wp-graphql/wp-graphql/pull/2457/files#diff-1442f03771f313c1d77fe582773ea81465d83a68581b4ba2047fc9d1ad6630ceL175) (in the cursor objects it was being overwritten by `before` and `after`, so it's gone now), and I deprecated 3rd-party [setting of `graphql_cursor_offset` altogether](https://github.com/wp-graphql/wp-graphql/pull/2457/files#diff-1219a5bac23dd517b3bf72c58cde884f2031d8963ef57a577e6420c9f20f5299R76-R85), but _this plugin_ is still setting them both in `{Comment|Post|Term|User}ConnectionResolver`.

Do we get rid of them altogether, or mark them as deprecated in a comment? I think it's more likely for people to be setting them then getting them in their own projects (and rare altogether), so I'm not that worried about removing them, but its still something to consider.

jasonbahl:
> but this plugin is still setting them both in {Comment|Post|Term|User}ConnectionResolver.

@justlevine you missed the link for _this plugin_. What plugin?

justlevine:
> > but this plugin is still setting them both in {Comment|Post|Term|User}ConnectionResolver.
> 
> @justlevine you missed the link for _this plugin_. What plugin?
This plugin meaning WPGraphQL core. _We're_ still setting  those custom query vars in all those connection resolvers, but only _using_ `graphql_cursor_compare` for post stability, and not using `graphql_cursor_offset` at all. I don't know if/how any third party code might be using them...

jasonbahl:
@justlevine ah! I thought you meant like an extension plugin. Sorry about that. 🤦🏻

jasonbahl:
> I don't know whether the fixes to last ordering should be considered a breaking change. I want to say it isn't, and definitely feel strongly about that for any regressions caused by https://github.com/wp-graphql/wp-graphql/pull/2294 , but if the order was backwards before 1.6.x, the question is whether this discrepancy between the plugin and Relay is a bug or expected behavior users have built around.

@justlevine I agree that this is a bugfix not a breaking change. Any queries that did change would have just changed the order of the items. Like 10-1 instead of 1-10. But the items will still be there.

PR #2464, opened 2022-08-02
fix: #1687 fix node resolver when page for posts is set

### LINKED ISSUE #1687: nodeByUri does not return ContentType for Posts Page.
Performing a query on ``nodeByUri`` for the page_for_posts uri returns the Page type instead of the ContentType type.

With v1.1.0 the below query works correctly:
```
query {
  nodeByUri(uri: "/blog") {
    ... on Page {
      id
      isPostsPage
      title
    }
    ... on ContentType {
      id
      name
      contentNodes(first: 3) {
        nodes {
          id
        }
      }
    }
  }
}
```
It will return 3 posts under contentNodes. When run on v1.1.2 it returns the Page (isPostsPage: true, title: "Blog") instead of the contentNodes.

jasonbahl:
This fixes a bug where the incorrect Type is returned when a page is set as the "page_for_posts" value. 

closes #1687 

Any other comments?
-------------------

When a Page is set as the "page_for_posts" setting under Reading Settings:

CleanShot 2022-08-02 at 16 39 43

Querying the path for that page should return a `ContentType` and not the `Page`

CleanShot 2022-08-02 at 16 47 38

Much like visiting the path in the browser returns the "blog" archive with the posts, instead of a static page: 

CleanShot 2022-08-02 at 16 43 49

If the reading settings do not have a `page_for_posts` set

CleanShot 2022-08-02 at 16 45 19

Then a `Page` type should be returned:

CleanShot 2022-08-02 at 16 47 07

Much like you would see when visiting the WordPress rendered URL:

CleanShot 2022-08-02 at 16 46 41

### DIFF
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -423,8 +423,8 @@ class NodeResolver {
 				return null;
 			}
 
-			if ( get_option( 'page_for_posts', 0 ) === $post->ID ) {
-				return $this->context->get_loader( 'post' )->load_deferred( $post->ID );
+			if ( (int) get_option( 'page_for_posts', 0 ) === $post->ID ) {
+				return $this->context->get_loader( 'post_type' )->load_deferred( 'post' );
 			}
 
 			return $this->context->get_loader( 'post' )->load_deferred( $post->ID );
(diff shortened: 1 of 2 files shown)


justlevine (approved):
tested locally with a `site_url` that isn't `localhost`, just to be safe.

PR #2471, opened 2022-08-14
feat: PHPCS: enhancements to the Coding Standards Setup

renatonascalves:

- Updates the `phpcs` setup to actually catch issues.
- Cache `phpcs` tests when possible
- Updates GH Actions
- Introduced the `styfle/cancel-workflow-action@0.5.0` action to immediately cancel already running actions
- Fixes a bunch of PHPCS issues
- Removed `automattic/vipwpcs` ruleset, it will be properly implemented at #1535


composer.json
```diff
@@ -36,7 +36,6 @@
 		"codeception/module-filesystem": "^1.0",
 		"codeception/module-cli": "^1.0",
 		"codeception/util-universalframework": "^1.0",
-		"automattic/vipwpcs": "^2.2",
```
renatonascalves, line 39: We are not currently checking the VIP Go ruleset. So I'm removing this unused package for now.

src/Data/Connection/ContentTypeConnectionResolver.php
```diff
@@ -18,20 +14,6 @@ class ContentTypeConnectionResolver extends AbstractConnectionResolver {
 	 */
 	protected $query;
 
-	/**
```
renatonascalves, line 21: This is useless code, but I'm totally happy to keep it if felt strongly. :)

justlevine (changes requested):
Hey @renatonascalves - awesome work!

Question: Why are we no longer caching composer deps in `.github/workflows`? It reduces the workflow time if composer.lock hasn't changed.

Also: The `build` folder is autogenerated. Can you exclude it from linting via phpcs.xml.dist?

renatonascalves:
@justlevine 

- They are cached. The new actions have caching support built in.
- Ops! Will remove them.

src/Admin/GraphiQL/GraphiQL.php
```diff
@@ -109,13 +109,13 @@ public function register_admin_page() {
...
 	/**
-	 * Render the markup to load GraphiQL to
+	 * Render the markup to load GraphiQL to.
 	 *
 	 * @return void
 	 */
 	public function render_graphiql_admin_page() {
 		$rendered = apply_filters( 'graphql_render_admin_page', '<div class="wrap"><div id="graphiql" class="graphiql-container">Loading ...</div></div>' );
-		echo $rendered;
+		echo $rendered; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
justlevine, line 118: This should be escaped, since the output can change via a filter.

src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -589,7 +588,7 @@ public function get_ids() {
 	 * @return int|mixed
 	 */
 	public function get_offset() {
-			_deprecated_function( __FUNCTION__, '1.9.0', get_class( $this ) . '::get_offset_for_cursor()' );
+		_deprecated_function( __FUNCTION__, '1.9.0', get_class( $this ) . '::get_offset_for_cursor()' ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
justlevine, line 591: Should this be escaped instead of ignored?

src/Data/Connection/ContentTypeConnectionResolver.php
justlevine: Nope. It's an oversight from #2457 and safe to remove

src/Data/Loader/UserLoader.php
```diff
@@ -74,7 +74,7 @@ public function get_public_users( array $keys ) {
 
 		global $wpdb;
 
-		$results = $wpdb->get_results(
+		$results = $wpdb->get_results( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.NoCaching,WordPress.DB.DirectDatabaseQuery.DirectQuery
```
justlevine, line 77: @jasonbahl should this be cached?

src/Data/NodeResolver.php
```diff
@@ -276,8 +276,8 @@ public function resolve_uri( string $uri, $extra_query_vars = '' ) {
 
 			if ( isset( $this->wp->extra_query_vars[ $wpvar ] ) ) {
 				$this->wp->query_vars[ $wpvar ] = $this->wp->extra_query_vars[ $wpvar ];
-			} elseif ( isset( $_GET[ $wpvar ] ) ) {
-				$this->wp->query_vars[ $wpvar ] = $_GET[ $wpvar ];
+			} elseif ( isset( $_GET[ $wpvar ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+				$this->wp->query_vars[ $wpvar ] = $_GET[ $wpvar ]; // phpcs:ignore
```
justlevine, line 280: Ideally the rule being ignored should be set explicitly.

src/Request.php
```diff
@@ -343,9 +343,9 @@ protected function has_authentication_errors() {
 			$nonce = null;
 
 			if ( isset( $_REQUEST['_wpnonce'] ) ) {
-				$nonce = $_REQUEST['_wpnonce'];
+				$nonce = $_REQUEST['_wpnonce']; // phpcs:ignore
```
justlevine, line 346: ideally the ignored rules should be explicitly listed

phpcs.xml.dist
```diff
@@ -1,16 +1,37 @@
...
 	<description>Coding standards for the WPGraphQL plugin</description>
 
-	<config name="ignore_errors_on_exit" value="0"/>
-	<config name="ignore_warnings_on_exit" value="0"/>
-	<arg name="colors"/>
-	<arg value="sp"/>
-	<arg name="basepath" value="."/>
-	<arg name="extensions" value="php"/>
+	<!-- Load WordPress Coding standards -->
+	<rule ref="WordPress" />
```
justlevine, line 6: Seems to be a duplicate of L74.

phpcs.xml.dist
```diff
@@ -1,16 +1,37 @@
...
 
-	<!-- Run the lints in parallel so they run faster. This will separate the files into 20 chunks to be linted -->
-	<arg name="parallel" value="20"/>
+	<!-- Load PHPCompatibilityWP standards -->
+	<rule ref="PHPCompatibilityWP" />
```
justlevine, line 9: Can we move this down to the rest of the rules?

phpcs.xml.dist
```diff
@@ -19,48 +40,58 @@
...
 
 		<!-- Should maybe Add Back Later -->
-		<exclude name="WordPress.WP.I18n.MissingTranslatorsComment"/>
 		<exclude name="PEAR.Functions.FunctionCallSignature.ContentAfterOpenBracket"/>
 		<exclude name="PEAR.Functions.FunctionCallSignature.CloseBracketLine"/>
 		<exclude name="PEAR.Functions.FunctionCallSignature.MultipleArguments"/>
+	</rule>
+
+	<rule ref="WordPress">
+		<exclude name="WordPress.NamingConventions.ValidFunctionName.MethodNameInvalid"/>
```
justlevine, line 75: Unless a decision is finalized that `MethodNameInvalid`, `date_date`, and `MissingTranslatorsComment` should stay excluded, we should probably continue to indicate such with comments.

phpcs.xml.dist
```diff
@@ -19,48 +40,58 @@
...
+		<exclude name="WordPress.DateTime.RestrictedFunctions.date_date"/>
+		<exclude name="WordPress.PHP.DisallowShortTernary.Found"/>
+		<exclude name="WordPress.CodeAnalysis.AssignmentInCondition.Found"/>
+		<exclude name="WordPress.WP.I18n.MissingTranslatorsComment"/>
 		<exclude name="WordPress.NamingConventions.ValidVariableName"/>
+		<exclude name="WordPress.DB.SlowDBQuery.slow_db_query_meta_value"/>
```
justlevine, line 81: Not sure how I feel about these 3 new exclusions:
`SlowDBQuery.slow_db_query_meta_value`, `SlowDBQuery.slow_db_query_meta_key`, and `GlobalVariablesOverride.Prohibited`.

I think it would be safer if these get `phpcs:ignore`d on a case-by-case basis to make sure usage is deliberate.

.gitignore
```diff
@@ -25,6 +25,8 @@ coverage/*
 # Generated Schema used in some tooling. Versioned Schema is uploaded as a Release artifact to Github.
 schema.graphql
 
+.phpcs/*.json
```
justlevine, line 28: Do we really need this folder in the repo? 
I don't remember if PHPCS requires the folder to already exist, but if it does, maybe we could just recycle `tests\_output\`.

src/Data/Connection/AbstractConnectionResolver.php
justlevine: Nvm, it seems fine since `get_class( $this )` isnt subject to abuse.

https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Tests/Security/EscapeOutputUnitTest.inc#L131

src/Admin/GraphiQL/GraphiQL.php
renatonascalves: 🤔 I think we can use `wp_kses_post` but I wonder if it is too powerful. `wp_kses` would not work since the html could be anything. and using `esc_html` could mess with the html.

Suggestions?!

justlevine: I _think_ `wp_kses_post()` is fine. My only concern is that it strips `data-*` attributes , which probably isnt a common use case for initializing a `GraphiQL` instance, and I wouldn't suggest allowing due to its potential for abuse (i.e. providing config values for a malicious script that gets enqueued somewhere else). And, if a user really needs to use a data attribute, it can still be allowlisted with `add_filter( 'wp_kses_allowed_html' ... )`.

If we want to be extra safe in case some users are _currently_ adding data attributes to the GraphiQL container, we could `phpcs:ignore` for now, and make the change `wp_kses_post` in a separate PR (so it gets its own line-item in the release changelog). 

@jasonbahl ?

justlevine (dismissed):
What remains from my review is
- sanitizing the results of `graphql_render_admin_page` ( [here](https://github.com/wp-graphql/wp-graphql/pull/2471#discussion_r945217708) )
- maybe caching `$wpdb->get_results()` ( [here](https://github.com/wp-graphql/wp-graphql/pull/2471#discussion_r945218430) ).

Neither of these are actual issues with this PR, but rather existing failed sniffs that this PR is silencing. And since passing those sniffs would require a change in functionality, they are probably better handled in a separate issue/PR (after Jason has a chance to weigh in).

src/Admin/GraphiQL/GraphiQL.php
jasonbahl: I think `wp_kses_post()` should work. 

The original intent of this was to allow for plugins (mainly my WPGraphiQL 2 Beta that was in progress at the time) to be able to override the base markup so a different React app could mount to it. 

I think `wp_kses_post()` should allow for custom markup to be output should folks need to do something similar again.

src/Data/Loader/UserLoader.php
jasonbahl: @justlevine seems reasonable to make this cached.

jasonbahl (changes requested):
I think we can add the `wp_kses_post()` to the filtered markup for GraphiQL to mount to. 

The comment about making the SQL query cached could probably happen in a separate PR as it might require more nuanced testing. 

Like, does it break things for logged in users vs. public users, etc?

🤔 

I think that's a different scope than this PR is trying to address?

jasonbahl:
@renatonascalves thanks so much for this PR! 🙌🏻 

Very much appreciate the time you put into this. 💜

src/Admin/GraphiQL/GraphiQL.php
renatonascalves: Updated with the suggestion! :)

src/Data/Loader/UserLoader.php
jasonbahl: Created issue so we don't lose this: https://github.com/wp-graphql/wp-graphql/issues/2476

jasonbahl (approved):

Issue #2421, opened 2022-06-24
Comment author avatar url is not correct when executing as a public user


umutesen:
I'm trying to load comments of a post along with comment author avatars. Avatar URL appears  to be incorrect when executing the query as a public user inside the GraphiQL IDE:

Executing as the logged in user returns:
```
"avatar": {
  "url": "http://2.gravatar.com/avatar/e8a00208a451c66eb52c105b9daa248a?s=96&d=mm&r=g"
}
```

Executing as a public user returns:
```
"avatar": {
  "url": "http://1.gravatar.com/avatar/?s=96&d=mm&r=g"
}
```

here is my query:
```
query NewQuery {
  postBy(slug: "my-post-slug") {
    comments {
      edges {
          node {
            id
            author {
              node {
                   avatar {
                    url
                  }
                }
            }
          }
        }
    }
  }
}
```

jasonbahl:
@umutesen can you provide a bit more information to reproduce? For example maybe specific information about the user data you have in place? 

Would be good if I could get in a spot where I could reproduce this.

jasonbahl:
@umutesen does this only happen when querying for a comment author? or does it happen when querying for a user directly?

The more information you can provide to help me reproduce this, the better. Thanks! 🙏🏻

umutesen:
Many thanks for your reply. Comments have been captured by default wordpress comment system. The users leave comments without creating accounts so I am unable to query users directly. 

I am currently migrating to a headless WP setup with Angular as the front end, using Apollo client. All calls to wpgraphql is made without authentication.

The default avatar under discussion settings is **Identicon (generated)**. This generates a random avatar where the comment author does not have a gravatar associated with their email address.

<img width="738" alt="Screenshot 2022-06-25 at 09 41 50" src="https://user-images.githubusercontent.com/4234033/175765699-7e750914-f42b-4d17-8eca-3859df52bc9c.png">

To reproduce the error inside WP admin IDE, I query comments along with author avatars in a postBy query, executing as the logged in user.
"http://2.gravatar.com/avatar/e8a00208a451c66eb52c105b9daa248a?s=96&d=identicon&r=g" is a valid url.

<img width="1035" alt="Screenshot 2022-06-25 at 09 43 47" src="https://user-images.githubusercontent.com/4234033/175765771-a0bbcbac-1a74-4f85-86d7-6c8a1c3bd4ce.png">

Switch to executing as a public user and all comment author avatars become:
[...]

PR #2472, opened 2022-08-15
fix: return CommentAuthor avatar urls to unauthenticated users

justlevine:
This PR ensures the `CommentAuthor` avatar url is visible to unauthenticated users.
It solves this by using the `commentID` (stored as the `CommentAuthor`'s `databaseId`) to fetch the email if the user can't access the CommentAuthor's email directly.

Additionally, Comments without a valid email will now have their `avatar` return `null`, instead of returning a malformed `avatar.url`.

#2421

image

Any other comments?
-------------------

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (Wsl2 + devilbox + php 8.0.19)

**WordPress Version:** 6.0.1

### DIFF
--- a/src/Type/ObjectType/CommentAuthor.php
+++ b/src/Type/ObjectType/CommentAuthor.php
@@ -2,8 +2,6 @@
 
 namespace WPGraphQL\Type\ObjectType;
 
-use WPGraphQL\Data\DataSource;
-
 class CommentAuthor {
 
 	/**
@@ -56,8 +54,10 @@ class CommentAuthor {
 
 						],
 						'resolve' => function ( $comment_author, $args, $context, $info ) {
+							// If the user cant see the CommentAuthor's email, grab it from the comment.
+							$comment_author_email = ! empty( $comment_author->email ) ? $comment_author->email : get_comment_author_email( $comment_author->databaseId );
 
-							if ( ! isset( $comment_author->email ) ) {
+							if ( empty( $comment_author_email ) ) {
 								return null;
 							}
 
@@ -77,7 +77,7 @@ class CommentAuthor {
 								$avatar_args['rating'] = esc_sql( $args['rating'] );
 							}
 
-							$avatar = get_avatar_data( $comment_author->email, $avatar_args );
+							$avatar = get_avatar_data( $comment_author_email, $avatar_args );
 
 							// if there's no url returned, return null
 							if ( empty( $avatar['url'] ) ) {
--- a/tests/wpunit/CommentObjectQueriesTest.php
+++ b/tests/wpunit/CommentObjectQueriesTest.php
@@ -1,6 +1,6 @@
 <?php
 
-class CommentObjectQueriesTest extends \Codeception\TestCase\WPTestCase {
+class CommentObjectQueriesTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase {
 
 	public $current_time;
 	public $current_date;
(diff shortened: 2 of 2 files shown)


tests/wpunit/CommentObjectQueriesTest.php
```diff
@@ -664,61 +672,76 @@ public function testQueryingAvatarOnUserAuthorsIsValidForPublicAndAuthenticatedR
...
 		$this->assertArrayNotHasKey( 'errors', $actual );
-		codecept_debug( $actual );
 
 		$typename = $actual['data']['comment']['author']['node']['__typename'];
 
 		$this->assertSame( 'CommentAuthor', $typename );
 		$this->assertNotEmpty( $actual['data']['comment']['author']['node']['avatar']['url'] );
 
-		$global_id = \GraphQLRelay\Relay::toGlobalId( 'comment', $comment_by_user_id );
+		// Ensure avatar is the same when logged in.
```
justlevine, line 710: I forgot to save this as a separate commit. 👆 is the new test

jasonbahl: @justlevine thanks for pointing this out. Makes reviewing much easier 🙏🏻

src/Type/ObjectType/CommentAuthor.php
```diff
@@ -56,8 +54,10 @@ public static function register_type() {
 
 						],
 						'resolve' => function ( $comment_author, $args, $context, $info ) {
+							// If the user cant see the CommentAuthor's email, grab it from the comment.
```
jasonbahl, line 57: We should update this comment to provide more information on why we're taking this approach. 

The intent of the Model layer is to keep private data from being used where it shouldn't be used. 

This bug is a result of the Model layer doing its job correctly, and I think we should clarify in this comment for users and our future selves why we're bypassing the Model layer and using a native function to get the email address instead. 

We've determined the `User->email` is a private field for public users. By default, we don't want public users to arbitrarily make queries and get user emails. 

This bug is a result of that. 

Of course, in this case, getting the email in the resolver and returning something (an avatar) should be considered "safe" because it doesn't expose the users email address publicly, instead it's transformed into an Avatar.

I want users to understand why we're bypassing the model layer here, so that anyone that comes across this code understands it's an exception, not the rule. 

It's still "best practice" to centralize permission checks in the Model layer. This is an exception to that best practice.

justlevine: How's this?
```php
/**
 * The model only returns emails for authenticated users, but we need a valid email to return an avatar for unauthenticated requests.
 * If the email isn't visible, we use the comment ID to retrieve it.
 */
 ```

jasonbahl (approved):

justlevine:
@jasonbahl did you update the comment? Because I was only suggested the change based on your feedback, I didn't commit it 😅

jasonbahl:
Ooph. I'll fix. My bad. Case of the "moving too fasts"

jasonbahl:
@justlevine updated: https://github.com/wp-graphql/wp-graphql/commit/b8ac8af9768d65b3e5a22a93dac0283c4edb9595

src/Type/ObjectType/CommentAuthor.php
romcik: I think it's a shame that you need to go to the database again for the email that the Model already contains. A typical use case is listing the last comments. This will lead to an unnecessary increase in the number of queries to the database for an email that we already know and is only "hidden". It makes sense to take email from the database only for commenters who commented as not logged in. In this case, the email is assigned to every single comment and not to the user.

If we make email available in the model in a field that is not defined in Type, then there is no risk of public exposure. I add this note to reflect on efficiency.

justlevine: @romcik according to the [WP developer documentation](https://developer.wordpress.org/reference/functions/get_comment/), the comment should already be cached, so this shouldn't cause an extra database hit. 

If I'm wrong, please feel free to open up a new issue to address this, as this PR was already merged _and released_ so it'll probably get lost and forgotten about if the discussion stays here 🙏

romcik:
Ok, you are right, we can rely on WP core in this case. Caching without
database hit sounds good :-)

út 6. 9. 2022 v 11:11 odesílatel Dovid Levine ***@***.***>
napsal:

Issue #2366, opened 2022-05-06
🎯 Tracking: Resolving nodes by `URI`


justlevine:
This is a tracker for all the issues with bad/missing results when resolving nodes by their URI.

The general problem stems from `NodeResolver`'s [attempt](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/NodeResolver.php#L75) to replicate [`WP::parse_request`](https://developer.wordpress.org/reference/classes/wp/parse_request/) with a provided `uri`. Assumingly, some of these are due to misordered conditionals, while others are missing edge cases.

## Proposed Work
As I see it, the ideal solve requires 3 steps:
- [x] Refactor `NodeResolver` to make the code less complex, and easier to test/iterate on. ( #2656 )
- [x] Add missing `wpunit` tests to `NodeByUriTest` to bring `NodeResolver::resolve_uri()` to 100%. ( #2659 )
- [x] Fix existing bugs in resolver ( #2680 )
- [ ] Research and add support for other edge cases, using the issues below as a baseline.

## Tracking:
- [x] #2066
- [x] #1850
- [x] #1687
- [x] #1639
- [x] #2190
- [x] #2178
- [x] #2394 
- [x] #1712 
- [x] #1910 
- [x] #1673
- [ ] #2514
- [ ] #2191

If I missed an issue or if you have a use case not referenced by an existing issue, **please share and I'll add it to the tracker**.

jmotes-lapero:
@justlevine I just opened #2394 that could be added here I think.

Ririshi:
A discussion thread I opened (#2714) falls under this category as well, as was pointed out by @justlevine.

justlevine:
Update: most the issues tracked have been resolved as of v1.14.0.

- #2191 is blocked by #2491
- #2513 and #2514 need to be revalidated to see if the use of the new WP_Query handling in v1.14.0 resolved the issue tangentially.
- #2392 might be `invalid` as native WP behavior.

justlevine:
Closing this out, as #2514 is only bug left on this list (as #2191 is more of an enhancement via #2491 at this point). Between that and the Issue Triage project sheet, this ticket is unnecessary for tracking progress/providing context.

PR #2482, opened 2022-08-18
feat: PHPCS: adding support for the WordPress VIP Go standard

### LINKED ISSUE #1535: Introduce WordPress VIP coding standards and fixes for existing errors

renatonascalves:
- Introduces support for the WordPress VIP Go standard
- It either silences or fixes the errors
- It removes duplicate composer packages already bundled in `automattic/vipwpcs`

fixes #1535 

Any other comments?
-------------------
Code reviewer: see my comments for some questions/considerations.

### DIFF
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -207,7 +207,7 @@ class NodeResolver {
 
 						if ( $wp_rewrite->use_verbose_page_rules && preg_match( '/pagename=\$matches\[([0-9]+)\]/', $query, $varmatch ) ) {
 							// This is a verbose page match, let's check to be sure about it.
-							$page = get_page_by_path( $matches[ $varmatch[1] ] );
+							$page = get_page_by_path( $matches[ $varmatch[1] ] ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.get_page_by_path_get_page_by_path
 							if ( ! $page ) {
 								continue;
 							}
@@ -389,7 +389,7 @@ class NodeResolver {
 				$post_type = $this->wp->query_vars['post_type'];
 			}
 
-			$post = get_page_by_path( $this->wp->query_vars['name'], 'OBJECT', $post_type );
+			$post = get_page_by_path( $this->wp->query_vars['name'], 'OBJECT', $post_type ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.get_page_by_path_get_page_by_path
 
 			unset( $this->wp->query_vars['uri'] );
 			$post = $post instanceof WP_Post ? $this->validate_post( $post ) : null;
@@ -411,7 +411,7 @@ class NodeResolver {
 
 			$post_type = isset( $this->wp->query_vars['post_type'] ) ? $this->wp->query_vars['post_type'] : \WPGraphQL::get_allowed_post_types();
 
-			$post = get_page_by_path( $this->wp->query_vars['pagename'], 'OBJECT', $post_type );
+			$post = get_page_by_path( $this->wp->query_vars['pagename'], 'OBJECT', $post_type ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.get_page_by_path_get_page_by_path
 
 			if ( ! $post instanceof WP_Post ) {
 				return null;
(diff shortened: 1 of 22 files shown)


src/Data/NodeResolver.php
```diff
@@ -207,7 +207,7 @@ public function resolve_uri( string $uri, $extra_query_vars = '' ) {
 
 						if ( $wp_rewrite->use_verbose_page_rules && preg_match( '/pagename=\$matches\[([0-9]+)\]/', $query, $varmatch ) ) {
 							// This is a verbose page match, let's check to be sure about it.
-							$page = get_page_by_path( $matches[ $varmatch[1] ] );
+							$page = get_page_by_path( $matches[ $varmatch[1] ] ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.get_page_by_path_get_page_by_path
```
renatonascalves, line 210: For this one, should we do a if/else strategy with `wpcom_vip_get_page_by_path`?

src/Router.php
```diff
@@ -392,17 +392,18 @@ public static function set_headers() {
...
-	 * @global string php://input Raw post data.
 	 * @since  0.0.5
+	 *
+	 * @global string php://input Raw post data.
+	 *
+	 * @return string|false Raw request data.
 	 */
 	public static function get_raw_data() {
-		$input = file_get_contents( 'php://input' );
+		$input = file_get_contents( 'php://input' ); // phpcs:ignore WordPressVIPMinimum.Performance.FetchingRemoteData.FileGetContentsRemoteFile
```
renatonascalves, line 402: VIP recommends:

> `file_get_contents()` is highly discouraged for remote requests, please use `wpcom_vip_file_get_contents()` or `vip_safe_wp_remote_get()`

src/Type/ObjectType/RootQuery.php
```diff
@@ -683,11 +683,11 @@ public static function register_post_object_fields() {
 								break;
 							case 'source_url':
 								$url     = $args['id'];
-								$post_id = attachment_url_to_postid( $url );
+								$post_id = attachment_url_to_postid( $url ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.attachment_url_to_postid_attachment_url_to_postid
```
renatonascalves, line 686: Another one where we could do a if/else strategy with `wpcom_vip_attachment_url_to_postid`

justlevine (commented):
Awesome work @renatonascalves !

Re adopting `vipcom_` functions, not saying we should, but _if_ we do: 
- let's wrap the `if( method_exists( 'vipcom_fn') {return vipcom_fn();} return wp_fn();` logic in a separate class (`Utils\Plugabble` or something) to keep things DRY.
*Edit:* we also might want to add a hook, or drop the namespace and wrap each method in its own `method_exists()`, so we can provide the same first-party parity with non-WPVip setups or plugins like `wp-graphql-smart-cache`
- A lot of those methods employ caching, so we should make sure they're unit tested to avoid issues we had with the ObjectCursor in #2457 
- Anything in `NodeResolver::resolve_uri()` should be left untouched (ie errors suppressed, no code change) until #2342 is merged. Currently most of it is lifted directly from `WP:parse_request()`, isn't unit tested, and is a source for a plethora of bugs ( 👀 #2366 ), so we should avoid changing it unless absolutely necessary.

renatonascalves:
@justlevine May I suggest then to move on silencing those examples in this issue? I'll create issues to address those separate changes.

Introducing unit tests, for example, would increase the scope a little bit since we would need to add support for the VIP `mu-plugins` into the unit tests and write the tests themselves.

renatonascalves:
@jasonbahl Could you code review this pull request when you have a chance?

jasonbahl (approved):

jasonbahl:
@renatonascalves I think I agree with @justlevine about adding vip support in a separate file. 

If any of these functions can't be filtered and replaced with the VIP equivalent, it's probably a good candidate for a new filter. 

Ideally, if do need to introduce new filters, I'd like to keep them central when possible. 

For example (doesn't apply here, but just to set the tone), if we needed to filter something in the UserLoader, if possible I'd prefer the `apply_filters()` call to exist in the AbstractDataLoader, not the UserLoader. That way _any_ loader could use the filter, not _just_ the user loader. (sometimes it can't be centralized, and that's ok, but when it can be, that's my preference).

renatonascalves:
Thank you! I'll proceed to merge this and create issues for the follow-up work.

Do you guys know what's this **Check code (8.0.15)** Github Action about?

jasonbahl:
@renatonascalves I think  I need to update the required checks. I think it's tied to an old PHP version that was updated. Looking now 👀

jasonbahl:
@renatonascalves ya, it was related to a change on the Github Workflow that checks PHPStan. 

I updated the branch settings. Good to go now.

renatonascalves:
Thank you, Jason!

Issue #2100, opened 2021-09-19
resetUserPassword returns null sometimes


web-programmer-here:
I am trying to use `resetUserPassword` mutation and sometimes it returns null when trying the update the password for the user. So something like this: 

```graphql
mutation RESET_USER_PASSWORD {
  resetUserPassword(input: {
    key: "n4hyZzDW9Of4zbTs",
    login: "user@user.com",
    password: "123456"
  }) {
    user {
      id
    }
  }
}
```

and I get this in response 

```
{
    "data": {
        "resetUserPassword": {
            "user": null
        }
    },
    "extensions": {
        "debug": [
            {
                "type": "DEBUG_LOGS_INACTIVE",
                "message": "GraphQL Debug logging is not active. To see debug logs, GRAPHQL_DEBUG must be enabled."
            }
        ]
    }
}
```

however, the password actually changes, and I get the email that the password has been changed but returned data is null. I am also facing same issue with `sendPasswordResetEmail` where the email is sent but the response is null. Any idea why this is happening?

web-programmer-here:
I noticed I dont see this issue when running this query in GraphQL IDE, here is a postman screenshot: 

<img width="906" alt="image" src="https://user-images.githubusercontent.com/81479051/133913638-3de85c3e-e8cc-4d7e-8571-562d38309963.png">

and here is another password reset using IDE: 

<img width="1048" alt="image" src="https://user-images.githubusercontent.com/81479051/133913678-025d85a6-9158-4a96-8a7c-44c594ec9423.png">

so when using the IDE, we dont face this issue (as far as I know when testing) - is there some settings I have to change?

jasonbahl:
@web-programmer-here The reason the response is `null` when you run the mutation outside of the IDE is because the request is not authenticated and the current user object cannot be returned in the payload. 

When executing in the GraphQL IDE in WordPress, the GraphQL request is authenticated, so the current user is returned.

web-programmer-here:
@jasonbahl but the password gets updated, it just returns null for the response, there is no way to fix this?

grazianodev:
+1 for this. I usually rely on looking at the data in the response to verify that the query/mutation worked as expected, so if the `sendPasswordResetEmail` and `resetUserPassword` mutations could return the user object even if not authenticated, or even just a flag such as `resetEmailSuccess` and `passwordResetSuccess` would be nice.

justlevine:
@grazianodev 
[...]

PR #2505, opened 2022-09-06
feat: login user after successful `resetUserPassword` mutation

justlevine:
This PR logs in the user after they successfully `resetUserPassword`, as they have successfully authenticated themselves.
This gives them permissions to see their own `user` data in the GraphQL response.

#2100 

Any other comments?
-------------------
Discussion may still be necessary to decide both if we should return the user at all and whether logging in the user is the best way to do it. But it was a simple PR, and I had some free time.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + PHP 8.0.19)

**WordPress Version:** 6.0.2

### DIFF
--- a/src/Mutation/ResetUserPassword.php
+++ b/src/Mutation/ResetUserPassword.php
@@ -101,6 +101,9 @@ class ResetUserPassword {
 			 */
 			reset_password( $user, $input['password'] );
 
+			// Log in the user, since they already authenticated with the reset key.
+			wp_set_current_user( $user->ID );
+
 			/**
 			 * Return the user ID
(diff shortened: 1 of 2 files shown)


jasonbahl (changes requested):
@justlevine I think this change makes sense. 

Looks like there's some failing tests, but I'm cool with proceeding with this change.

justlevine:
I see.... weird that its not failing consistently. Will get this polished off tomorrow 🤞

jasonbahl (approved):

Issue #2466, opened 2022-08-04
Add post type / taxonomy args for defining the GraphQL model


justlevine:
### What problem does this address?

There are numerous cases where a custom post type / taxonomy needs to use a custom model. Currently implementing a custom model is a complicated and imperfect process, as the model is usually hard-coded into the data loader and connection resolver. This requires devs to extend/replace those classes in addition to all the `resolve` callbacks that use them.

### What is your proposed solution?

Continuing the work done in #2399 , a custom property should be added to `WP_Post_Type` and `WP_Taxonomy` that allows a user to set a custom model. 

This model should then be used by the various DataLoader/ConnectionResolvers, falling back to the defaults when none is set.

### Additional Context

- The plugin should dogfood this approach with `nav_menu` (taxonomy) and `nav_menu_items` (post_type).
- Assumption:  If done correctly, there should be rarely (if ever) need to be a reason to create a custom DataLoader for a CPT/Tax term, as they mostly exist to work around the [hardcoded `Model` instantiation in core](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Loader/PostObjectLoader.php#L53). This is premised on my own experience, as well as browsing the source code of [WooGraphQL](https://github.com/wp-graphql/wp-graphql-woocommerce/blob/develop/includes/data/loader/class-wc-cpt-loader.php) and [WPGraphQL for BP](https://github.com/wp-graphql/wp-graphql-buddypress/tree/main/src/Data/Loader) . @kidunot89 , @renatonascalves if I'm missing a use case, please chime in 🙏
- The same assumption _cannot_ be made about Connection Resolvers. Maybe in the future (after #1738, #2027) we can use the connection interface to determine what the resolver does with query args, but for now we probably need a separate WP_Post_Type / WP_Taxonomy parameter to supply a custom Resolver class so users dont need to filter every possible resolve callback to avoid issues like #2447

renatonascalves:
> This is premised on my own experience, as well as browsing the source code...

BuddyPress actually does not use custom post types. So our custom loaders are necessary.

justlevine:
> BuddyPress actually does not use custom post types or taxonomies.

Thanks for clarifying. Not sure why I thought `Groups` was a custom tax 😅

renatonascalves:
No problem. I do like the idea though. Whatever we can do to make it easier, the better. But it needs to be flexible enough that one can hook into it, via callbacks or hooks.

PR #2519, opened 2022-09-14
feat: Add QueryAnalyzer and Expose "X-GraphQL-Keys" in headers

### LINKED ISSUE #2516: Return Query ID and GraphQL Keys in response headers

We've been working on some functionality in the WPGraphQL Smart Cache plugin that I believe should become core WPGraphQL functionality. 

When a GraphQL Query is executed via HTTP request, WPGraphQL Smart Cache analyzes the query document and tracks resolution, then compiles a list of "GraphQL Keys" that are then returned in the request headers.

For example:

CleanShot 2022-09-12 at 09 53 29

These keys can be used by caching clients to "tag" the cached document, and then better understand how to evict the cached document when associated events occur. 

I propose that the logic used in WPGraphQL Smart Cache to generate the keys and return the headers be ported to and maintained in WPGraphQL Core. 

We're currently maintaining this logic in a separate plugin. I believe that the benefits of having this functionality enabled in core WPGraphQL outweigh the cost of maintaining it in a separate plugin. It's also a feature that can benefit _all_ users of WPGraphQL, vs a subset of the userbase (like WPGraphQL for ACF, for example, which only benefits users of ACF)

We can ensure that _all_ users of WPGraphQL benefit from this feature, and then ideally benefit from caching clients using this feature with less friction than having to seek out and install an additional plugin. 

I also believe the maintenance burden will be significantly lower and the quality will remain higher if this functionality is part of the core WPGraphQL plugin. 

### Additional Context

_No response_

jasonbahl:
This introduces a new "QueryAnalyzer" that tracks Types, Models and Nodes that are queried and returned in any given GraphQL Request, then returns those values in the headers. 

For example, making a query like so `/graphql?query={posts(first:5){nodes{id,title}}}` will return headers like so: 

```bash
X-GraphQL-Keys: list:post cG9zdDoxMTY3 cG9zdDoxMTM1 cG9zdDoxMTIw cG9zdDoxMTA5 cG9zdDoxMDk3 graphql:Query d43c3d17aeaf76b21e747f5350a36262
X-GraphQL-Keys-Size: 121
X-GraphQL-List-Types: list:post
X-GraphQL-Nodes: cG9zdDoxMTY3 cG9zdDoxMTM1 cG9zdDoxMTIw cG9zdDoxMTA5 cG9zdDoxMDk3
X-GraphQL-Operation-Type: Query
X-GraphQL-Query-ID: d43c3d17aeaf76b21e747f5350a36262
```

CleanShot 2022-09-14 at 14 38 42

closes #2516 

Any other comments?
-------------------
The goal of this is to centralize collecting info that can be used to identify queries via logs or cache clients (varnish/fastly, etc)

With the information returned in these headers, caches can "tag" the cached query with this information and then corresponding events can evict said caches based on these tags.

Additionally error logging, performance tracking, etc can be better organized by the data returned in these headers.

**Note**: Much of this functionality was originally written in the WPGraphQL Smart Cache plugin. This brings it to the core plugin to help with maintenance overhead and ultimately benefit a wider audience of users.

### DIFF
--- a/src/Data/Connection/PostObjectConnectionResolver.php
+++ b/src/Data/Connection/PostObjectConnectionResolver.php
@@ -334,7 +334,7 @@ class PostObjectConnectionResolver extends AbstractConnectionResolver {
 						}
 					}
 
-					$query_args['orderby'][ esc_sql( $orderby_input['field'] ) ] = esc_sql( $order );
+					$query_args['orderby'][ $orderby_input['field'] ] = (string) $order;
 				}
 			}
 		}
--- a/src/Data/Connection/PostObjectConnectionResolver.php
+++ b/src/Data/Connection/PostObjectConnectionResolver.php
@@ -334,7 +334,7 @@ class PostObjectConnectionResolver extends AbstractConnectionResolver {
 						}
 					}
 
-					$query_args['orderby'][ $orderby_input['field'] ] = (string) $order;
+					$query_args['orderby'][ esc_sql( $orderby_input['field'] ) ] = esc_sql( $order );
 				}
 			}
 		}
--- a/src/Mutation/MediaItemUpdate.php
+++ b/src/Mutation/MediaItemUpdate.php
@@ -45,7 +45,7 @@ class MediaItemUpdate {
 						'non_null' => 'ID',
 					],
 					// translators: the placeholder is the name of the type of post object being updated
-					'description' => sprintf( __( 'The ID of the %1$s object', 'wp-graphql' ), $post_type_object->graphql_single_name ?? 'MediaItem' ),
+					'description' => sprintf( __( 'The ID of the %1$s object', 'wp-graphql' ), $post_type_object->graphql_single_name ),
 				],
 			]
 		);
--- a/src/Mutation/MediaItemUpdate.php
+++ b/src/Mutation/MediaItemUpdate.php
@@ -6,6 +6,7 @@ use Exception;
 use GraphQL\Error\UserError;
 use GraphQL\Type\Definition\ResolveInfo;
 use GraphQLRelay\Relay;
+use WP_Post_Type;
 use WPGraphQL\AppContext;
 use WPGraphQL\Data\MediaItemMutation;
 use WPGraphQL\Utils\Utils;
@@ -34,7 +35,7 @@ class MediaItemUpdate {
 	 * @return array
 	 */
 	public static function get_input_fields() {
-		/** @var \WP_Post_Type $post_type_object */
+		/** @var WP_Post_Type $post_type_object */
 		$post_type_object = get_post_type_object( 'attachment' );
 		return array_merge(
 			MediaItemCreate::get_input_fields(),
@@ -44,7 +45,7 @@ class MediaItemUpdate {
 						'non_null' => 'ID',
 					],
 					// translators: the placeholder is the name of the type of post object being updated
-					'description' => sprintf( __( 'The ID of the %1$s object', 'wp-graphql' ), $post_type_object->graphql_single_name ),
+					'description' => sprintf( __( 'The ID of the %1$s object', 'wp-graphql' ), $post_type_object->graphql_single_name ?? 'MediaItem' ),
 				],
 			]
 		);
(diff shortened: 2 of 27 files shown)


src/Data/Connection/PostObjectConnectionResolver.php
```diff
@@ -334,7 +334,7 @@ public function get_query_args() {
 						}
 					}
 
-					$query_args['orderby'][ esc_sql( $orderby_input['field'] ) ] = esc_sql( $order );
+					$query_args['orderby'][ $orderby_input['field'] ] = (string) $order;
```
justlevine, line 337: Not sure whether enum values need to be escaped, but for consistency we should do the same thing in other places: https://github.com/wp-graphql/wp-graphql/blob/8dc66601a825f3616d6701a5d875e0f509a15dbd/src/Data/Connection/PostObjectConnectionResolver.php#L324

src/Type/ObjectType/Avatar.php
```diff
@@ -14,6 +16,7 @@ public static function register_type() {
 			'Avatar',
 			[
 				'description' => __( 'Avatars are profile images for users. WordPress by default uses the Gravatar service to host and fetch avatars from.', 'wp-graphql' ),
+				'model'       => AvatarModel::class,
```
justlevine, line 19: Possibly beyond the scope of this PR, but maybe we should abstract this to a filterable getter() instead of hardcoding the model type?

I know this PR is just focused on displaying them in query analyzer, but the logical next step would be to have `TypeRegistry->register_type()` actually use this define the model for the object. 

cc #2466

justlevine (commented):
Looks awesome! Left a few comments, _none_ of which are actual blockers.

src/Utils/QueryAnalyzer.php
```diff
@@ -0,0 +1,489 @@
...
+	 *
+	 * @param ?Schema $schema The WPGraphQL Schema
+	 * @param ?string $query  The query string
+	 *
+	 * @return array
+	 * @throws SyntaxError|Exception
+	 */
+	public function set_list_types( ?Schema $schema, ?string $query ): array {
+
+		/**
```
justlevine, line 207: Not sure if the docblocks are used for auto-generation, but if so I think this is a bit more intuitive:
```php
/**
 * @param array|null $list_types. Defaults to null.
 * @param ?Schema $schema The WPGraphQL Schema for the current request
 * @param ?string $query The query string being requested
 */
 $pre_get_list_types = apply_filters( 'graphql_pre_query_analyzer_get_list_types, null, $shema, $query );
 ```
 
 Same for the the other filters using `$null` in this file.
 
 Eitther way the first param should be `array|null` for advanced static analysis.

File: src/Mutation/MediaItemUpdate.php [hunk 4]
```diff
@@ -44,7 +45,7 @@ public static function get_input_fields() {
 						'non_null' => 'ID',
 					],
 					// translators: the placeholder is the name of the type of post object being updated
-					'description' => sprintf( __( 'The ID of the %1$s object', 'wp-graphql' ), $post_type_object->graphql_single_name ),
+					'description' => sprintf( __( 'The ID of the %1$s object', 'wp-graphql' ), $post_type_object->graphql_single_name ?? 'MediaItem' ),
```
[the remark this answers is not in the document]
jasonbahl: This was a PHPStan thing. 🤔 

Let me take another look at _why_ it was complaining here. 👀

src/Type/ObjectType/Avatar.php
jasonbahl: Ya, we can definitely iterate on this. Right now the intent of this is to be able to identify the relationship between a node that was resolved and the Types that were asked for. 

Sometimes Models are loaded, but not returned. 

For example, a query for `{ posts { nodes { id, title } } }` will load the User Model for the author(s) of the posts (to determine ownership, etc), but not return those models because the query didn't ask for them. 

So, if each Type identifies the Model that's used to resolve its node, as is done here, then we can identify the intersection between Models/Nodes that were loaded and those that were asked for in the query. 

Ultimately we can use this Model definition elsewhere in resolvers/loaders, etc.

PR #2521, opened 2022-09-17
chore: refactor `AbstractConnectionResolver::$args` handling

### LINKED ISSUE #998: Update Resolvers for all ID inputs to behave consistently

justlevine:
This PR deprecates the `AbstractConnectionResolver::getArgs()` in favor of `AbstractConnectionResolver::get_args()`, which is passed through the `graphql_connection_args` WP filter on instantiation.

This PR replaces part of #2340 , and lays the foundation to create the individual PRs necessary to close #998 in a piecemeal approach, by allowing child `ConnectionResolver` classes and 3rd-party code to manipulate connection arg values. before they are used in the resolver.

…

Any other comments?
-------------------
- As the functionality will be tested in future PRs, I opted to avoid the redundant work of adding unit tests.

Where has this been tested?
---------------------------

**WordPress Version:** 6.0.2

### DIFF
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -143,11 +143,6 @@ abstract class AbstractConnectionResolver {
 		$this->source = $source;
 
 		/**
-		 * Set the args for the resolver
-		 */
-		$this->args = $args;
-
-		/**
 		 * Set the context of the resolver
 		 */
 		$this->context = $context;
@@ -161,6 +156,22 @@ abstract class AbstractConnectionResolver {
 		 * Get the loader for the Connection
 		 */
 		$this->loader = $this->getLoader();
+
+		/**
+		 * Set the args for the resolver
+		 */
+		$this->args = $args;
+
+		/**
+		 *
+		 * Filters the GraphQL args before they are used in get_query_args().
+		 *
+		 * @param array                      $args                The GraphQL args passed to the resolver.
+		 * @param AbstractConnectionResolver $connection_resolver Instance of the ConnectionResolver
+		 *
+		 * @since @todo
+		 */
+		$this->args = apply_filters( 'graphql_connection_args', $this->getArgs(), $this );
 
 		/**
 		 * Determine the query amount for the resolver.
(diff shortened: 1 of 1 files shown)


src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -162,6 +157,22 @@ public function __construct( $source, array $args, AppContext $context, ResolveI
...
+		/**
+		 *
+		 * Filters the GraphQL args before they are used in get_query_args().
+		 *
+		 * @param array                      $args                The GraphQL args passed to the resolver.
+		 * @param AbstractConnectionResolver $connection_resolver Instance of the ConnectionResolver
+		 *
+		 * @since @todo
+		 */
+		$this->args = apply_filters( 'graphql_connection_args', $this->get_args(), $this );
```
jasonbahl, line 174: I think I'm ok with this filter being added, but it might be a _bit_ naiive as it only targets connections, and not _any_ type of field. 

I was thinking it would make sense to allow filtering args on _any_ field, not just connections. 

Like described here: https://github.com/wp-graphql/wp-graphql/issues/965#issuecomment-539719493

I think it would be nice to be able to filter args _before_ they're used in resolution, but _also_ do something with the args _after_ resolution but _before_ the response is returned. 

Anyway, I think we can approach the bigger picture separately. This works for connections and has immediate benefits for working with connections.

src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -212,12 +223,28 @@ protected function getLoader() {
 		return $this->context->get_loader( $name );
 	}
 
-	/**
+		/**
 	 * Returns the $args passed to the connection
 	 *
+	 * Deprecated in favor of $this->get_args();
+	 *
+	 * @deprecated @todo
```
jasonbahl, line 231: We can update the deprecated version/message when release is prepped. 👍🏻

jasonbahl (approved):

PR #2532, opened 2022-09-21
fix: regression to menuItem queries when parentId is 0

### LINKED ISSUE #2530: regression: Querying for Menu Items with parentId = 0 returns all menu items
### Description

Previously, querying for menuItems with parentId=0 only returned top level items. Now it's returning all items.

### Steps to reproduce

see: https://github.com/wp-graphql/wp-graphql/pull/2523#issuecomment-1253642196

### Additional context

This is a regressions with release v1.11.0

### WPGraphQL Version

v1.11.0

### WordPress Version

6.0

### PHP Version

8.0

_No response_

- [X] Yes

- [X] Yes

jasonbahl:
This fixes a regression where using 0 as the `parentId` for menuItem queries is no longer returning only menuItems with no parent set. 

closes #2530 

Any other comments?
-------------------

Failing test run: https://github.com/wp-graphql/wp-graphql/actions/runs/3100423158/jobs/5020687861#step:7:270
Failing test commit: https://github.com/wp-graphql/wp-graphql/pull/2532/commits/0575f93646418488c27140c4c301084b29fa22ad

Passing test run: https://github.com/wp-graphql/wp-graphql/actions/runs/3100510553/jobs/5020872900
Passing test commit: https://github.com/wp-graphql/wp-graphql/pull/2532/commits/667558891c7fa9c8a6ebd4c8cae88b1923d2daf1

### DIFF
--- a/src/Data/Connection/MenuItemConnectionResolver.php
+++ b/src/Data/Connection/MenuItemConnectionResolver.php
@@ -50,7 +50,7 @@ class MenuItemConnectionResolver extends PostObjectConnectionResolver {
 			$query_args['meta_value'] = (int) $this->args['where']['parentDatabaseId']; // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
 		}
 
-		if ( isset( $this->args['where']['parentId'] ) ) {
+		if ( ! empty( $this->args['where']['parentId'] ) || ( isset( $this->args['where']['parentId'] ) && 0 === (int) $this->args['where']['parentId'] ) ) {
 			$query_args['meta_key']   = '_menu_item_menu_item_parent'; // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_key
 			$query_args['meta_value'] = $this->args['where']['parentId']; // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
 		}
--- a/src/Data/Connection/MenuItemConnectionResolver.php
+++ b/src/Data/Connection/MenuItemConnectionResolver.php
@@ -50,10 +50,9 @@ class MenuItemConnectionResolver extends PostObjectConnectionResolver {
 			$query_args['meta_value'] = (int) $this->args['where']['parentDatabaseId']; // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
 		}
 
-		if ( ! empty( $this->args['where']['parentId'] ) ) {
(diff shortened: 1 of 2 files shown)


src/Data/Connection/MenuItemConnectionResolver.php
```diff
@@ -50,10 +50,9 @@ public function get_query_args() {
 			$query_args['meta_value'] = (int) $this->args['where']['parentDatabaseId']; // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
 		}
 
-		if ( ! empty( $this->args['where']['parentId'] ) ) {
-
-				$query_args['meta_key']   = '_menu_item_menu_item_parent'; // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_key
-				$query_args['meta_value'] = $this->args['where']['parentId']; // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
+		if ( isset( $this->args['where']['parentId'] ) ) {
```
justlevine, line 53: I think this conditional needs to be more explicit, since `$this->get_args()` runs before `get_query_args()`, and a bad (vs falsy) input value will cause `Utils::get_database_id_from_id()` to return `false`

jasonbahl: ok, so probably true with the `isset` in place a few lines above as well?

justlevine: No because `parentDatabaseId` doesn't get passed through that Utils function, so `isset()` there is fine.

I think `( 0 === $this->args[...] || ! empty( $this->args[...] ) )` logic we want.

jasonbahl: ok, I can run with that.

jasonbahl: > I think ( 0 === $this->args[...] || ! empty( $this->args[...] ) ) logic we want.

We need to check `isset( $this->args['where']['parentId'] )` before evaluating if it's equal to zero. 

So it'll need to be something like `! empty( $... ) || ( isset( $... ) && 0 === $... )`

jasonbahl: see: https://github.com/wp-graphql/wp-graphql/pull/2532/commits/6bac3841342bf36413f979721bfe8f2291301030

justlevine (approved):

PR #2534, opened 2022-09-22
feat: Add graphql_resolve_uri_query_vars for extra query vars

### LINKED ISSUE #2515: `graphql_pre_resolve_uri` does not receive extra query vars

I am trying to write custom resolution logic for single post queries based on slug.

Spec: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/NodeResolver.php#L75

```
public function resolve_uri( string $uri, $extra_query_vars = '' ) {

/**
* When this filter return anything other than null, it will be used as a resolved node
* and the execution will be skipped.
*
* This is to be used in extensions to resolve their own nodes which might not use
* WordPress permalink structure.
*
* @param mixed|null $node The node, defaults to nothing.
* @param string $uri The uri being searched.
* @param AppContext $content The app context.
* @param WP $wp WP object.
*/
$node = apply_filters( 'graphql_pre_resolve_uri', null, $uri, $this->context, $this->wp);
```

However I want to preserve all of the logic, but in our situation we have a meta field, per object (posts including custom post types and users) that specifies a geographic region, and we want each user to only have access to content for their specific region.

But the hook doesn't receive `$extra_query_vars` and as a result, it is not aware of which content type is being passed; hence much of the resolution logic defaults to **post** or **page**, and fails for custom post types.

Ideally, if this hook is to substitute the original resolution, it should also receive the original arguments passed to the resolver function.

```
$node = apply_filters( 'graphql_pre_resolve_uri', null, $uri, $this->context, $this->wp, $extra_query_vars);
```

I have considered:
1. Using numeric IDs exclusively in single queries, but while it should have been the initial approach, it's too much of a refactor.
2. re-declaring RootQuery single types, originally declared in `register_post_object_fields` and routing them to my own resolution function, but this defeats the point.
3. Other manners of monkey patching which seemed evil (ie. "pluggable" WP Core functions to disambiguate on the resolution).
4. A custom "where" GraphQL arg on single queries, inelegant.

And suggesting other ways/hooks to break page resolution down. 

For example: a filter or function that "proxies" `get_page_by_path` (WP Core) and enables more granular implementations of custom singles resolution.

Depending on feasibility, this last suggestion might be the best long term option, but in the interim, I think simply passing *$extra_query_vars* through could do the job.

### Additional Context

_No response_

markkelnar:
Add filter during the resolve_uri function to filter the request with it's extra query vars to allow a node to be returned based on the filter.

closes #2515 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -339,6 +339,20 @@ class NodeResolver {
 		/**
 		 * Filters the array of parsed query variables.
 		 *
+		 * @param mixed|null $node The node, defaults to nothing.
+		 * @param string $uri The uri being searched.
+		 * @param AppContext $content The app context.
+		 * @param WP $wp WP object.
+		 * @param mixed|array|string $extra_query_vars Any extra query vars to consider
+		 */
+		$node = apply_filters( 'graphql_resolve_uri_query_vars', null, $uri, $this->context, $this->wp, $extra_query_vars );
+		if ( ! empty( $node ) ) {
+			return $node;
+		}
+
+		/**
+		 * Filters the array of parsed query variables.
+		 *
 		 * @param array $query_vars The array of requested query variables.
 		 *
 		 * @since 2.1.0
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -345,7 +345,7 @@ class NodeResolver {
 		 * @param WP $wp WP object.
 		 * @param mixed|array|string $extra_query_vars Any extra query vars to consider
 		 */
-		$node = apply_filters( 'graphql_resolve_uri_query_vars', null, $uri, $this->context, $this->wp, $extra_query_vars );
+		$node = apply_filters( 'graphql_pre_resolve_uri_with_query_vars', null, $uri, $this->context, $this->wp, $extra_query_vars );
 		if ( ! empty( $node ) ) {
 			return $node;
 		}


src/Data/NodeResolver.php
```diff
@@ -336,6 +336,20 @@ public function resolve_uri( string $uri, $extra_query_vars = '' ) {
...
+		/**
+		 * Filters the array of parsed query variables.
+		 *
+		 * @param mixed|null $node The node, defaults to nothing.
+		 * @param string $uri The uri being searched.
+		 * @param AppContext $content The app context.
+		 * @param WP $wp WP object.
+		 * @param mixed|array|string $extra_query_vars Any extra query vars to consider
+		 */
+		$node = apply_filters( 'graphql_resolve_uri_query_vars', null, $uri, $this->context, $this->wp, $extra_query_vars );
```
jasonbahl, line 348: @markkelnar When using this pattern of having a `null` and then checking if it's not null and returning, we use `graphql_pre_` as the prefix, stemming from patterns in WordPress core. 

I think this looks good, but I think we should change the name to something like: `graphql_pre_resolve_uri_with_query_vars`

jasonbahl, line 348:
```suggestion
		$node = apply_filters( 'graphql_pre_resolve_uri_with_query_vars', null, $uri, $this->context, $this->wp, $extra_query_vars );
```

jasonbahl (approved):

jasonbahl (changes requested):
Let's get a test. 

Here's one scenario we can test: 

```
create a post
create another post

query for 1st post by uri

assert first post is returned

add filter that returns post 2 instead

run the same query

assert post 2 is now returned
```

justlevine:
Sorry I know @jasonbahl suggested naming the filter `graphql_pre_resolve_uri_with_query_vars`, but I gotta disagree.

Not just because [I think that `graphql_pre_resolve_uri` should also pass `$extra_query_vars` ](https://github.com/wp-graphql/wp-graphql/issues/2515#issuecomment-1257076618), but also because its not communicating what makes this filter useful to the user: it happens towards the end of the `parse_request()` logic once `$this->wp` has been populated.

Perhaps naming the filter just `graphql_resolve_uri` makes sense here?

For similar reasons, I dont think that we need to or even should pass `$extra_query_vars` at this point of the method, since it's already been set (and possibly manipulated) on `$this->wp` .

Also, I think we should move it a few lines down to _after_ `do_action_ref_array( 'parse_request', [ &$this->wp ] );`, since thats also part of the `parse_request()` logic and the `$this->wp` population.

(Ideally, Id suggest putting `$node = apply_filters( 'graphql_pre_parse_request', null, $uri, $this->context, $this->wp, $extra_query_vars );` at the top of the method, and moving the existing `$node = apply_filters( 'graphql_pre_resolve_uri, null, $uri, $this->context, $this->wp );`  to after the `do_action_ref_array( ... )`, since IMO that makes the most sense semantically, but I'm not sure whether that would be considered a breaking - or even an unjustifiably annoying - change ).

jasonbahl:
closing as #2582 was merged

PR #2538, opened 2022-09-24
chore: update phpstan and fix surfaced issues

justlevine:
This PR updates the `phpstan` and `phpstan-wordpress` Composer dependencies to their latest versions, and fixes newly surfaced static analysis issues due to this update.

Specific details of these changes are in the comment review.

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (WSL2 + devilbox + PHP 8.0.19)

**WordPress Version:** 6.0.2

### DIFF
--- a/src/Data/Connection/PostObjectConnectionResolver.php
+++ b/src/Data/Connection/PostObjectConnectionResolver.php
@@ -308,7 +308,12 @@ class PostObjectConnectionResolver extends AbstractConnectionResolver {
 		 */
 		if ( isset( $this->args['where']['orderby'] ) && is_array( $this->args['where']['orderby'] ) ) {
 			$query_args['orderby'] = [];
+
 			foreach ( $this->args['where']['orderby'] as $orderby_input ) {
+				if ( empty( $orderby_input['field'] ) ) {
+					continue;
+				}
+
 				/**
 				 * These orderby options should not include the order parameter.
 				 */
@@ -322,20 +327,22 @@ class PostObjectConnectionResolver extends AbstractConnectionResolver {
 					true
 				) ) {
 					$query_args['orderby'] = esc_sql( $orderby_input['field'] );
-				} elseif ( ! empty( $orderby_input['field'] ) ) {
 
-					$order = $orderby_input['order'];
-
-					if ( isset( $query_args['graphql_args']['last'] ) && ! empty( $query_args['graphql_args']['last'] ) ) {
-						if ( 'ASC' === $order ) {
-							$order = 'DESC';
-						} else {
-							$order = 'ASC';
-						}
-					}
-
-					$query_args['orderby'][ esc_sql( $orderby_input['field'] ) ] = esc_sql( $order );
+					// If we're ordering explicitly, there's no reason to check other orderby inputs.
+					break;
 				}
+
+				$order = $orderby_input['order'];
+
+				if ( isset( $query_args['graphql_args']['last'] ) && ! empty( $query_args['graphql_args']['last'] ) ) {
+					if ( 'ASC' === $order ) {
+						$order = 'DESC';
+					} else {
+						$order = 'ASC';
+					}
+				}
+
+				$query_args['orderby'][ esc_sql( $orderby_input['field'] ) ] = esc_sql( $order );
 			}
 		}
 
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -96,12 +96,26 @@ class NodeResolver {
 
 		$parsed_url = wp_parse_url( $uri );
 
+		if ( false === $parsed_url ) {
+			graphql_debug( __( 'Cannot parse provided URI', 'wp-graphql' ), [
+				'uri' => $uri,
+			] );
+			return null;
+		}
+
 		if ( isset( $parsed_url['host'] ) ) {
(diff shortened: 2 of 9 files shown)


src/Data/Connection/PostObjectConnectionResolver.php
```diff
@@ -322,20 +327,22 @@ public function get_query_args() {
 					true
 				) ) {
 					$query_args['orderby'] = esc_sql( $orderby_input['field'] );
-				} elseif ( ! empty( $orderby_input['field'] ) ) {
 
-					$order = $orderby_input['order'];
+					// If we're ordering explicitly, there's no reason to check other orderby inputs.
+					break;
```
justlevine, line 332: The point of this conditional is that `post__in` etc overwrites all other sorting criteria. As such, we should bail early instead of overwriting it if the user unwittingly orders by this _and_ other fields.

src/Data/Connection/PostObjectConnectionResolver.php
```diff
@@ -322,20 +327,22 @@ public function get_query_args() {
...
-					$query_args['orderby'][ esc_sql( $orderby_input['field'] ) ] = esc_sql( $order );
+				if ( isset( $query_args['graphql_args']['last'] ) && ! empty( $query_args['graphql_args']['last'] ) ) {
+					if ( 'ASC' === $order ) {
+						$order = 'DESC';
+					} else {
+						$order = 'ASC';
+					}
 				}
+
+				$query_args['orderby'][ esc_sql( $orderby_input['field'] ) ] = esc_sql( $order );
```
justlevine, line 345: Before the [above change](https://github.com/wp-graphql/wp-graphql/pull/2538/files?diff=split&w=1#r979216342), this was throwing an error since `$query_args['orderby']` could be a `string` if a previous `$orderby_input['field']` was `post__in` etc.

src/Data/NodeResolver.php
```diff
@@ -96,12 +96,26 @@ public function resolve_uri( string $uri, $extra_query_vars = '' ) {
 
 		$parsed_url = wp_parse_url( $uri );
 
+		if ( false === $parsed_url ) {
+			graphql_debug( __( 'Cannot parse provided URI', 'wp-graphql' ), [
```
justlevine, line 100: it's rare, but `wp_parse_url()` can sometimes fail to parse if it gets malformed input. This also prevents a handful of php errors from accessing array keys on an `array|false`

src/Data/NodeResolver.php
```diff
@@ -115,7 +129,7 @@ public function resolve_uri( string $uri, $extra_query_vars = '' ) {
 		$this->wp->query_vars = [];
 		$post_type_query_vars = [];
 
-		if ( isset( $parsed_url['query'] ) && '/' === $parsed_url['path'] ) {
+		if ( isset( $parsed_url['query'] ) && ( empty( $parsed_url['path'] ) || '/' === $parsed_url['path'] ) ) {
```
justlevine, line 132: if `$uri = '?myparam';` then `$parsed_url['path']` is not smart enough to know it refers to the home page.

src/Data/NodeResolver.php
```diff
@@ -144,7 +158,7 @@ public function resolve_uri( string $uri, $extra_query_vars = '' ) {
 			$pathinfo         = str_replace( '%', '%25', $pathinfo );
 
 			list( $req_uri ) = explode( '?', $pathinfo );
-			$home_path       = trim( wp_parse_url( home_url(), PHP_URL_PATH ), '/' );
+			$home_path       = trim( (string) wp_parse_url( home_url(), PHP_URL_PATH ), '/' );
```
justlevine, line 161: We know `home_url()` will give us a valid value, so we can coerce it instead of worrying about `string|false|null` like we need to [above](https://github.com/wp-graphql/wp-graphql/pull/2538/files?diff=split&w=1#r979216937)

src/Type/Enum/TimezoneEnum.php
```diff
@@ -41,30 +43,30 @@ public static function register_type() {
 			load_textdomain( 'continents-cities', $mofile );
 			$mo_loaded = true;
 		}
+	
 		$zonen = [];
 		foreach ( timezone_identifiers_list() as $zone ) {
 			$zone = explode( '/', $zone );
+
 			if ( ! in_array( $zone[0], $continents, true ) ) {
```
justlevine, line 51: This check means `Etc` is never even a possibility.

src/Type/ObjectType/TermObject.php
```diff
@@ -56,7 +56,8 @@ public static function register_taxonomy_object_type( WP_Taxonomy $tax_object )
 							$url = $term->link;
 							if ( ! empty( $url ) ) {
 								$parsed = wp_parse_url( $url );
```
justlevine, line 58: TBH we should probably stop using `wp_parse_url()` to get a `uri` throughout the codebase. Much better to use [`wp_make_relative_uri()`](https://developer.wordpress.org/reference/functions/wp_make_link_relative/) and strip the possible localhost port. Good candidate for a Utils function.

jasonbahl (approved):

jasonbahl:
I merged this into the v1.12.0 release branch and resolved conflicts.

Release will go out Monday.

PR #2541, opened 2022-09-24
feat: Obfuscate `SendPasswordResetEmail` response.

### LINKED ISSUE #2100: resetUserPassword returns null sometimes

justlevine:
This PR refactors `SendPasswordResetEmail` to both provide a valid indicator of the mutation's `success`, as well as obfuscate any errors that could be used to enumerate valid users on the site, by logging them with `graphql_debug()` instead of throwing an exception.

### Details
- chore: Refactor `SendPasswordResetEmail` registration config to use static methods.
- feat: add the `success` field to `SendPasswordResetEmailPayload`.
- feat: deprecate the `user` field from `SendPasswordResetEmailPayload`.
- feat: log `SendPasswordResetEmail` mutation errors with `graphql_debug()` to prevent user enumeration.

closes #2100

Any other comments?
-------------------
- `SendPasswordResetEmailPayload.user` currently returns `public` user data if the user name/email is valid. I think even before the field is removed entirely we might want to at least limit this to logged-in users with the `list_users` cap to prevent email reverse lookups. 

Where has this been tested?
---------------------------

**WordPress Version:** 6.0.2

### DIFF
--- a/src/Mutation/SendPasswordResetEmail.php
+++ b/src/Mutation/SendPasswordResetEmail.php
@@ -33,7 +33,7 @@ class SendPasswordResetEmail {
 	 *
 	 * @return array
 	 */
-	public static function get_input_fields() {
+	public static function get_input_fields() : array {
 		return [
 			'username' => [
 				'type'        => [
@@ -49,7 +49,7 @@ class SendPasswordResetEmail {
 	 *
 	 * @return array
 	 */
-	public static function get_output_fields() {
+	public static function get_output_fields() : array {
 		return [
 			'user'    => [
 				'type'              => 'User',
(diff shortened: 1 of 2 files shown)


justlevine:
Note: was failing on `PHP < 8` because of a trailing comma in a function param list. I'm assuming this wasnt picked up by PHPstan since its run on PHP 8.x. We might need to lower it to PHP 7.3 ( after #2538 is merged) to ensure we catch future issues like this one.

jasonbahl (approved):

PR #2546, opened 2022-09-26
feat: add `register_graphql_edge_fields()` and `register_graphql_connection_where_args()` access functions

### LINKED ISSUE #1214: Add register_graphql_edge_field() method
Currently, in order to register a field to an Edge you have to know the full Edge name, like `RootQueryToPostConnectionEdge` and that's cumbersome.

It would be nice to have something like: `register_graphql_edge_field( $fromType, $toType, $fieldName, $config );`

I think it's easier to know the from/to type than know the fully qualified edge name.

### LINKED ISSUE #1522: Add tests for register_graphql_fields function
There is currently no explicit test for the `register_graphql_fields` function. The `register_graphql_field` function does have tests, but it would be a good idea to explicitly test `register_graphql_fields` as well.

justlevine:
This PR adds the `register_graphql_edge_field()`, `register_graphql_edge_fields()`, `register_graphql_connection_where_arg()` and `register_graphql_connection_where_args()` access functions, making it easier to add fields to Connection `edges` and `where` args without needing to know the generated GraphQL type names.

It also cleans up some extraneous conditional logic in `TypeRegistry::get_fields()`, and backfills a WPUnit test for `register_graphql_fields()`

Closes #1522, closes #1214

Any other comments?
-------------------
- While #1214 doesnt call for a `register_graphql_connection_input()` access function, the logic to justify it is the same as `register_graphql_edge_field()`.
- ~Re naming: I chose to mirror[`WPConnectionType::register_connection_input()`](https://github.com/wp-graphql/wp-graphql/blob/6e8bb30892d877d61961d1eec9783d2e7ddac333/src/Type/WPConnectionType.php#L463), instead of something like `register_graphql_connection_where_arg()` for consistency, but I'm open to calling it whatever.~ 
- This PR is based off of #2545, which should be merged first

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (Wsl2 + devilbox + PHP 8.0.19)

**WordPress Version:** 6.0.2

### DIFF
--- a/access-functions.php
+++ b/access-functions.php
@@ -361,7 +361,7 @@ function register_graphql_edge_fields( string $from_type, string $to_type, array
  *
  * @since @todo
  */
-function register_graphql_connection_input( string $from_type, string $to_type, string $field_name, array $config ) : void {
+function register_graphql_connection_where_arg( string $from_type, string $to_type, string $field_name, array $config ) : void {
 	$connection_name = ucfirst( $from_type ) . 'To' . ucfirst( $to_type ) . 'ConnectionWhereArgs';
 
 	add_action(
@@ -382,7 +382,7 @@ function register_graphql_connection_input( string $from_type, string $to_type, 
  *
  * @since @todo
  */
-function register_graphql_connection_inputs( string $from_type, string $to_type, array $fields ) : void {
+function register_graphql_connection_where_args( string $from_type, string $to_type, array $fields ) : void {
 	$connection_name = ucfirst( $from_type ) . 'To' . ucfirst( $to_type ) . 'ConnectionWhereArgs';
 
 	add_action(
--- a/access-functions.php
+++ b/access-functions.php
@@ -306,6 +306,83 @@ function register_graphql_fields( string $type_name, array $fields ) {
 	);
 }
 
+/**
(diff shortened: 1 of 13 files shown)


jasonbahl:
> Re naming: I chose to mirror[WPConnectionType::register_connection_input()](https://github.com/wp-graphql/wp-graphql/blob/6e8bb30892d877d61961d1eec9783d2e7ddac333/src/Type/WPConnectionType.php#L463), instead of something like register_graphql_connection_where_arg() for consistency, but I'm open to calling it whatever.

Ok, so I think I would actually prefer `register_graphql_connection_where_args()` vs `register_graphql_connection_input()`

The reason it's `input` on mutations is that the actual field is called `input`. 

For Connections, it's called `WhereArgs`, so the registration should follow. 

In the future I intend to deprecate `whereArgs` in favor of 2 new (better structured) root fields: `filter` and `sort`, and at that time we could introduce `register_graphql_connection_filter()` and `register_graphql_connection_sort()` or something to that tune.

jasonbahl (changes requested):
see: https://github.com/wp-graphql/wp-graphql/pull/2546#issuecomment-1312268408

jasonbahl (approved):

PR #2555, opened 2022-09-29
fix: #2554 reduce header output add debug output

### LINKED ISSUE #2554: Max Header Size, still a limitation even after 1.11.2
### Description

1.11.0 introduced the QueryAnalyzer functionality which outputs `X-GraphQL-Keys` headers including information such as list types, query id, and node IDs that were resolved in the query. 

This information can be used by caching clients, loggers, etc to help understand the data that is being resolved. Caching clients can use it to "tag" the cached documents and then purge the documents based on those tags when future events occur. 

A bug was reported stating that the max header length was causing problems building in Gatsby and Next. 

This is because the Max Header Length allowed by Node is 8192 bytes. 

Foolishly, I thought that meant _each_ header, and thus chunked the headers into multiple headers below that limit. 

v1.11.2 added these chunked headers. 

Yet, here we are still experiencing `Parse Error: Header overflow` errors. 

Upon more investigation, I discovered it's not a 8,192 limit PER header, it's for _all_ headers combined. 😱 

see: https://nodejs.org/en/blog/vulnerability/november-2018-security-releases/#denial-of-service-with-large-http-headers-cve-2018-12121

> The total size of HTTP headers received by Node.js now must not exceed 8192 bytes.

Instead of chunking the headers, we should truncate the keys at a conservative limit (4,000? to allow room for other headers as well), expose another header that shows how many keys were truncated, and (optionally) expose the truncated keys in the "extensions" payload of the body. 

### Steps to reproduce

Add this filter to make sure the keys are a massive list of keys: 

```php

add_filter( 'graphql_query_analyzer_get_runtime_nodes', function ( $nodes ) {

	$keys = [];
	for ( $i = 1; $i <= 2000; $i++ ) {
		$keys[] = Relay::toGlobalId( 'post', $i );
	}
	return array_merge( $keys, $nodes );

});
```

Try and build a Next site (this is the one I used: https://github.com/jasonbahl/headless-wp-template-hierarchy). 

See the `Parse Error: Header overflow` errors ensue. 

### Additional context

_No response_

### WPGraphQL Version

1.11.2

### WordPress Version

6.0.2

### PHP Version

8.0.2

_No response_

- [X] Yes

- [ ] Yes

jasonbahl:
This PR changes the behavior of the QueryAnalyzer. 

Instead of chunking the X-GraphQL-Keys into multiple headers, this truncates to a default value of 4000 characters, then takes any additional keys and places them in "$query_analyzer->skipped_keys". 

The "X-GraphQL-Skipped-Keys-Count" header exposes the number of keys that were skipped as a header. 

#### If keys are within the limits:

CleanShot 2022-09-29 at 15 10 14

#### If keys are over the limit:

CleanShot 2022-09-29 at 15 32 42

Additionally, the Skipped Keys are output in the "extensions" payload if GraphQL Debug mode is enabled, or the filter `graphql_should_show_query_analyzer_in_extensions` returns `true`.

CleanShot 2022-09-29 at 14 57 38

closes #2554 

Any other comments?
-------------------
This addresses an issue with NODE clients failing with `Parse Error: Header overflow`.

### DIFF
--- a/src/Utils/QueryAnalyzer.php
+++ b/src/Utils/QueryAnalyzer.php
@@ -105,6 +105,12 @@ class QueryAnalyzer {
 	public function init(): void {
 
 		// allow query analyzer functionality to be disabled
+		/**
+		 * Filters whether to analyze queries or not
+		 *
+		 * @param bool $should_analyze_queries Whether to analyze queries or not. Default true
+		 * @param QueryAnalyzer $query_analyzer The QueryAnalyzer instance
+		 */
 		$should_analyze_queries = apply_filters( 'graphql_should_analyze_queries', true, $this );
 
 		// If query analyzer is disabled, bail
@@ -112,18 +118,23 @@ class QueryAnalyzer {
 			return;
 		}
 
-		// Filter the header key limit. Default is 8000 (8k) to play nice
-		// with common clients such as:
-		//
-		// Next: https://vercel.com/docs/concepts/functions/edge-functions/limitations#limits-on-requests
-		// Varnish: https://varnish-cache.org/docs/4.1/reference/varnishd.html?highlight=storage%20types#http-req-hdr-len,
-		// Node: https://nodejs.org/api/cli.html#--max-http-header-sizesize
-		// Fastly: https://docs.fastly.com/en/guides/resource-limits#request-and-response-limits
-		//
-		// Next, Varnish, Node, Fastly and others. (some support 16k).
-		// 8k will be the default, but filter away.
-		//
-		// and others.
+
+		/**
+		 * Filter the header key limit. Default is 8000 (8k) to play nice
+		 * with common clients such as:
+		 *
+		 * Next: https://vercel.com/docs/concepts/functions/edge-functions/limitations#limits-on-requests
+		 * Varnish: https://varnish-cache.org/docs/4.1/reference/varnishd.html?highlight=storage%20types#http-req-hdr-len,
+		 * Node: https://nodejs.org/api/cli.html#--max-http-header-sizesize
+		 * Fastly: https://docs.fastly.com/en/guides/resource-limits#request-and-response-limits
+		 *
+		 * Next, Varnish, Node, Fastly and others. (some support 16k).
+		 * 8k will be the default, but filter away.
+		 *
+		 * and others.
+		 *
+		 * @param int $header_length_limit The max limit headers should be. Longer than this, they're chunked into multiple headers.
+		 */
 		$this->header_length_limit = apply_filters( 'graphql_query_analyzer_header_length_limit', 8192 );
 
 		// track keys related to the query
@@ -295,6 +306,10 @@ class QueryAnalyzer {
 		$map = array_values( array_unique( array_filter( $type_map ) ) );
 
 		// @phpcs:ignore
+
+		/**
+		 * Filter teh list types
+		 */
 		return apply_filters( 'graphql_cache_collection_get_list_types', $map, $schema, $query, $type_info );
 	}
 
--- a/src/Utils/QueryAnalyzer.php
+++ b/src/Utils/QueryAnalyzer.php
@@ -654,12 +654,12 @@ class QueryAnalyzer {
 	 * @param mixed       $response
 	 * @param WPSchema    $schema         The WPGraphQL Schema
 	 * @param string|null $operation_name The operation name being executed
-	 * @param string      $request        The GraphQL Request being made
+	 * @param string|null $request        The GraphQL Request being made
 	 * @param array|null  $variables      The variables sent with the request
 	 *
 	 * @return array|object|null
 	 */
-	public function show_query_analyzer_in_extensions( $response, WPSchema $schema, ?string $operation_name, string $request, ?array $variables ) {
+	public function show_query_analyzer_in_extensions( $response, WPSchema $schema, ?string $operation_name, ?string $request, ?array $variables ) {
 
 		$should = \WPGraphQL::debug();
 
--- a/src/Utils/QueryAnalyzer.php
+++ b/src/Utils/QueryAnalyzer.php
@@ -668,7 +668,7 @@ class QueryAnalyzer {
 		 * @param mixed       $response       The response of the WPGraphQL Request being executed
 		 * @param WPSchema    $schema         The WPGraphQL Schema
 		 * @param string|null $operation_name The operation name being executed
-		 * @param string      $request        The GraphQL Request being made
+		 * @param string|null      $request        The GraphQL Request being made
 		 * @param array|null  $variables      The variables sent with the request
 		 */
 		$should_show_query_analyzer_in_extensions = apply_filters( 'graphql_should_show_query_analyzer_in_extensions', $should, $response, $schema, $operation_name, $request, $variables );
--- a/src/Utils/QueryAnalyzer.php
+++ b/src/Utils/QueryAnalyzer.php
@@ -88,7 +88,7 @@ class QueryAnalyzer {
 	protected $header_length_limit;
 
 	/**
-	 * @var string The keys that were skipped from being returned in the X-GraphQL-Keys header
+	 * @var string The keys that were skipped from being returned in the X-GraphQL-Keys header.
 	 */
 	protected $skipped_keys = '';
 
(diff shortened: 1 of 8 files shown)


src/Utils/QueryAnalyzer.php
```diff
@@ -483,46 +512,42 @@ public function track_nodes( $model ) {
...
+
+		$keys        = [];
+		$return_keys = '';
 
 		if ( $this->get_query_id() ) {
 			$keys[] = $this->get_query_id();
 		}
 
 		if ( ! empty( $this->get_root_operation() ) ) {
-			$headers['X-GraphQL-Operation-Type'] = $this->get_root_operation();
```
chriszarate, line 503: no longer setting this header. intentional?

src/Utils/QueryAnalyzer.php
```diff
@@ -483,46 +512,42 @@ public function track_nodes( $model ) {
...
-			$keys[]                              = 'graphql:' . $this->get_root_operation();
+			$keys[] = 'graphql:' . $this->get_root_operation();
 		}
 
 		if ( ! empty( $this->get_list_types() ) && is_array( $this->get_list_types() ) ) {
-			$headers['X-GraphQL-List-Types'] = implode( ' ', array_unique( array_values( $this->get_list_types() ) ) );
```
chriszarate, line 508: no longer setting this header. intentional?

src/Utils/QueryAnalyzer.php
```diff
@@ -483,46 +512,42 @@ public function track_nodes( $model ) {
...
-			$keys                            = array_merge( $keys, $this->get_list_types() );
-
+			$keys = array_merge( $keys, $this->get_list_types() );
 		}
 
 		if ( ! empty( $this->get_runtime_nodes() ) && is_array( $this->get_runtime_nodes() ) ) {
-			$headers['X-GraphQL-Nodes'] = implode( ' ', array_unique( array_values( $this->get_runtime_nodes() ) ) );
```
chriszarate, line 514: same here

src/Utils/QueryAnalyzer.php
```diff
@@ -531,16 +556,95 @@ public function get_headers( array $headers = [] ): array {
...
+
+		$keys = $this->get_graphql_keys();
+
+		if ( ! empty( $keys ) ) {
+
+			$headers['X-GraphQL-Query-ID']  = $this->query_id ?: null;
+			$headers['X-GraphQL-Keys']      = $keys['keys'] ?: null;
+			$headers['X-GraphQL-Keys-Size'] = strlen( $keys['keys'] ) ?: 0;
+			if ( 0 < absint( $keys['skippedKeysCount'] ) ) {
+				$headers['X-GraphQL-Skipped-Keys-Count'] = $keys['skippedKeysCount'] ?: 0;
```
chriszarate, line 597: I might quibble with using `Size` for the total and `Count` for the skip count. 

Maybe use count for both?
`X-GraphQL-Keys-Count`
`X-GraphQL-Skipped-Keys-Count`

src/Utils/QueryAnalyzer.php
```diff
@@ -80,6 +81,16 @@ class QueryAnalyzer {
 	 */
 	protected $header_length_limit;
 
+	/**
+	 * @var string The keys that were skipped from being returned in the X-GraphQL-Keys header
+	 */
+	protected $skipped_keys = '';
+
+	/**
+	 * @var array
```
justlevine, line 90:
```suggestion
	 * @var array The GraphQL keys to return in the X-GraphQL-Keys header.
```

src/Utils/QueryAnalyzer.php
```diff
@@ -80,6 +81,16 @@ class QueryAnalyzer {
 	 */
 	protected $header_length_limit;
 
+	/**
+	 * @var string The keys that were skipped from being returned in the X-GraphQL-Keys header
```
justlevine, line 85:
```suggestion
	 * @var string The keys that were skipped from being returned in the X-GraphQL-Keys header.
```

src/Utils/QueryAnalyzer.php
```diff
@@ -105,33 +116,52 @@ public function get_request(): Request {
 	public function init(): void {
 
 		// allow query analyzer functionality to be disabled
```
justlevine, line 118:
```suggestion
```

src/Utils/QueryAnalyzer.php
```diff
@@ -531,16 +556,95 @@ public function get_headers( array $headers = [] ): array {
...
+	 *
+	 * @param mixed    $response
+	 * @param WPSchema $schema         The WPGraphQL Schema
+	 * @param string   $operation_name The operation name being executed
+	 * @param string   $request        The GraphQL Request being made
+	 * @param array    $variables      The variables sent with the request
+	 *
+	 * @return array|mixed
+	 */
+	public function show_query_analyzer_in_extensions( $response, $schema, $operation_name, $request, $variables ) {
```
justlevine, line 614: We should add the param types to the method siguature.

src/Utils/QueryAnalyzer.php
```diff
@@ -531,16 +556,95 @@ public function get_headers( array $headers = [] ): array {
...
+		// If the query analyzer output is disabled,
+		// don't show the output in the response
+		if ( false === $should_show_query_analyzer_in_extensions ) {
+			return $response;
+		}
+
+		$keys = $this->get_graphql_keys();
+
+		if ( ! empty( $response ) ) {
+			if ( is_array( $response ) ) {
```
justlevine, line 638: Been looking at [the `graphql_request_results` filter](https://github.com/jasonbahl/wp-graphql/blob/522dcbcbfc4d540912972ebd87640c65b9c594c5/src/Request.php#L563). I'm not entire clear under which circumstances a `$request` is an `object` and not an `array`, but we should probably update the docblocks there that currently only allow for an `array` type.

(doesnt have to happen in _this_ pr, but we should at least update the `@return` here to `array|object|null`).

justlevine (commented):
Left some suggestions re inline docs .

I'm a bit confused at the overall `diff`, The PR is based off `develop` and being merged into `develop`, but all the `0.11.2` release bumps are shown as changes here too. Probably nothing that actually needs worrying over 🤷‍♂️.

src/Utils/QueryAnalyzer.php
jasonbahl: Ya, I was trying to reduce the headers. This one is probably not a big deal to stay, but the `X-GraphQL-Nodes` is the main culprit of the `X-GraphQL-Keys` header being too long. Def don't want it exposed 2x.

src/Utils/QueryAnalyzer.php
jasonbahl: Ya, similar to `X-GraphQL-Nodes`, this data is already grouped in the `X-GraphQL-Keys` and with the limits, I don't want to output redundant data.

src/Utils/QueryAnalyzer.php
jasonbahl: This is the one that really seems to build up the most and is already output in the `X-GraphQL-Keys` header, so removed this to reduce the overall header size. It was output separately for convenience over necessity.

src/Utils/QueryAnalyzer.php
chriszarate: If it's only useful for debugging and we're worried about header size, maybe move this kind of info to the response body `extensions`.

I think we should limit response headers to things that _need_ to be present for systems interconnections, e.g., Varnish, etc.

jasonbahl:
@chriszarate I'm going to come back to this on Monday. 

After some more discussion, I think it's important to at least add keys that "summarize" what was skipped, allowing purge calls to still provide correctness. 

Here's what I'm thinking: 

- analyze the skipped keys
- determine what types of nodes were skipped (i.e. posts, users, terms, etc)
- add `skipped:post skipped:user`, etc to the `X-GraphQL-Keys` header


That way when actions occur that purge caches for individual nodes, we can purge by the node's `id` and `skipped:type`

In an ideal world, nothing is cached and tagged with `skipped:type` so calling purge on that should do nothing. 

But, when folks have queried 1500 nodes and we must skip some keys, we can at least identify broadly the type(s) being skipped. 

Then an action like updating a post would purge any caches containing that post, but also any caches tagged with `skipped:post`

Issue #1733, opened 2021-02-23
Deregister core connections and mutations


stubbies:
It would be great if we could deregister some core connections, in our case we are using only a small set of queries for posts, custom posts, menus. When generating types on the js-client we get a lot of unused types like enum `TimezoneEnum`.

CodeProKid:
Hey @stubbies you can in fact deregister the fields & connections that come with the plugin by default. Check out the `deregister_graphql_field` function: https://www.wpgraphql.com/functions/deregister_graphql_field/.

ramyareye:
I think I need same thing, and it's not deregister field but deregister model(object)!
I mean as @stubbies said remove other models from hole schema. any idea? something like `deregister_graphql_model` or `deregister_graphql_object`

justlevine:
While reviewing #874 , I couldn't find a deregister any GraphQL _types_ ( `deregister_graphql_field()` only unhooks the field from the parent type ), a prerequisite for any type-based deregistration (e.g objects or mutations) .

This isn't an easy feat, because in order to _fully_ remove a type from the schema you also need to:
1. Remove any fields that resolve to that type.
2. Remove any connections that resolve to that type.
3. Prevent any interfaces/unions from resolving to that that type.
4. Clean up any types that no longer have any fields or interfaces/unions that no longer have any types to resolve to (repeating 1-4 for each).

Before dealing with the recursive logic, here's what I gleaned so far:

### Deregistering mutations
This is more of a disclaimer. `graphql_deregister_field( 'RootMutation', $mutation )` seems to deregister `${mutation}Input` and `${mutationPayload}` but I cant figure out how/why. Understanding this might provide a alternatives/solutions to what follows:

### Deregistering the type itself
At the top of `TypeRegistry->register_type()`, we can bail early if `graphql_should_register_type` filter is false.
```php
public function register_type( string $type_name, $config ) {
  /**
   * Filter whether the GraphQL type should be registered to the schema.
   *
   * @param bool $should_register Whether the GraphQL type should be registered to the schema.
   * @param string $type_name The name of the the GraphQL type to register.
   * @param mixed|array|Type $config The config for the type.
   */
  if ( ! apply_filters( 'graphql_should_register_type', true, $type_name, $config ) ) {
    return;
  }
  // rest of function
}
```

[...]

Issue #874, opened 2019-06-06
add deregister_graphql_mutation API


mhinton:
When using deregister_graphql_field the field name argument is wrong. The field name arg has to be lowercase. 

This is what should work

`deregister_graphql_field('TypeName', 'menuOrder');`

but that doesn't remove the field.

This is does remove the field 

`deregister_graphql_field('TypeName', 'menuorder');`

I am on the release/v0.2.3 branch.

betweenbrain:
Adding that this seems to still be an issue with v0.3.2

CodeProKid:
This may have been fixed, need to reproduce and close / update this issue.

jasonbahl:
Ok, I've confirmed this is an issue. When you do use `deregister_graphql_field()` to remove a mutation from `RootMutation`, it _does_ remove the field as intended. 

However, it doesn't remove the registered Types that are used by the mutation. 

We should offer a `deregister_graphql_mutation()`, which will cleanup the mutation field, but also the input type(s) and output type(s) that are registered along with it.

hsimah:
I am looking to prevent some CPT mutations from existing (they're created as side effects of other actions). Instead of adding a deregister mutation and removing the input types etc, maybe a config options `graphql_mutation`? Config options:

|Param|Outcome|
|--|--|
|`true`|Normal behaviour (default)|
|`false`|No mutations generated|
|`'create'`|Generates creation mutation only|
|`'update'`|Generates update mutation only|

desmondhoward:
any update on this? a `deregister_graphql_mutation` function is crucial as the typings for previously registered mutations stick around forever and can't be updated.

jasonbahl:
@desmondhoward It's not been a priority to work on yet. If it's causing issues, I'd love to see a PR tackling this. 

Let me know if you're interested in working on a PR?

desmondhoward:
unfortunately not willing to work on a PR since I am not well versed in PHP.
please let me know if this will be added to an upcoming milestone though.

justlevine:
@jasonbahl is there currently a way to deregister a GraphQL `type` (not a `field` on a type) that I'm missing, or do we need to add a new filter to short circuit `TypeRegistry::register_{type}()`?

jasonbahl:
> @jasonbahl is there currently a way to deregister a GraphQL type (not a field on a type) that I'm missing, or do we need to add a new filter to short circuit TypeRegistry::register_{type}()?

[...]

PR #2557, opened 2022-10-01
feat: Add `deregister_graphql_type()` access function and corresponding `graphql_excluded_types` filter. 

justlevine:
The PR adds a new access method: `deregister_graphql_type( string $type_name ) : void;` for removing a GraphQL type _and all of its references_ from the schema.

The function uses on a new `graphql_excluded_types` filter that prevents types (or fields that resolve to that type) from being registered, and then subsequently removes the type from the list of inherited interfaces by use of the existing `graphql_type_interfaces` filter.

### Changelog
- feat: added new `deregister_graphql_type()` access function.
- feat: add the `graphql_excluded_types` filter.
- feat: add the `graphql_wp_connection_type` action.
- dev: Call `WPConnectionType::register_connection()` from within the class constructor. (see below)
- fix: Only register connections from a GraphQL Object `$config` if the type itself is valid. (see below).

Prerequisite for #1733 (and to a lesser extent, #874) 

Any other comments?
-------------------
1. To make this PR work, I had to move the `TypeRegisty::register_connections_from_config()` call in `TypeRegistry::register_type()` to [_after_ the type name validation](https://github.com/AxeWP/wp-graphql/blob/6f0ecf3c283214de543d010f1185c9930621d04f/src/Registry/TypeRegistry.php#L614). I believe the previous order of operations to have been a bug (i.e. the connections shouldnt register if the type name is invalid).

2. To make this PR work, `WPConnectionType::register_connection()` is now called from within `WPConnectionType::__construct()`. I don't think this should considered be a breaking change (afaik there isnt a use case for `WPConnectionType::register_connection()` _outside_ of the core plugin. If we _are_ worried about usage, we can deprecate `register_connection()` in favor of a more generic `WPConnectionType::register()`.

3. ~Currently, the logic uses a `TypeRegistry::get_excluded_types()` as a wrapper around the `graphql_excluded_types` filter. It would be more performant if we only called the filter itself once (ideally in the constructor), but due to the way the TypeRegistry lifecycle works, I wanted to get a working PoC first.~

4. `deregister_graphql_type()` is a "last resort" function, and not intended for those who don't know what they are doing. Two important caveats to keep in mind are that:
    - Unlike `deregister_graphql_field()`, you cannot register a new type with a deregistered type name.
    - This doesn't affect the internal resolve/ resolveType logic. This is demonstrated in `AccessFunctionsTest::testDeregisterObjectType`. <br />
    
    Most user cases will likely benefit from less 'destructive' removal methods, like `deregister_graphql_field()` (can be used to replace a field with another that resolves to a different type), `$wp_object_type->show_in_graphql = false` (for preventing auto-registering of CPT/Taxonomy types, or the eventual `deregister_graphql_mutation()` and `deregister_graphql_connection()`. 

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + php 8.1.19)

**WordPress Version:** 6.1

### DIFF
--- a/access-functions.php
+++ b/access-functions.php
@@ -402,6 +402,59 @@ function register_graphql_scalar( string $type_name, array $config ) {
 }
 
 /**
+ * Given a Type Name, this removes the type from the entire schema
+ *
+ * @param string $type_name The name of the Type to remove.
+ *
+ * @return void
+ */
+function deregister_graphql_type( string $type_name ) : void {
+	// Prevent the type from being registered to the scheme directly.
+	add_filter(
+		'graphql_excluded_types',
+		function ( $excluded_types ) use ( $type_name ) : array {
+			// Transform the type name, just in case a user provided graphql_single_name instead.
+			$type_name = ucfirst( $type_name );
+			// If the type isn't already excluded, add it to the array.
+			if ( ! in_array( $type_name, $excluded_types, true ) ) {
+				$excluded_types[] = $type_name;
+			}
+
+			return $excluded_types;
+		},
+		10
+	);
+
+	// Prevent the type from being inherited as an interface.
+	add_filter(
+		'graphql_type_interfaces',
+		function ( $interfaces ) use ( $type_name ) : array {
+			$key = array_search( $type_name, $interfaces, true );
+			if ( false !== $key ) {
+				unset( $interfaces[ $key ] );
+			}
+
+			return $interfaces;
+		},
+		10
+	);
+
+	// Prevent the type from being added as possible type in a union.
+	add_filter(
+		'graphql_union_possible_types',
+		function ( $possible_types ) use ( $type_name ) : array {
+			$key = array_search( $type_name, $possible_types, true );
+			if ( false !== $key ) {
+				unset( $possible_types[ $key ] );
+			}
+
+			return $possible_types;
+		},
+		10
+	);
+}
+
+/**
  * Given a Type Name and Field Name, this removes the field from the TypeRegistry
  *
  * @param string $type_name  The name of the Type to remove the field from
--- a/access-functions.php
+++ b/access-functions.php
@@ -413,8 +413,8 @@ function deregister_graphql_type( string $type_name ) : void {
 	add_filter(
 		'graphql_excluded_types',
 		function ( $excluded_types ) use ( $type_name ) : array {
-			// Transform the type name, just in case a user provided graphql_single_name instead.
-			$type_name = ucfirst( $type_name );
+			// Normalize the types to prevent case sensitivity issues.
+			$type_name = strtolower( $type_name );
 			// If the type isn't already excluded, add it to the array.
 			if ( ! in_array( $type_name, $excluded_types, true ) ) {
 				$excluded_types[] = $type_name;
@@ -429,7 +429,13 @@ function deregister_graphql_type( string $type_name ) : void {
 	add_filter(
 		'graphql_type_interfaces',
 		function ( $interfaces ) use ( $type_name ) : array {
-			$key = array_search( $type_name, $interfaces, true );
+			// Normalize the needle and haystack to prevent case sensitivity issues.
+			$key = array_search(
+				strtolower( $type_name ),
+				array_map( 'strtolower', $interfaces ),
+				true
+			);
+			// If the type is found, unset it.
 			if ( false !== $key ) {
 				unset( $interfaces[ $key ] );
 			}
--- a/access-functions.php
+++ b/access-functions.php
@@ -406,7 +406,7 @@ function register_graphql_scalar( string $type_name, array $config ) {
  *
  * @param string $type_name The name of the Type to remove.
  *
- * @return void
+ * @since @todo
  */
 function deregister_graphql_type( string $type_name ) : void {
 	// Prevent the type from being registered to the scheme directly.
--- a/access-functions.php
+++ b/access-functions.php
@@ -438,20 +438,6 @@ function deregister_graphql_type( string $type_name ) : void {
 		},
 		10
 	);
-
-	// Prevent the type from being added as possible type in a union.
-	add_filter(
-		'graphql_union_possible_types',
-		function ( $possible_types ) use ( $type_name ) : array {
-			$key = array_search( $type_name, $possible_types, true );
-			if ( false !== $key ) {
-				unset( $possible_types[ $key ] );
-			}
-
-			return $possible_types;
-		},
-		10
-	);
 }
 
 /**
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -1202,8 +1202,9 @@ class TypeRegistry {
 	public function get_excluded_types() : array {
 		/**
 		 * Filter the list of GraphQL types to exclude from the schema.
-		 *
-		 * Useful for excluding types that are registered by plugins or themes
+		 * 
+		 * Note: using this filter directly will NOT remove the type from being referenced as a possible interface or a union type.
+		 * To remove a GraphQL from the schema **entirely**, please use deregister_graphql_type();
 		 *
 		 * @param array $excluded_types The names of the GraphQL Types to exclude.
 		 */
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -1198,6 +1198,8 @@ class TypeRegistry {
 
 	/**
 	 * Get the list of GraphQL type names to exclude from the schema.
+	 *
+	 * @since @todo
 	 */
 	public function get_excluded_types() : array {
 		/**
@@ -1207,8 +1209,10 @@ class TypeRegistry {
 		 * To remove a GraphQL from the schema **entirely**, please use deregister_graphql_type();
 		 *
 		 * @param array $excluded_types The names of the GraphQL Types to exclude.
+		 *
+		 * @since @todo
 		 */
-		return apply_filters( 'graphql_excluded_types', [ 'Post' ] );
+		return apply_filters( 'graphql_excluded_types', [] );
 	}
 
 	/**
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -1199,6 +1199,8 @@ class TypeRegistry {
 	/**
 	 * Get the list of GraphQL type names to exclude from the schema.
 	 *
+	 * Type names are normalized using `strtolower()`, to avoid case sensitivity issues.
+	 *
 	 * @since @todo
 	 */
 	public function get_excluded_types() : array {
@@ -1214,7 +1216,7 @@ class TypeRegistry {
 		 */
 		$excluded_types = apply_filters( 'graphql_excluded_types', [] );
 
-		// Normalize the types to be lowercase, to avoid case-sensitive when comparing.
+		// Normalize the types to be lowercase, to avoid case-sensitivity issue when comparing.
 		return ! empty( $excluded_types ) ? array_map( 'strtolower', $excluded_types ) : [];
 	}
 
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -598,14 +598,8 @@ class TypeRegistry {
 	 * @return void
 	 */
 	public function register_type( string $type_name, $config ) {
-
-		if ( is_array( $config ) && isset( $config['connections'] ) ) {
-			$config['name'] = ucfirst( $type_name );
-			$this->register_connections_from_config( $config );
-		}
-
 		/**
-		 * If the Type Name starts with a number, prefix it with an underscore to make it valid
+		 * If the Type Name starts with a number, skip it.
 		 */
 		if ( ! is_valid_graphql_name( $type_name ) ) {
 			graphql_debug(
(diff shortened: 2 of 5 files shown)


tests/wpunit/AccessFunctionsTest.php
```diff
@@ -701,4 +701,380 @@ public function testRegisterMutationWithIsPrivatePreventsPublicMutationExecution
...
+		}
+		';
+
+		$actual = $this->graphql( compact( 'query') );
+
+		$this->assertIsValidQueryResponse( $actual );
+		$this->assertArrayHasKey( 'errors', $actual );
+		$this->assertStringContainsString( 'GraphQL Interface Type `ContentNode` returned `null', $actual['errors'][0]['debugMessage'] );
+		$this->assertNull( $actual['data']['contentNodes']['nodes'][0] );
+		$this->assertContains( 'post', array_column( $actual['data']['contentTypes']['nodes'], 'name' ) );
```
justlevine, line 855: This is an important caveat to the functionality. See note 4in the PR description comments.

tests/wpunit/AccessFunctionsTest.php
```diff
@@ -701,4 +701,380 @@ public function testRegisterMutationWithIsPrivatePreventsPublicMutationExecution
...
+
+		$actual = $this->graphql( compact( 'query' ) );
+
+		$this->assertIsValidQueryResponse( $actual );
+		$this->assertArrayNotHasKey( 'errors', $actual );
+		$this->assertCount( 1, $actual['data']['posts']['nodes'] );
+	}
+
+	public function testDeregisterUnionType() {
+		deregister_graphql_type( 'MenuItemObjectUnion' );
```
justlevine, line 1044: It would be better to test a non-deprecated Union type, but im not exactly sure what can be used.

src/Registry/TypeRegistry.php
```diff
@@ -856,14 +873,35 @@ protected function prepare_field( $field_name, $field_config, $type_name ) {
...
+				if ( empty( $arg ) ) {
+					unset( $field_config['args'][ $arg_name ] );
+					continue;
+				}
+
+				$field_config['args'][ $arg_name ] = $arg;
+			}
+			// Ensure the array still isn't empty.
+			if ( empty( $field_config['args'] ) ) {
+				unset( $field_config['args'] );
```
justlevine, line 904: There's probably a more efficient way to do this...

access-functions.php
```diff
@@ -401,6 +401,45 @@ function ( TypeRegistry $type_registry ) use ( $type_name, $config ) {
...
+ *
+ * @since @todo
+ */
+function deregister_graphql_type( string $type_name ) : void {
+	// Prevent the type from being registered to the scheme directly.
+	add_filter(
+		'graphql_excluded_types',
+		function ( $excluded_types ) use ( $type_name ) : array {
+			// Transform the type name, just in case a user provided graphql_single_name instead.
+			$type_name = ucfirst( $type_name );
```
jasonbahl, line 417: I think elsewhere in the registry we convert the strings to all lowercase to help avoid nuances of `typeName` vs `TypeName` vs `typename`. 

I'm thinking maybe this should do the same, instead of `ucfirst()`

jasonbahl: Types are added to the `type_loader` using the `format_key()` function: https://github.com/wp-graphql/wp-graphql/blob/6de28e35389891781b8107e0404fa8e0c415c844/src/Registry/TypeRegistry.php#L630

Which just maps to `strtolower()`: https://github.com/wp-graphql/wp-graphql/blob/6de28e35389891781b8107e0404fa8e0c415c844/src/Registry/TypeRegistry.php#L179-L181

src/Type/WPConnectionType.php
```diff
@@ -140,6 +139,14 @@ public function __construct( array $config, TypeRegistry $type_registry ) {
 
 		$this->validate_config( $config );
 
+		/**
+		 * Bail if one of the connection types has been excluded from the schema.
+		 */
+		$excluded_types = $type_registry->get_excluded_types();
+		if ( in_array( ucfirst( $config['fromType'] ), $excluded_types, true ) || in_array( ucfirst( $config['toType'] ), $excluded_types, true ) ) {
```
jasonbahl, line 146: instead of `ucfirst()` here, again, I think we should do `strtolower()` as it normalizes the same way the type loader does in the type registry

tests/wpunit/AccessFunctionsTest.php
```diff
@@ -701,4 +701,380 @@ public function testRegisterMutationWithIsPrivatePreventsPublicMutationExecution
...
+
+		$this->assertIsValidQueryResponse( $actual );
+		$this->assertArrayHasKey( 'errors', $actual );
+		$this->assertStringContainsString( 'GraphQL Interface Type `ContentNode` returned `null', $actual['errors'][0]['debugMessage'] );
+		$this->assertNull( $actual['data']['contentNodes']['nodes'][0] );
+		$this->assertContains( 'post', array_column( $actual['data']['contentTypes']['nodes'], 'name' ) );
+	}
+
+	public function testDeregisterEnumType() {
+		deregister_graphql_type( 'ContentTypeEnum' );
```
jasonbahl, line 859: I think the `strtolower()` normalization would help if someone called: 

```
deregister_graphql_type( 'Contenttypeenum' )
deregister_graphql_type( 'contentTypeenum' )
deregister_graphql_type( 'contenttypeEnum' )
etc
```

access-functions.php
justlevine: My understanding is lowercase patterns are used for _internal_ array keys. but we need to be concerned about what users are going to pass to [for the `graphql_excluded_types` filter](https://github.com/wp-graphql/wp-graphql/pull/2557/files#diff-fea80346fad1aa40b8c8872a4d3a8f73a343b677fe4e8410e1224fcfb502a1a5R1186).

If you want we can normalize it in TypeRegistry::get_excluded_types(), and then safely use lowercase across the PR. Your call 😎

jasonbahl: I think I'd lean toward normalizing to lowercase as I think it will help prevent any edge cases. 

Like dynamically generated names too. 

Like

```php
foreach ( $post_types as $post_type ) {
  deregister_graphql_type( 'RootQueryTo' . $post_type . 'Connection' );
 
}
```

it wouldn't de-register. But, if the strings were normalized to all lowercase, it _would_ deregister.

justlevine: They'd probably want to use the upcoming `deregister_graphql_connection()`  ;-)
```php
foreach( $post_type as $post_type ) {
  deregister_graphql_connection( 'RootQuery', $post_type, $maybe_an_options_array); 
  } 
```

More seriously (and possibly better suited for its own discussion/issue), I think _coding_ for `strtolower()` is a bad pattern.

We should be forgiving of user input of course, but all the types/fields except for CPT/Terms (aka the vast majority of them) use  `KebabCase` or `camelCase`. Abstracting them to `lowercase` doesnt just drop our ability to recreate the `snake_case` used programmatically in PHP, but also means WPGraphQL core needs to be coding defensively every single time type names _are used_ (hence all the `ucFirst()`s in this PR), instead of just on public method param types that user's might use in their own code.

I actually had turning the internal CPT/Tax names into `KebabCase` and moving as many of the transforms to the public api footprint instead of internal logic on my PR todo list ;-)

justlevine: (To clarify, I agree with the suggestion to use strtolower when comparing on top-level - going to go through the PR to hit those, just want to make sure the above delineation makes sense to you)

jasonbahl: > They'd probably want to use the upcoming deregister_graphql_connection() ;-)

lol. Ya, that was just _one_ example. 

This type of pattern exists in other things like ACF Flex Fields, likely Gutenberg Blocks when we go down that road, etc. 

For someone trying to extend the plugin, to have something like the middle word of a concatenated string _not_ do what they thought feels weird to me.

I think we want to encourage folks to use proper casing for Fields & Types when possible, but would hate for someone not adding `ucfirst()` in an example like above, cause their code to not work. It would be frustrating / difficult to debug, and probably even difficult to pinpoint when helping someone debug.

jasonbahl: Basically, in my mind, it makes sense to to the `strtolower()` at the deepest part of the codebase only. 

Like we do here: https://github.com/wp-graphql/wp-graphql/blob/6de28e35389891781b8107e0404fa8e0c415c844/src/Registry/TypeRegistry.php#L630

Only when the Type is formally being added to the registry is it normalized to all lowercase

I think this is ultimately where we should lowercase the excluded types and do what's needed to leave them out of the Type registry.

src/Registry/TypeRegistry.php
```diff
@@ -596,14 +596,15 @@ protected function register_connections_from_config( array $config ) {
...
-			$config['name'] = ucfirst( $type_name );
-			$this->register_connections_from_config( $config );
+		/**
+		 * If the type should be excluded from the schema, skip it.
+		 */
+		if ( in_array( strtolower( $type_name ), $this->get_excluded_types(), true ) ) {
+			return;
 		}
-
+
```
justlevine, line 605:
```suggestion

```

justlevine:
@jasonbahl I updated the code to use `strtolower()`. With this change I think its _imperative_ that we stash the results of `TypeRegistry::get_excluded_types()`  - see my 3rd note in the PR description. Any suggestions?

jasonbahl:
>@jasonbahl I updated the code to use strtolower(). With this change I think its imperative that we stash the results of TypeRegistry::get_excluded_types() - see my 3rd note in the PR description. Any suggestions?

@justlevine I think we can add a property to the TypeRegistry class like so: 

```php
	/**
	 * @var array
	 */
	protected $excluded_types;
``` 

Then, I we can update `get_excluded_types()` like so: 

```php
	/**
	 * Get the list of GraphQL type names to exclude from the schema.
	 *
	 * Type names are normalized using `strtolower()`, to avoid case sensitivity issues.
	 *
	 * @since @todo
	 */
	public function get_excluded_types() : array {

		if ( null === $this->excluded_types ) {

			/**
			 * Filter the list of GraphQL types to exclude from the schema.
			 *
			 * Note: using this filter directly will NOT remove the type from being referenced as a possible interface or a union type.
			 * To remove a GraphQL from the schema **entirely**, please use deregister_graphql_type();
			 *
			 * @param string[] $excluded_types The names of the GraphQL Types to exclude.
			 *
			 * @since @todo
			 */
			$this->excluded_types = apply_filters( 'graphql_excluded_types', [] );

		}

		// Normalize the types to be lowercase, to avoid case-sensitivity issue when comparing.
		return ! empty( $this->excluded_types ) ? array_map( 'strtolower', $this->excluded_types ) : [];
	}

```

Then, we can test to make sure this is doing what we hoped, with this snippet: 

```php
$counter = 0;

add_filter( 'graphql_excluded_types', function( $excluded ) use ( &$counter ) {
	$counter++;
	return $excluded;
} );

add_action( 'graphql_return_response', function() use ( &$counter ) {
	wp_send_json( [ 'count' => $counter_yo ]);
});
```

## Current Behavior

The filter is called 854 times (as you suspected):

CleanShot 2022-11-11 at 16 25 00


## With suggested changes:

CleanShot 2022-11-11 at 16 24 06

jasonbahl (changes requested):
see my suggestion here: https://github.com/wp-graphql/wp-graphql/pull/2557#issuecomment-1312262774

I was going to push up these changes, but I don't have permission to push to your branch

justlevine:
> I was going to push up these changes, but I don't have permission to push to your branch

TIL that "Allow edits from maintainers" [doesnt work on organization forks](https://github.com/orgs/community/discussions/5634) - will make sure future PR's are from my personal account 😅

src/Registry/TypeRegistry.php
```diff
@@ -1198,23 +1206,29 @@ public function list_of( $type ) {
...
-		 * 
-		 * Note: using this filter directly will NOT remove the type from being referenced as a possible interface or a union type.
-		 * To remove a GraphQL from the schema **entirely**, please use deregister_graphql_type();
-		 *
-		 * @param string[] $excluded_types The names of the GraphQL Types to exclude.
-		 *
-		 * @since @todo
-		 */
-		$excluded_types = apply_filters( 'graphql_excluded_types', [] );
+		if ( ! isset( $this->excluded_types ) ) {
```
justlevine, line 1214: `! isset()` is future compatible with typed properties. (`null ===` will throw a ` Typed property must not be accessed before initialization` error)

src/Registry/TypeRegistry.php
```diff
@@ -1198,23 +1206,29 @@ public function list_of( $type ) {
...
+			 * To remove a GraphQL from the schema **entirely**, please use deregister_graphql_type();
+			 *
+			 * @param string[] $excluded_types The names of the GraphQL Types to exclude.
+			 *
+			 * @since @todo
+			 */
+			$excluded_types = apply_filters( 'graphql_excluded_types', [] );
+
+			// Normalize the types to be lowercase, to avoid case-sensitivity issue when comparing.
+			$this->excluded_types = ! empty( $excluded_types ) ? array_map( 'strtolower', $excluded_types ) : [];
```
justlevine, line 1228: I figure we dont want to `array_map()` 854 times either.

jasonbahl (approved):

PR #2565, opened 2022-10-06
feat: add X-GraphQL-URL to headers of GraphQL Responses

### LINKED ISSUE #2564: Expose GraphQL Endpoint in headers

multi-tenant caching clients using tagged based cache keys (i.e. surrogate keys/x-key) need to be able to differentiate tags from tenant to tenant. 

See: https://github.com/varnish/varnish-modules/issues/41#issuecomment-244747800

In order to do this, we need a baseline value to identify the site. 

Since WordPress can be interacted with from multiple domains (i.e. `content.wphost.com` & `production-domain.com` ) thee http.host can't be used as the value to determine the prefix for caching and cache invalidation. 

We need a constant value that can be used to prefix cache tags for both caching and purging. 

Expose `X-GraphQL-Endpoint` header on responses for HTTP requests to GraphQL. 

This endpoint can be used as a constant value to prefix cache tags in caching clients. 

Purge events can also send the same value in payloads which can then be used to identify tagged caches to purge.

I've considered handling this in a plugin, but I believe having it centralized in the core plugin will help lead to more standardization.

### Additional Context

_No response_

jasonbahl:
This exposes a new `X-GraphQL-Endpoint` header to be returned in HTTP requests for WPGraphQL.

closes #2564

CleanShot 2022-10-06 at 10 55 39

### DIFF
--- a/access-functions.php
+++ b/access-functions.php
@@ -620,6 +620,23 @@ function get_graphql_setting( string $option_name, $default = '', $section_name 
 }
 
 /**
+ * Return the GraphQL Endpoint for the site.
+ *
+ * @return string
+ */
+function graphql_get_endpoint() {
+
+	$graphql_endpoint = trailingslashit( site_url() ) . get_graphql_setting( 'graphql_endpoint', 'graphql' );
+
+	/**
+	 * Filter the endpoint before returning
+	 *
+	 * @param string $graphql_endpoint The graphql endpoint for the site
+	 */
+	return apply_filters( 'graphql_get_endpoint', $graphql_endpoint );
+}
+
+/**
  * Polyfill for PHP versions below 7.3
  *
  * @return int|string|null
--- a/access-functions.php
+++ b/access-functions.php
@@ -620,20 +620,27 @@ function get_graphql_setting( string $option_name, $default = '', $section_name 
 }
 
 /**
+ * Get the endpoint route for the WPGraphQL API
+ *
+ * @return mixed|void
+ */
+function get_graphql_endpoint() {
+
+	$endpoint = get_graphql_setting( 'graphql_endpoint', 'graphql' );
+
+	/**
+	 * @param string $endpoint The relative endpoint that graphql can be accessed at
+	 */
+	return apply_filters( 'graphql_endpoint', $endpoint );
+}
+
+/**
  * Return the full url for the GraphQL Endpoint.
  *
  * @return string
  */
 function graphql_get_endpoint_url() {
-
-	$graphql_endpoint = site_url( Router::$route );
-
-	/**
-	 * Filter the endpoint before returning
-	 *
-	 * @param string $graphql_endpoint The graphql endpoint for the site
-	 */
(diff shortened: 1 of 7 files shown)


access-functions.php
```diff
@@ -619,6 +619,23 @@ function get_graphql_setting( string $option_name, $default = '', $section_name
...
+
+	$graphql_endpoint = trailingslashit( site_url() ) . get_graphql_setting( 'graphql_endpoint', 'graphql' );
+
+	/**
+	 * Filter the endpoint before returning
+	 *
+	 * @param string $graphql_endpoint The graphql endpoint for the site
+	 */
+	return apply_filters( 'graphql_get_endpoint', $graphql_endpoint );
+}
```
justlevine, line 637: Should this maybe be a wrapper for a method in the `Router` class (and possibly called `graphql_get_endpoint_url()` instead)?

We've got similar functionality [in `Router::init()`](https://github.com/wp-graphql/wp-graphql/blob/57302c08ab655ac0720263c42272318e493a45af/src/Router.php#L60), as well as a similarly named filter that actually alters the endpoint used by WPGraphQL instead of just what gets output on the header. 

Thinking maybe a single source of truth and/or more clarity in the method/filter names would be helpful to dx.

jasonbahl: @justlevine pushed an update.

The `Router::$route` deals with the endpoint route specifically, not the full url. I need the full url to be consistent here.

jasonbahl: Perhaps the new filter is unnecessary 🤔

justlevine: GH isnt showing me the new commit yet.

What I _blindly_ want to suggest is to instead create a new method (either in Utils or a top-level access function) something like:

```php
public static function get_graphql_endpoint() : string {
  $filtered_endpoint = apply_filters( 'graphql_endpoint', null );
  return  filtered_endpoint ? $filtered_endpoint : get_graphql_setting( 'graphql_endpoint', 'graphql' );
}
```
then use that in both [`Router::init()`](https://github.com/wp-graphql/wp-graphql/blob/0e6618a5da8282ab443c5ddebbc0d86fdd99730c/src/Router.php#L60), [`Setting::register_settings()`,](https://github.com/wp-graphql/wp-graphql/blob/0e6618a5da8282ab443c5ddebbc0d86fdd99730c/src/Admin/Settings/Settings.php#L95) _and_ our new `get_graphql_endpoint_url()` (which probably doesn't need a filter), and _then_ use that to populate our `$headers` which could either be filtered with the existing [`graphql_response_headers_to_send` filter](https://github.com/wp-graphql/wp-graphql/blob/0e6618a5da8282ab443c5ddebbc0d86fdd99730c/src/Router.php#L377), or a new filter if we need to modify headers [before they get sent to the query analyzer](https://github.com/wp-graphql/wp-graphql/blob/0e6618a5da8282ab443c5ddebbc0d86fdd99730c/src/Router.php#L353).

jasonbahl: ah, ya. makes sense.

PR #2566, opened 2022-10-06
release: v1.12.0

jasonbahl:
# Release Notes

## Upgrading

This release removes the `ContentNode` and `DatabaseIdentifier` interfaces from the `NodeWithFeaturedImage` Interface. 

This is considered a breaking change for client applications using a `...on NodeWithFeaturedImage` fragment that reference fields applied by those interfaces. If you have client applications doing this (or are unsure if you do) you can use the following filter to bring back the previous behavior:

```php
add_filter( 'graphql_wp_interface_type_config', function( $config ) {
	if ( $config['name'] === 'NodeWithFeaturedImage' ) {
		$config['interfaces'][] = 'ContentNode';
		$config['interfaces'][] = 'DatabaseIdentifier';
	}
	return $config;
}, 10, 1 );
```

## New Features

- ([#2399](https://github.com/wp-graphql/wp-graphql/pull/2399)): New Schema Customization options for register_post_type and register_taxonomy. Thanks @justlevine!
- ([#2565](https://github.com/wp-graphql/wp-graphql/pull/2565)): Expose X-GraphQL-URL header.


justlevine:
#2573 will fix the phpstan error

justlevine:
#2574 updates the `@since`/`@deprecated` tags to v1.12.0

justlevine:
My only other thought is whether we want to move the PostType/Taxonomy specific code from out of `WPGraphQL.php` and into `Registry/Utils/{PostObject|TermObject}.php`, before we expose a bunch of public methods that we'll need to deprecate before relocating if/when `WPGraphQL.php` keeps growing longer.
🤷‍♂️

Issue #2370, opened 2022-05-09
ACF Sections for Flexible content not showing in the IDE since 1.7+


nhankla-kl:
Found a bug in the updated WpGraphQL 1.7.1 and ACF
We use Flexible Fields a lot and the updated WPGraphQL does not show the flexible fields in the scheme:
Vs 1.6.10

1.6.10
![Screen Shot 2022-03-08 at 9 40 55 AM](https://user-images.githubusercontent.com/67442303/167448732-a98f74eb-5ac5-49a9-bb43-9373c83b2ba2.png)

1.7+
![image](https://user-images.githubusercontent.com/67442303/167448737-1cfe838c-3aa4-488d-968a-d74100c74fff.png)

rodrigo-arias:
Leaving a comment here, as this is still an issue.

jensmercedesbenz:
Described issue still present in WPGraphQL 1.9.1

@justlevine What does the [status: awaiting reviewer response](https://github.com/wp-graphql/wp-graphql/labels/status%3A%20awaiting%20reviewer%20response) mean/refer to?

justlevine:
@jensmercedesbenz in this context, it means someone with a deeper understanding of the codebase needs to figure what's causing this issue, so it can be addressed. 

Everyone is encouraged to do their own sleuthing to help us get to the bottom of this and ship a fix, the label just lets committees/maintainers know this could benefit from their 👀

jensmercedesbenz:
Described issue still present in WPGraphQL 1.10.0 and with latest update to WPGraphQL for Advanced Custom Fields 0.6.1

I've tried doing some poking around on my own but without any real progress. Any good ideas or pointers on a good entry point or process to take on such an endeavour @justlevine ?

justlevine:
@jensmercedesbenz I'm personally not all that familiar with `wp-graphql-acf`. In general, issues like these (i.e. missing GraphQL fields when upgrading to v1.7.0+) are due to the way types -especially those on an interface/union) get lazy-loaded. 
Depending on the specific implementation, the solution is usually to change the logic used to register the field/type, or if that fails (since it comes at a performance cost), to set the relevant config to `'eagerlyLoadType' => true`.

@jasonbahl should we move this issue to the ACF repo, so it doesn't get overlooked? (I don't have the necessary permissions or id do it myself).

rodrigo-arias:
[...]

Issue #2287, opened 2022-03-09
In GraphiQL Query Composer No Sub-Selection Available


jadamconnor:
When composing a query in GraphiQL, the composer does not recognize certain selections as having sub-selections. I'm using Advanced Custom fields with the WPGraphQL for Advanced Custom Fields plugin, and I'm referring specifically to these fields. Interestingly, that plugin's page is no longer there. Does this mean that WPGrapQL no longer supports this plugin?

In the screen below, executiveTeam should have an arrow right icon to indicate that there are sub-selections available. However, as you can see, there is a checkbox instead. If we check the box and run the query we get the error, `"Field \"executiveTeam\" of type \"[Page_Aboutpagefields_ExecutiveTeam]\" must have a sub selection."`

<img width="1164" alt="Screen Shot 2022-03-09 at 8 53 24 AM" src="https://user-images.githubusercontent.com/23176631/157469087-6c6fa685-0d01-4d90-a170-dd9cea847392.png">

Notably, only Flexible Content fields seem to have this problem. As you can see in the image above, heroImage, which is field of type image, is expandable. If I create a test field of the field type repeater, sub-selection is available.

<img width="940" alt="Screen Shot 2022-03-09 at 9 08 57 AM" src="https://user-images.githubusercontent.com/23176631/157470400-4e5267cd-abd3-4c55-9a5b-a6af9731b945.png">

Finally, I did not have this issue before updating the plugin, and I only updated the plugin to resolve a perpetual loading spinner.

jasonbahl:
@jadamconnor could you provide an export of your ACF Field config so I can try to reproduce?

I believe this is probably just a UI issue with the GraphiQL Query Composer (the checkbox UI on the left site), and not an issue with the actual Schema. I'd like to confirm this for sure though. 

I believe you should be able to find the GraphQL Types and Fields in the Schema by searching the Schema docs at the top right of GraphiQL, and by typing in the query editor to build your queries, instead of using the Query Composer checkbox UI.

jadamconnor:
Hello, thanks for the prompt reply. You are correct that the schema is available in the schema docs. Here's the config:

```
[
    {
        "key": "group_61d64f7b52ff5",
        "title": "About Page Fields",
        "fields": [
            {
                "key": "field_61d64f8244be3",
                "label": "Intro",
                "name": "intro",
                "type": "flexible_content",
                "instructions": "",
                "required": 0,
                "conditional_logic": 0,
[...]

PR #2572, opened 2022-10-07
fix: GraphiQL Query Composer update to FieldView for union type

chrisherold:
Discussed in #2287, Flexible Content and Relationship fields that are Union types don't have sub fields but rather sub types. There's existing code in the Query Composer FieldView component for Union types but that only gets executed in the context of sub fields https://github.com/wp-graphql/wp-graphql/blob/develop/packages/graphiql-query-composer/components/FieldView.js#L402. Adding that logic as an `else if` after the `if` statement for sub fields resolves the issue and correctly displays the Union type fields.

There appears to be a typo in `_addAllFieldsToSelections` where the function arg camel casing doesn't match the references in the function. Updated both cases to avoid confusion.

Also, made one small change to the packages/readme as I had to run `npm run start` as opposed `npm run dev` to get the GraphiQL app running in dev mode

#2287, #2370, wp-graphql/wp-graphql-acf#312, wp-graphql/wp-graphql-acf#319

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** MacOS Monterey 12.6

**WordPress Version:** 5.9.3


jasonbahl:
@chrisherold checking this out now. 👀 Thanks so much!

jasonbahl:
@chrisherold just pulled this and tested it with ACF and with the (deprecated) `MenuItem.connectedObject` union field, and it's working great!

Very much appreciate you taking the time to work through this. I know it will help a lot of users! 👏🏻

jasonbahl:
for the record, here's a before/after:

## Before

There's no sub-selection for `flexFieldGroup` field. There _should_ be the different Types available for the Union.

CleanShot 2022-10-07 at 14 33 23



## After

The possible types for the union are visible and selectable, as expected:

CleanShot 2022-10-07 at 14 33 01

jasonbahl:
@chrisherold this will go out with 1.12.0 on Monday! 👏🏻 

Again, thanks so much for your work on this! Many people will appreciate this fix!

chrisherold:
Awesome, glad to be able to contribute

Issue #2543, opened 2022-09-25
Idea: SemVer auto-update enforcement


justlevine:
### What problem does this address?

WordPress loves auto-updates; Headless apps that rely on a strongly-typed schema, not so much.
As we get closer to closing out the non-breaking backlog and start on a v2.0, we should consider a way to enforce SemVer on the WP backend.

### What is your proposed solution?

At a bare minimum, we should disable automatic updates for **X**.y.z releases, and place some sort of notice/confirmation dialog when trying to update the plugin manually to a major version in the background.

Ideally, we would take this a step further, and let users (either by filter or by admin setting) set a different autoupdate policy (E.g. only autoupdate x.y.**Z** versions).

As a bonus, we could even let users opt-in to a beta/next channel, as (based on the gh issue tracker), there's a lot of fundamental/breaking work we have slated for v2.

Major stretch goal would be to allow WPGraphQL extensions to provide a `tested up to` label, so we could also warn about possible incompatibilities with a new major release.

I would recommend implementing this _well before_ we start work on v2, so there can be a series of releases to ensure there's no kinks before game-day.

### Additional Context

- The [`auto_update_{type}` hook](https://developer.wordpress.org/reference/classes/wp_automatic_updater/should_update/#hooks)
- WC's [Plugin Update logic](https://github.com/woocommerce/woocommerce/tree/2343d5201af878a394b800a5664adeb82c37bae7/plugins/woocommerce/includes/admin/plugin-updates) (used to display update warnings based on versions and tested compatibility.
- WC's [Beta Channel logic](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce-beta-tester/includes/class-wc-beta-tester.php)

jasonbahl:
@Blakewilson has done some work in the realm of semver updates for WordPress plugins and might have some ideas to contribute. (see: https://github.com/blakewilson/semantic-versioning-plugin)

justlevine:
Per today's triage session: we'll put together some POCs for feedback.

blakewilson:
@jasonbahl @justlevine Last night I updated my semantic versioning plugin to be a standalone class instead that you can install via Composer. Here's more info: https://github.com/blakewilson/wp-enforce-semver#usage

justlevine:
Todo: wrangle this to an actionable list (@justlevine )

PR #2577, opened 2022-10-13
feat: Root Entry points for GuestCommenter

### LINKED ISSUE #1757: CommentAuthor should have RootQuery entry points in the Graph
The `CommentAuthor` is a Type in the graph that is used to represent authors of comments that are not users. 

When querying comments and their authors, you can query like so: 

```
{
  comments {
    nodes {
      id
      content
      author {
         node {
           __typename
           id
          ...on User {
            firstName
            lastName
          }
          ...on CommentAuthor {
             name
          }
         }
      }
    }
  }
}
```

Currently, the only access point to get to an object of the `CommentAuthor` type is to fetch it as a relationship of a comment. There's no direct Root entry points in the Graph to get a `CommentAuthor`.

But, since the `CommentAuthor` implements the `Node` Interface, it should have Root entry points into the Graph. 

I propose adding a `commentAuthors { ... }` Root entry point as well as a singular `commentAuthor( id: $id ) { ... }` entry point to the graph. 

----

Related: https://github.com/gatsbyjs/gatsby/issues/29998

moonmeister:
Makes comment authors who are not `Users` (aka `GuestCommenters`), available to be queried from the RootQuery.

`GuestCommenter` type has replaced `CommentAuthor` type in hopes of clarifying the difference between `User` and `non-User` comment authors. This also helps pave the way for `commenter(s)` entry points that can get `Users` or `GuestCommenters`.

closes #1757 

Any other comments?
-------------------
As discussed with @jasonbahl we'll need to provide a filter to revert the name change on this Type for those affected by this breaking change with the release.

I also changed how the `ID` filed got generated on this type, meaning it's prefixed with `guest_commenter` now instead of `comment_author`. This means if anyone has hard coded IDs they will change. Are we okay with this? Does this need to be reverted?

I have not written any tests, I can work on them if they're needed. 

Where has this been tested?
---------------------------
**Operating System:** Linux

**WordPress Version:** 6.0.2


justlevine:
> As discussed with @jasonbahl we'll need to provide a filter to revert the name change on this Type for those affected by this breaking change with the release.

@moonmeister @jasonbahl Im seeing breaking changes to GraphQL types, PHP classes (and assumingly files bc of PSR-4), and  dataloader keys. Are all/any of these essential just to expose commenters to the `RootQuery`?

I assume we want this feature merged before a major release, but without #2543 this will break pretty much every single site using comments (and likely a few plugins) overnight. 

Even if users could fix their site after-the-fact with a filter (a lot of what's included _cant_ be reverted with existing filters), in the past we've only relied on such a 'workaround' when the breaking changes were absolutely necessary. I worry this is a very bad precedent just to (albeit justifiably) improve the semantic naming of a GraphQL type that's been in use since 2019, unless it's otherwise essential to shipping these changes.

Thoughts?

jasonbahl:
@moonmeister @justlevine I think there is a backward compatible way to do this. 

I think we can introduce a new `GuestCommenter` type to the Schema, but leave the `CommentAuthor` type for now.

I believe the only field that resolves to `CommentAuthor` currently is the `comment.author.node` field. 

If we deprecate that field (but leave it), and introduce a new field, something like `comment.authoredBy.node` which then resolves to either `User` or `GuestCommenter` then we can do this in a back compatible way. 

Existing users querying for `comment.author` can continue to resolve to either `User` or `CommentAuthor` and we can encourage folks to start querying for `comment.authoredBy` and expecting a `User` or `GuestCommenter` type. 

Then, sometime in the future when we believe most users have likely switched to using the new `comment.authoredBy` field, we could remove the `comment.author` field and the `CommentAuthor` type, but still provide a snippet that makes it possible to bring it back for users that _really_ need it.

moonmeister:
All makes sense. The type chances are not necessary. I can change it to Jason's suggestion or just not worry about it and we can make the changes later during a major release.

justlevine:
I like @jasonbahl 's suggestion too, that way we can continue to iterate on the new type without getting held up by back-compat.

Speaking in broad terms, I suggest we use the new/future names for the new PHP classes/types, but keep the existing dataloader key. As far as the existing PHP classes, instead of renaming them entirely, I think they should extend the new classes with a deprecation warning, so users can prepare for the eventual breaking change at their leisure.

moonmeister:
Okay. SO new Types create, new interfaces and new `authoredBy` field has been  created with those new types. All old types/interfaces/fields remain. though `author` has been marked as depreciated. That isn't working correctly due to `connections` not implementing `deprecationReason` correctly...Jason is working on a quick fix for that.

justlevine:
@moonmeister would you mind fixing the phpcs / phpstan issues?

> I have not written any tests, I can work on them if they're needed.

Looks like we're going to need them, especially around the id type resolution. I'm happy to lend a hand if you want 😎

moonmeister:
@justlevine Help with Tests would be great. I'm working on that last phpcs comment warning. Needed to revisit the filters anyway.

moonmeister:
Where are we at on this? Is it waiting for tests or something else?

justlevine:
@moonmeister I just tried reviewing /rebasing this locally in order to start working on tests, but I'm not entirely sure which of these changes are intentional vs cosmetic, or what is actually is supposed to be happening since the force-push vs what was originally described in the PR description. Seems there's also still some breaking changes like the removal of `DataSource::resolve_comment_author()`. 

If you can provide some direction I'd love to give it another attempt 🤞

bin/run-docker.sh
```diff
@@ -82,7 +82,7 @@ case "$subcommand" in
                 WP_VERSION=${WP_VERSION} PHP_VERSION=${PHP_VERSION} docker compose up app
                     ;;
                 t )
-                docker-compose run --rm \
+                docker compose run --rm \
```
moonmeister, line 85: Docker changed how compose is executed. I'm not sure if I was the only one having the issue, but if y'all aren't having issues this probably doesn't need to carry over.

src/AppContext.php
```diff
@@ -106,8 +107,9 @@ public function __construct() {
 			'comment'             => new CommentLoader( $this ),
 			'enqueued_script'     => new EnqueuedScriptLoader( $this ),
 			'enqueued_stylesheet' => new EnqueuedStylesheetLoader( $this ),
-			'plugin'              => new PluginLoader( $this ),
+			'guest_commenter'     => new GuestCommenterLoader( $this ),
 			'nav_menu_item'       => new PostObjectLoader( $this ),
+			'plugin'              => new PluginLoader( $this ),
```
moonmeister, line 112: This change was cosmetic to keep with the alphabetical list.

src/Mutation/CommentDelete.php
```diff
@@ -105,8 +105,8 @@ public static function mutate_and_get_payload() {
 			} else {
 				// Get the current user id
 				$current_user_id = absint( get_current_user_id() );
-				// If the current user ID is the same as the comment author's ID, then the
-				// current user is the comment author and can delete the comment
+				// If the current user ID is the same as the commenter's ID, then the
```
moonmeister, line 108: Cosmetic. I think I was trying to keep consistent language through out the code base.

src/Type/InterfaceType/CommenterInterface.php
moonmeister, line 1: Looks cosmetic for consistent language

src/Type/ObjectType/Comment.php
```diff
@@ -26,10 +26,11 @@ public static function register_type() {
 				'model'       => CommentModel::class,
 				'interfaces'  => [ 'Node', 'DatabaseIdentifier' ],
 				'connections' => [
-					'author' => [
+					'author'     => [
```
moonmeister, line 29: cosmetic

src/Type/ObjectType/Comment.php
```diff
@@ -26,10 +26,11 @@ public static function register_type() {
...
 						'toType'               => 'Commenter',
 						'connectionInterfaces' => [ 'CommenterConnection' ],
 						'description'          => __( 'The author of the comment', 'wp-graphql' ),
+						'deprecationReason'    => __( 'Deprecated in favor of `authoredBy`', 'wp-graphql' ),
```
moonmeister, line 33: cosmetic for consistent language

src/Type/ObjectType/RootQuery.php
```diff
@@ -630,13 +630,46 @@ function ( $post ) use ( $allowed_post_types ) {
 
 						},
 					],
-					'viewer'      => [
+					'viewer'         => [
```
moonmeister, line 633: Tabs are all cosmetic

src/Type/WPConnectionType.php
moonmeister, line 1: looks cosmetic.

moonmeister (commented):
@justlevine Here are some comments. Hopefully, this is helpful. my memory ain't so good. No comments meant it seemed important. Maybe not correct...but 🤷🏻

PR #2582, opened 2022-10-14
fix(noderesolver): adding extra_query_vars in graphql_pre_resolve_uri…

yanmorinokamca:
From tag v1.8.7

Solve wp-graphql/wp-graphql #2515
Solve LEDE-740

Where has this been tested?
---------------------------
**WordPress Version:** 5.8.4
wp-graphql 1.8.7
sitepress-multingual-cms: 4.5.8 (wpml)

### DIFF
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -86,7 +86,7 @@ class NodeResolver {
 		 * @param AppContext $content The app context.
 		 * @param WP $wp WP object.
 		 */
-		$node = apply_filters( 'graphql_pre_resolve_uri', null, $uri, $this->context, $this->wp );
+		$node = apply_filters( 'graphql_pre_resolve_uri', null, $uri, $this->context, $this->wp, $extra_query_vars );
 
 		if ( ! empty( $node ) ) {
 			return $node;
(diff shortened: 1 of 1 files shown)


justlevine:
As I noted in https://github.com/wp-graphql/wp-graphql/issues/2515#issuecomment-1257076618, I still think this is the best solve for #2515, and can stand alone from implementing a new filter later in the lifecycle (via #2534).

src/Data/NodeResolver.php
```diff
@@ -86,7 +86,7 @@ public function resolve_uri( string $uri, $extra_query_vars = '' ) {
 		 * @param AppContext $content The app context.
 		 * @param WP $wp WP object.
```
justlevine, line 87:
```suggestion
		 * @param WP $wp WP object.
		 * @param mixed|array|string $extra_query_vars Any extra query vars to consider.
```

justlevine (changes requested):
Thanks for this PR @yanmorinokamca ! Looks like this is compatible with `develop` even if it was based off of v1.8.7 💪
All I see missing is the param in the filter doc-block. Would you mind accepting my suggestion?

PR #2593, opened 2022-10-25
feat: use sha256 instead of md5 for hashing the queryId

### LINKED ISSUE #2584: use sha256 instead of md5 for X-GraphQL-Query-ID

Currently we use [md5](https://github.com/wp-graphql/wp-graphql/blob/0af1a98c341ded9935ac5111402a5ad6baeb33e2/src/Utils/Utils.php#L22) to hash the query string and expose it as the X-GraphQL-Query-ID header.

In order to be more consistent with the wider community (i.e. [Apollo's Automated Persisted Queries](https://www.apollographql.com/docs/react/api/link/persisted-queries/#usage)), think we should instead use sha256 as the default for hashing the id. 

Change the hash from `md5( $query_string )` to `hash( 'sha256', $query_string )`

_No response_

### Additional Context

_No response_

jasonbahl:
This changes the hash for the queryId to use sha256 by default instead of md5. 

The algorithm is filterable to support any algorithm compatible with the php `hash()` function.

This plays nicer with the existing "standard" set by Apollo's Automated Persisted Queries functionality

closes #2584 

Any other comments?

## Before (md5)
CleanShot 2022-10-25 at 06 50 08

## After (sha256)

CleanShot 2022-10-25 at 06 51 11

### DIFF
--- a/src/Utils/Utils.php
+++ b/src/Utils/Utils.php
@@ -16,10 +16,17 @@ class Utils {
 	 */
 	public static function get_query_id( string $query ) {
 
+		/**
+		 * Filter the hash algorithm to allow different algorithms.
+		 *
+		 * @string $algorithm Default is sha256. Possible values are those that work with the PHP hash() function. See: https://www.php.net/manual/en/function.hash-algos.php
+		 */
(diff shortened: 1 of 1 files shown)


markkelnar (approved):
👌

PR #2617, opened 2022-11-22
feat: introduce common interfaces (again)

### LINKED ISSUE #1738: Common Interfaces that other Interfaces can Implement

jasonbahl:

This introduces common interfaces. This PR is a fresh take on #2022 

Interfaces introduced in this PR are: 

- [x] Connection
- [x] ConnectionEdge
- [x] Edge
- [x] OneToOneConnection
- [x] ContentNodeConnection
- [x] ContentTypeConnection
- [x] CommentConnection
- [x] CommenterConnection
- [x] ContentNodeConnection
- [x] HierarchicalNode
- [x] MenuConnection
- [x] MenuItemConnection
- [x] MenuItemLinkableConnection
- [x] Previewable
- [x] TaxonomyConnection
- [x] TermNodeConnection
- [x] UserConnection

closes #1738 

Any other comments?
-------------------
These connections add a lot of value to Schema organization, code reusability, and introspection. 

One can now use GraphiQL to find all instances of Connections in the Schema like so: 

connection-interface-implementations

One could also look for all Connections that implement a more specific Connection Interface, such as `ContentNodeConnection`. This interface is applied to any node that has a connection to any type of Content Node. So a User to Post connection is a `ContentNodeConnection` and a User to Page connection is also a ContentNodeConnection. 

content-node-connection-implementations

These shared connection interfaces allow client developers to have a better understanding of shared properties of connections and can allow for reduced code when querying similar data from the Graph. 

On the server side, developers can more easily add features to specific connections or general connection interfaces.


justlevine:
Really cool stuff. Do we want to keep the old (admittedly poorly-named) `ContentRevisionUnion` instead of `ContentRevision` for back-compat?

jasonbahl:
> Really cool stuff. Do we want to keep the old (admittedly poorly-named) ContentRevisionUnion instead of ContentRevision for back-compat?

Ya, perhaps we keep the `ContentRevisionUnion` Type, and deprecate the `revisions` connections that resolve to it, then introduce a new field with a new name that resolves as a `ContentNode` connection. . .then remove the deprecated type and field later.

Thoughts on a new name for the field? 

```
post.revisions @deprecated
post.contentRevisions?
```

jasonbahl:
There's only 2 places the `ContentRevisionUnion` was referenced: 

```
RootQuery.revisions
User.revisions
```

I'm thinking we can deprecate these and introduce: 

```
RootQuery.revisedNodes
User.revisedNodes
```
```

**Comment by justlevine** [comment]
Not sure about `revisedNodes`, since semantically it implies nodes that have been revised and not individual `revision` snapshots of the particular node... Idk naming things is hard 😵‍💫

**Comment by jasonbahl** [comment]
@justlevine ah! I remember why I had to make this change. 

With the introduction of the `Connection` and `Edge` interfaces applied to all connections, we get errors like this now: 

```
Interface field Connection.nodes expects type [Node!]! but UserToRevisionsConnection.nodes is type [ContentRevisionUnion!]!.
```

I think the best way forward is to make this "technically" breaking change. 

For most use (99%+?) cases, it won't actually be a breaking change. 

Because the previous connection returned a Union instead of an Interface, the only way to query was like so (using the `...on` syntax for every single type)

```
{
  viewer {
    revisions {
      nodes {
        __typename
        ... on Post {
          id
          uri
          isRevision
        }
        ... on Page {
          id
          uri
          isRevision
        }
      }
    }
  }
  revisions {
    nodes {
      __typename
      ... on Post {
        id
        uri
        isRevision
      }
      ... on Page {
        id
        uri
        isRevision
      }
    }
  }
}
```

Since the new connection is an Interface, the same queries will still work. 

### v1.12.2
Here's the same query on WPGraphQL v1.12.2 (ContentRevisionUnion)

CleanShot 2022-11-23 at 14 23 46


### Current PR
And here's the same query with the changes from this PR applied

CleanShot 2022-11-23 at 14 22 23@2x

----

The only way it would cause a break would be if someone were querying like so: 

```graphql
{
  viewer {
    revisions {
      ... on UserToContentRevisionUnionConnection {
        nodes {
          __typename
          ... on Post {
            id
            uri
            isRevision
          }
          ... on Page {
            id
            uri
            isRevision
          }
        }
      }
    }
  }
  revisions {
    ... on RootQueryToContentRevisionUnionConnection {
      nodes {
        __typename
        ... on Post {
          id
          uri
          isRevision
        }
        ... on Page {
          id
          uri
          isRevision
        }
      }
    }
  }
}
```

Which seems highly unlikely, but something we can call out in the release notes.

jasonbahl:
fwiw, working on this makes me want to re-visit https://github.com/wp-graphql/wp-graphql/pull/2355. I still don't quite now how to resolve that "problem" but it keeps haunting me. 

I have several one-off utility post types that I don't want to be lumped together as "content" but they keep showing up in my `contentNodes { ... }` queries. lol

jasonbahl:
@justlevine we could potentially add something to the connection config for exluding the `Connection` and `Edge` interfaces too 🤔 

like:

```php
register_graphql_connection( [
  'fromType' => 'RootQuery',
  'toType'               => 'ContentRevisionUnion',
  ...
  'exclude_interfaces' => [ 'Connection' ],
  'exclude_edge_interfces' => [ 'Edge' ],
] );
```

🤔

PR #2627, opened 2022-11-29
fix: removed connection classes as deprecated wrappers (regression)

justlevine:
This PR restores the `WPGraphQL\Connection` classes removed in #2617 , as backwards-compatible deprecation wrappers for the new classes.

[Slack](https://wp-graphql.slack.com/archives/CCYJRDN4A/p1669679113213969)

Any other comments?
-------------------
- Since the class is static, each method needs to be deprecated (vs placing one notice in the constructor in an instance class).
- Not entirely sure if this is _worth_ doing, but it was an easy-enough PR to write up.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (WSL2 + devilbox + php 8.0.19)

**WordPress Version:** 6.1.1

### DIFF
--- a/src/Connection/Comments.php
+++ b/src/Connection/Comments.php
@@ -9,30 +9,30 @@ class Comments extends \WPGraphQL\Type\Connection\Comments {
 	/**
 	 * {@inheritDoc}
 	 *
-	 * @deprecated @todo
+	 * @deprecated 1.13.0
 	 */
 	public static function register_connections() {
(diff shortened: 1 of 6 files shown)


jasonbahl (approved):

jasonbahl:
> Not entirely sure if this is worth doing, but it was an easy-enough PR to write up.

@justlevine it's worth doing. We can remove at a later date. This is minor technical debt to prevent fatals when folks update like we just saw happen.

PR #2638, opened 2022-11-30
Bug/#2394 node by uri not working with woocommerce

### LINKED ISSUE #2394: NodeByUri will not query WooCommerce products since v1.8.1

jasonbahl:

This fixes a bug where the `NodeByUri` queries weren't properly working with the WPGraphQL for WooCommerce plugin. 

The WooCommerce plugin makes use of the `graphql_post_entities_allowed_post_types ` filter to prevent certain Types from being registered to the Schema so that it can register them differently. 

However, those Types still need to be resolved, and the use of the filter was causing those Types to not be resolvable. 

**FAILING TEST:** https://github.com/wp-graphql/wp-graphql/actions/runs/3579069808/jobs/6019868327#step:7:356
**PASSING TEST:**

closes #2394 

### Before

Null results returned when querying a product by URI

CleanShot 2022-11-29 at 16 59 48

### After

The product is returned when querying by URI

CleanShot 2022-11-29 at 17 00 21

Any other comments?
-------------------
I think we might need to explore 2 methods for `WPGraphQL::get_allowed_post_types()`

I think there should be a way to determine what post types should be auto-registered to the Schema and what post types should be resolve-able. 

The problem we're facing here is that we assume they're the same. 

In the Schema generation side, I've left the use of `WPGraphQL::get_allowed_post_types()`, but in the resolver side I moved back to calling `get_post_types( [ 'show_in_graphql' => true ]);` instead. 

In the case of WooCommerce, the type `Product` should still be allowed to resolve, even if it's prevented from automatically registering in the Schema. 

I think some of the work we've done with adding post_type/taxonomy args for `graphql_kind`, etc should also help here too, so WooCommerce wouldn't need to filter the `graphql_post_entities_allowed_post_types`, but that'll take some time to implement on the WooGraphQL side.

### DIFF
--- a/src/Data/Loader/PostObjectLoader.php
+++ b/src/Data/Loader/PostObjectLoader.php
@@ -87,7 +87,7 @@ class PostObjectLoader extends AbstractDataLoader {
 		 * to the count of the keys provided. The query must also return results
 		 * in the same order the keys were provided in.
 		 */
-		$post_types = \WPGraphQL::get_allowed_post_types();
+		$post_types = get_post_types( [ 'show_in_graphql' => true ] );
 		$post_types = array_merge( $post_types, [ 'revision', 'nav_menu_item' ] );
 		$args       = [
 			'post_type'           => $post_types,
--- a/src/Data/Loader/PostTypeLoader.php
+++ b/src/Data/Loader/PostTypeLoader.php
@@ -29,7 +29,7 @@ class PostTypeLoader extends AbstractDataLoader {
 	 * @throws Exception
 	 */
 	public function loadKeys( array $keys ) {
-		$post_types = \WPGraphQL::get_allowed_post_types( 'objects' );
+		$post_types = get_post_types( [ 'show_in_graphql' => true ], 'objects' );
 
 		$loaded = [];
 		if ( ! empty( $post_types ) && is_array( $post_types ) ) {
--- a/src/Data/Loader/UserLoader.php
+++ b/src/Data/Loader/UserLoader.php
@@ -50,9 +50,10 @@ class UserLoader extends AbstractDataLoader {
 		// Get public post types that are set to show in GraphQL
 		// as public users are determined by whether they've published
 		// content in one of these post types
-		$post_types = \WPGraphQL::get_allowed_post_types( 'names', [
+		$post_types = get_post_types( [
+			'show_in_graphql' => true,
 			'public' => true,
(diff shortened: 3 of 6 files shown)


justlevine:
> I think there should be a way to determine what post types should be auto-registered to the Schema and what post types should be resolve-able.
>
> The problem we're facing here is that we assume they're the same.

> In the Schema generation side, I've left the use of WPGraphQL::get_allowed_post_types(), but in the resolver side I moved back to calling get_post_types( [ 'show_in_graphql' => true ]); instead.

> In the case of WooCommerce, the type Product should still be allowed to resolve, even if it's prevented from automatically registering in the Schema.

I'm not sure I agree with this logic/solution.

First, WPGraphQL::get_allowed_post_types() is just a wrapper for `get_post_types( 'show_in_graphql' => true )` with some validation (including the mentioned filter). If its working normally except when the CPT is excluded via `graphql_post_entities_allowed_post_types`, that means the filter is doing its intended job.

If a CPT doesnt want to be included in the schema automatically, but still benefit from all the automatically generated types and connections, then it should need to opt into the functionality it _wants_ to use (in the case of `nodeByUri` via the `graphql_pre_resolve_uri` filter).

Specifically regarding Woocommerce - @kidunot89 should correct me if im wrong, but my understanding is that it registers WC types the way it does because there were no other options before the recent schema registration API improvements. As such I think it would make more sense to fix the issue there instead of strengthening (what is in my opinion only) an antipattern. (PS: my schedule clears up mid December and Im happy to volunteer on this then)

justlevine:
Less philosophical and more concrete cause for concern are developers using ` graphql_post_entities_allowed_post_types as intended` to exclude a CPT registered with GraphQL support. NodeByUri will now try to resolve them and at best throw an missing GraphQL type error, at worst expose data that was meant to be excluded by the schema. This could be particularly problematic with the arising non-determinate query pattern thats quickly catching on (e.g. Faust).

jasonbahl:
Ya, I think the best way forward is to close this, and address the issue in the WooGraphQL repo. 

I agree that if you opt out, it's on you to opt-back in to the functionality you need. 

We can work on getting WooGraphQL updated to play nice with the register_post_type/register_taxonomy updates.

PR #2642, opened 2022-12-01
fix(resolver): check if `name` is an array

BronsonQuick:

This is a fairly big edge case but we're using WPGraphQL with Next.js. Our application replaces a bunch of different CMS and as a result there are backlinks from other sites to legacy URLs. Most of those result in successful 404's which we can handle on the frontend however we noticed one edge case with an old URL which has the following URL pattern:

`/api/getServices?name[]=<slug>`

Essentially any request that has `?name=[]` will end up hitting this issue.

Currently this causes an error for us and we'd prefer if it just returned a `null` value for the page without an Internal server error.

No. 

Before:

before

After:

after

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:**
* Ubuntu Jammy Jellyfish

**WordPress Version:**

* 6.1.1

### DIFF
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -134,6 +134,10 @@ class NodeResolver {
 			if ( isset( $this->wp->query_vars['post_type'] ) && in_array( $this->wp->query_vars['post_type'], $allowed_post_types, true ) ) {
 				$post_type = $this->wp->query_vars['post_type'];
 			}
+			
+			if ( is_array( $this->wp->query_vars['name'] ) ) {
+				$this->wp->query_vars['name'] = implode( $this->wp->query_vars['name'] );
+			}
 
 			$post = get_page_by_path( $this->wp->query_vars['name'], 'OBJECT', $post_type ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.get_page_by_path_get_page_by_path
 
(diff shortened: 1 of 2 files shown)


jasonbahl:
@BronsonQuick would you be interested in writing a test case for this so we can ensure we don't cause a regression in the future?

BronsonQuick:
@jasonbahl Sure thing. I'll try and carve out some time on the weekend for that seeing I don't have unit tests setup for it yet.

jasonbahl:
@BronsonQuick sweet! let me know if you need any help or whatever. I think the docs for running tests are updated, but if you have trouble you should be able to follow the steps in the github testing workflow to get tests running in Docker.

BronsonQuick:
@jasonbahl Sorry for the delay on adding a test. I was busy last week. I've had a crack at adding a test for this. Let me know if that's what you were after!

justlevine:
Hey @BronsonQuick , would you mind testing if the `resolve_uri()` refactor solves this as well, or if we need to retool it to incorporate this edge case?

The test you wrote is passing, but since im not 100% your setup I want to confirm.

BronsonQuick:
Hey @justlevine, I just pulled down https://github.com/wp-graphql/wp-graphql/pull/2680 and I can confirm that I'm getting the `null` value that I am expecting that that PR. So I'm happy to close this one out seeing it's handle there. Thanks!

PR #2652, opened 2022-12-04
feat: add `deregister_graphql_connection()` and corresponding `graphql_excluded_connections` filter 

### LINKED ISSUE #1733: Deregister core connections and mutations

justlevine:
This PR adds the `deregister_graphql_mutation()` access function, and the corresponding `TypeRegistry::deregister_mutation()` method.

Similar to #2651 this approach uses the new `graphql_excluded_connections` filter to prevent the connection (and their associated WhereArgs / Edge objects) from being registered as well.)

closes #1733 (along with #2651 )

Any other comments?
-------------------
- `@since` tags need to be populated. Marked with `@todo`
- `deregister_graphql_connection()` is dumb, and only takes a fully-qualified connection name. While there's room for an access function that deregisters connections based on a `fromType` and `toType` (including one-to-many deregistrations), I'm not sure how that method signature should look to allow custom-named connections and those based on `fromFieldName` instead of `fromType`. That's additive, so this doesn't feel incomplete without it to me.

Where has this been tested?
---------------------------

**WordPress Version:** 6.1.1


jasonbahl:
Ok, so we'll need to make sure we update some docs about usage here. 

This doesn't work:

```php

	deregister_graphql_connection( 'FavoritePosts' );

	register_graphql_connection([
		'fromType' => 'RootQuery',
		'toType' => 'Post',
		'connectionTypeName' => 'FavoritePosts',
		'fromFieldName' => 'favoritePosts',
	]);

} );
```

But this does:

```php
add_action( 'graphql_register_initial_types', function() {

	deregister_graphql_connection( 'FavoritePosts' );

});

	register_graphql_connection([
		'fromType' => 'RootQuery',
		'toType' => 'Post',
		'connectionTypeName' => 'FavoritePosts',
		'fromFieldName' => 'favoritePosts',
	]);

} );
```

The `deregister_*` calls need to be called _before_ the `register_*` calls to ensure the things being `deregistered` are placed in the proper `excluded` lists that prevent registration. 

If the `deregister_*` function call happens in the same action as the register, it would be too late and the connection would still be registered.

PR #2657, opened 2022-12-06
feat: `original_args` parameter added to `graphql_post_object_connection_args` hook

kidunot89:

🚨 Please review the [guidelines for contributing](/.github/CONTRIBUTING.md) to this repository.

- [x] Make sure your PR title follows Conventional Commit standards. See: [https://www.conventionalcommits.org/en/v1.0.0/#specification](https://www.conventionalcommits.org/en/v1.0.0/#specification)
- [x] Make sure you are making a pull request against the **develop branch** (left side). Also you should start *your branch* off *our master*.

-->

Adds `original_args` parameter to hook for bypassing WP_Query argument sanitation.

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -168,14 +168,11 @@ abstract class AbstractConnectionResolver {
 		 *
 		 * @param array                      $args                The GraphQL args passed to the resolver.
 		 * @param AbstractConnectionResolver $connection_resolver Instance of the ConnectionResolver.
-		 * @param mixed                      $source              Source passed down from the resolve tree.
-		 * @param array                      $all_args            Array of arguments input in the field as part of the GraphQL query.
-		 * @param AppContext                 $context             Object containing app context that gets passed down the resolve tree.
-		 * @param ResolveInfo                $info                Info about fields passed down the resolve tree.
+		 * @param array                      $unfiltered_args     Array of arguments input in the field as part of the GraphQL query.
 		 *
 		 * @since 1.11.0
 		 */
-		$this->args = apply_filters( 'graphql_connection_args', $this->get_args(), $this, $this->source, $this->args, $this->context, $this->info );
+		$this->args = apply_filters( 'graphql_connection_args', $this->get_args(), $this, $args );
 
 		/**
 		 * Determine the query amount for the resolver.
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -167,11 +167,15 @@ abstract class AbstractConnectionResolver {
 		 * Filters the GraphQL args before they are used in get_query_args().
 		 *
 		 * @param array                      $args                The GraphQL args passed to the resolver.
-		 * @param AbstractConnectionResolver $connection_resolver Instance of the ConnectionResolver
+		 * @param AbstractConnectionResolver $connection_resolver Instance of the ConnectionResolver.
+		 * @param mixed                      $source              Source passed down from the resolve tree.
+		 * @param array                      $all_args            Array of arguments input in the field as part of the GraphQL query.
+		 * @param AppContext                 $context             Object containing app context that gets passed down the resolve tree.
+		 * @param ResolveInfo                $info                Info about fields passed down the resolve tree.
 		 *
 		 * @since 1.11.0
 		 */
-		$this->args = apply_filters( 'graphql_connection_args', $this->get_args(), $this );
+		$this->args = apply_filters( 'graphql_connection_args', $this->get_args(), $this, $this->source, $this->args, $this->context, $this->info );
 
 		/**
 		 * Determine the query amount for the resolver.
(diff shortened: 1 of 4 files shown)


src/Data/Connection/PostObjectConnectionResolver.php
```diff
@@ -601,11 +601,12 @@ public function get_args(): array {
 		 * Filters the GraphQL args before they are used in get_query_args().
 		 *
 		 * @param array                        $args                The GraphQL args passed to the resolver.
+		 * @param array                        $original_args       Unsanitized args passed to the resolver.
 		 * @param PostObjectConnectionResolver $connection_resolver Instance of the ConnectionResolver
 		 *
 		 * @since 1.11.0
 		 */
-		return apply_filters( 'graphql_post_object_connection_args', $args, $this );
+		return apply_filters( 'graphql_post_object_connection_args', $args, $this->args, $this );
```
jasonbahl, line 609: The filter should be able to access the args from the 2nd argument already, eh?

jasonbahl: Also, this would be a breaking change. We'd need to add new args after the existing args.

kidunot89: Nope because the `$args` is a protected parameter.

jasonbahl: ah, and calling `$this->get_args()` is too late because of [this](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Connection/AbstractConnectionResolver.php#L174), eh?

jasonbahl: starting to remember what we went through last week 😆

jasonbahl: ok, ya, I think we could pass `$this->args, $this->context, $this->info` like we do in this filter: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Connection/AbstractConnectionResolver.php#L445


We need to add new args after the existing args though, otherwise it breaks existing code using the filter

jasonbahl (changes requested):

src/Data/Connection/PostObjectConnectionResolver.php
kidunot89: I'll update the code then. One sec.

kidunot89: Done.

justlevine:
My only concern is ConnectionResolver filter parity. I think we should either add these new params to the filters at the end of _all_ the `get_args()` methods, or move the change to `graphql_connection_args` [in AbstractConnectionResolver](https://github.com/wp-graphql/wp-graphql/blob/790c1efe9dca7740520520afb63a1b1b0942975c/src/Data/Connection/AbstractConnectionResolver.php#L174).

Thoughts?

jasonbahl:
@justlevine I think your suggestion makes sense. 

Add the additional args here: https://github.com/wp-graphql/wp-graphql/blob/790c1efe9dca7740520520afb63a1b1b0942975c/src/Data/Connection/AbstractConnectionResolver.php#L198

Looks like we've already added these args to some of the other specific connection resolver filters: https://github.com/wp-graphql/wp-graphql/blob/790c1efe9dca7740520520afb63a1b1b0942975c/src/Data/Connection/CommentConnectionResolver.php#L147

kidunot89:
@jasonbahl Additional parameters added to each hook.

jasonbahl:
@kidunot89 I opened this PR to your branch with more updates to filters for consistency: https://github.com/kidunot89/wp-graphql/pull/4

justlevine:
@jasonbahl `$source`, `$args`, `$context`, and `$info`, all have _public_ getters. Do we really need to be adding them to all the filters? (keeping in mind that to maintain back-compat, the `$resolver` param is already passed, and earlier than these )?

The goal of this PR is to shoehorn an `$original_args` into an existing filter to allow WooGraphQL to maintain back-compat with its v1.13.x changes, in a way we'll probably want to deprecate ourselves once we're not time-pressured to deliver a solution. Imo we should limit our changes to that.

jasonbahl:
take 2: https://github.com/kidunot89/wp-graphql/pull/5

src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -191,11 +192,15 @@ public function __construct( $source, array $args, AppContext $context, ResolveI
...
-		 *                                                                 This should take in the GraphQL args and return args for use in fetching the data.
-		 * @param AbstractConnectionResolver $connection_resolver          Instance of the ConnectionResolver
+		 * @param array                      $query_args          The query args to be used with the executable query to get data.
+		 *                                                        This should take in the GraphQL args and return args for use in fetching the data.
+		 * @param AbstractConnectionResolver $connection_resolver Instance of the ConnectionResolver
+		 * @param mixed                      $source              Source passed down from the resolve tree.
+		 * @param array                      $all_args            Array of arguments input in the field as part of the GraphQL query.
+		 * @param AppContext                 $context             Object containing app context that gets passed down the resolve tree.
+		 * @param ResolveInfo                $info                Info about fields passed down the resolve tree.
```
justlevine, line 201:
```suggestion
		 * @param array                      $query_args          The query args to be used with the executable query to get data.
		 *                                                        This should take in the GraphQL args and return args for use in fetching the data.
		 * @param AbstractConnectionResolver $connection_resolver Instance of the ConnectionResolver
		 * @param array                      $unfiltered_args Array of arguments input in the field as part of the GraphQL query.
```

src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -191,11 +192,15 @@ public function __construct( $source, array $args, AppContext $context, ResolveI
...
 		 */
-		$this->query_args = apply_filters( 'graphql_connection_query_args', $this->get_query_args(), $this );
+		$this->query_args = apply_filters( 'graphql_connection_query_args', $this->get_query_args(), $this, $this->source, $this->args, $this->context, $this->info );
```
justlevine, line 203:
```suggestion
		$this->query_args = apply_filters( 'graphql_connection_query_args', $this->get_query_args(), $this, $args );
```

justlevine (changes requested):

justlevine (approved):

jasonbahl (approved):

PR #2659, opened 2022-12-07
tests: backfill `nodeByUri` query testing

justlevine:
What does this implement/fix? Explain your changes.
---------------------------------------------------
…

Does this close any currently open issues?
------------------------------------------
…

Any relevant logs, error output, GraphiQL screenshots, etc?
-------------------------------------
(If it’s long, please paste to https://ghostbin.com/ and insert the link here.)

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

Issue #2491, opened 2022-08-30
Introduce Archive Interfaces


jasonbahl:
### What problem does this address?

Currently, Archives for post types return a `ContentType` Type. 

This is difficult to use as it treats _every_ Archive the same, when there are often differences. 

For example, the Archive for the "post" post type can have a formal association with the "Posts Page" as set in "Reading Settings", and that relationship should be exposed on the Post Archive, but not on all "ContentType" types. (see: https://github.com/wp-graphql/wp-graphql/issues/2486#issuecomment-1232005508)

### What is your proposed solution?

We _might_ be able to start by converting `ContentType` into an Interface and then introducing new `PostArchive`, `HouseArchive`, `CarArchive`, etc types for each post type that has archive support. 

ex: 

```graphql
type PostArchive implements ContentType {
...
}

type CarArchive implements ContentType {
...
}
```

We probably actually want to introduce a new Interface in-between, like `ContentArchive`

So then it might be: 

```
interface ContentArchive implements ContentType {
...
}

interface PostArchive implements ContentArchive & ContentType { 
... 
}
```

Where all post types (with archive support or not) are considered a "ContentType", but only post types that have archive support would implement the ContentArchive interface.

justlevine:
Don't expect it to be a problem, and it might go withoit saying, but we should design this with the 'Latest Posts' Home Page, and CPTs that use a page as an archive endpoint (WC's `shop`, TEC's `events` etc) in mind. 

The schema for querying / routing to an archive should be the same regardless of whether the backend sets it to an explicit page ID or an autogenerated endpoint.

justlevine:
go away stalebot

jasonbahl:
Just some more ideation here: 

```graphql
type PostArchive implements Archive, ContentType {
  ...fieldsInheritedFromContentType
  # Field that exposes a connection between the "ContentType" and the associated "Page" Node (i.e. `page_for_posts` )
  archivePage: ArchiveToPageConnection 
}
```

Then a query could look something like: 

```graphql
{
  nodeByUri( "/blog" ) {
    __typename
    uri
    isPostsPage
    isFrontPage
    isArchive
    ...on ContentType {
      # fields from the ContentType type
    }
    ...on PostArchive {
      archivePage {
        node {
           __typename
           id
           title
           # someMetaFieldOnThePage
        }
      }
    }
  }
}
```

PR #2680, opened 2022-12-17
feat: Refactor NodeResolver::resolve_uri() to use WP_Query

### LINKED ISSUE #1910: Can't query `/shop/` page by contentNode with uri argument when woocommerce is enabled (no wp-graphql-woocommerce required)

### LINKED ISSUE #2178: NodeByUri not resolving properly for media items
When using the path to a media item in a `nodeByUri` query, I get a `null` response. 

This isn't happening on every environment I've tested, so it might be permalink related, but I cat replicate here: 

### Media Item Url: 
http://wpgraphql-content.flywheelstaging.com/docs/debugging/screen-shot-2020-11-22-at-7-50-12-am/

### Query for said url

returns null

Screen Shot 2021-12-14 at 11 24 30 AM

----

### Works on my machine!

This behavior works on my local setup: 

Here's an image visited directly: 

Screen Shot 2021-12-14 at 12 39 10 PM

And the same image queried by uri in GraphQL:

Screen Shot 2021-12-14 at 12 40 09 PM

----

## Theories:

I would guess that this could have to do with a permalink setting. I've not compared my permalink rules across environments yet. 

Another thing might be the model layer. It's possible that the Model Layer might be preventing the images from returning in some environments? 

Not quite sure the root of the issue and still need to understand the exact steps to reproduce since I'm getting different behavior in different environments.

### LINKED ISSUE #2190: Post Format archives don't resolve with nodeByUri query

justlevine:
This pr refactors `NodeResolver::resolve_uri()` to use `WP_Query`, instead of the current approach where we try to replicate the class logic locally.

This brings the resolution process more in line with `WP::main()`, mitigating edge cases and reducing code complexity.

In addition, it adds the following WP filters to the method:
- `graphql_resolve_uri_query_class` . Used to swap out WP_Query with a specialized class, e.g. `WC_Query` from woo.
- `graphql_resolve_uri` . Used to short-circuit `resolve_uri()` after the query has been run, but before our own attempts to resolve it to a WPGraphQL `node`, e.g. if we want a CPT to use a special dataloader.
- `graphql_post_resolve_uri`. Used to provide a WPGraphQL `node` if `resolve_uri()` is unable to do so, e.g. a plugin-specific Edge case.

fixes #2178
fixes #2190
closes #1910 
(possibly others)

Part of #2366 

Any other comments?
-------------------
- This PR includes #2659, which should be merged first. 
- Since we're no longer routing via our own conditionals, `$extra_query_vars['nodeType']` and `$extra_query_vars['taxonomy']` are now set in more places to let us validate the type we're receiving is compatible with the GraphQL type we resolve to.
- I removed `nodeByUri` tests for when pretty permalinks are disabled _for now_. The reason is that `parse_request()` (from before this PR) marks those as a 404, which we were able to ignore when we were using our own logic before calling WP_Query, but no longer. I _believe_ this is a quirk with Codeception, and not an issue with the code itself, and I am looking into workarounds to reinstate those tests.
- Before merging, we should test with popular WPGraphQL Extensions, just in case they're doing something funky to work around the existing bugs.
- While I dont expect there to be any performance hits (if anything, maybe a small perf improvement because `WP_Query` results are usually cached, and we're hitting it earlier in the lifecycle), but we should do some profiling to be 100% sure. This is beyond my skillset.
- Date archives _still_ resolve null, since without #2491 we have nowhere to resolve them to. We _could_ put a `graphql_debug()` saying theyre not currently supported, and fallback to the `ContentType`, but there isnt precedent for that in the codebase.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + php8.0.19)

**WordPress Version:** 6.1.1

### DIFF
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -77,13 +77,33 @@ class NodeResolver {
 	 * @return \WP_Term|null
 	 */
 	public function validate_term( \WP_Term $term ) {
-		if ( ! isset( $this->wp->query_vars[ $term->taxonomy ] ) ) {
-			return null;
-		}
-		
 		if ( ! $this->is_valid_node_type( 'TermNode' ) ) {
 			return null;
 		}
+
+		// Check if an array has a key that starts with a string.
+		$is_same_term = false;
+
+		$taxonomy = $term->taxonomy;
+
+		// Categories and Tags can use a shorthand in the query vars.
+		if ( 'category' === $taxonomy ) {
+			$taxonomy = 'cat';
+		} elseif ( 'post_tag' === $taxonomy ) {
+			$taxonomy = 'tag';
+		}
+
+		foreach ( array_keys( $this->wp->query_vars ) as $key ) {
+			if ( strpos( $key, $taxonomy ) === 0 ) {
+				$is_same_term = true;
+				break;
+			}
+		}
+
+		if ( ! $is_same_term ) {
+			return null;
+		}
+
 
 		return $term;
 	}
@@ -237,11 +257,6 @@ class NodeResolver {
 
 		// Resolve Terms.
 		if ( $queried_object instanceof \WP_Term ) {
-			// Bail if we're explictly requesting a different GraphQL type.
-			if ( ! $this->is_valid_node_type( 'TermNode' ) ) {
-				return null;
-			}
-
 			// Validate the term before returning it.
 			if ( ! $this->validate_term( $queried_object ) ) {
 				return null;
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -50,6 +50,10 @@ class NodeResolver {
 			return $post;
 		}
 
+		/**
+		 * Disabling the following code for now, since add_rewrite_uri() would cause a request to direct to a different valid permalink.
+		 */
+		/* phpcs:disable
 		$permalink    = get_permalink( $post );
 		$parsed_path  = $permalink ? wp_parse_url( $permalink, PHP_URL_PATH ) : null;
 		$trimmed_path = $parsed_path ? rtrim( ltrim( $parsed_path, '/' ), '/' ) : null;
@@ -57,6 +61,7 @@ class NodeResolver {
 		if ( $trimmed_path !== $uri_path ) {
 			return null;
 		}
+		phpcs:enable */
 
 		return $post;
 	}
@@ -151,6 +156,11 @@ class NodeResolver {
 				}
 
 				return ! empty( $post_type_object->name ) ? $this->context->get_loader( 'post_type' )->load_deferred( $post_type_object->name ) : null;
+			}
+
+			// Validate the post before returning it.
+			if ( ! $this->validate_post( $queried_object ) ) {
+				return null;
 			}
 
 			return ! empty( $queried_object->ID ) ? $this->context->get_loader( 'post' )->load_deferred( $queried_object->ID ) : null;
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -93,11 +93,23 @@ class NodeResolver {
 			return $node;
 		}
 
+		/**
+		 * Try to resolve the URI with WP_Query.
+		 * 
+		 * This is the way WordPress native permalinks are resolved.
+		 *
+		 * @see \WP::main()
+		 */
+
+		// Parse the URI and sets the $wp->query_vars property.
 		$uri = $this->parse_request( $uri, $extra_query_vars );
 
-		// Resolve the homepage.
-		if ( '/' === $uri ) {
+		$query = new \WP_Query( $this->wp->query_vars );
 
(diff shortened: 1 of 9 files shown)


justlevine:
Verifying fix for #1910 

1. Install and Activate Woocommece
2. Add some dummy products.
3. Query `nodeByUri` for "\shop". This returns null, since by default products are not exposed to the schema
```graphql
{
  nodeByUri(uri: "/shop") {
    __typename
    ... on ContentType {
      name
      contentNodes {
        nodes {
          __typename
          databaseId
        }
      }
    }
  }
}
```
Results:
image

4. Expose WC Products to the graphql schema:
```php
add_filter( 'register_post_type_args',
  function( array $args, string $post_type ) {
      $args['show_in_graphql'] = true;
      $args['graphql_single_name'] = 'product';
      $args['graphql_plural_name'] = 'products';
    }
    return $args;
  },
10, 2 );
```

5. Rerun the query. This will return a `ContentType`, since `/shop` is a product archive.
Result: 
image

6. Attempt to fetch the query as a `Page`. This will return `null`, since `/shop` is a product archive.
```graphql
{
  page(id: "/shop", idType: URI) {
    __typename
    databaseId
  }
}
```
Result:
image

jasonbahl (approved):

PR #2689, opened 2023-01-11
fix: prevent infinite recursion when an interface implements itself 

### LINKED ISSUE #2687: If an interface attempts to implement itself, infinite recursion can occur
### Description

When registering an Interface, if the interface itself is listed as an Interface to be applied to itself, infinite recursion can occur. 

### Steps to reproduce

Using the following code, register an Interface that implements itself as an Interface: 

```php
add_action(  'graphql_register_types', function() {

	register_graphql_interface_type( 'InterfaceA', [
		'interfaces' => [ 'Node', 'InterfaceA' ],
		'fields' => [ 'fieldA' => [ 'type' => 'String' ] ],
		'resolveType' => function() {
			return 'Post';
		}
	]);


	register_graphql_interfaces_to_types( [ 'InterfaceA' ], [ 'Post' ] );

});
```

See a fatal error: 

```php
Fatal error: Uncaught Error: Maximum function nesting level of '1000' reached, aborting!
in ... on line 164
```

### Additional context

If an interface tries to implement itself (intentionally or accidentally) we should prevent this by unsetting itself as an implemented interface. 

### WPGraphQL Version

1.13.7

### WordPress Version

6.1.1

### PHP Version

7.4.1

_No response_

- [X] Yes

- [X] Yes

jasonbahl:
- prevent the interface being registered from implementing itself
- **Failing Test**: https://github.com/wp-graphql/wp-graphql/actions/runs/3894753346/jobs/6649175397
- **Passing Test**: https://github.com/wp-graphql/wp-graphql/actions/runs/3894832627/jobs/6649360575

closes #2687 

With the following code in place:

```php
add_action( 'graphql_register_types', function () {

	register_graphql_interface_type( 'InterfaceA', [
		'interfaces' => [ 'Node', 'InterfaceA' ],
		'fields' => [ 'fieldA' => [ 'type' => 'String' ] ],
		'resolveType' => function() {
			return 'Post';
		}
	]);

	register_graphql_interfaces_to_types( [ 'InterfaceA' ], [ 'Post' ] );

});
```

## Before

When trying to open the GraphiQL IDE, we get infinite recursion leading to max nested function errors, and the Schema never loads:

CleanShot 2023-01-11 at 09 45 51

## After

Schema loads fine and we can use GraphiQL:

CleanShot 2023-01-11 at 09 48 10

Any other comments?
-------------------

We're outputting a `graphql_debug` message informing the user that the interface attempted to implement itself, but we don't throw an error, as that would break existing queries that were working.

### DIFF
--- a/src/Type/WPInterfaceTrait.php
+++ b/src/Type/WPInterfaceTrait.php
@@ -1,9 +1,7 @@
 <?php
 namespace WPGraphQL\Type;
 
-use Exception;
 use GraphQL\Type\Definition\InterfaceType;
-use WPGraphQL\Registry\TypeRegistry;
 
 /**
  * Trait WPInterfaceTrait
@@ -21,9 +19,7 @@ trait WPInterfaceTrait {
 	 *
 	 * @return array
 	 */
-	protected function get_implemented_interfaces() {
-
-		$new_interfaces = [];
+	protected function get_implemented_interfaces(): array {
 
 		if ( ! isset( $this->config['interfaces'] ) || ! is_array( $this->config['interfaces'] ) || empty( $this->config['interfaces'] ) ) {
 			$interfaces = parent::getInterfaces();
@@ -40,6 +36,19 @@ trait WPInterfaceTrait {
 		 */
 		$interfaces = apply_filters( 'graphql_type_interfaces', $interfaces, $this->config, $this );
 
+		if ( empty( $interfaces ) || ! is_array( $interfaces ) ) {
+			return $interfaces;
+		}
+
+		// check if an interface is attempting to implement itself, and if so unset it
+		$key = array_search( strtolower( $this->config['name'] ), array_map( 'strtolower', $interfaces ), true );
+		if ( false !== $key ) {
+			graphql_debug( sprintf( __( 'The "%1$s" Interface attempted to implement the "%2$s" Interface, which is not allowed', 'wp-graphql' ), $interfaces[ $key ], $interfaces[ $key ] ) );
+			unset( $interfaces[ $key ] );
+		}
+
+		$new_interfaces = [];
+
 		foreach ( $interfaces as $interface ) {
 			if ( ! is_string( $interface ) ) {
 				continue;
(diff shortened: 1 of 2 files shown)


justlevine:
Accidently left comments on the [commit](https://github.com/wp-graphql/wp-graphql/commit/ca9f50f5a4765e2f7b8ee11280e74cdc80193213#comments) instead of the PR

src/Type/WPInterfaceTrait.php
```diff
@@ -40,6 +36,19 @@ protected function get_implemented_interfaces() {
...
+
+		// check if an interface is attempting to implement itself, and if so unset it
+		$key = array_search( strtolower( $this->config['name'] ), array_map( 'strtolower', $interfaces ), true );
+		if ( false !== $key ) {
+			graphql_debug( sprintf( __( 'The "%s" Interface attempted to implement itself, which is not allowed', 'wp-graphql' ), $interfaces[ $key ] ) );
+			unset( $interfaces[ $key ] );
+		}
+
+		$new_interfaces = [];
+
```
justlevine, line 51: (Moving https://github.com/wp-graphql/wp-graphql/commit/ca9f50f5a4765e2f7b8ee11280e74cdc80193213#r95832065 here for better discoverability)

> If I'm not mistaken, `in_array()` wouldn't work here because of the case sensitivity.
> 
> For better or worse you could implement an interface like:
> 
> `$interfaces = [ 'node', 'NODE', 'Node' ];`
> 
> And `unset( $interfaces[ 'Node' ] )` wouldn't unset `NODE`.
> 
> So the array search ensures we get the key for the interface regardless of case sensitivity.

It's precisely the `[ 'node', 'Node', 'NODE' ]` situation that is why I suggested suggesting ~`in_array()`~, since `array_search()` will only return the first match. We would still normalize the values.

edit: as of writing realized we dont need `in_array()` at all, and just rely on the existing foreach loop.
In untested psuedocode:

```php
$interfaces = apply_filters( 'graphql_type_interfaces', $interfaces, $this->config, $this );
// is_empty() on an array is handled by foreach() so we dont need it, but a different type returned is a coding error.
if ( ! is_array( $interfaces ) {
  graphql_debug( 'doing it wrong.....' );
  return []; // returning $interface if its something else breaks the method signature and would still throw a PHP error.
}

$normalized_interface_name = strtolower( $this->config['name'] );

foreach ( $interfaces as $interface ) {
  if ( ! is_string( $interface ) {
    continue;
  }

  // now lets test for dupes
  if ( in_array( strtolower( $interface), $normalized_interface_key, true ) {
    graphql_debug( 'Recursive interfaces are bad or whatever' );
    continue;
  }
  // rest of loop
}
```

PR #2706, opened 2023-02-03
fix: update PHPStan and fix newly-surfaced sniffs

justlevine:
This PR updates PHPStan and its associated extensions to their latest versions, and fixes newly-surfaced sniffs.

The majority of the newly-failed sniffs were due to action callbacks trying to return a value instead of `void`.

…

Any other comments?
-------------------
I ignored bc05256314b430e5d163f5ded0c2a47f7531ee6b instead of "fixing", since I'm unclear whether this is a code or a typedef error (and if `[]` is a valid logical path ).

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + PHP 8.026)

**WordPress Version:** 6.1.1

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -677,8 +677,6 @@ class Request {
 			return $this->after_execute( $response );
 
 		}
-
-		return []; // @phpstan-ignore-line
 	}
 
 	/**
--- a/src/Request.php
+++ b/src/Request.php
@@ -678,7 +678,7 @@ class Request {
 
 		}
 
-		return [];
+		return []; // @phpstan-ignore-line
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


src/Request.php
```diff
@@ -678,7 +678,7 @@ public function execute() {
 
 		}
 
-		return [];
+		return []; // @phpstan-ignore-line
```
justlevine, line 681: `$this->data` is defined as `OperationParams`, and `WPHelper::parseRequestParams` returns `OperationParams|OperationParams[]`, but since they're only defined in phpDoc and not in actual code, I'm wary to make the change without a second set of eyes.

jasonbahl (approved):

jasonbahl (approved):

Issue #2734, opened 2023-02-24
Debug output is enabled when tests run with DEBUG=0


markkelnar:
### Description

When setting the Docker environment variable DEBUG to 0 (zero), the debug output is enabled and prints when tests are run using `composer run-test`.  The expectation is that debug output is not printed when DEBUG=0.

### Steps to reproduce

`DEBUG=0 composer run-test`

### Additional context

When `DEBUG=`, behavior acts as expected.

### WPGraphQL Version

1.13.9

### WordPress Version

6.1.1

### PHP Version

8.1

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

### Please confirm that you have disabled ALL plugins except for WPGraphQL.

- [ ] Yes
- [ ] My issue is with compatibility with a specific WordPress plugin, and I have listed all my installed plugins (and version info) above.

jasonbahl:
closed by #2735

PR #2735, opened 2023-02-24
fix: Explicitly check for DEBUG enabled value for tests

markkelnar:
When running tests, specifying any value for DEBUG env variable, enables codeception debug mode.  Allow DEBUG=0 to not enable debug mode for tests.  Actually, only enable debug when DEBUG=1 or DEBUG=true.

…[issue](https://github.com/wp-graphql/wp-graphql/issues/2734)

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/docker/testing.entrypoint.sh
+++ b/docker/testing.entrypoint.sh
@@ -7,7 +7,8 @@ run_tests() {
     if [[ -n "$COVERAGE" ]]; then
         local coverage="--coverage --coverage-xml"
     fi
-    if [[ -n "$DEBUG" ]]; then
+    if [[ "1" -eq "$DEBUG" || "true" == "$DEBUG" ]]; then
+        echo "Debug enabled"
         local debug="--debug"
     fi
 


jasonbahl:
@markkelnar this will only apply to verions of WPGraphQL going forward, eh? like any plugins that are testing on previous versions of WPGraphQL wouldn't get this change?

Would we have to do some sort of manual re-deploy of older docker versions to get them all to respect this change? 

If it's only for the Docker version going forward, I think it'll just add more confusion because it would be hard to write a matrix for this and keep things straight

jasonbahl:
@markkelnar I'm concerned with plugins that use this Docker for their tests and how they can write test matrixes. 

I don't think it's that big of a deal though. 

Right now we're not using DEBUG in the other plugin matrixes. Like smart cache is just running in DEBUG always: https://github.com/wp-graphql/wp-graphql-smart-cache/blob/main/.github/workflows/tests-wordpress.yml#L61

And in the new ACF repo we're not using DEBUG in the matrix at all (https://github.com/wp-graphql/wp-graphql-acf-2/blob/main/.github/workflows/testing-integration.yml#L21-L32 / https://github.com/wp-graphql/wp-graphql-acf-2/blob/main/.github/workflows/testing-integration.yml#L78)

So it's not like this would actually _change_ behavior like I thought it might.

I think we can proceed.

jasonbahl (approved):

PR #2737, opened 2023-02-24
feat: allow fields to be registered with underscores

### LINKED ISSUE #2736: Allow fields to be registered with underscores

When registering a Type to the schema, it's possible to add fields that have underscores and WPGraphQL allows this. 

For example: 

```php
register_graphql_object_type( 'MyTestType', [
  'eagerlyLoadType' => true,
  'fields' => [
     'field_with_underscores' => [
       'type' => 'String',
     ]
  ]
] );
```

We can introspect and see the type has a field `field_with_underscores`:

CleanShot 2023-02-24 at 16 14 58

But, if we register a field using `register_graphql_field()` WPGraphQL strips the underscores. 

for example: 

```php
register_graphql_field( 'MyTestType', 'another_field_with_underscores', [ 
    'type' => 'String' 
] );
```

If we introspect again, we'll see one field has underscores preserved, and one does not:

CleanShot 2023-02-24 at 16 17 09

I would like to be able to use the `register_graphql_field()` function to add a field to the Schema that has underscores.

I propose that we add a new (optional) argument to the `Utils::format_field_name()` function to `allow_underscores`, and if true, the underscores will not be stripped. 

Then, we can modify the `register_graphql_field` function to check the field `$config` to see if `$config['allow_field_underscores']` is set to true, and if so, we pass that to the `Utils::format_field_name()`, thus allowing the field to be added to the Schema with underscores. 

This would allow existing functionality to be the same. 

**Existing Fields**

Fields that are registered like so: 

```php
register_graphql_field( 'MyType', 'my_field_with_underscore', [ 
  'type' => 'String'
] );
``` 

would continue mapping to the schema as `myFieldWithUnderscore`

**New Fields / Opt-in Fields**

New fields, or fields that want to opt-in to the new behavior could do the following: 

```php
register_graphql_field( 'MyType', 'my_field_with_underscore_allowed', [ 
  'type' => 'String', 
  'allow_field_underscore' => true 
] );
```

And now the field would be added to the schema as `my_field_with_underscore_allowed` (instead of `myFieldWithUnderscoreAllowed`)

### Connections, Mutations, etc

This behavior could then also be expanded to `register_graphql_mutation`, `register_graphql_connection` and other access functions that call `register_graphql_field` under the hood.

For example, if I wanted to add a connection with the field name `my_favorite_foods` I cannot do that today. 

```php
register_graphql_connection([
  'fromType' => 'RootQuery',
  'toType' => 'Post',
  'fromFieldName' => 'my_favorite_foods',
  'allow_field_underscores' => true,
]);
```

gets mapped to the schema today as `myFavoriteFoods`

but we could allow: 

```php
register_graphql_connection([
  'fromType' => 'RootQuery',
  'toType' => 'Post',
  'fromFieldName' => 'my_favorite_foods',
  'allow_field_underscores' => true, // new field
]);
```

To map to the Schema as `my_favorite_foods`. 

Same with mutations, etc.

In an ideal world, both paths (register_graphql_field, and adding fields directly on objects or via `graphql_$typeName_fields` filter) would behave identically, but to do that in a more central way, we would almost certainly cause breaking changes to many users. 

For example, if we removed the stripping of underscores from the `format_field_name` function, than anyone who was benefitting from that would have a breaking change to their schema. Custom post types with `"graphql_single_name" => "my_custom_type"` that currently map to the schema as `myCustomType` would now have underscores and their schema would break. 

Same with the inverse, if we fully forced underscore stripping across the board, we'd have a regression like we saw here: https://github.com/wp-graphql/wp-graphql/issues/2670 

I believe the best path forward (until we're ready for a major breaking release) is to allow folks to opt-in as proposed.

In the future, I think we should fully remove the underscore stripping as the GraphQL Spec allows underscores and WPGraphQL was going maybe a bit too far in being opinionated about field name formatting. 

### Additional Context

I came across this issue while working on the re-architecture of WPGraphQL for Advanced Custom Fields.

jasonbahl:
This allows fields, connections and mutations to be registered with underscores and have the underscores be preserved when the field is added to the Schema. 

closes #2736 

Given the following snippet: 

```php
register_graphql_object_type( 'TestObject', [
	'eagerlyLoadType' => true,
	'fields' => [
		'test' => [ 'type' => 'String' ],
	]
]);

register_graphql_field( 'TestObject', 'my_field_with_underscores', [
	'type' => 'String',
	'description' => 'Default behavior with underscores stripped',
] );

register_graphql_field( 'TestObject', 'my_field_with_underscores', [
	'type' => 'String',
	'description' => 'Field configured to allow field underscores',
	'allow_field_underscores' => true,
] );
```

We can introspect the `TestObject` and see that both fields have been added, one without underscores (default) and one with underscores (opted-in):

CleanShot 2023-02-24 at 16 50 37

Any other comments?
-------------------
I'd _love_ to solve this in a more central way, but I don't see a good way to do it without breaking a lot of Schemas for a lot of users, something I don't think is worth the tradeoff right now.

Also, this PR was branched of the branch I was working on to prep for the v1.14.0 release, so it might have a bit of extra changes coming for the ride 🤦🏻‍♂️

### DIFF
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -704,7 +704,7 @@ class TypeRegistry {
 	 * Add an Interface Type to the registry
 	 *
 	 * @param string $type_name The name of the type to register
-	 * @param array $config he configuration of the type
+	 * @param array $config The configuration of the type
 	 *
 	 * @throws \Exception
 	 * @return void
@@ -1005,7 +1005,7 @@ class TypeRegistry {
 	 * @return void
 	 * @throws \Exception
 	 */
-	public function register_fields( string $type_name, array $fields = [], ?bool $allow_underscores_in_field_names = false ): void {
+	public function register_fields( string $type_name, array $fields = [] ): void {
 		if ( ! empty( $fields ) ) {
 			foreach ( $fields as $field_name => $config ) {
 				if ( is_string( $field_name ) && ! empty( $config ) && is_array( $config ) ) {
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -634,7 +634,7 @@ class TypeRegistry {
 	 *
 	 * @return void
 	 */
-	public function register_type( string $type_name, $config ) {
+	public function register_type( string $type_name, $config ): void {
 		/**
 		 * If the type should be excluded from the schema, skip it.
 		 */
@@ -695,7 +695,7 @@ class TypeRegistry {
 	 * @throws \Exception
 	 * @return void
 	 */
-	public function register_object_type( string $type_name, array $config ) {
+	public function register_object_type( string $type_name, array $config ): void {
 		$config['kind'] = 'object';
 		$this->register_type( $type_name, $config );
 	}
@@ -709,7 +709,7 @@ class TypeRegistry {
 	 * @throws \Exception
 	 * @return void
 	 */
-	public function register_interface_type( string $type_name, array $config ) {
+	public function register_interface_type( string $type_name, array $config ): void {
 		$config['kind'] = 'interface';
 		$this->register_type( $type_name, $config );
 	}
@@ -723,7 +723,7 @@ class TypeRegistry {
 	 * @return void
 	 * @throws \Exception
 	 */
-	public function register_enum_type( string $type_name, array $config ) {
+	public function register_enum_type( string $type_name, array $config ): void {
 		$config['kind'] = 'enum';
 		$this->register_type( $type_name, $config );
 	}
@@ -737,7 +737,7 @@ class TypeRegistry {
 	 * @return void
 	 * @throws \Exception
 	 */
-	public function register_input_type( string $type_name, array $config ) {
+	public function register_input_type( string $type_name, array $config ): void {
 		$config['kind'] = 'input';
 		$this->register_type( $type_name, $config );
 	}
@@ -752,7 +752,7 @@ class TypeRegistry {
 	 *
 	 * @throws \Exception
 	 */
-	public function register_union_type( string $type_name, array $config ) {
+	public function register_union_type( string $type_name, array $config ): void {
 		$config['kind'] = 'union';
 		$this->register_type( $type_name, $config );
 	}
@@ -836,7 +836,7 @@ class TypeRegistry {
 	 *
 	 * @return bool
 	 */
-	public function has_type( string $type_name ) {
+	public function has_type( string $type_name ): bool {
 		return isset( $this->type_loaders[ $this->format_key( $type_name ) ] );
 	}
 
@@ -845,7 +845,7 @@ class TypeRegistry {
 	 *
 	 * @return array
 	 */
-	public function get_types() {
+	public function get_types(): array {
 
 		// The full map of types is merged with eager types to support the
 		// rename_graphql_type API.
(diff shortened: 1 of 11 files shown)


src/Registry/TypeRegistry.php
```diff
@@ -1001,8 +1003,9 @@ public function setup_type_modifiers( $type ) {
 	 * @param array  $fields    Fields to register
 	 *
 	 * @return void
+	 * @throws \Exception
 	 */
-	public function register_fields( string $type_name, array $fields = [] ) {
+	public function register_fields( string $type_name, array $fields = [], ?bool $allow_underscores_in_field_names = false ): void {
```
justlevine, line 1008:
```suggestion
	public function register_fields( string $type_name, array $fields = [], bool $allow_underscores_in_field_names = false ): void {
```

I don't think we want to have to control for the possibility of `null`.

jasonbahl: correct. will update 🙏🏻

jasonbahl: ah, actually that arg needs removed. I changed the approach to happen in each field's `$config`. Removing that arg from the plural `register_fields()`

src/Utils/Utils.php
```diff
@@ -105,11 +105,12 @@ public static function prepare_date_response( string $date_gmt, $date = null ) {
...
 	 * Given a field name, formats it for GraphQL
 	 *
-	 * @param string $field_name The field name to format
+	 * @param string $field_name         The field name to format
+	 * @param ?bool   $allow_underscores  Whether the field should be formatted with underscores allowed. Default false.
 	 *
 	 * @return string
 	 */
-	public static function format_field_name( string $field_name ) {
+	public static function format_field_name( string $field_name, ?bool $allow_underscores = false ): string {
```
justlevine, line 113:
```suggestion
	public static function format_field_name( string $field_name, bool $allow_underscores = false ): string {
```

justlevine:
Just confirming (having a bit of trouble with the diff on mobile): existing custom connections/mutations will remain `camelCase`d unless the new `allow_field_underscores` is added to the `$config` array, and set to `true`?

src/Registry/TypeRegistry.php
```diff
@@ -1015,18 +1018,25 @@ public function register_fields( string $type_name, array $fields = [] ) {
...
+	public function register_field( string $type_name, string $field_name, array $config ): void {
+
+
 		add_filter(
 			'graphql_' . $type_name . '_fields',
 			function ( $fields ) use ( $type_name, $field_name, $config ) {
 
-				$field_name = Utils::format_field_name( $field_name );
+				// Whether the field should be allowed to have underscores in the field name
+				$allow_field_underscores = isset( $config['allow_field_underscores'] ) && true === $config['allow_field_underscores'];
```
justlevine, line 1037: Irony notwithstanding, isnt the convention for `$config` keys themselves to currently use `camelCase`?

justlevine: (same in WPConnectionType and WPMutationType)

jasonbahl:
> Irony notwithstanding, isnt the convention for $config keys themselves to currently use camelCase?

@justlevine ya, I'm super inconsistent with this. WordPress prefers PHP keys in arrays to be `snake_case` and I'm just so used to JS / GraphQL world being `camelCase` so we have a bad mix of both right now and probably should standardize. 

🤔

jasonbahl:
closing in favor of: https://github.com/wp-graphql/wp-graphql/pull/2745

since I accidentally branched off the 1.14.0 release branch when working on the PR 🤦🏻

PR #2741, opened 2023-03-01
fix: Remove space from plugin name

josephfusco:

This WordPress plugin's header currently has the [Plugin Name set as **WP GraphQL**](https://github.com/wp-graphql/wp-graphql/blob/develop/wp-graphql.php#L3) (w/space),
with mostly all other references reading as **WPGraphQL** (w/o space).

These changes **ONLY** update the Plugin Name within the plugin's [header fields](https://developer.wordpress.org/plugins/plugin-basics/header-requirements/#header-fields) (+  Appsero),
which will be displayed in the Plugins list in the WordPress Admin.

currently

proposed

Note that this name is returned when using the WordPress core function [get_plugin_data()](https://developer.wordpress.org/reference/functions/get_plugin_data/),
however I did not find any use of that internally within this plugin.

If a 3rd party plugin is using the plugin name, they should be encouraged to use a more reliable field for a unique identifier, such as "Text Domain", which is traditionally the same as the plugin directory.

Where has this been tested?
---------------------------
**Operating System:**
macOS v13.2.1

**WordPress Version:**
v6.1

### DIFF
--- a/wp-graphql.php
+++ b/wp-graphql.php
@@ -72,7 +72,7 @@ function graphql_init_appsero_telemetry() {
 		return;
 	}
 
-	$client   = new Appsero\Client( 'cd0d1172-95a0-4460-a36a-2c303807c9ef', 'WP GraphQL', __FILE__ );
+	$client   = new Appsero\Client( 'cd0d1172-95a0-4460-a36a-2c303807c9ef', 'WPGraphQL', __FILE__ );
 	$insights = $client->insights();
 
 	// If the Appsero client has the add_plugin_data method, use it
--- a/wp-graphql.php
+++ b/wp-graphql.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Plugin Name: WP GraphQL
+ * Plugin Name: WPGraphQL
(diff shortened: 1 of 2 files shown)


wp-graphql.php
```diff
@@ -72,7 +72,7 @@ function graphql_init_appsero_telemetry() {
 		return;
 	}
 
-	$client   = new Appsero\Client( 'cd0d1172-95a0-4460-a36a-2c303807c9ef', 'WP GraphQL', __FILE__ );
+	$client   = new Appsero\Client( 'cd0d1172-95a0-4460-a36a-2c303807c9ef', 'WPGraphQL', __FILE__ );
```
josephfusco, line 75: I'm not sure what the effects of changing the 2nd argument here would be, but wanted to update to discuss within this PR.

jasonbahl: @josephfusco, ironically Appsero actually already says this should be WPGraphQL without the space. . .so I'll merge this and monitor to make sure it doesn't impact Appsero tracking in a negative way. 

Thanks 🙌🏻

Issue #2715, opened 2023-02-08
Acquiring the WP_Query instance within a wp-graphql filter causes post hooks to fire again


ToughCrab24:
### Description

When acquiring the WP_Query instance inside of wp-graphql filters, it causes post hooks such as `posts_pre_query` to fire a second time within the wp-graphql request lifecyle.

**Example**
```php
function filter_graphql_connection_page_info(
  $page_info,
  AbstractConnectionResolver $resolver,
) {
  $query = $resolver->get_query();
  return $query->query_args;
}

add_filter(
  'graphql_connection_page_info',
  ['filter_graphql_connection_page_info'],
  10,
  2
);
```
**Explanation**

In the case the `AbstractConnectionResolver` is an instance of `PostObjectConnectionResolver `
we can see that the `WP_Query` class is instantiated

https://github.com/wp-graphql/wp-graphql/blob/9ee971695cdb848d7b51afdd2df837bdf111df76/src/Data/Connection/PostObjectConnectionResolver.php#L94-L107

Digging into WordPress core we can see the side-effect of instantiating a new instance of `WP_Query` calls `$this->query( $query )` in its constructor and then calls `$this->get_posts()`

https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-query.php#L3787

This results in applying the `posts_pre_query` and other hooks https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-query.php#L3136

### Steps to reproduce

Define hooks:
```php
function filter_graphql_connection_page_info(
  $page_info,
  AbstractConnectionResolver $resolver,
) {
  $query = $resolver->get_query();
  return $query->query_args;
}

add_filter(
  'graphql_connection_page_info',
  ['filter_graphql_connection_page_info'],
  10,
  2
);

function filter_posts_pre_query($posts, WP_Query $query) {
   graphql_debug("Triggered posts_pre_query");
   return $posts;
}

add_filter('posts_pre_query', 'filter_posts_pre_query', 10, 2);
```

With wp-graphql debug enabled run the query 

```graphql
{
  posts {
    nodes {
      title
    }
  }
}
```

### WPGraphQL Version

1.13.8

### WordPress Version

6.1.1

### PHP Version

7.4 & 8

### Additional enviornment details

Observed in wp-graphql-offset-pagination https://github.com/valu-digital/wp-graphql-offset-pagination/blob/master/src/Loader.php#L128

[...]

Issue #2678, opened 2022-12-15
Overriding/subclassing the default object connection types


maurocolella:
### What problem does this address?

Hi.

I have a situation where we are using paginated queries, but need to apply complex access control logic.

This section from `PostObjectConnectionResolver` illustrates the issue:

```
  /**
   * {@inheritDoc}
   */
  public function get_ids_from_query()
  {
    $ids = !empty($this->query->posts) ? $this->query->posts : [];

    // FIXME: this is where we should filter entries according to security rules
    foreach($ids as $id) {

    }

    // If we're going backwards, we need to reverse the array.
    if (!empty($this->args['last'])) {
      $ids = array_reverse($ids);
    }

    return $ids;
  }
  ```
  
Existing filters let us either filter the `WP_Query` upstream (but our logic can't be expressed as a meta query), or the returned ids _after_ cursors have been applied to them. 

Currently, the `AbstractConnectionResolver` applies cursors immediately to the output of the above function.

### What is your proposed solution?

A filter after ids have been retrieved, but before cursors are applied to them.

### What alternatives have you considered?

I have tried to override the `PostObjectConnectionResolver`:

```
register_graphql_connection(
          [
            'fromType'       => 'RootQuery',
            'toType'         => ucfirst($post_type->graphql_single_name),
            'queryClass'     => 'WP_Query',
            'fromFieldName'  => $post_type->graphql_plural_name,
            'connectionArgs' => PostObjects::get_connection_args(),
            'resolve'        => function ($root, $args, $context, $info) use ($post_type) {
              $resolver = new \WPGraphQL\Data\Connection\CustomPostObjectConnectionResolver($root, $args, $context, $info, $post_type);

              return $resolver->get_connection();
            },
          ]
        );
 ```
 
 In:
 
 ```
 graphql_register_types
 ```
 
 But obviously, this results in a complaint about duplicate types.
 
 Is something like this possible currently?

jasonbahl:
> I have a situation where we are using paginated queries, but need to apply complex access control logic.

@maurocolella can you provide a bit more information about what logic you need to apply?

If it's logic about individual posts and whether they should be returned or not, the Model layer might be the right place to apply this logic. 

For example, [this filter](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Model/Model.php#L238) could be used to determine whether the node should be considered private or not. 

If you return `true` from that filter, then the node would be considered private, and the node would be removed from the results of the query. 

[...]

Issue #1588, opened 2020-11-16
Filter API inconsistencies 


smthomas:
In the `graphql_map_input_fields_to_wp_query` filter, the $post_type is the last parameter. This is inconsistent with the other `graphql_map_input_fields_*` filters. For example in `graphql_map_input_fields_to_get_terms` the $taxonomy parameter is the third parameter (while the $context & $info parameters are the last two in all of the `graphql_map_input_fields_*` filters). For consistency sake, it may make sense to update this filter.

wp_query filter: https://github.com/wp-graphql/wp-graphql/blob/560bd4ab94ffbc939014a90c4ffa9ca5da1880fa/src/Data/Connection/PostObjectConnectionResolver.php#L431

get_terms filter: https://github.com/wp-graphql/wp-graphql/blob/560bd4ab94ffbc939014a90c4ffa9ca5da1880fa/src/Data/Connection/TermObjectConnectionResolver.php#L225

justlevine:
Related: #331

justlevine:
Removing `not stale`, now that stalebot is behaving.

While filter consistency is still an issue, the specific inconsistencies will need to be reviewed before we can figure out an implementation plan (and what can be addressed via breaking/nonbreaking changes).

tihom4537:
I have generated a pull request. Do consider it for the review .

justlevine:
Hi @tihom4537 

The effort is appreciated. However, as I mentioned [above](https://github.com/wp-graphql/wp-graphql/issues/1588#issuecomment-1837441050), this is currently on hold. It likely won't be merged until we start working on v2.0, and that's assuming no other changes invalidate it.

PR #2749, opened 2023-03-04
dev!: [WIP] Exploring a ConnectionResolver 2.0 

justlevine:
This PR explores what an "ideal" `AbstractConnectionResolver` would look like if we didnt need to worry about breaking changes.

The `AbstractConnectionResolver` is an essential yet complex part of the codebase and API. The existing class complicates dx and makes it hard to extend, and also obfuscates what a holistic Connection Resolver would look like without the tech debt.

The goal is to is to use this as a springboard to map out a path to a theoretical breaking release and the nonbreaking improvements we can already make.

### Details
The gist of these changes boil down to splitting existing methods into those that handle the actual logic, and wrappers that filter the data and set the properties. A working set of public getter methods allows for simpler yet more powerful interaction with the resolver instance (either in a filter or in a `resolve` callback). And a [doc](https://github.com/justlevine/wp-graphql/blob/9c2ed83af746adcba7f11025d8fb9acce71a9fe3/src/Data/Connection/connection-resolvers.md) to explain the tricky lifecycle and best ways to interact with it.

### To do
- [ ] Review method visibility.
- [ ] Decide if existing getters that doesnt use class properties should be refactored to store the values on the instance.
- [ ] Review the internals of the child resolver classes (as of now I didnt really pay much attention beyond the method signatures), expose new getters and extend-friendly methods, and improve filter parity.
- [ ] Review implications of and on #2466 (including possible ways to instantiate the class indeterminately from the `resolveFn()` ).
- [ ] Take a look at 3rd party Connection Resolvers and see if there's anything else we can do to make their lives easier.

- [ ] Fix/update doc-blocks.
- [ ] Select candidates for cherry-picking.

- #2715 
- #2678
- #1588 
and hopefully a few others both here and downstream.

Any other comments?
-------------------
WIP

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + php 8.0.26)

**WordPress Version:** 6.1.1


justlevine:
aad01061f1c734ace2d86abd2e1838a5f901bdcd refactors the ConnectionResolver to support overloading the `$query_class` on a resolver level. This differs from #2821 in several ways:

1. The `$query_class` can be set in the resolver with the `query_class()` method. It returns `null` by default, and should be overloaded to return the name of the query class in resolvers that require one.
```php
protected function query_class() ?: string {
  return 'WC_Query';
}
```

2. The property is set on the constructor, using `get_query_class()`. That method will use `$context->queryClass` before defaulting to `query_class()`, and wraps it in the `graphql_connection_query_class` filter.

3. `$resolver->set_query_class()` can be used to overload the query class on a specific Resolver instance. You can pass an empty string to reset the query class to the default (i.e. ignoring the `$context->queryClass` or the results of the filter).
```php
...$connectionConfig,
'resolve' => function( $source, $args, $context, $info ) {
  $resolver = new PostObjectConnectionResolver( $source, $args, $context, $info );
  
  $resolver->set_query_class( WC_Query::class ); // If "" is passed, revert to default query class.
  
},
```

4. When `get_query()` is run, a check will be made to `is_valid_query_class()` to make sure that:
  - There is only a query class is the connection requires it (i.e. if `query_class()` is not `null`).
  - The query class exists.
  - The query class is compatible with our Connection Resolver. This is checked with `is_valid_query_class()`.
```php
protected function is_valid_query_class( string $query_class ) : bool {  
  return method_exists( $query_class, 'get_results' ); // Instead of the default `WP_Query::query()`
}
```
  
As a result of these changes:
- `query()` is no longer a required abstract method; however if a ConnectionResolver does not use a `query_class`, an invariant error will be thrown if the method is not overloaded.
- Extending ConnectionResolver is now DRYer, since in most cases developers will only need to overload the `query_class` method, instead of the entire `query()`.

justlevine:
rebased on `develop`.

justlevine:
Even with the indeterminate model/dataloader pattern unsolved, I'd love to start backporting some of the nonbreaking changes (and close several existing Issues).

@jasonbahl What can I do to make this proposal easier for you to review and give feedback on?

justlevine:
Rebased back onto `develop`, and moved the ConnectionResolver changes back inline to `AbstractConnectionResolver`, to make it easy to review.

For a quick code overview of the breaking changes, you can compare the `ConnectionResolverShim` to the _old_ `AbstractConnectionResolver`.

justlevine:
I believe I've backported everything I can without creating an impossible mess of merge conflicts.

As @jasonbahl reviews/merges the existing PRs, I'll be able to start reviewing more.

You can see the diff of what _remains_ to be backported here: https://github.com/wp-graphql/wp-graphql/pull/2749/files/7415eca0ebd3f24c6a5c5e64289be655431bac50..217135b426d039fb95a407d31ed0e574608a533b

justlevine:
The next batch of backports is up, focusing on improving DX and instantiating class properties only once:
- #3120
- #3121
- #3122
- #3123
- #3124
- #3125

Once this batch is merged, I'll do one more sweep, but I think we're getting close to being able to backport as many of these changes as possible without breaking backwards-compatibility.

cc @jasonbahl

josephfusco:
Hey @justlevine, it looks like most of this has been reviewed and merged as separated pull requests. Is this PR still relevant or would it be safe to close?

justlevine:
> Hey @justlevine, it looks like most of this has been reviewed and merged as separated pull requests. Is this PR still relevant or would it be safe to close?

Almost, still on the todo list are:

- [ ] Centralized hooks/methods for mapping/sanitizing where args.
- [ ] Docs.

Once that's done, I'll rebase and close this, so there's a reference for when we're ready to implement breaking changes.

PR #2758, opened 2023-03-08
fix: Allow post types and taxonomies to be registered without graphql_plural_name

### LINKED ISSUE #2754: Documentation and application validation do not match
### Description

The documentation lists `graphql_plural_name` as optional: https://www.wpgraphql.com/docs/custom-post-types

> graphql_plural_name | string | (optional): camel case string with no punctuation or spaces. Needs to start with a letter (not a number).

>     $args['graphql_plural_name'] = 'documents'; # Don't set, and it will default to `all${graphql_single_name}`, i.e. `allDocument`.

However, this validation step throws errors pretty aggressively if it is not set:
https://github.com/wp-graphql/wp-graphql/blob/ead0acb6d9ea1ad3eaaf6a36850c33ba0720bb6a/src/WPGraphQL.php#L622-L630

I'm not sure which should be updated, documentation or the validation, but they are currently in conflict with one another.

### Steps to reproduce

Create a post type like so:

```
   register_post_type( 'docs', [
      'show_ui' => true, # whether you want the post_type to show in the WP Admin UI. Doesn't affect WPGraphQL Schema.
      'labels'  => [
        //@see https://developer.wordpress.org/themes/functionality/internationalization/
        'menu_name' => __( 'Docs', 'your-textdomain' ), # The label for the WP Admin. Doesn't affect the WPGraphQL Schema.
      ],  
      'hierarchical' => true, # set to false if you don't want parent/child relationships for the entries
      'show_in_graphql' => true, # Set to false if you want to exclude this type from the GraphQL Schema
      'graphql_single_name' => 'document',
      'public' => true, # set to false if entries of the post_type should not have public URIs per entry
      'publicly_queryable' => true, # Set to false if entries should only be queryable in WPGraphQL by authenticated requests
   ] );
} );
```

And watch all graphQL queries return a fatal error.

### Additional context

_No response_

### WPGraphQL Version

1.14.0-rc

### WordPress Version

6.1.1

### PHP Version

8.2.3

_No response_

- [X] Yes

- [X] Yes

jasonbahl:
This allows post types/taxonomies to be registered to show_in_graphql without requiring `graphql_plural_name` (graphql_single_name) is still required.

closes #2754 

Registering a post type like so: 

```php
	register_post_type( 'test_type', [
		'label' => 'Test Type',
		'public' => true,
		'show_in_graphql' => true,
		'graphql_single_name' => 'testType',
	]);
});
```

## Before

CleanShot 2023-03-08 at 13 39 20

## After

CleanShot 2023-03-08 at 13 41 47

Any other comments?
-------------------
This PR also converts the hard UserError that's thrown and changes it to a `graphql_debug()` message. 

This allows the Schema to still generate even if a post type is registered with `show_in_graphql => true`, but is missing a graphql_single_name. 

ex: 

```php
	register_post_type( 'test_type', [
		'label' => 'Test Type',
		'public' => true,
		'show_in_graphql' => true,
		// oops, forgot to add a graphql_single_name
	]);
});
```

Results in the Schema loading in GraphiQL, but a debug message being output:

CleanShot 2023-03-08 at 14 45 46

Previously this would've prevented the Schema / GraphiQL from loading.

### DIFF
--- a/src/WPGraphQL.php
+++ b/src/WPGraphQL.php
@@ -605,32 +605,7 @@ final class WPGraphQL {
 				'objects'
 			);
 
-			$post_type_names = [];
-
-			foreach ( $post_type_objects as $post_type_object ) {
-
-				// if the graphql_single_name was provided, but not the plural name
-				// use the single name as the plural name
-				if ( empty( $post_type_object->graphql_plural_name ) && ! empty( $post_type_object->graphql_single_name ) ) {
-					$post_type_object->graphql_plural_name = $post_type_object->graphql_single_name;
-				}
-
-				/**
-				 * Validate that the post_types have a graphql_single_name and graphql_plural_name
-				 */
-				if ( empty( $post_type_object->graphql_single_name ) || empty( $post_type_object->graphql_plural_name ) ) {
-					throw new UserError(
-						sprintf(
-							/* translators: %s will replaced with the registered type */
-							__( 'The %s post_type isn\'t configured properly to show in GraphQL. It needs a "graphql_single_name" and a "graphql_plural_name"', 'wp-graphql' ),
-							$post_type_object->name
-						)
-					);
-				}
-
-				// Save allowed post type names so they can be filtered by the user.
-				$post_type_names[] = $post_type_object->name;
-			}
+			$post_type_names = wp_list_pluck( $post_type_objects, 'name' );
 
 			/**
 			 * Pass through a filter to allow the post_types to be modified.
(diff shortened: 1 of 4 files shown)


src/WPGraphQL.php
```diff
@@ -640,33 +622,45 @@ public static function get_allowed_post_types( $output = 'names', $args = [] ) {
...
-			if ( $post_type_names !== $allowed_post_type_names ) {
-				// Maybe they're just out of order.
-				sort( $post_type_names );
-				sort( $allowed_post_type_names );
-
-				if ( $post_type_names !== $allowed_post_type_names ) {
-					$post_type_objects = array_filter($post_type_objects, function ( $obj ) use ( $allowed_post_type_names ) {
-						return in_array( $obj->name, $allowed_post_type_names, true );
-					});
+			$post_type_objects = array_filter( $post_type_objects, static function ( $obj ) use ( $allowed_post_type_names ) {
```
justlevine, line 625: The upcoming conditions can be simplified by early returning if `graphql_single_name` is empty.

src/WPGraphQL.php
```diff
@@ -640,33 +622,45 @@ public static function get_allowed_post_types( $output = 'names', $args = [] ) {
...
+
+				if ( empty( $obj->graphql_plural_name ) && ! empty( $obj->graphql_single_name ) ) {
+					$obj->graphql_plural_name = $obj->graphql_single_name;
```
justlevine, line 628: Can we drop a comment here pointing to where it gets renamed to `all{graphql_single_name}` to improve readability?

src/WPGraphQL.php
```diff
@@ -640,33 +622,45 @@ public static function get_allowed_post_types( $output = 'names', $args = [] ) {
...
 				}
-			}
+
+				/**
+				 * Validate that the post_types have a graphql_single_name and graphql_plural_name
+				 */
+				if ( empty( $obj->graphql_single_name ) || empty( $obj->graphql_plural_name ) ) {
```
justlevine, line 634: The new logic means we only need to check for `graphql_single_name`

src/WPGraphQL.php
```diff
@@ -640,33 +622,45 @@ public static function get_allowed_post_types( $output = 'names', $args = [] ) {
...
+					graphql_debug(
```
justlevine, line 635: We used to throw a UserError. Is this change to graphql_debug() intentional, and are there downstream considerations we need to worry about?

justlevine (commented):
Left some feedback on allowed_post_types, but it applies to allowed_taxonomies to. (Apologies for the berevity, reviewing from mobile)

src/WPGraphQL.php
jasonbahl: It is intentional. 

With so many non-code ways to add post types (CPT UI, ACF (upcoming), Toolset Types, etc) it feels harsh to blow up the entire schema with an error for a missing key. Instead, we simply exclude the post type because it's missing information and can inform the user of the mistake via graphql_debug() while still allowing the existing queries, etc to continue working.

src/WPGraphQL.php
justlevine: I'm worried that setting it there and not transforming it is going to cause issues with people using `$obj->graphql_plural_name` in their own code, since its no longer the source of truth 🤔

src/WPGraphQL.php
```diff
@@ -605,25 +605,7 @@ public static function get_allowed_post_types( $output = 'names', $args = [] ) {
...
-							__( 'The %s post_type isn\'t configured properly to show in GraphQL. It needs a "graphql_single_name" and a "graphql_plural_name"', 'wp-graphql' ),
-							$post_type_object->name
-						)
-					);
-				}
-
-				// Save allowed post type names so they can be filtered by the user.
-				$post_type_names[] = $post_type_object->name;
-			}
+			$post_type_names = wp_list_pluck( $post_type_objects, 'name' );
```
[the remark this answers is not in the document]
jasonbahl: well, it's still just post types set to `show_in_graphql`, because we call `get_post_types( [ 'show_in_graphql' => true ]);` above. 

So, we initially start with a list of post types set to `show_in_graphql` (even if they mistakenly forget to set graphql_single_name / graphql_plural_name)

Then, we apply this filter over that already reduced list of post types.

After the filter is applied we validate that the post type has at least the graphql_single_name set (and set graphql_plural_name fallback if needed), then also make sure the post type is allowed by comparing to the post-filtered `$allowed_post_types` list. 

I believe it's the same intended functionality as before, but slightly cleaner and (in my opinion) a better user experience for gracefully failing when a post type / taxonomy is registered incorrectly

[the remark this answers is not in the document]
justlevine: 💯 just wanted to leave a note in case the change in behavior requires a local code change.

PR #2770, opened 2023-03-27
fix: wrong order in term and post ancestors.

creative-andrew:

This PR fixes the wrong order in the ancestor's resolvers for terms and posts.

- Terms: The default order in [TermObjectConnectionResolver](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Connection/TermObjectConnectionResolver.php#L79) is `$query_args['orderby'] = 'name';`. The [ancestors resolver](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Registry/Utils/TermObject.php#L179) uses `$resolver->set_query_arg( 'include', $ancestor_ids );` that requires `$resolver->set_query_arg( 'orderby', 'include' );` in other to preserve the [initial order from](https://developer.wordpress.org/reference/functions/get_ancestors/) `get_ancestors` function.

- Post: The default order in [PostObjectConnectionResolver](https://github.com/wp-graphql/wp-graphql/blob/44fbc682ec0d62e011447941c38cb74d1ef591de/src/Type/Connection/PostObjects.php#L140) is `$resolver->set_query_arg( 'orderby', 'post__in' );`. The [ancestors resolver](https://github.com/wp-graphql/wp-graphql/pull/2770/files#diff-eae106d899fe458092e3a45740f6a69b7a99f259222c25dfb1d4d67ae45b3356R161) uses `$resolver->set_query_arg( 'post__in', $ancestors );` that requires `$resolver->set_query_arg( 'orderby', 'post__in' );` in other to preserve the [initial order from](https://developer.wordpress.org/reference/functions/get_ancestors/) `get_ancestors` function.

Props to @italodr for identifying the bug.

### DIFF
--- a/src/Registry/Utils/TermObject.php
+++ b/src/Registry/Utils/TermObject.php
@@ -177,6 +177,7 @@ class TermObject {
 
 					$resolver = new TermObjectConnectionResolver( $term, $args, $context, $info, $tax_object->name );
 					$resolver->set_query_arg( 'include', $ancestor_ids );
+					$resolver->set_query_arg( 'orderby', 'include' );
 
 					return $resolver->get_connection();
 				},
--- a/src/Type/Connection/PostObjects.php
+++ b/src/Type/Connection/PostObjects.php
@@ -159,6 +159,7 @@ class PostObjects {
 				}
 				$resolver = new PostObjectConnectionResolver( $post, $args, $context, $info );
 				$resolver->set_query_arg( 'post__in', $ancestors );
+				$resolver->set_query_arg( 'orderby', 'post__in' );
 
 				return $resolver->get_connection();
 			},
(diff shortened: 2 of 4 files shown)


jasonbahl:
@creative-andrew would you be willing to write tests to ensure we don't cause a regression to this?

creative-andrew:
> * TermObjectConnectionResolver

Will do!

creative-andrew:
> @creative-andrew would you be willing to write tests to ensure we don't cause a regression to this?

@jasonbahl  Will do!

jasonbahl:
@creative-andrew I was able to reproduce the scenario for post ancestors being ordered wrong.

Given the following hierarchy: 

- Parent
  - Child
    - Grandchild

If the "Child" has a publish date prior to the parent, then the "ancestors" of the "grandchild" were indeed coming in an unexpected order. 

Going to take a stab at testing the other scenario now.

creative-andrew:
@jasonbahl  thanks for taking the time to look into this!

PR #2777, opened 2023-04-08
chore: update composer dev-deps (not PHPStan)

justlevine:
What does this implement/fix? Explain your changes.
---------------------------------------------------
This PR updates the Composer `dev-dependencies` to their latest (compatible) versions.

Does this close any currently open issues?
------------------------------------------
No

Any relevant logs, error output, GraphiQL screenshots, etc?
-------------------------------------
(If it’s long, please paste to https://ghostbin.com/ and insert the link here.)

Any other comments?
-------------------
This _does not_ include updating `PHPStan` and related deps, since thats throwing code smells. A separate PR will handle that.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + php8.0.26)

**WordPress Version:** 6.2

PR #2778, opened 2023-04-08
fix: Update PHPStan and fix smells

justlevine:
This PR updates PHPStan (and phpstan-wordpress, etc) to the latest versions, fixing the newly-surfaced code smells.

Requires #2777 to avoid composer.lock merge conflicts, but can be rebased if necessary.

…

Any other comments?
-------------------
The only _potential_ user-facing change is the new `Exception` in `Request::execute()` if `Request->$params` isnt an `array` or 
`OperationParams`, but I'm pretty sure from the code that's not an actually possibility.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + php8.0.26)

**WordPress Version:** 6.2


src/Request.php
```diff
@@ -626,58 +626,58 @@ public function execute() {
 			}, $this->params);
 		}
 
-		if ( $this->params instanceof OperationParams ) {
+		// If $this->params isnt an array or an OperationParams instance, then something probably went wrong.
+		if ( ! $this->params instanceof OperationParams ) {
```
justlevine, line 630: The PHPStan-requested change here is that `execute()` doesn't return anything if `$this->params` is neither an array or an `OperationParams` instance.

This is solved by "returning early" instead of wrapping the logic in a conditional.

I _believe_ throwing an exception is the right approach here, but because of the method recursion and the lack of `array` shape for `$this->params`/`$this->data`, this could use a second set of eyes.

cc: @jasonbahl

src/Type/ObjectType/CommentAuthor.php
```diff
@@ -84,7 +84,7 @@ public static function register_type() {
 							}
 
 							if ( ! empty( $args['rating'] ) ) {
-								$avatar_args['rating'] = esc_sql( $args['rating'] );
+								$avatar_args['rating'] = esc_sql( (string) $args['rating'] );
```
justlevine, line 87: This is me being lazy, since we know `$args['rating']` is a string and I didnt want to be bothered defining the entire shape of `$args`.

jasonbahl: 👌🏻

jasonbahl (approved):

PR #2792, opened 2023-04-19
fix: uri field is null when querying the page for posts uri

### LINKED ISSUE #2791: Posts page not properly returning from nodeByUri query
### Description

When using the nodeByUri query to fetch the posts page (blog archive) the response is not what I would expect. 

Any node that can be queried via a URI should output a uri on the node, however when I query for the posts page, currently the `uri` is returning null. 

see: 
CleanShot 2023-04-19 at 10 11 35

I would expect the uri field to output `/sample-page/` instead of null

### Steps to reproduce

1. go to Settings > Reading
2. Set the "posts page" setting to a page (for me it's "Sample Page")
3. execute the following query (changing the uri input to that of the page you set as the posts page):

```graphql
{
  nodeByUri(uri: "/sample-page/") {
    __typename
    uri
  }
}
```

4. See that the `__typename` is `ContentType` (expected)
5. See that the `uri` is `null` (unexpected)

### Additional context

_No response_

### WPGraphQL Version

1.14.1

### WordPress Version

6.2

### PHP Version

8.1

_No response_

- [X] Yes

- [ ] Yes

jasonbahl:

This fixes a bug in the NodeResolver where querying the posts page prevents the node from fully resolving properly. 

failing test: https://github.com/wp-graphql/wp-graphql/actions/runs/4745816863/jobs/8428653391?pr=2792#step:8:816
passing test: https://github.com/wp-graphql/wp-graphql/actions/runs/4745911977/jobs/8428890823?pr=2792#step:8:813

closes #2791

Any other comments?
-------------------

**BEFORE**

I can query a node by URI and get the node back, but the node returns `null` for the uri field 🤔. This is a bug. If a node can be identified by a uri, the node has a uri. 

CleanShot 2023-04-19 at 10 43 30

**AFTER**

The uri field is output as expected 👏🏻 

CleanShot 2023-04-19 at 10 44 28

### DIFF
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -201,8 +201,7 @@ class NodeResolver {
 					return null;
 				}
 
-				/** @todo resolve to an Archive Type. */
-				$post_type_object = get_post_type_object( $queried_object->post_type );
+				$post_type_object = get_post_type_object( 'post' );
 
 				if ( ! $post_type_object ) {
 					return null;
--- a/tests/wpunit/NodeByUriTest.php
+++ b/tests/wpunit/NodeByUriTest.php
@@ -1084,13 +1084,13 @@ class NodeByUriTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase {
 		codecept_debug( $uri );
 
 		$term = get_term_by('slug', 'post-format-aside', 'post_format');
-		
+
 		/**
 		 * NodeResolver::parse_request() will generate the following query vars:
 		 * uri => /type/aside/
 		 * post_format => post-format-aside
 		 * post_type => [ post ]
-		 * 
+		 *
 		 */
 		$actual = $this->graphql([
 			'query'     => $query,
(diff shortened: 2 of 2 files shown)


src/Data/NodeResolver.php
```diff
@@ -201,8 +201,7 @@ public function resolve_uri( string $uri, $extra_query_vars = '' ) {
 					return null;
 				}
 
-				/** @todo resolve to an Archive Type. */
-				$post_type_object = get_post_type_object( $queried_object->post_type );
+				$post_type_object = get_post_type_object( 'post' );
```
justlevine, line 204: I'm not by my desk to check, but won't this break CPTs?

jasonbahl: @justlevine nope, this is within this condition `if ( $query->is_posts_page ) {` which we know is only for the `post` post type

justlevine: Okay phew!

PR #2799, opened 2023-04-24
fix: querying posts by slug fails when custom permalinks are set

### LINKED ISSUE #2798: Querying posts by slug fails when custom permalinks are set
### Description

When custom permalinks are set, querying for posts by slug returns null, even if the slug is a valid slug. 

### Steps to reproduce

1. set your permalinks to a custom structure. I set mine to `/articles/%postname%/`
2. Publish a post. I published a post to `/articles/kitchen-sink`
3. Query that post by slug: 
```graphql
query GetPostsBySlug {
  post(id: "kitchen-sink", idType: SLUG) {
    __typename
  }
}
```
4. See that there is a null result for the post, even though the slug is valid. 
5. You can even query for a list of posts at the same time, and see the null result while also seeing the slug is valid:

CleanShot 2023-04-24 at 17 07 04
`

### Additional context

@justlevine needs to go to sleep.

### WPGraphQL Version

1.14.2

### WordPress Version

6.2

### PHP Version

8.1

_No response_

- [X] Yes

- [ ] Yes

jasonbahl:

This fixes a bug where querying for posts (of a non-hierarchical post type) by Slug would return null when the site had custom permalinks set. 

closes #2798 2798

Any other comments?
-------------------

**BEFORE**

CleanShot 2023-04-24 at 17 07 04

**AFTER**

CleanShot 2023-04-24 at 17 15 10

Notes for future reference (written at 2:30AM by @justlevine):

- The original issue only occurs when querying via the non-pretty url (index.php?graphql) like in GraphiQL, not the pretty permalink (/graphql).
- The root cause is that unlike WP native's routing, our parse_request() doesn't match slugs (since the rewrite rules are now prefixed with a custom base), leaving $query_vars['error'] set to true.
- The reason that it only affects non-pretty urls is because our parse_request() matches requests via the pretty permalink to the pretty permalink, instead of our query. This is because we're reusing the $wp global instead of creating a new one. This is a bug
- The 'fix' in this PR works by employing that bug to our advantage. We pretend we matched the /graphql rewrite rules, essentially bypassing the rule matching in parse_request() so there's no difference in handling between requests to the prettied/naked urls.
- An actual fix in the long term would be reviewing the differences between our parse_request() and native WP's, so we're accurately able to match and rely on WP's rewrite rules. (Or alternatively to ditch more parts of parse_request() that are seemingly unnecessary).

### DIFF
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -506,7 +506,7 @@ class NodeResolver {
 				if ( isset( $post_type_query_vars[ $wpvar ] ) ) {
 					$this->wp->query_vars['post_type'] = $post_type_query_vars[ $wpvar ];
 					$this->wp->query_vars['name']      = $this->wp->query_vars[ $wpvar ];
-				}           
+				}
 			}
 		}
 
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -506,8 +506,7 @@ class NodeResolver {
 				if ( isset( $post_type_query_vars[ $wpvar ] ) ) {
 					$this->wp->query_vars['post_type'] = $post_type_query_vars[ $wpvar ];
 					$this->wp->query_vars['name']      = $this->wp->query_vars[ $wpvar ];
-				}
-
+				}           
 			}
 		}
 
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -2,12 +2,11 @@
 
 namespace WPGraphQL\Data;
 
-use Exception;
 use GraphQL\Deferred;
-use WP;
 use WP_Post;
 use WPGraphQL\AppContext;
 use GraphQL\Error\UserError;
+use WPGraphQL\Router;
 
 class NodeResolver {
 
@@ -30,8 +29,9 @@ class NodeResolver {
 	 */
 	public function __construct( AppContext $context ) {
 		global $wp;
-		$this->wp      = $wp;
-		$this->context = $context;
+		$this->wp               = $wp;
+		$this->wp->matched_rule = Router::$route . '/?$';
+		$this->context          = $context;
 	}
 
 	/**
@@ -435,6 +435,7 @@ class NodeResolver {
 			}
 
 			if ( ! empty( $this->wp->matched_rule ) ) {
+
 				// Trim the query of everything up to the '?'.
 				$query = preg_replace( '!^.+\?!', '', $query );
 
@@ -506,6 +507,7 @@ class NodeResolver {
 					$this->wp->query_vars['post_type'] = $post_type_query_vars[ $wpvar ];
 					$this->wp->query_vars['name']      = $this->wp->query_vars[ $wpvar ];
 				}
+
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


src/Data/NodeResolver.php
```diff
@@ -505,7 +506,7 @@ public function parse_request( string $uri, $extra_query_vars = '' ) {
 				if ( isset( $post_type_query_vars[ $wpvar ] ) ) {
 					$this->wp->query_vars['post_type'] = $post_type_query_vars[ $wpvar ];
 					$this->wp->query_vars['name']      = $this->wp->query_vars[ $wpvar ];
-				}
+				}
```
justlevine, line 509:
```suggestion
				}
```

tests/wpunit/NodeByUriTest.php
```diff
@@ -1948,4 +1948,45 @@ public function testQueryPostsPageByUriReturnsExpectedUriOnNode() {
...
+		';
+
+		$actual = $this->graphql([
+			'query' => $query,
+			'variables' => [
+				'id' => $slug,
+			]
+		]);
+
+		self::assertQuerySuccessful( $actual, [
```
justlevine, line 1985:
```suggestion
		$this->assertQuerySuccessful( $actual, [
```

justlevine:
Notes for future reference (written at 2:30AM. Will review and edit in the morning):
- The original issue only occurs when querying via the non-pretty url (`index.php?graphql`) like in GraphiQL, not the pretty permalink `(/graphql)`.
- The root cause is that unlike WP native's routing, our `parse_request()` doesn't match slugs (since the rewrite rules are now prefixed with a custom base), leaving `$query_vars['error']` set to true.
- The reason that it only affects non-pretty urls is because _our_ `parse_request()` matches requests via the pretty permalink _to_ the pretty permalink, instead of our query. This is because we're reusing the `$wp` global instead of creating a new one. **This is a bug**
- The 'fix' in this PR works by employing that bug to our advantage. We pretend we matched the `/graphql` rewrite rules, essentially bypassing `parse_request()` so there's no difference in handling between requests to the prettied/naked urls.
- An actual fix in the long term would be reviewing the differences between our `parse_request()` and native WP's, so we're accurately able to match and rely on WP's rewrite rules. (Or alternatively to ditch _more_ parts of `parse_request()` that are _seemingly_ unnecessary).

tests/wpunit/NodeByUriTest.php
```diff
@@ -1948,4 +1948,45 @@ public function testQueryPostsPageByUriReturnsExpectedUriOnNode() {
...
+		$this->set_permalink_structure( '/articles/%postname%/' );
+
+		$slug = 'test-slug';
+
+		$postId = $this->factory()->post->create([
+			'post_type'   => 'post',
+			'post_status' => 'publish',
+			'post_name' => $slug,
+			'post_title' => 'post-title'
+ 		]);
```
justlevine, line 1962: Do we want to create a second post here, just to confirm it's actually matching and not returning the most recent?

jasonbahl: @justlevine good call. updated the test.

PR #2808, opened 2023-05-08
fix: fallback to template filename if sanitized name is empty

### LINKED ISSUE #2492: GraphQL IDE got broken when site locale is CJK
### Description

If my wp site locale setting is CJK(Chinese/Japanese/Korean)(Non-english) it will throw GraphQLError.

### Steps to reproduce

1. Download & install WPGraphQL from WordPress Plugin Directory.
2. Switch to GraphiQL IDE.
3. Nothing show but keep loading.
4. Console shows
```
Uncaught (in promise) GraphQLError: Names must only contain [_a-zA-Z0-9] but "Template_链接" does not.
    a app.js:39730
    a app.js:44958
    q app.js:45227
    h app.js:47708
    h app.js:47767
    keyValMap app.js:42081
    buildClientSchema app.js:47697
    yg app.js:23742
app.js:39691
```
5. When set site locale to English, it works fine.

### Additional context

screenshot

It look likes it's not unusual. I found a issue comment has the same problem. [Issue Comment](https://github.com/wp-graphql/wp-graphql/issues/2313#issuecomment-1108232058)

### WPGraphQL Version

1.9.1

### WordPress Version

6.0.2

### PHP Version

7.4

_No response_

- [X] Yes

- [X] Yes

### LINKED ISSUE #2807: When using non-ascii characters, generated an invalid schema.
### Description

Here, it is assumed that there is a template file with meta information set for the following.

```php
<?php
/*
Template Name: 会社概要ページ トップ
*/

/* some contents is returned */

```

In this case, generated schema file's `ContentTemplate` node is here.

```grahql
"""The template assigned to the node"""
type Template_会社概要ページトップ implements ContentTemplate {
  """The name of the template"""
  templateName: String
}
```

This is invalid schema, because schema type is accepted only ascii characters. So GraphQL IDE show error message in browser console as below.

```
Uncaught (in promise) GraphQLError: Names must only contain [_a-zA-Z0-9] but "Template_会社概要ページトップ" does not.
```

### Steps to reproduce

1. Create a template file and set the template name to non-ASCII characters(ex: jp, zh-CN, ar).
```php
<?php
/*
Template Name: 会社概要ページ トップ
*/

/* some contents is returned */

```

2. Run wordpress with the aforementioned template

```graphql
query Reproduction {
  post(id: "0") {
    template {
      ... on Template_会社概要ページトップ {
        templateName
      }
    }
  }
}
```

#### Related code

https://github.com/wp-graphql/wp-graphql/blob/e37f2e06ded15647630a71c830d35c3f46f1d2b7/src/Type/InterfaceType/ContentTemplate.php#L39-L47

This code use `Template Name`. If use file name without extension, it will be repaired. But if file name was used non-ascii characters, there's nothing we can do about it 😇 

However, almost non-ascii character country(at least in japan), programmatic code's file name is usually used ascii characters.

#### Related issue

- https://github.com/wp-graphql/wp-graphql/issues/1658

### WPGraphQL Version

1.14.3

### WordPress Version

6.2

### PHP Version

8.2

_No response_

- [X] Yes

- [X] Yes

mecaota:
When we use non-ascii characters in all of template name, non-ascii characters type generated.

This PR is fix the issue by using file name.

fixes #2807, fixes #2492

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** docker

**WordPress Version:** 6.2


src/Type/InterfaceType/ContentTemplate.php
```diff
@@ -53,6 +53,10 @@ public static function register_content_template_types() {
 
 			if ( ! empty( $replaced_name ) ) {
 				$name = $replaced_name;
+			} else {
+				// If replaced_name is empty, use the file name
+				$file_name = explode( '.', $file )[0] ?? '';
+				$name      = preg_replace( '/[^\w]/', '', $file_name );
```
mecaota, line 59: The template file is prefixed with "page". Should I remove it?

jasonbahl:
@mecaota so if we change the Types that are registered, we also need to change how the Template resolves here: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Model/Post.php#L580-L584

We need to make sure we add some tests around this too. 

If you're not comfortable writing tests, it would be good to at least write detailed steps to reproduce the current issue + the behavior after the changes, then we can help get some tests written

jasonbahl:
I think it would actually be nice if we could add support for something like a `GraphQL Type Name` field on templates. 

like: 

```php
<?php
/**
 * Template Name: My Templaté
 * GraphQL Type Name: MyTemplate
 */
 ```

**Comment by mecaota** [comment]
But, I am busy at the moment and it will be a while before I can fix....
Would you like to make an issue or work on it instead?

**Comment by mecaota** [comment]
@justlevine OK, I'll leave the corrections in this PR ready for merging.

**Comment by mecaota** [comment]
@jasonbahl 

> We need to make sure we add some tests around this too.
If you're not comfortable writing tests, it would be good to at least write detailed steps to reproduce the current issue + the behavior after the changes, then we can help get some tests written

I'm sorry, I'm not familiar with PHP and I'm not familiar with writing integration tests, so I'd appreciate any advice.

Here are the steps I would like to test: I think it will be some sort of test to check if the generated schema is correct. This is because the error currently mentioned in each issue comes from an invalid schema.

1. Make template php file with non-ascii template name.
*1 If wordpress have a create template function, it use, but I can't find.
*2 This example use `カスタムテンプレート`、 it means Custom Template in Japanese, so it can be set other language like `自定義模板`, `맞춤 템플릿`, and so on.

like this `page-custom.php`
```page-custom.php
<?php
/*
Template Name: カスタムテンプレート
*/

?>

<section>
</section>

```

2. Run `WPGraphQL::get_schema()`, check this schema is valid.

odrowonz (approved):
It is worked for me (#2881).

mecaota:
@jasonbahl 

This is reproduce method→ https://github.com/wp-graphql/wp-graphql/pull/2808#issuecomment-1609030185

If I need to add a test for this, I think I'll add a test for non-ASCII character slugs to sample posts, categories, taxonomies, etc. However, this will cause many errors. I think it would be better to add the non-ASCII slug test after merging this.

Additionally, multiple similar issues have cropped up in the last few months since this PR was created, so it's clear that some people are in trouble. Please respond as soon as possible.

justlevine:
Rebased and linted to fix the phpcs issue

src/Type/InterfaceType/ContentTemplate.php
```diff
@@ -53,6 +53,10 @@ public static function register_content_template_types() {
 
 			if ( ! empty( $replaced_name ) ) {
 				$name = $replaced_name;
+			} else {
+				// If replaced_name is empty, use the file name
+				$file_name = explode( '.', $file )[0] ?? '';
```
justlevine, line 58: This is getting flagged (and rightly so) by PHPStan, because we no longer have access to `$file` once the previous loop ends.

justlevine (approved):
@jasonbahl I did a bit of cleanup, and then added the necessary tests in d5c72b6c7b392307437beadf4479b2f07ed54469.

Should be good to merge once you review

tests/_data/templates/custom-i18n.php
```diff
@@ -0,0 +1,6 @@
+<?php
```
jasonbahl, line 1: @justlevine let's add one more test where the file name also contains non-ascii characters. 

I believe in that case the schema will still error as the fallback will still be an invalid GraphQL name 🤷🏻‍♂️ 

If it is invalid by time we `register_graphql_object_type()`, let's `graphql_debug()` and `continue;` instead of registering it.

justlevine (changes requested):
https://github.com/wp-graphql/wp-graphql/pull/2808#discussion_r1350792576

(changing status so I dont forget)

tests/_data/templates/custom-i18n.php
justlevine: Just putting a `continue;` here punted the Fatal Error to querying for a bad template, so I extracted the formatting logic to a Utils method, and the `__typename` passed by the `Post` model now falls back to `DefaultTemplate` if the formatted name isnt compatible with GraphQL.

justlevine:
@jasonbahl I made the requested changes, this should be good to go 🤞

imjlk:
Well, the issue is definitely fixed for WordPress site using the GraphiQL IDE, but when working with the integration using various GraphQL clients(in my case, graphql-codegen features), I still encounter issues when the site locale is in non-ASCII format.
People can find the issue(#2492) and know about it, but wouldn't it be nice if there was a mention of it in the official docs?

jasonbahl:
@imjlk can you open another issue with steps to reproduce? If there's still a bug with locales, we should track it and work toward a fix (or at least document it)

I'd love to know more about what you're experiencing and how to reproduce consistently

PR #2821, opened 2023-05-23
feat: AbstractConnectionResolver:set_query_class() created

kidunot89:
- Adds `query_class` and `set_query_class()` to **AbstractConnectionResolver** class.
- Updates **Comment**, **Post**, **User**, and **Term** connection resolver classes to source the `$query_class`.
- Adds `testRegisteringConnectionsWithCustomQueryClasses` to the `ConnectionRegistrationTest` case 
- **WP_Query_Custom** added to `tests/_support/utils/class-wp-query-custom.php` for testing.

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -119,6 +119,13 @@ abstract class AbstractConnectionResolver {
 	protected $query_amount;
 
 	/**
+	 * The query class used to fetch the nodes.
+	 *
+	 * @var mixed
+	 */
+	protected $query_class;
+
+	/**
 	 * ConnectionResolver constructor.
 	 *
 	 * @param mixed       $source  source passed down from the resolve tree
@@ -304,6 +311,17 @@ abstract class AbstractConnectionResolver {
 		$this->query_args[ $key ] = $value;
 
 		return $this;
+	}
+
+	/**
+	 * Sets the connection Query class.
+	 *
+	 * @param string $class  Query class name.
+	 *
+	 * @return void
+	 */
+	public function set_query_class( $class ) {
+		$this->query_class = $class;
 	}
 
 	/**
--- a/src/Data/Connection/CommentConnectionResolver.php
+++ b/src/Data/Connection/CommentConnectionResolver.php
@@ -150,13 +150,18 @@ class CommentConnectionResolver extends AbstractConnectionResolver {
 	/**
 	 * Get_query
 	 *
-	 * Return the instance of the WP_Comment_Query
+	 * Return the instance of the Query class defaults to \WP_Comment_Query.
 	 *
-	 * @return \WP_Comment_Query
+	 * @return mixed
 	 * @throws \Exception
 	 */
 	public function get_query() {
-		return new WP_Comment_Query( $this->query_args );
+		// Get query class.
+		$queryClass = ! empty( $this->query_class )
+			? $this->query_class
+			: '\WP_Comment_Query';
+
+		return new $queryClass( $this->query_args );
 	}
 
 	/**
--- a/tests/wpunit/ConnectionRegistrationTest.php
+++ b/tests/wpunit/ConnectionRegistrationTest.php
@@ -1,5 +1,7 @@
 <?php
 
+use Test\WPGraphQL\Utils\WP_Query_Custom;
+
 class ConnectionRegistrationTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase {
 	public $connection_config;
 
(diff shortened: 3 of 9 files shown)


justlevine:
First off, love this! (@renatonascalves [just suggested something like this](https://wp-graphql.slack.com/archives/CDE8Y3MN3/p1684809738546109?thread_ts=1684072239.099649&cid=CDE8Y3MN3) so the timing is perfect!

Couple questions (for content 😬) before I look at the actual code:

>While it started as a fix for the queryClass connection registration argument which is broken atm of creation in WPGraphQL v1.14.3.

1. Can you be more specific about what's currently broken?

> There is some functionality being lost here
2. What specifically are we losing (and do you mean vs what's currently _working_ or what's currently _supposed to work_ but is anyway broken (per 1).

The last several attempts at isolated fixes/enhancements to the Connection Resolver all resulted in regressions around cases we don't currently have unit tests for. To try and avoid that from happening again, I've been working to improve the class holistically in #2749 from which we would then backport the individual changes in a nonbreaking manner.

My recommendation for this would be the same if it's just an enhancement and not a regression.

src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -306,6 +313,17 @@ public function set_query_arg( $key, $value ) {
...
 	}
 
+	/**
+	 * Sets the connection Query class.
+	 *
+	 * @param string $class  Query class name.
+	 *
+	 * @return void
+	 */
+	public function set_query_class( $class ) {
```
justlevine, line 323: I've got concerns/questions from a "holistic" POV - _nothing_ that follows should be considering blocking since the same issues exist with the current method of setting the query class via `WPConnectionType`:

I love the fact that this pattern makes local overloading in a resolver possible, and brings Comment/User connection queries the same extensibility patten, but I'm concerned about it's availability on `AbstractConnectionResolver`, when only 4 (including those in this PR) out of the bundled 13 Connection resolvers actually use a `$query_class`. 

Similarly, I'm concerned around the lack of checks around the provided `$query_class`. We dont want to enforce a `WP_Query` child, but we do need to make sure that the provided class string (`SWP_Query`, `WC_Query`, `Tribe_Query`, `FWP_Query`, etc etc) is actually compatible with the Connection resolver relying on it.

The latter issue can be solved with an overloadable `is_valid_query_class( string $class )`  that could for example check for `method_exists( $class, 'query' )`. 

We could even in theory use the same `is_valid_query_class()` to throw an exception if `! empy( $this->query_class )` on the Connection Resolvers that don't use it (would need to figure out backwards-compatibility), but I think we should first confirm that `AbstractConnectionResolver` is really the ideal home for this instead of adding what's _possibly_ tech debt to the majority of the included connection resolvers.

src/Data/Connection/CommentConnectionResolver.php
```diff
@@ -148,15 +148,18 @@ public function get_query_args() {
...
-	 * Return the instance of the WP_Comment_Query
-	 *
-	 * @return \WP_Comment_Query
-	 * @throws \Exception
+	 * @return mixed
 	 */
 	public function get_query() {
-		return new WP_Comment_Query( $this->query_args );
+		// Get query class.
+		$queryClass = ! empty( $this->query_class )
```
justlevine, line 158: Seeing this alone makes me way happier than it has any right to. 🙌🙌🙌

tests/wpunit/ConnectionRegistrationTest.php
```diff
@@ -677,4 +679,45 @@ protected function assertValidTypes( $actual ) : void {
 		$this->assertEquals( 'TestObject', $actual['data']['testConnection']['nodes'][0]['__typename'] );
 	}
 
+	public function testRegisteringConnectionsWithCustomQueryClasses() {
```
justlevine, line 682: We should also have tests for some common edge cases: bad class name, incompatible class, and using the wrong `WP_{$type}_Query` are the ones that jump to mind.

tests/wpunit/_bootstrap.php
```diff
@@ -0,0 +1,3 @@
+<?php
+
+require __DIR__ . '/../_support/Utils/class-wp-query-custom.php';
```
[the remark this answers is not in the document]
kidunot89: Tried too, and it didn't work. It was namespaced and everything. For some reason GraphQL wouldn't see the class and would throw so I eventually devolve into that in the bootstrap file, and I just removed all my PSR dev configuration.

src/Data/Connection/AbstractConnectionResolver.php
kidunot89: Actually it's more like 6 or 7 out of 13, because they extend one of the 4 connection resolvers and not the AbstractConnectionResolver class directly, Some of them don't even have a `get_query()` function and default to parent `get_query()` function in one of the 4. I agree there should be `class_exists( $this->query_class )` check somewhere and a `is_valid_query_class()` function is a good idea. I'll add it immediately.

justlevine: > because they extend one of the 4 connection resolvers and not the AbstractConnectionResolver class directly, Some of them don't even have a get_query() function and default to parent get_query() function in one of the 4.

Actual numbers aside ☝️is the crux of my concern: we're exposing a setter that semantically should apply on any connection but will be ignored by most of them - something that becomes less obvious from the code the better we are at DRYing out the implementing classes. But I don't have a good alternative 🤔

justlevine: Or breaking up AbstractConnectionResolver into traits, but I think either is bad DX in this particular situation since it extenders then need to know which intermediate class / trait is right for a particular WP type, but I don't see a way to have those classes/traits to use existing WP namings. 🤷‍♂️

kidunot89: Exactly, but the connection resolvers not using `$query_class` are less likely to be extended for usage. `EnqueuedScripts`, `EnqueuedStylesheet`, `Plugin`, `Taxonomy`, `Theme`, `UserRole`, and `ContentType`

justlevine:
> Couple questions (for context 😬) before I look at the actual code:
> 
> > While it started as a fix for the queryClass connection registration argument which is broken atm of creation in WPGraphQL v1.14.3.
> 
> 1. Can you be more specific about what's currently broken?
> 
> > There is some functionality being lost here
> 
> 2. What specifically are we losing (and do you mean vs what's currently _working_ or what's currently _supposed to work_ but is anyway broken (per 1).
> 

Summarized from a Slack Convo with @kidunot89 (correct me if I made a mistake):

1. In v1.14.3, the custom `queryClass` passed via `AppContext` is ignored. This is a *regression* (for which I believe there _is not_ an existing GH issue).
2. This PR doesn't address the regression itself, since using `AppContext` to overload the query type is "hacky".  Instead it provides an *enhancement*, offering a better way to overload the query class (from within the field resolver for the individual connection).

We should probably still address the actual regression in a separate PR.

With this in mind, I'm going to port this over to #2749 - even if we don't decide to wait on that PR merging this, I'm betting the effort will still unearth some code suggestions we can implement here independently.

justlevine:
I've updated #2749 to implement this pattern.
[This comment](https://github.com/wp-graphql/wp-graphql/pull/2749#issuecomment-1563078712) details the differences between this PR vs the holistic approach over there.

(If this PR goes forward independently, I highly recommend backporting at least some of the changes).

PR #2834, opened 2023-06-23
fix: improve tracking of query analyzer list types

jasonbahl:
This modifies the QueryAnalyzer to only track `list:$type_name` for RootConnections.

Nested connections will not tag the query with `list:$type_name` as it was leading to over-purging. 

see: https://github.com/wp-graphql/wp-graphql-smart-cache/issues/225

related: https://github.com/wp-graphql/wp-graphql-smart-cache/issues/225

Any other comments?
-------------------

This should significantly reduce the amount of queries that are purged from common events such as publishing posts.

## Before

Here we can see that a query for a Category and it's related posts will return a `list:post` key, meaning it will invalidate anytime a new post is published. 

CleanShot 2023-06-23 at 13 12 00

## After

Here we do not see the `list:post` key, meaning that this query will not invalidate whenever a new post is published. 

However, if a new post is published and associated with the "child" category, then this query will be purged because that will call purge on the ID of the child category.

CleanShot 2023-06-23 at 13 11 38

### DIFF
--- a/src/Utils/QueryAnalyzer.php
+++ b/src/Utils/QueryAnalyzer.php
@@ -7,6 +7,7 @@ use GraphQL\Error\SyntaxError;
 use GraphQL\Language\Parser;
 use GraphQL\Language\Visitor;
 use GraphQL\Server\OperationParams;
+use GraphQL\Type\Definition\FieldDefinition;
 use GraphQL\Type\Definition\InterfaceType;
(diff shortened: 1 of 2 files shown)


jasonbahl:
NOTE: The Schema Linter is failing because the 1.14.4 release failed to upload the schema because the mariadb version in the workflow was having issues. The workflow has now been pinned and the next release should fix it.

PR #2840, opened 2023-06-29
fix: adjust validation of createMediaItem mutation

jasonbahl:
This adds additional validation to the `createMediaItem` mutation to prevent invalid input to the `filePath` argument.

### DIFF
--- a/src/Mutation/MediaItemCreate.php
+++ b/src/Mutation/MediaItemCreate.php
@@ -159,7 +159,7 @@ class MediaItemCreate {
 			$check_file = wp_check_filetype( $uploaded_file_url );
 
 			// if the file doesn't pass the check, throw an error
-			if ( ! $check_file['ext'] || ! $check_file['type'] ) {
+			if ( ! $check_file['ext'] || ! $check_file['type'] || ! wp_http_validate_url( $uploaded_file_url ) ) {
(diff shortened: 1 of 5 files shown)


src/Mutation/MediaItemCreate.php
```diff
@@ -151,8 +151,26 @@ public static function mutate_and_get_payload() {
...
+			$check_file = wp_check_filetype( $uploaded_file_url );
+
+			// if the file doesn't pass the check, throw an error
+			if ( ! $check_file['ext'] || ! $check_file['type'] || ! wp_http_validate_url( $uploaded_file_url ) ) {
+				throw new UserError( sprintf( __( 'Invalid filePath "%s"', 'wp-graphql' ), $input['filePath'] ) );
+			}
+
+			// prevent the filePath from being submitted with a non-allowed protocols
+			$allowed_protocols = [ 'https', 'http', 'file' ];
+			$allowed_protocols = apply_filters( 'graphql_media_item_create_allowed_protocols', $allowed_protocols );
```
justlevine, line 168: Would it make sense to initialize the current `$protocol` earlier and pass it through the filter?

Pros:
- make it easier to support more complex validation without a dedicated action/filter.

Cons:
- make it easier to ~support more complex validation~ _introduce unwanted edge cases_  without a dedicated action/filter

jasonbahl: yes, good call. Will also pass `$input`, `$context` and `$info`.  🙌🏻

PR #2865, opened 2023-07-24
fix: user roles should return empty 

j3ang:
When my user does not have any roles assigned.  wp-graphql is returning all existing roles instead of null or empty roles.
…

…

CleanShot 2023-07-24 at 15 43 49

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** aws lightsail bitnami wordpress

**WordPress Version:** 6.2

### DIFF
--- a/src/Type/ObjectType/User.php
+++ b/src/Type/ObjectType/User.php
@@ -67,6 +67,9 @@ class User {
 
 							if ( isset( $user->roles ) && ! empty( $user->roles ) ) {
 								$resolver->set_query_arg( 'slugIn', $user->roles );
+							} else {
+								// If the user has no roles, we don't want to return any roles
+								$resolver->set_query_arg('slugIn', []);
 							}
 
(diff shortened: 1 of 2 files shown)


jasonbahl:
@j3ang would you be willing to write a test for this to ensure we don't have regressions on this in the future? 

Something similar to the tests here: https://github.com/wp-graphql/wp-graphql/blob/develop/tests/wpunit/UserRoleConnectionQueriesTest.php

we'd want a test that: 

- Creates a user with no rule
- Executes the query you reported this issue with
- Asserts that no roles are returned for that user

The test should fail on current latest WPGraphQL and pass with your change in place.

jasonbahl:
@j3ang thanks for adding the test 🙌🏻

jasonbahl:
I made some modifications locally, but Git isn't letting me push to your branch, so I will merge this and follow-up with the changes

Issue #2644, opened 2022-12-01
Cursor pagination on queries sorted by custom field broken


ndigenpcc:
### Description

I am currently making queries from my NextJS frontend using Apollo client and queries that include sorting by an ACF field return the first page fine, however when I use fetchMore from Apollos useLazyQuery, the next page is empty. Then, when using fetchMore to go back to the previous page, it is a totally different one from the initial load.

### Steps to reproduce
1. Query partners sorting by PARTNER_TIER ex.

```
where = {...where, orderby: [{order: "DESC", field: "PARTNER_TIER"}, {order: "ASC", field: "TITLE"}]};
```

```
const GET_POSTS = gql`
query NewQuery($where: RootQueryToPartnerConnectionWhereArgs = {}) {
  partners(first: 10, where: $where) {
    nodes {
      partnerDetails {
        certifiedTrainingPartner
        childCompanyName
        contactEmail
        description
        discountAssistance
        fieldGroupName
        ordersInterface
        partnerCategory
        partnerLogo {
          altText
          sourceUrl
        }
        partnerNsInternalid
        partnerSolutionCategory
        partnerTier
        partnerWebsite
        stateProvince
      }
      id
      title
      uri
      slug
    }
    pageInfo {
      hasNextPage
      hasPreviousPage
      endCursor
      startCursor
    }
  }
}
`;
```

2. Use apollo client fetchMore to get the next page of partners, it will be blank.

3. Use apollo client fetchMore to get previous page, it will not be the original partners.

### Additional context

Correct results on first load.
<img width="1792" alt="correct initial load" src="https://user-images.githubusercontent.com/110477935/205137326-8eff7848-6fce-4450-9189-34af905c33f3.png">

Empty page after clicking next button
<img width="1788" alt="next page" src="https://user-images.githubusercontent.com/110477935/205137443-7e649ccf-482d-4861-96ad-e8e16b89bed3.png">

Trying to return to first page using previous button
<img width="1683" alt="prev page" src="https://user-images.githubusercontent.com/110477935/205137481-6f999544-1f7a-47fd-b0eb-f84bc701180d.png">

### WPGraphQL Version

1.13.4

### WordPress Version

6.0.3

### PHP Version

8.0.25

### Additional enviornment details

NextJS frontend using Apollo client

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

### Please confirm that you have disabled ALL plugins except for WPGraphQL.

- [X] Yes
[...]

PR #2874, opened 2023-07-27
fix: PostObjectCursor's support for meta fields improved

kidunot89:
- Fixes support for sorting post connection results by multiple meta.

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Data/Cursor/PostObjectCursor.php
+++ b/src/Data/Cursor/PostObjectCursor.php
@@ -169,13 +169,13 @@ class PostObjectCursor extends AbstractCursor {
 				break;
 		}
 
-		$value = $this->cursor_node->{$by} ?? null;
+		$value     = $this->cursor_node->{$by} ?? null;
+		$cursor_id = $this->cursor_node->ID ?? null;
 
 		/**
 		 * Compare by the post field if the key matches a value
 		 */
-		if ( ! empty( $value ) && ! metadata_exists( 'post', $this->cursor_node->ID, $by ) ) {
-
+		if ( ! empty( $value ) && ! empty( $cursor_id ) && ! metadata_exists( 'post', $cursor_id, $by ) ) {
 			$this->builder->add_field( "{$this->wpdb->posts}.{$by}", $value, null, $order );
 
 			return;
@@ -224,7 +224,7 @@ class PostObjectCursor extends AbstractCursor {
 			}
 
 			$meta_keys = array_column( $meta_query, 'key' );
-			$index = array_search( $meta_key, $meta_keys, true );
+			$index     = array_search( $meta_key, $meta_keys, true );
 
 			if ( 1 < count( $meta_query ) && false !== $index ) {
 				$key = "mt{$index}.meta_value";
--- a/src/Data/Cursor/PostObjectCursor.php
+++ b/src/Data/Cursor/PostObjectCursor.php
@@ -174,7 +174,7 @@ class PostObjectCursor extends AbstractCursor {
 		/**
 		 * Compare by the post field if the key matches a value
 		 */
-		if ( ! empty( $value ) ) {
+		if ( ! empty( $value ) && ! metadata_exists( 'post', $this->cursor_node->ID, $by ) ) {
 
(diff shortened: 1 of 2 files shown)


src/Data/Cursor/PostObjectCursor.php
```diff
@@ -169,13 +169,13 @@ private function compare_with( $by, $order ) {
...
-		$value = $this->cursor_node->{$by} ?? null;
+		$value     = $this->cursor_node->{$by} ?? null;
+		$cursor_id = $this->cursor_node->ID ?? null;
 
 		/**
 		 * Compare by the post field if the key matches a value
 		 */
-		if ( ! empty( $value ) ) {
-
+		if ( ! empty( $value ) && ! empty( $cursor_id ) && ! metadata_exists( 'post', $cursor_id, $by ) ) {
```
justlevine, line 178: I might be wrong but I think `metadata_exists()` adds an extra db query.
If so perhaps there's a way to break down this conditional so it's only called when necessary?

tests/wpunit/PostObjectCursorTest.php
```diff
@@ -16,7 +16,68 @@ public function setUp(): void {
 		$this->admin            = $this->factory()->user->create( [
 			'role' => 'administrator',
 		] );
+		$this->start_count      = 100;
+		$this->start_time       = time();
 		$this->created_post_ids = $this->create_posts();
+
+		register_graphql_fields(
```
justlevine, line 23: We shouldn't be doing this on all the tests, just `testPostOrderingByMultipleMeta()`

tests/wpunit/PostObjectCursorTest.php
```diff
@@ -16,7 +16,68 @@ public function setUp(): void {
...
+				'testMetaNumber' => [
+					'type'    => 'Number',
+					'resolve' => function( $source ) {
+						return get_post_meta( $source->ID, 'test_meta_number', true );
+					}
+				],
+			]
+		);
+
+		add_filter(
```
justlevine, line 41: If you `add_filter` in `setUp()`, you need to `remove_filter` in `tearDown()` to prevent side effects in later tests.

justlevine (commented):
Code change itself looks good to me (although I'm worried about a perf hit on non-custom meta). Wondering if maybe this deserves it's own test class instead of shoehorning into the existing post cursor test. 

See comments for details.

tests/wpunit/PostObjectCursorTest.php
kidunot89: Either way, I moved it

tests/wpunit/PostObjectCursorTest.php
kidunot89: Moved it.

src/Data/Cursor/PostObjectCursor.php
kidunot89: It should not, because the `cursor_node` is already cached with it's meta. https://github.com/WordPress/wordpress-develop/blob/6.2/src/wp-includes/meta.php#L751

justlevine (approved):

jasonbahl (approved):

justlevine:
> There's still some issues re: meta queries and pagination that this doesn't fully solve, for example, if multiple posts have the same meta value and one is the `endCursor` the next page will skip the other items with the same value.
> 
> This helps unblock a certain set of users and we can circle back on the remaining bugs later.

Is there an existing issue for this? That's a complex edge case, wanna make sure we dont lose it.

kidunot89:
@justlevine There is, using certain meta query structure when ordering by a meta field don't get processed correctly by the `PostObjectCursor` and the same in the `TermObjectCursor` class too. This PR fixes the issue for WP_Query instances with a `meta_query` field with at least two entries that aren't `relation`.

justlevine:
Okay, my misunderstanding - the remaining edge cases that are unresolved are _in_ #2644, not in addition to that GH issue. 👍

PR #2882, opened 2023-08-04
fix: Config and Cursor classes refactor.

### LINKED ISSUE #732: Sticky Posts are repeated on every page
When you have one ore more sticky posts they are repeated on every page in the pagination.

Reproduction is simple:

- have more posts than your `first` setting in the query
- set one (or more) posts to stick to the homepage
- query the posts by having after set to the endCursor of the first (or any subsequent) page

In this example post `18339` is set to sticky.

Query 1:

```graphql
query POSTS_WITH_STICKY {
  posts(first: 5, after:null) {
    pageInfo {
      hasNextPage
      endCursor
    }
    nodes {
      postId
    }
  }
}
```

Result:
```json
{
  "data": {
    "posts": {
      "pageInfo": {
        "hasNextPage": true,
        "endCursor": "YXJyYXljb25uZWN0aW9uOjE4MzAx"
      },
      "nodes": [
        {
          "postId": 18339
        },
        {
          "postId": 18741
        },
        {
          "postId": 18745
        },
        {
          "postId": 18744
        },
        {
          "postId": 18301
        }
      ]
    }
  }
}
```

Query for second page: 

```graphql
query POSTS_WITH_STICKY {
  posts(first: 5, after:"YXJyYXljb25uZWN0aW9uOjE4MzAx") {
    pageInfo {
      hasNextPage
      endCursor
    }
    nodes {
      postId
    }
  }
}
```

Result: 

```json
{
  "data": {
    "posts": {
      "pageInfo": {
        "hasNextPage": true,
        "endCursor": "YXJyYXljb25uZWN0aW9uOjE3NzMw"
      },
      "nodes": [
        {
          "postId": 18339
        },
        {
          "postId": 18030
        },
        {
          "postId": 18282
        },
        {
          "postId": 18082
        },
        {
          "postId": 17730
        }
      ]
    }
  }
}
```

The sticky post `18339` is repeated on the second page and every subsequent page as well.

edit: I pasted the wrong result for query 2, fixed now.

### LINKED ISSUE #2894: Allow GraphQL Cursor Pagination Modifications to be skipped by resolvers

Currently WPGraphQL filters the SQL "where" clauses on WP_Query, WP_Comment_Query, WP_Users_Query and WP_Terms_Query to add support for cursor pagination. 

These filters are applied in [/src/Data/Config.php ](https://github.com/wp-graphql/wp-graphql/blob/v1.14.10/src/Data/Config.php) and are currently applied to all queries that are executed as a graphql request. 

Individual resolvers cannot opt-out of this behavior, nor can plugins override/customize this behavior. 

As a plugin/theme developer creating custom solutions, I would like to be able to override this behavior globally and/or for individual resolvers.

For plugins that use custom tables, etc, it's critical to have the ability to override the default behavior WPGraphQL executes for cursor support.

## Pre filters

I believe we should add some `pre_*` filters to the filter callbacks in Config.php, for example within this callback: https://github.com/wp-graphql/wp-graphql/blob/v1.14.10/src/Data/Config.php#L198

Currently this callbacks return early if the request is not a GraphQL request, but then it proceeds to execute the default logic always with no way to tell WPGraphQL to skip / alter the behavior.

The `pre_*` filter(s) would allow 3rd party plugin/theme code to override the default behavior from a central location. We could follow the pre_ pattern we have here: https://github.com/wp-graphql/wp-graphql/blob/ad5a4e97f9d9701c55917cbec99d08c215e38b10/src/Model/Model.php#L239

If we add pre_ filters at the beginning of the function, 3rd party plugin/theme code could determine if certain criteria has been met and could customize the behavior instead of letting the default behavior execute. 

## Custom Query Arg

Similar to how the callbacks return early if the request is not a GraphQL Request, we could add an additional default check for a specific query arg. 

For example, if (name tbd) `$query->query_vars['graphql_apply_cursor_pagination'] === false`, we could bail early. 

By default, this argument would be set to true for connection resolvers (i.e. somewhere in here: https://github.com/wp-graphql/wp-graphql/blob/v1.14.10/src/Data/Connection/PostObjectConnectionResolver.php#L173)

This would allow individual resolvers to do something like (pseudo): 

```php
$resolver = new PostObjectConnectionResolver();
$resolver->set_query_arg( 'graphql_apply_cursor_pagination', false );
```

Resulting in individual resolvers being able to opt-out of the default cursor pagination logic at the resolver level. 

_No response_

### Additional Context

@kidunot89 is trying to fix some issues between WPGraphQL and WPGraphQL for The Events Calendar which has led to the need to customize how WPGraphQL handles cursor support, as TEC uses some custom SQL stuff themselves so the default behavior from WPGraphQL doesn't always jive with what TEC is doing.

kidunot89:
- Refactors and support for custom unique threshold fields all ***Cursor** classes, to be utilized thru hooks or respective query vars.
- Fixes numerous result sorting bugs in all ***Cursor** classes.
- Adds stricter conditionals to **Config** class callback to ensure non-connection defined query are not affected by **Cursor** classes.
- Adds pre-filter hooks to ***Cursor** classes and **Config** class.  
- Adds the `graphql_post_object_cursor_meta_key` for targeting fields in custom tables.

Here are a few example to get a better picture of how it works and how to best utilize it.

The following is a simple use-case to create a post connection that's ordering defaults to the slug column instead of the date column when no `orderby` is provided.

```php
// Register new posts connection.
register_graphql_connection(
	[
		'fromType'       => 'RootQuery',
		'toType'         => 'Post',
		'fromFieldName'  => 'postsOrderedBySlug',
		'resolve'        => function ( $source, $args, $context, $info ) {
			global $wpdb;
			$resolver = new PostObjectConnectionResolver( $source, $args, $context, $info, 'post' );

			// Get cursor node
			$cursor  = $args['after'] ?? null;
			$cursor  = $cursor ?: ( $args['before'] ?? null );
			$post_id = substr( base64_decode( $cursor ), strlen( 'arrayconnection:' ) );
			$post    = get_post( $post_id );

			// Get order.
			$order   = ! empty( $args['last'] ) ? 'ASC' : 'DESC';

			$resolver->set_query_arg(
				'graphql_cursor_threshold_fields',
				[
					[
						'key'   => "{$wpdb->posts}.post_name",
						'value' => null !== $post ? $post->post_name : null,
						'type'  => 'CHAR',
						'order' => $order,
					],
				]
			);

			// Set default ordering.
			if ( empty( $args['where']['orderby'] ) ) {
				$resolver->set_query_arg( 'orderby', 'post_name' );
			}

			if ( empty( $args['where']['order'] ) ) {
				$resolver->set_query_arg( 'order', $order );
			}

		}
	]
);
```

Here is a slighty more complex example by utilizing the [get_query_args](https://github.com/the-events-calendar/ql-events/blob/fix/events-query-fix/includes/data/connection/class-event-connection-resolver.php#L77-L294)() in the Event_Connection_Resolver from an WIP version of QL-Events.  The query vars used in this example:

- `custom_id_key` is provided. This is to provided a different SQL index field, so when other plugins that alter WP's SQL statement with to the point a new index is needed. Utilizing the `graphql_cursor_id_key` we can provided the table column or alias the SQL statement to the new index. 
- `graphql_cursor_id_value` works much the same way but provides the cursor node's value of index field, that why it's in the conditional `( false !== $cursor_offset )`. 
- The  `graphql_cursor_compare_by_{$by}_(key|value)` are exactly like the last two vars except the don't target the index field, but the custom ordering fields that `Cursor` class doesn't know about or can't retrieve the value for with a simple `field` name, e.g. `$this->cursor_node->{$by}`.

```php
// set the graphql cursor args
$query_args['graphql_cursor_compare']          = ( ! empty( $last ) ) ? '>' : '<';
$query_args['graphql_after_cursor']            = $this->get_after_offset();
$query_args['graphql_before_cursor']           = $this->get_before_offset();
$query_args['graphql_cursor_id_key']           = Occurrences::table_name() . '.occurrence_id';

if ( false !== $cursor_offset ) {
	/**
	 * If the starting offset is not 0 sticky posts will not be queried as the automatic checks in wp-query don't
	 * trigger due to the page parameter not being set in the query_vars, fixes #732
	 */
	$query_args['ignore_sticky_posts'] = true;

	// Configure post object cursor.
	$cursor_id = $cursor_offset;
	if ( class_exists( '\TEC\Events\Custom_Tables\V1\Models\Occurrence' ) ) {
		$cursor_id = \TEC\Events\Custom_Tables\V1\Models\Occurrence::normalize_id( $cursor_offset );
	}

	$query_args['graphql_cursor_id_value'] = $cursor_id;

	$cursor_node = tribe_get_event( $cursor_offset );
	$query_args['graphql_cursor_threshold_fields'] = [
		[
			'key'   => Occurrences::table_name() . '.start_date_utc',
			'value' => $cursor_node->start_date_utc,
			'type'  => 'DATETIME',
		]
	];

	$query_args['graphql_cursor_compare_by__EventStartDateUTC_key']   = Occurrences::table_name() . '.start_date_utc';
	$query_args['graphql_cursor_compare_by__EventStartDateUTC_value'] = $cursor_node->start_date_utc;
	$query_args['graphql_cursor_compare_by__EventEndDateUTC_key']     = Occurrences::table_name() . '.end_date_utc';
	$query_args['graphql_cursor_compare_by__EventEndDateUTC_value']   = $cursor_node->end_date_utc;
}
```

Closes #2894 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …


jasonbahl (approved):

jasonbahl:
I'm afk this weekend but looks good from my phone

src/Data/Cursor/PostObjectCursor.php
```diff
@@ -230,6 +230,17 @@ private function compare_with_meta_field( string $meta_key, string $order ) {
...
+		/**
+		 * Allow filtering the meta key used for cursor based pagination
+		 *
+		 * @param string $key       The meta key to use for cursor based pagination
+		 * @param string $meta_key  The original meta key
+		 * @param string $meta_type The meta type
+		 * @param string $order     The order direction
+		 * @param object $cursor    The PostObjectCursor instance
+		 */
+		$key = apply_filters( 'graphql_post_object_cursor_meta_key', $key, $meta_key, $meta_type, $order, $this );
```
jasonbahl, line 243: @kidunot89 can you use this existing filter? https://github.com/wp-graphql/wp-graphql/blob/develop/src/Data/Cursor/CursorBuilder.php#L58-L68

I _believe_ you should be able to accomplish the same thing and we wouldn't have to add new filters for each Cursor class.

justlevine: +1 for keeping this generic.

Not sure the original `meta_key` has a strong use case this late in the lifecycle, or at least not one that that justifies a bad signature (appending the param after the instance) or a breaking change.

If there is something I'm missing then a _long term_ goal would be to place the filter in the `AbstractCursor`, so any custom or extended cursors would maintain interoperability without tech debt, but I don't see an easy nonbreaking path so I'm down with duplication as a short term solve and adding a refactor to my evergrowing 2.0 wishlist

jasonbahl (changes requested):
see: https://github.com/wp-graphql/wp-graphql/pull/2882/files#r1287324943

src/Data/Config.php
```diff
@@ -165,24 +165,39 @@ static function ( $query ) {
...
+		if ( null !== $pre_orderby ) {
+			return $pre_orderby;
+		}
 
-		// If the cursor "graphql_cursor_compare" arg is not in the query,
-		// default to using ID DESC as the stabilizer
-		if ( ! isset( $wp_query->query['graphql_cursor_compare'] ) ) {
-			return "{$orderby}, {$wpdb->posts}.ID DESC ";
+		// Bail early if disabled by connection.
+		if ( isset( $wp_query->query_vars['graphql_apply_cursor_pagination_orderby'] )
```
justlevine, line 180: Is the change here and everywhere else from `$wp_query->query` to `$wp_query->query_vars` intentional?

This _seems_ like a good change, I just want to make sure there isn't some unintended consequence downstream I'm not thinking of.

src/Data/Config.php
```diff
@@ -165,24 +165,39 @@ static function ( $query ) {
...
+		}
+
+		global $wpdb;
+
 		// Check the cursor compare order
-		$order = '>' === $wp_query->query['graphql_cursor_compare'] ? 'ASC' : 'DESC';
+		$order = '>' === $wp_query->query_vars['graphql_cursor_compare'] ? 'ASC' : 'DESC';
+
+		// Get Cursor ID key.
+		$cursor = new PostObjectCursor( $wp_query->query_vars );
```
justlevine, line 196: Can we make the cursor class filterable instead of hard coding as `PostObjectCursor`?

src/Data/Config.php
```diff
@@ -196,19 +211,30 @@ public function graphql_wp_query_cursor_pagination_stability( string $orderby, W
...
 	public function graphql_wp_query_cursor_pagination_support( string $where, WP_Query $query ) {
-
-		// if it's not a graphql request, return early
+		// Bail early if it's not a GraphQL Request.
 		if ( true !== is_graphql_request() ) {
 			return $where;
 		}
 
-		// apply the after cursor to the query
+		// If pre-filter hooked, return $pre_where.
```
justlevine, line 219: Let's fully doc-block these (and any other ) filter comments + params.

src/Data/Config.php
justlevine: Also in `*_cursor_pagination_support()`. Otherwise this feels like a half measure to me.

src/Data/Config.php
```diff
@@ -165,24 +165,39 @@ static function ( $query ) {
...
 	 */
 	public function graphql_wp_query_cursor_pagination_stability( string $orderby, WP_Query $wp_query ) {
-
+		// Bail early if it's not a GraphQL Request.
 		if ( true !== is_graphql_request() ) {
 			return $orderby;
 		}
 
-		global $wpdb;
+		// If pre-filter hooked, return $pre_orderby.
```
justlevine, line 173: We should fully doc-block all these filters.

src/Data/Config.php
justlevine, line 180: Is the change from `$query->query` to `$query->query_vars` intentional.

(This seems like a valid change to me, just wondering about any downstream breaks).

src/Data/Config.php
justlevine, line 196: Do we want to filter all these cursor classes before we call them, so devs can use their own?

src/Data/Cursor/AbstractCursor.php
```diff
@@ -55,13 +56,22 @@ abstract class AbstractCursor {
...
+
 	/**
 	 * The constructor
 	 *
-	 * @param array $query_vars
-	 * @param string|null $cursor
+	 * @param array       $query_vars         Query variable for the query to be executed.
+	 * @param string      $id_key             .
+	 * @param string|null $cursor             Cursor type. Either 'after' or 'before'.
+	 * @param array       $initial_threshold
```
justlevine, line 72: Add a comment explaining what this is please 🙏

src/Data/Cursor/AbstractCursor.php
```diff
@@ -134,6 +144,99 @@ protected function is_valid_offset_and_node() {
...
+	 * error because query execution is guaranteed to fail.
+	 *
+	 * @param array $field  Threshold configuration.
+	 * 
+	 * @throws InvariationViolation Invalid configuration format.
+	 * 
+	 * @return void
+	 */
+	protected function is_valid_threshold_field( $field ) {
+		$class = get_called_class();
```
justlevine, line 158: We only need the $class for the errors it seems. Let's `sprintf() it into the error instead of calling it here.

src/Data/Cursor/AbstractCursor.php
```diff
@@ -134,6 +144,99 @@ protected function is_valid_offset_and_node() {
...
+		// Throw if an array not provided.
+		if ( ! is_array( $field ) ) {
+			$type  = gettype( $field );
+			throw new InvariantViolation( __( "Invalid value provided for {$class} threshold field. Expected Array, ${type} given.", 'ql-events' ) );
```
justlevine, line 162: (here and below), your translation namespace is leaking. Should be `wp-graphql`

justlevine (changes requested):
Did a quick/partial review from my phone. Will look more in depth over the weekend.

composer.json
```diff
@@ -45,7 +45,8 @@
 		"simpod/php-coveralls-mirror": "^3.0",
 		"slevomat/coding-standard": "^8.9",
 		"szepeviktor/phpstan-wordpress": "~1.3.0",
-		"wp-graphql/wp-graphql-testcase": "~2.3"
+		"wp-graphql/wp-graphql-testcase": "~2.3",
+		"wp-cli/wp-cli-bundle": "^2.8"
```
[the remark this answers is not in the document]
kidunot89: I typically add for testing then remove it afterwards, just forgot to remove it this time. I wasn't sure if anyone else needed it besides me.

src/Data/Config.php
kidunot89: I figured this is the right change, because this is never not set for GraphQL connections.

composer.json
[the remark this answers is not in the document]
justlevine: No its a known bug with test env (might not have an issue but its come up in a few prs). If you have wp-cli installed globally, then `bash bin/install-test-env.sh` works as a workaround.

src/Data/Config.php
kidunot89: No cuz I don't encourage that behavior, the preferred behaviors for them to hook into filters in the classes and alter the behavior that way. And there's a good chance that if they filter and replace one instance of the class in the `orderby` callback but not the `where` callback for the same type, that would likely result in a fatal error.

src/Data/Cursor/AbstractCursor.php
kidunot89: 🤦🏿‍♂️Crap I didn't it. I'll put it back.

src/Data/Cursor/AbstractCursor.php
kidunot89: 🤷🏿‍♂️

src/Data/Config.php
justlevine: > And there's a good chance that if they filter and replace one instance of the class in the orderby callback but not the where callback for the same type, that would likely result in a fatal error.

We'd just use the same filter in both places. But I hear the desire to lock this down.

src/Data/Config.php
```diff
@@ -281,23 +377,49 @@ public function graphql_wp_user_query_cursor_pagination_support( $where, \WP_Use
...
 
-		// apply the after cursor
+		// Apply the after cursor, moving forward through results.
 		if ( ! empty( $args['graphql_after_cursor'] ) ) {
 			$after_cursor    = new TermObjectCursor( $args, 'after' );
 			$pieces['where'] = $pieces['where'] . $after_cursor->get_where();
 		}
 
-		// apply the before cursor
+		// Apply the after cursor, moving backward through results.
```
justlevine, line 422:
```suggestion
		// Apply the before cursor, moving backward through results.
```

src/Data/Config.php
```diff
@@ -226,15 +269,48 @@ public function graphql_wp_query_cursor_pagination_support( string $where, WP_Qu
...
 
-			// phpcs:ignore
-			return "{$orderby}, {$wpdb->users}.ID DESC ";
+		// Bail early if the cursor "graphql_cursor_compare" arg is not in the query,
+		if ( ! isset( $query->query_vars['graphql_cursor_compare'] ) ) {
+			return $orderby;
 		}
 
-		return $orderby;
+		global $wpdb;
```
justlevine, line 304: This one too

src/Data/Config.php
```diff
@@ -317,23 +439,45 @@ public function graphql_wp_term_query_cursor_pagination_support( array $pieces,
...
 			$pieces['where'] .= $after_cursor->get_where();
 		}
 
+		// Apply the after cursor, moving backward through results.
 		if ( ! empty( $query->query_vars['graphql_before_cursor'] ) ) {
 			$before_cursor    = new CommentObjectCursor( $query->query_vars, 'before' );
 			$pieces['where'] .= $before_cursor->get_where();
 		}
 
+		\codecept_debug( compact( 'pieces' ) );
```
justlevine, line 479: This leaked in

src/Data/Cursor/AbstractCursor.php
```diff
@@ -99,7 +108,7 @@ public function __construct( $query_vars, $cursor = 'after' ) {
 	 * @return mixed|null
 	 */
 	public function get_query_var( string $name ) {
-		return ! empty( $this->query_vars[ $name ] ) ? $this->query_vars[ $name ] : null;
+		return isset( $this->query_vars[ $name ] ) ? $this->query_vars[ $name ] : null;
```
justlevine, line 111: Confirming the intended change in behavior here, since this doesn't just return `false`, but empty strings and arrays. Alternatively we can use a string 'skip' value instead of an explicit 'false'.

Not sure what it is offhand, but we should follow whatever WP_Query does internally.

src/Data/Config.php
```diff
@@ -196,19 +220,38 @@ public function graphql_wp_query_cursor_pagination_stability( string $orderby, W
...
+		// Bail early if it's not a GraphQL Request.
 		if ( true !== is_graphql_request() ) {
 			return $where;
 		}
 
-		// apply the after cursor to the query
+		/**
+		 * If pre-filter hooked, return $pre_where.
+		 * 
+		 * @param null|mixed $pre_where The pre-filtered WHERE clause of the query.
```
justlevine, line 231:
```suggestion
		 * @param null|string $pre_where The pre-filtered WHERE clause of the query.
```

This also needs to be changed on the `@return` tag, and on the other filters.

justlevine (changes requested):
Reviewed via my phone (so just looked at the diff, and didn't inspect the actual conditional logic) and flagged a few insignificant code smells.

One thing more generally: what exactly _is_ a threshold field? Afaik this isn't common terminology, and since the cursor stuff is already complex, we might want to be a bit more specific (vs self referencing 'threshold fields') in the doc blocks and comments.

src/Data/Cursor/AbstractCursor.php
kidunot89: This adds support for `[]` as a value

kidunot89:
@justlevine The threshold fields are one or more fields that act as unique key and boundary for the cursor SQL statement

src/Data/Cursor/AbstractCursor.php
kidunot89: @justlevine So `''`, `[]`, and `false` aren't valid values for SQL statements? Or just `''`

kidunot89: @justlevine @jasonbahl Changes made 👍🏿

justlevine: My comment here is not about SQL statements, it's what WordPress/WPGraphQL considers as a set `query_var`, and what our code is expecting from `AbstractCursor::get_query_var()`.

- Before this change, WPGraphQL would treat _any_ falsey value as if it wasn't set.
- After this change, `''`, `[]` and `false`, are all returned instead of `null`. This is a change in behavior that _may_ be breaking (eg, `if ( isset( $cursor->get_query_var() )` now returns true).
I'm calling this out to ensure someone reviews all the uses of `AbstractCursor::get_query_var()` to confirm none of those possible return values cause the code to behave unexpectedly.
- WordPress core [treats an `''` value the same as if it's unset](https://github.com/WordPress/wordpress-develop/blob/6.3/src/wp-includes/class-wp-query.php#L1833-L1839), so at the bare minimum we should at least continue to return `null` in that case. I.e. `isset( $this->query_vars[ $name ] ) && $this->query_vars[ $name ] !== '' ) ? $this->query_vars[ $name ] : null`

kidunot89: Understood. The proper changes have been made. 👍🏿

tests/wpunit/CommentObjectCursorTest.php
```diff
@@ -3,6 +3,9 @@
 
 class CommentObjectCursorTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase {
     public function testThresholdFieldsQueryVar() {
+        // Skip until bugs in CommentObjectCursor class are resolved.
```
justlevine, line 6: What bugs? Do they have existing tickets?

kidunot89: It's an ordering bug and no I haven't written a ticket on it yet.

src/Data/Config.php
```diff
@@ -162,28 +162,50 @@ static function ( $query ) {
...
 
-		global $wpdb;
+		/**
+		 * If pre-filter hooked, return $pre_orderby.
+		 * 
+		 * @param null|mixed $pre_orderby The pre-filtered ORDER BY clause of the query.
+		 * @param string     $orderby     The ORDER BY clause of the query.
+		 * @param \WP_Query  $query       The WP_Query instance (passed by reference).
+		 * 
+		 * @return null|mixed
```
justlevine, line 182:
```suggestion
		 * @param null|string $pre_orderby The pre-filtered ORDER BY clause of the query.
		 * @param string      $orderby     The ORDER BY clause of the query.
		 * @param \WP_Query   $query       The WP_Query instance (passed by reference).
		 * 
		 * @return null|string
```

src/Data/Cursor/AbstractCursor.php
```diff
@@ -132,6 +144,153 @@ protected function is_valid_offset_and_node() {
...
+	 * Validates threshold field configuration. Validation failure results in a fatal
+	 * error because query execution is guaranteed to fail.
+	 *
+	 * @param array $field  Threshold configuration.
+	 * 
+	 * @throws \GraphQL\Error\InvariantViolation Invalid configuration format.
+	 * 
+	 * @return void
+	 */
+	protected function is_valid_threshold_field( $field ) {
```
justlevine, line 157: Can we rename this to `protected function validate_threshold_field( $field ) : void {}`? More self documenting (`is_*` usually implies a bool return).

src/Data/Cursor/AbstractCursor.php
```diff
@@ -132,6 +144,153 @@ protected function is_valid_offset_and_node() {
 		return true;
 	}
 
+	/**
+	 * Validates threshold field configuration. Validation failure results in a fatal
+	 * error because query execution is guaranteed to fail.
+	 *
+	 * @param array $field  Threshold configuration.
```
justlevine, line 151:
```suggestion
	 * @param array|mixed $field  Threshold configuration.
```

src/Data/Cursor/AbstractCursor.php
```diff
@@ -132,6 +144,153 @@ protected function is_valid_offset_and_node() {
 		return true;
 	}
 
+	/**
+	 * Validates threshold field configuration. Validation failure results in a fatal
```
justlevine, line 148: This is an example of my [earlier feedback](https://github.com/wp-graphql/wp-graphql/pull/2882#pullrequestreview-1578197213). Ideally this doc-block would say _what_ a valid threshold field configuration looks like.

src/Data/Cursor/AbstractCursor.php
```diff
@@ -132,6 +144,153 @@ protected function is_valid_offset_and_node() {
...
+	/**
+	 * Applies threshold fields to the cursor cutoff.
+	 * 
+	 * @param array $fallback  Default threshold fields.
+	 * 
+	 * @throws \GraphQL\Error\InvariantViolation Invalid configuration format.
+	 *
+	 * @return void
+	 */
+	protected function compare_with_threshold_fields( $fallback = [] ) {
```
justlevine, line 238:
```suggestion
	 * @throws \GraphQL\Error\InvariantViolation Invalid configuration format.
	 */
	protected function compare_with_threshold_fields( $fallback = [] ): void {
```

src/Data/Cursor/AbstractCursor.php
```diff
@@ -132,6 +144,153 @@ protected function is_valid_offset_and_node() {
...
+		/**
+		 * Get threshold fields from query vars.
+		 * 
+		 * @var array|null
```
justlevine, line 242:
```suggestion
		 * @var array|null $threshold_fields
```

src/Data/Cursor/AbstractCursor.php
```diff
@@ -132,6 +144,153 @@ protected function is_valid_offset_and_node() {
...
+			);
+		}
+
+		// Check if only one threshold field provided, wrap it in an array.
+		if ( ! empty( $threshold_fields ) && is_array( $threshold_fields ) && ! isset( $threshold_fields[0] ) ) {
+			$threshold_fields = [ $threshold_fields ];
+		}
+
+		foreach ( $threshold_fields as $field ) {
+			$this->is_valid_threshold_field( $field );
```
justlevine, line 265:
```suggestion
		// Bail early if no threshold fields.
		if ( empty ( $threshold_fields ) ) {
			return;
		}

		if ( ! is_array( $threshold_fields ) ) {
			throw new InvariantViolation(
				sprintf(
					/* translators: %s: value type. */
					__( 'Invalid value provided for graphql_cursor_threshold_fields. Expected Array, %s given.', 'wp-graphql' ), 
					gettype( $threshold_fields )
				)
			);
		}

		// Check if only one threshold field provided, wrap it in an array.
		if ( ! isset( $threshold_fields[0] ) ) {
			$threshold_fields = [ $threshold_fields ];
		}

		foreach ( $threshold_fields as $field ) {
			$this->is_valid_threshold_field( $field );
```

src/Data/Cursor/AbstractCursor.php
```diff
@@ -132,6 +144,153 @@ protected function is_valid_offset_and_node() {
...
+			$this->builder->add_field( $key, $value, $type, $order );
+		}
+	}
+
+	/**
+	 * Applies ID field to the cursor builder.
+	 *
+	 * @return void
+	 */
+	protected function compare_with_id_field() {
```
justlevine, line 281:
```suggestion
	/**
	 * Applies ID field to the cursor builder.
	 */
	protected function compare_with_id_field(): void {
```

jasonbahl:
I pulled this down to test. I succeeded with one test (https://github.com/wp-graphql/wp-graphql/issues/2644#issuecomment-1686700472)

But, when I tried another, I did not find the same success. 

Here's steps to reproduce the issue I'm facing: 

- register an "event" post type
- use ACF to add a "event_date" field
- create posts with the following data: 

| title       | event_date (meta) | price (meta) |
|-------------|-------------------|--------------|
| AAA January | 20230101          | 1.00         |
| BBB January | 20230101          | 66.00        |
| CCC Jauary  | 20230101          | 1.00         |
| DDD January | 20230101          | 1.00         |
| February    | 20230201          | 99.00        |
| March       | 20230301          | 22.00        |
| April       | 20230401          | 22.00        |

- add the following PHP snippet: 

```php
add_filter('graphql_PostObjectsConnectionOrderbyEnum_values', function ($values) {

	$values['EVENT_DATE'] = [
		'value' => 'event_date',
		'description' => __('The event date (stored in meta)', 'wp-graphql'),
	];

	$values['EVENT_PRICE'] = [
		'value' => 'price',
		'description' => __('The event price (stored in meta)', 'wp-graphql'),
	];

	return $values;
});

add_filter('graphql_post_object_connection_query_args', function ($query_args, $source, $input) {

	global $wpdb;

	if ( isset($input['where']['orderby'] ) && is_array($input['where']['orderby'] ) ) {

		$new_orderby = isset( $query_args['orderby'] ) ? $query_args['orderby'] : [];

		if ( ! is_array( $new_orderby ) ) {
			$new_orderby = [ $query_args['orderby'] ];
		}

		foreach ( $input['where']['orderby'] as $orderby ) {

			if ( ! isset( $orderby['field'] ) || ! in_array( $orderby['field'], [ 'event_date', 'price' ], true ) ) {
				continue;
			}

			$meta_query = isset( $query_args['meta_query'] ) ? $query_args['meta_query'] : [];

			// set the relation
			$meta_query['relation'] = isset( $meta_query['relation'] ) ? $meta_query['relation'] : 'AND';

			$type = 'NUMERIC';

			if ( 'event_date' === $orderby['field'] ) {
				$type = 'DATETIME';
			}

			if ( 'price' === $orderby['field'] ) {
				$type = 'DECIMAL';
			}

			// Add the claus
			$meta_query[ $orderby['field'] . '_claus' ] = [
				'key' => $orderby['field'],
				'compare' => 'EXISTS',
				'type' => $type,
			];


			$new_orderby[ $orderby['field'] . '_claus' ] = $orderby['order'];

			$query_args['orderby'] = $new_orderby;
			$query_args['meta_query'] = $meta_query;

		}
	}
	
	return $query_args;
}, 10, 3);
```

- execute the following query: 

```graphql
query GetEvents($cursor: String = "") {
  events(
    first: 10
    after: $cursor
    where: {orderby: [{field: EVENT_PRICE, order: DESC}, {field:TITLE, order:DESC}]}
  ) {
    pageInfo {
      startCursor
      endCursor
      hasNextPage
      hasPreviousPage
    }
    edges {
      cursor
      node {
        id
        databaseId
        title
        eventDetails {
          eventDate
          price
        }
      }
    }
  }
}
```

See the following results:

```js
{
  ...
  "edges": [
        {
          "cursor": "YXJyYXljb25uZWN0aW9uOjM1NTE=",
          "node": {
            "id": "cG9zdDozNTUx",
            "databaseId": 3551,
            "title": "February",
            "eventDetails": {
              "eventDate": "01/02/2023",
              "price": 99
            }
          }
        },
        {
          "cursor": "YXJyYXljb25uZWN0aW9uOjM1NTQ=",
          "node": {
            "id": "cG9zdDozNTU0",
            "databaseId": 3554,
            "title": "bbb-january",
            "eventDetails": {
              "eventDate": "01/01/2023",
              "price": 66
            }
          }
        },
        {
          "cursor": "YXJyYXljb25uZWN0aW9uOjM1NTI=",
          "node": {
            "id": "cG9zdDozNTUy",
            "databaseId": 3552,
            "title": "March",
            "eventDetails": {
              "eventDate": "01/03/2023",
              "price": 22
            }
          }
        },
        {
          "cursor": "YXJyYXljb25uZWN0aW9uOjM1NTM=",
          "node": {
            "id": "cG9zdDozNTUz",
            "databaseId": 3553,
            "title": "April",
            "eventDetails": {
              "eventDate": "01/04/2023",
              "price": 22
            }
          }
        },
        {
          "cursor": "YXJyYXljb25uZWN0aW9uOjY3NDk=",
          "node": {
            "id": "cG9zdDo2NzQ5",
            "databaseId": 6749,
            "title": "ddd-january",
            "eventDetails": {
              "eventDate": "01/01/2023",
              "price": 1
            }
          }
        },
        {
          "cursor": "YXJyYXljb25uZWN0aW9uOjM1NDk=",
          "node": {
            "id": "cG9zdDozNTQ5",
            "databaseId": 3549,
            "title": "ccc-january",
            "eventDetails": {
              "eventDate": "01/01/2023",
              "price": 1
            }
          }
        },
        {
          "cursor": "YXJyYXljb25uZWN0aW9uOjY3NTA=",
          "node": {
            "id": "cG9zdDo2NzUw",
            "databaseId": 6750,
            "title": "AAA January",
            "eventDetails": {
              "eventDate": "01/01/2023",
              "price": 1
            }
          }
        }
      ],
      ...
}
```

So far, so good. 

The query is ordered by PRICE descending, then Title Descending. Great 🙌🏻 

- Execute the query again, but this time do `first: 1`
- see the following result: 

```js
{
  ...
  "edges": [
        {
          "cursor": "YXJyYXljb25uZWN0aW9uOjM1NTE=",
          "node": {
            "id": "cG9zdDozNTUx",
            "databaseId": 3551,
            "title": "February",
            "eventDetails": {
              "eventDate": "01/02/2023",
              "price": 99
            }
          }
        },
      ],
      ...
}
```

Still good! 🙌🏻 

### PROBLEM:

- Take the `cursor` and use it in the `$cursor` of the query and execute again, repeating with each cursor.

```graphql
query GetEvents($cursor: String = "YXJyYXljb25uZWN0aW9uOjM1NTE=") {
  events(
    first: 1
    after: $cursor
    where: {orderby: [{field: EVENT_PRICE, order: DESC}, {field:TITLE, order:DESC}]}
  ) {
    ...
```

If we keep paginating, we should expect to get the nodes in the same order as the larger list above. But, instead we get the following nodes returned: 

- `first: 1, cursor: ""`: February (expected)
- `first: 1, cursor: "YXJyYXljb25uZWN0aW9uOjM1NTE=" (the cursor representing february): bbb-January (expected)
- `first: 1, cursor: "YXJyYXljb25uZWN0aW9uOjM1NTQ=" (cursor represenging bbb-january): April (not expected. Expected "March")
- `first 1, cursor: "YXJyYXljb25uZWN0aW9uOjM1NTM=" (cursor representing April): AAA January(not expected, expected DDD January)


The SQL generated under the hood is: 

raw: 

```sql
\n\t\t\tSELECT   wp_posts.ID\n\t\t\tFROM wp_posts  INNER JOIN wp_postmeta ON ( wp_posts.ID = wp_postmeta.post_id )\n\t\t\tWHERE 1=1  AND ( \n  wp_postmeta.meta_key = 'price'\n) AND wp_posts.post_type = 'event' AND ((wp_posts.post_status = 'publish')) AND  wp_posts.post_title <= \"April\" AND ( wp_posts.post_title < \"April\" OR (  wp_postmeta.meta_value <= \"22.00\" AND ( wp_postmeta.meta_value < \"22.00\" OR (  wp_posts.ID < 3553  ) )  ) ) \n\t\t\tGROUP BY wp_posts.ID\n\t\t\tORDER BY CAST(wp_postmeta.meta_value AS DECIMAL) DESC, wp_posts.post_title DESC, CAST(wp_postmeta.meta_value AS DECIMAL) DESC, wp_posts.ID DESC\n\t\t\tLIMIT 0, 2\n\t\t"
```


cleaned up by chat-gpt:

```sql
SELECT wp_posts.ID
FROM wp_posts
INNER JOIN wp_postmeta ON (wp_posts.ID = wp_postmeta.post_id)
WHERE 1=1
    AND (wp_postmeta.meta_key = 'price')
    AND wp_posts.post_type = 'event'
    AND (wp_posts.post_status = 'publish')
    AND wp_posts.post_title <= "April"
    AND (wp_posts.post_title < "April"
        OR (wp_postmeta.meta_value <= "22.00"
            AND (wp_postmeta.meta_value < "22.00"
                OR (wp_posts.ID < 3553)
            )
        )
    )
GROUP BY wp_posts.ID
ORDER BY CAST(wp_postmeta.meta_value AS DECIMAL) DESC,
         wp_posts.post_title DESC,
         CAST(wp_postmeta.meta_value AS DECIMAL) DESC,
         wp_posts.ID DESC
LIMIT 0, 2;
```

jasonbahl:
I believe this is the SQL we want: 

```sql
SELECT wp_posts.ID, wp_posts.post_title
FROM wp_posts
INNER JOIN wp_postmeta ON (wp_posts.ID = wp_postmeta.post_id)
WHERE 1=1
  AND (wp_postmeta.meta_key = 'price')
  AND wp_posts.post_type = 'event'
  AND (wp_posts.post_status = 'publish')
  AND (wp_postmeta.meta_value < "22.00")
    OR ( wp_postmeta.meta_value = "22.00" AND wp_posts.post_title < "April")
    OR ( wp_postmeta.meta_value = "22.00" AND wp_posts.post_title = "April" AND wp_posts.ID > 3553 )
ORDER BY wp_posts.post_title DESC, CAST(wp_postmeta.meta_value as DECIMAL) DESC
limit 0, 2
```

This returns: 

- ddd-january
- ccc-january

which is what I would expect if I were trying to paginate after the "April" event

CleanShot 2023-08-22 at 10 27 41

jasonbahl:
ok, this actually looks like the query we want: 

```sql
SELECT *
FROM wp_posts
INNER JOIN wp_postmeta AS mt0 ON wp_posts.ID = mt0.post_id
INNER JOIN wp_postmeta AS mt1 ON wp_posts.ID = mt1.post_id
WHERE mt0.meta_key = 'price'
  AND mt1.meta_key = 'event_date'
  AND wp_posts.post_type = 'event'
  AND wp_posts.post_status = 'publish'
  AND (mt0.meta_value, wp_posts.post_title, mt1.meta_value) < ('22.00', 'March', '20230301')
ORDER BY CAST(mt0.meta_value AS DECIMAL) DESC,
  wp_posts.post_title DESC,
  CAST(mt1.meta_value AS DATETIME) DESC,
  wp_posts.ID DESC
LIMIT 0, 2;
```

justlevine:
@jasonbahl is there a way to distill the [above](https://github.com/wp-graphql/wp-graphql/pull/2882#issuecomment-1688439017)  into a WPUnit test sans ACF? 

I'm not entirely clear if it's just the cursor iteration that's the problem or if it's also the specific choice of field types, but either way it would probably be helpful preventing future regressions from slipping in.

jasonbahl:
@justlevine ya, ACF isn't necessary, just was the fastest way for me to reproduce. 

it was just a fast way for me to `update_post_meta()`

I could have done it with cli or anything. Just wanted to record the steps I took. 

But yes, I can work on a test, too

jasonbahl:
fwiw, I'm making decent progress getting the SQL to generate the way I would expect. 

Also, for reference, this article has been helpful to reference SQL and cursor pagination: https://medium.com/@george_16060/cursor-based-pagination-with-arbitrary-ordering-b4af6d5e22db

wp-graphql.php
```diff
@@ -85,3 +85,70 @@ function graphql_init_appsero_telemetry() {
...
+			$new_orderby[ $orderby['field'] . '_claus' ] = $orderby['order'];
+
+			$query_args['orderby'] = $new_orderby;
+			$query_args['meta_query'] = $meta_query;
+
+		}
+	}
+
+	return $query_args;
+}, 10, 3);
```
kidunot89: @jasonbahl Is this suppose to be here 🤣

jasonbahl:
I pushed up a few commits. the first one was an accident. I had 2 branches, 1 for debugging and 1 cleaned up for pushing up, but I pushed the wrong one 🤦🏻‍♂️ 

I pushed follow-up commits to fix that. 

Ultimately, the change I ended up making was: https://github.com/wp-graphql/wp-graphql/pull/2882/files#diff-ffd0e341c15c329e46bb25d9d9caa195975fb8477af13a986d2e4a5021aa572bL224-R253

The SQL generated was creating a meta alias of `mt0` but WordPress doesn't alias for the first meta query, only the 2nd and higher. 

So, we were ending up with SQL such as: 

```sql
SELECT wp_posts.ID
FROM wp_posts
INNER JOIN wp_postmeta ON wp_posts.ID = wp_postmeta.post_id
INNER JOIN wp_postmeta AS mt1 ON wp_posts.ID = mt1.post_id
WHERE 
    wp_postmeta.meta_key = 'price' 
    AND mt1.meta_key = 'event_date'
    AND wp_posts.post_type = 'event' 
    AND wp_posts.post_status = 'publish' 
    AND mt0.meta_value <= "99.00"
    AND (
        mt0.meta_value < "99.00" 
        OR (
            wp_posts.post_title <= "February" 
            AND (
                wp_posts.post_title < "February" 
                OR (
                    mt1.meta_value <= "20230201" 
                    AND (
                        mt1.meta_value < "20230201" 
                        OR (
                            wp_posts.ID < 3551
                        )
                    )
                )
            )
        )
    )
GROUP BY wp_posts.ID
ORDER BY 
    CAST(wp_postmeta.meta_value AS CHAR) DESC, 
    wp_posts.post_title DESC, 
    CAST(mt1.meta_value AS DATETIME) DESC, 
    wp_posts.ID DESC
LIMIT 0, 101;
```

But wanted (note, the lack of `mt0`):

```sql
SELECT wp_posts.ID
FROM wp_posts
INNER JOIN wp_postmeta ON wp_posts.ID = wp_postmeta.post_id
INNER JOIN wp_postmeta AS mt1 ON wp_posts.ID = mt1.post_id
WHERE 
    wp_postmeta.meta_key = 'price' 
    AND mt1.meta_key = 'event_date'
    AND wp_posts.post_type = 'event' 
    AND wp_posts.post_status = 'publish' 
    AND wp_postmeta.meta_value <= "99.00"
    AND (
        wp_postmeta.meta_value < "99.00" 
        OR (
            wp_posts.post_title <= "February" 
            AND (
                wp_posts.post_title < "February" 
                OR (
                    mt1.meta_value <= "20230201" 
                    AND (
                        mt1.meta_value < "20230201" 
                        OR (
                            wp_posts.ID < 3551
                        )
                    )
                )
            )
        )
    )
GROUP BY wp_posts.ID
ORDER BY 
    CAST(wp_postmeta.meta_value AS CHAR) DESC, 
    wp_posts.post_title DESC, 
    CAST(mt1.meta_value AS DATETIME) DESC, 
    wp_posts.ID DESC
LIMIT 0, 101;
```

jasonbahl (approved):

PR #2890, opened 2023-08-09
fix: Use hostname for graphql cache header url for varnish

markkelnar:
The graphql url cache key header returned includes http/https as well as path.  But this differs from integration with WPE Varnish cache, where this currently runs.  Bring this to consistent/expected behavior where the varnish rule uses and invalidates only the hostname.

part of [wp-graphql/wp-graphql-smart-cache/issues/246](https://github.com/wp-graphql/wp-graphql-smart-cache/issues/246)

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Router.php
+++ b/src/Router.php
@@ -326,6 +326,9 @@ class Router {
 			]
 		);
 
+		// For cache url header, use the domain without protocol or path.
+		$endpoint_hostname = parse_url( graphql_get_endpoint_url() )['host'];
+
 		$headers = [
 			'Access-Control-Allow-Origin'  => '*',
 			'Access-Control-Allow-Headers' => implode( ', ', $access_control_allow_headers ),
@@ -334,7 +337,7 @@ class Router {
 			'Content-Type'                 => 'application/json ; charset=' . get_option( 'blog_charset' ),
 			'X-Robots-Tag'                 => 'noindex',
 			'X-Content-Type-Options'       => 'nosniff',
-			'X-GraphQL-URL'                => graphql_get_endpoint_url(),
+			'X-GraphQL-URL'                => $endpoint_hostname,
 		];
 
 
--- a/src/Router.php
+++ b/src/Router.php
@@ -327,7 +327,8 @@ class Router {
 		);
 
 		// For cache url header, use the domain without protocol or path.
-		$endpoint_hostname = parse_url( graphql_get_endpoint_url() )['host'];
+		$graphql_endpoint = parse_url( graphql_get_endpoint_url() );
+		$endpoint_hostname = isset( $graphql_endpoint['host'] ) ? $graphql_endpoint['host'] : graphql_get_endpoint_url();
 
 		$headers = [
 			'Access-Control-Allow-Origin'  => '*',
--- a/src/Router.php
+++ b/src/Router.php
@@ -327,7 +327,7 @@ class Router {
 		);
 
 		// For cache url header, use the domain without protocol or path.
-		$graphql_endpoint = parse_url( graphql_get_endpoint_url() );
+		$graphql_endpoint  = wp_parse_url( graphql_get_endpoint_url() );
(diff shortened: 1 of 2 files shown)


src/Router.php
```diff
@@ -326,6 +326,10 @@ protected static function get_response_headers() {
 			]
 		);
 
+		// For cache url header, use the domain without protocol or path.
+		$graphql_endpoint  = wp_parse_url( graphql_get_endpoint_url() );
+		$endpoint_hostname = isset( $graphql_endpoint['host'] ) ? $graphql_endpoint['host'] : graphql_get_endpoint_url();
```
justlevine, line 331: `graphql_get_endpoint_url()` applies a filter, so we should avoid calling it twice if we can.

```suggestion
		$graphql_endpoint  = graphql_get_endpoint_url() ;
		$endpoint_hostname = wp_parse_url( $graphql_endpoint, PHP_URL_HOST ) ?: $graphql_endpoint;
```

justlevine (approved):
LGTM 🚀

The lockfile version changes included here are explicit in #2891, so i'm not concerned that they slip in here too.

src/Router.php
```diff
@@ -326,6 +326,10 @@ protected static function get_response_headers() {
 			]
 		);
 
+		// For cache url header, use the domain without protocol or path.
+		$graphql_endpoint  = graphql_get_endpoint_url() ;
```
jasonbahl, line 330:
```suggestion
		$graphql_endpoint  = graphql_get_endpoint_url();
```

jasonbahl:
@markkelnar might need to revert this. 

The full endpoint is returned to allow for multi-tenant caches to not accidentally purge the incorrect thing. 

see: https://github.com/varnish/varnish-modules/issues/41#issuecomment-244747800

For WordPress instances that are set up as sub-directory multisites (`site.com/site-1/graphql`, `site.com/site-2/graphql`), caching clients (Varnish, etc) can prefix the cache keys with the graphql endpoint to ensure they're cached uniquely. 

If we use the X-GraphQL-Url to prefix our keys, and the value is only the host, not the full endpoint, then we could have collisions. 

For example a query from `site-1` and a query from `site-2` could both return the same node (`post:1`, for example). 

Currently they would be tagged as `site.com/site-1/graphql:post:1` and `site.com/site-2/graphql:post:1` respectively. 

When a post in site 1 changes, it would then purge `site.com/site-1/graphql:post:1` but not purge the site-2 post, and vis-versa. 

If we go back to just the host, instead of the full endpoint, then we end up tagging both as: `site.com:graphql:post:1` and an update to `post:1` on either site will purge the cache for both sites.

If we expand to a multisite network of hundreds of sites that each have `post:1` the problem gets worse. 

I believe, unless I'm missing something about how Varnish + XKey, etc works, I think we probably need to keep the full endpoint here.

----

Some more context:

- Original Issue that lead to the X-GraphQL-Url being output: https://github.com/wp-graphql/wp-graphql/issues/2564
- Original PR: https://github.com/wp-graphql/wp-graphql/pull/2565

markkelnar:
I see site_url() used in graphql_get_endpoint_url().  We should confirm that is indeed returning the value for public/queryable hostname/path, ex. domain.com/pathy/graphql/, that you are expecting and not the path to wp-admin path.  We might want to use home_url() instead for the url of where requests are sent.  

It might also make sense not to include the protocol (http/https) in the url cache header.

Without knowing those, we might not have a change in behavior for anyone using caching.  If you are more comfortable reverting out this change and figuring that out before we try again, we can do that.

jasonbahl:
@markkelnar I set up a multisite using localwp and set it to use sub-directories. 

My site is cleverly named `multisite.local`. 

I created a 2nd site, cleverly named `site-2`. This site's GraphQL endpoint can be accessed at `multisite.local/site-2/graphql`

When I query data from site-2, the expected fully qualified endpoint is output in the headers: 

CleanShot 2023-08-10 at 14 33 33

I believe we will cause a regression for subdomain multisites if we make this change.

PR #2896, opened 2023-08-14
fix: Deprecated: trim(): Passing null to parameter #1 () of type stri…

izzygld:
**Error:** Deprecated:` trim():` Passing null to parameter #1 `($string)` of type string is deprecated
**Goal:** To return empty string if no excerpt, (instead of just null)

Screenshot 2023-08-14 at 11 49 09

Where has this been tested?
---------------------------
**Plugin Version:** 1.14.10

**WordPress Version:** 6.3

### DIFF
--- a/src/Model/Post.php
+++ b/src/Model/Post.php
@@ -502,11 +502,16 @@ class Post extends Model {
 					},
 					'capability' => isset( $this->post_type_object->cap->edit_posts ) ? $this->post_type_object->cap->edit_posts : 'edit_posts',
 				],
-				'excerptRendered'           => function () {
+				'excerptRendered' => function () {
 					$excerpt = ! empty( $this->data->post_excerpt ) ? $this->data->post_excerpt : null;
 					$excerpt = apply_filters( 'get_the_excerpt', $excerpt, $this->data );
-
-					return $this->html_entity_decode( apply_filters( 'the_excerpt', $excerpt ), 'excerptRendered' );
+			
+					if ($excerpt !== null) {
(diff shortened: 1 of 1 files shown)


src/Model/Post.php
```diff
@@ -502,11 +502,16 @@ protected function init() {
 					},
 					'capability' => isset( $this->post_type_object->cap->edit_posts ) ? $this->post_type_object->cap->edit_posts : 'edit_posts',
 				],
-				'excerptRendered'           => function () {
+				'excerptRendered' => function () {
 					$excerpt = ! empty( $this->data->post_excerpt ) ? $this->data->post_excerpt : null;
```
justlevine, line 506: My understanding is that we want to apply the filters no matter what (so WordPress for example can populate an empty excerpt), and it seems `get_the_excerpt` [also expects a string](https://developer.wordpress.org/reference/functions/get_the_excerpt/).

I think if we just change `$excerpt` to fall back to an empty `''` instead of `null` and revert everything else, we can fix this bug and leave intended behavior intact.


I.e.
```suggestion
					$excerpt = ! empty( $this->data->post_excerpt ) ? $this->data->post_excerpt : '';
```

Thoughts @izzygld ?

izzygld: yes! even better idea.

thsanks @justlevine 

made a [pr for you.](pr - https://github.com/wp-graphql/wp-graphql/pull/2897)

justlevine: Looks great! When I have a minute (unless you beat me to it) I'll copy over the PR title and description and mark it for review.

PR #2935, opened 2023-09-19
fix: admin notice if composer dependencies not installed

### LINKED ISSUE #1831: Warning message when activating wp-graphql plugin from master branch
Experiencing the following when activating or deactivating wp-graphql.

PHP Warning:  call_user_func_array() expects parameter 1 to be a valid callback, function 'graphql_activation_callback' not found or invalid function name in /var/www/html/wp-includes/class-wp-hook.php on line 287
Warning: call_user_func_array() expects parameter 1 to be a valid callback, function 'graphql_activation_callback' not found or invalid function name in /var/www/html/wp-includes/class-wp-hook.php on line 287

To reproduce, install the plugin found at https://github.com/wp-graphql/wp-graphql/archive/master.zip
Steps to reproduce:
1. Download plugin zip from https://github.com/wp-graphql/wp-graphql/archive/master.zip
2. Open wp-admin.
2. Go to plugins
3. Click Add New
4. Select upload file.
5. Select the file and upload.
6. Find the apache error log for the WP and notice the warning.

Steps to reproduce using wp-cli:
1.  Deactivate the plugin if it is active. `wp --allow-root plugin deactivate wp-graphql`
2. Delete the pliugin if it is installed. `wp --allow-root plugin delete wp-graphql`
3. Install and activate the plugin `wp --allow-root plugin install https://github.com/wp-graphql/wp-graphql/archive/master.zip --activate`

You will see the message in standard error output.

I do not see the error when installed the tagged version 1.3.3 from https://github.com/wp-graphql/wp-graphql/releases/download/v1.3.3/wp-graphql.zip or https://github.com/wp-graphql/wp-graphql/releases/tag/v1.3.3.

jasonbahl:
This fixes a regression to the admin error that displays if Composer dependencies have not been installed. 

Now, if the plugin is cloned from Github or downloaded as a .zip from Github the error will show in the admin explaining that the composer dependencies have not been installed. 

If the plugin is installed from composer (packagist / wpackagist) or the WordPress.org repo, the admin error will not be displayed, as the dependencies are bundled in those distributions. 

**TO TEST:**

- Install the plugin by downloading the .zip from Github or by cloning from Github and _not_ running `composer install`. 

closes #1831 

### When activating the plugin without composer dependencies: 

Error is displayed and GraphQL functionality is not loaded. No GraphiQL IDE, etc.

CleanShot 2023-09-19 at 10 03 51

### When activating the plugin with composer dependencies:

No error is displayed, and GraphQL functionality is active

CleanShot 2023-09-19 at 10 05 30


josephfusco (approved):
Looks good!
I can see the admin notice when vendor/ is absent.
Screenshot 2023-09-25 at 2 43 53 PM

jasonbahl:
For anyone that comes across the above screenshot and is curious what the "Symlinked Branch" flags are on the plugin page, they're from @josephfusco sweet plugin [Symlinked Plugin Branch](https://github.com/josephfusco/symlinked-plugin-branch)

wp-graphql.php
```diff
@@ -41,13 +48,73 @@
...
+	 * WPGRAPHQL_AUTOLOAD can be set to "false" to prevent the autoloader from running.
+	 * In most cases, this is not something that should be disabled, but some environments
+	 * may bootstrap their dependencies in a global autoloader that will autoload files
+	 * before we get to this point, and requiring the autoloader again can trigger fatal errors.
+	 *
+	 * The codeception tests are an example of an environment where adding the autoloader again causes issues
+	 * so this is set to false for tests.
+	 */
+	// @phpstan-ignore-next-line: this is ignored as the constant could be defined in wp-config, prior to being defined above
+	if ( defined( 'WPGRAPHQL_AUTOLOAD' ) && false === WPGRAPHQL_AUTOLOAD ) {
```
justlevine, line 66: @jasonbahl this change is breaking for me on bedrock (and my assumption is _any_ setup using composer + normal packagist), since the autoloader isnt in `wp-graphql/vendor`.

image

jasonbahl: If we check for class_exists( `\WPGraphQL` ) here, does that fix it? 

Like if autoload is false _or_ WPGraphQL class already exists?

justlevine: Not by my desk to play around, and I'm not entirely sure the source of the bug this PR is fixing, but yeah I think the previous (intended) logical flow was more robust:

1. If `WPGraphQL` exists, we clearly have the autoloader generated which means either composer was run, it's being autoloaded somewhere else, it's dist version, or someone shimmed the class.
2. If the class doesn't, then we should try and and load the autoloader, and gracefully let people know it's missing.
3. If there's _intentionally_ no autoloader and WPGraphQL still doesn't exist, then the user should manually set a constant that tells WPGraphQL not to error out. (At least that's been my assumption what the autoload constant is for).

jasonbahl: Ya so it's tricky. 

We had an admin notice that warned folks that cloned the plugin or downloaded the zip from GitHub (not from the release page) that the dependencies don't exist. 

That admin notice was being called from a class, a class that didn't exist if the dependencies didn't exist, so the users we were trying to warn, never saw the warning 😂 

So, it's trying to fix that.

PR #2937, opened 2023-09-25
fix: Support asPreview by URI/SLUG Id Type

### LINKED ISSUE #1673: Querying for preview via asPreview argument doesn't work when querying by slug, uri idTypes
According to the switch statement at https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/Object/RootQuery.php#L503, when querying a post preview via slug or uri, the resolver functions return early before checking the `asPreview` parameter and returning the revision. 

This seems like it would be considered a bug because I would expect the same behavior regardless of which IdType is being used, also because I don't see any documentation that specifically explains why this would be intentional.

blakewilson:

This PR fixes previews when querying by `URI` ID Type in a `ContentNode` with `AsPreview` set to `true` or by a post type (eg. `post`, `page`) with `asPreview` set to `true` and with ID Types of `URI` or `SLUG`.

The following queries can be used for testing:

```graphql
query GetPreview($slug: ID!, $asPreview: Boolean!) {
  post(id: $slug, idType: SLUG, asPreview: $asPreview) {
    ... on NodeWithTitle {
      title
    }
    ... on NodeWithContentEditor {
      content
    }
    date
  }
}
```

```graphql
query GetPreview($uri: ID!, $asPreview: Boolean!) {
  contentNode(id: $uri, idType: URI, asPreview: $asPreview) {
    ... on NodeWithTitle {
      title
    }
    ... on NodeWithContentEditor {
      content
    }
    date
  }
}
```

Closes #1673

`contentNode` test with `asPreview` set to `false`
Screenshot 2023-09-25 at 2 32 45 PM

`contentNode` test with `asPreview` set to `true`
Screenshot 2023-09-25 at 2 32 56 PM

`post` test using `URI` for ID Type with `asPreview` set to `false`
Screenshot 2023-09-25 at 2 33 08 PM

`post` test using `URI` for ID Type with `asPreview` set to `true`
Screenshot 2023-09-25 at 2 33 17 PM

`post` test using `SLUG` for ID Type with `asPreview` set to `false`
Screenshot 2023-09-25 at 2 34 04 PM

`post` test using `SLUG` for ID Type with `asPreview` set to `true`
Screenshot 2023-09-25 at 2 34 13 PM

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** MacOS 13.1

**WordPress Version:** 6.3.1

**WPGraphQL Version:** 1.16.0


src/Data/NodeResolver.php
```diff
@@ -582,6 +588,30 @@ public function parse_request( string $uri, $extra_query_vars = '' ) {
...
 	}
 
+	/**
+	 * Given a WP Post or post ID, this method attempts to resolve a preview post ID.
+	 * 
+	 * @param Int|WP_Post $post The WP Post object or Post ID
+	 * 
+	 * @return Int A preview post ID if one exists, the current post ID if one doesn't exist.
+	 */
+	public static function resolve_node_preview( $post ) {
```
blakewilson, line 598: I'm not sure if this `static` method belongs here. Is there a better spot for this util? cc @jasonbahl @justlevine

blakewilson:
Tests will be created soon 🤞

src/Data/NodeResolver.php
jasonbahl, line 598:
```suggestion
	public static function get_post_preview_id( $post ) {
```

src/Type/ObjectType/RootQuery.php
```diff
@@ -236,15 +238,7 @@ public static function register_type() {
...
-								$revisions = wp_get_post_revisions(
-									$post_id,
-									[
-										'posts_per_page' => 1,
-										'fields'         => 'ids',
-										'check_enabled'  => false,
-									]
-								);
-								$post_id   = ! empty( $revisions ) ? array_values( $revisions )[0] : $post_id;
+								$post_id = NodeResolver::resolve_node_preview($post_id);
```
jasonbahl, line 241:
```suggestion
								$post_id = NodeResolver:: get_post_preview_id($post_id);
```

src/Type/ObjectType/RootQuery.php
```diff
@@ -735,15 +731,7 @@ public static function register_post_object_fields() {
...
-							$revisions = wp_get_post_revisions(
-								$post_id,
-								[
-									'posts_per_page' => 1,
-									'fields'         => 'ids',
-									'check_enabled'  => false,
-								]
-							);
-							$post_id   = ! empty( $revisions ) ? array_values( $revisions )[0] : $post_id;
+							$post_id = NodeResolver::resolve_node_preview($post_id);
```
jasonbahl, line 734:
```suggestion
							$post_id = NodeResolver::get_post_preview_id($post_id);
```

src/Data/NodeResolver.php
```diff
@@ -232,7 +232,13 @@ public function resolve_uri( string $uri, $extra_query_vars = '' ) {
 				return null;
 			}
 
-			return ! empty( $queried_object->ID ) ? $this->context->get_loader( 'post' )->load_deferred( $queried_object->ID ) : null;
+			$post_id = $queried_object->ID;
+
+			if ( isset( $extra_query_vars['asPreview'] ) && true === $extra_query_vars['asPreview'] ) {
+				$post_id = $this->resolve_node_preview($post_id);
```
jasonbahl, line 238:
```suggestion
				$post_id = self::get_post_preview_id($post_id);
```

src/Data/NodeResolver.php
```diff
@@ -582,6 +588,30 @@ public function parse_request( string $uri, $extra_query_vars = '' ) {
...
 	}
 
+	/**
+	 * Given a WP Post or post ID, this method attempts to resolve a preview post ID.
+	 * 
+	 * @param Int|WP_Post $post The WP Post object or Post ID
+	 * 
+	 * @return Int A preview post ID if one exists, the current post ID if one doesn't exist.
+	 */
+	public static function get_post_preview_id( $post ) {
```
justlevine, line 598: Now that this is a public static getter, I think we should move this to `WPGraphQL\Utils`.

It's not directly related to NodeResolver, it's the only static method on the class, and as a public static, keeping it with other utils methods will increase refactorability.

jasonbahl:
> Tests will be created soon 🤞

@blakewilson were you waiting on me for anything before you write tests? let me know 🙏🏻

blakewilson:
> > Tests will be created soon 🤞
> 
> @blakewilson were you waiting on me for anything before you write tests? let me know 🙏🏻

No, I'm currently trying to resolve some failing tests. It looks like the changes in this PR have some unintended side effects, so trying to get those resolved before writing new tests.

blakewilson:
@jasonbahl Just added appropriate tests for previews by URI and SLUG! Sorry, it's kind of a lot of test changes 😬

jasonbahl:
@blakewilson nothing to be sorry about! Appreciate the contributions. Will review it shortly! 🙏

tests/wpunit/PreviewTest.php
```diff
@@ -110,206 +110,529 @@ public function get_query() {
...
+		    node {
+		      databaseId
+		    }
+		  }
+		}
+		';
+
+	}
+
+	public function get_query_by_slug() {
```
jasonbahl, line 159: @blakewilson I don't think we need to make this 2 queries. We could just make the `idType` a variable and use the same query with different variable inputs, eh?

blakewilson: Doh! Yeah good call. I was in automate/copy/paste mode. I'll take care of that right now.

src/Utils/Utils.php
```diff
@@ -261,4 +261,28 @@ public static function get_node_type_from_id( $id ) {
 		$id_parts = Relay::fromGlobalId( $id );
 		return $id_parts['type'] ?: null;
 	}
+
+	/**
+	 * Given a WP Post or post ID, this method attempts to resolve a preview post ID.
+	 * 
+	 * @param Int|\WPGraphQL\Utils\WP_Post $post The WP Post object or Post ID
```
justlevine, line 268:
```suggestion
	 * @param int|\WPGraphQL\Utils\WP_Post $post The WP Post object or Post ID
```

src/Utils/Utils.php
```diff
@@ -261,4 +261,28 @@ public static function get_node_type_from_id( $id ) {
...
+	 * 
+	 * @return Int A preview post ID if one exists, the current post ID if one doesn't exist.
+	 */
+	public static function get_post_preview_id( $post ) {
```
justlevine, line 272: It's a new method, lets be strict here.
```suggestion
	 */
	public static function get_post_preview_id( $post ): int {
```

justlevine: Whoops. accidentally removed the entire return comment. Updated suggestion.

jasonbahl:
uh oh. I was working on updating tests to use WPGraphQLTestCase, and now a bunch of things are broken. Probably my fault 🤦🏻‍♂️

justlevine:
> uh oh. I was working on updating tests to use WPGraphQLTestCase, and now a bunch of things are broken. Probably my fault 🤦🏻‍♂️

The last action to run before your latest commits also wasnt working (https://github.com/wp-graphql/wp-graphql/actions/runs/6398962072/job/17370176424) , so the cause might still be outside that diff.

blakewilson:
> > uh oh. I was working on updating tests to use WPGraphQLTestCase, and now a bunch of things are broken. Probably my fault 🤦🏻‍♂️
> 
> The last action to run before your latest commits also wasnt working (https://github.com/wp-graphql/wp-graphql/actions/runs/6398962072/job/17370176424) , so the cause might still be outside that diff.

Weird. When I ran these independently they worked, wonder if something is conflicting. I can look into this

jasonbahl:
@blakewilson I am working with @josephfusco to debug the test suite issues. We believe it's something to do with the setup/tear down of some other suite(s)

jasonbahl:
@blakewilson

I added this to both of Blakes tests:

`$this->set_permalink_structure( '' );`

(and copied that method from here: https://github.com/wp-graphql/wp-graphql/blob/develop/tests/wpunit/PostObjectQueriesTest.php#L88-L93)

and they pass.

So it's a conflict with permalinks. Probably need to manually investigate why it works with default permalinks but not other pretty permalinks.

blakewilson:
> @blakewilson
> 
> I added this to both of Blakes tests:
> 
> `$this->set_permalink_structure( '' );`
> 
> (and copied that method from here: https://github.com/wp-graphql/wp-graphql/blob/develop/tests/wpunit/PostObjectQueriesTest.php#L88-L93)
> 
> and they pass.
> 
> So it's a conflict with permalinks. Probably need to manually investigate why it works with default permalinks but not other pretty permalinks.

Awesome work @jasonbahl @josephfusco! Thanks for looking into this.

So does this mean that there is something else in WPGraphQL that needs to be looked at to fix the permalinks issue? Does this affect all "query by URI" things, or just things relating to this PR? Any suggestions on places to look first?

jasonbahl:
@blakewilson im not sure yet. We diagnosed the issue but didn't have a chance to manually test further with different permalinks. 

My hunch is that the resolve_by_uri() needs some work to resolve properly when asPreview is true and various permalink structures are in play

jasonbahl:
@blakewilson ok, so I manually tested and it looks like the issue is related to draft posts. Since draft posts aren't public, they don't have a "pretty permalink" issued (WordPress doesn't make them publicly visible by URI) so I think `idType: URI` won't work with draft posts. 

We might be able to fix it under the hood, and I'll keep digging, but here's the gist of what's happening:

## Published Post

If I click "preview" on a published post, I get a URL like so: `/blog/2023/10/09/published/?preview_id=6866&preview_nonce=1e626c85ac&preview=true&_thumbnail_id=3390`

I can take that URI and pass it to the query like so: 

```graphql
{
  post(id: "/blog/2023/10/09/published/?preview_id=6866&preview_nonce=1e626c85ac&preview=true&_thumbnail_id=3390", asPreview: true, idType: URI) {
    id
    isPreview
    databaseId
    uri
  }
}
```

And I get the preview node as expected:

CleanShot 2023-10-11 at 12 45 34

## Draft Posts

If I click "preview" on a draft post, I get a link like so (no pretty permalink, even though I have pretty permalinks enabled): `/?p=6874&preview=true`

If I pass this URI to the query:

```graphql
{
  post(id: "/?p=6874&preview=true", asPreview: true, idType: URI) {
    id
    isPreview
    databaseId
    uri
  }
}
```

I get null: 

CleanShot 2023-10-11 at 12 47 34

----

The tests are making assertions with pretty permalinks enabled (carried over from the PostObjectQueriesTest and never reset), which I think has exposed that this feature isn't fully working as expected. 

That said, I might argue that instead of passing the following args to the query: `id: "/?p=6874&preview=true", asPreview: true, idType: URI` the client might be better off parsing the query params and passing the following query: 

```graphql
{
  post(id: "6874", asPreview: true, idType: DATABASE_ID ) {
    id
    isPreview
    databaseId
    uri
  }
}
```

Which _does_ work: 

CleanShot 2023-10-11 at 12 51 17

So, as a user I could click "preview" on a draft post.

The client app (Faust, etc) could parse query params, if there are any, and decide what query to send. 

For example, Faust could determine that `preview=true` exists and set the `asPreview: true` in the query then determine that `p=${id}` or `preview_id=${id}` exists and use those as the values for `id: $p` or `id: $preview_id`, and set `idType: DATABASE_ID`.

jasonbahl:
@blakewilson ok, I think I might have figured it out. 

2 things: 

1. The tests were using `get_permalink( $draft_id )` which doesn't match the link you get when you click `preview` on a draft, so I updated to `get_preview_post_link( $draft_id )`. 

2. There was a bug where the WP_Query was receiving both `'p' => $id` and `'pagename' => ''` and this was causing WP_Query to generate SQL queries like: 

```sql
SELECT
    wp_posts.*
FROM
    wp_posts
WHERE
    1=1
    AND (wp_posts.ID = '0')
    AND wp_posts.ID = 6874
    AND wp_posts.post_type = 'post'
ORDER BY
    wp_posts.post_date DESC;
```

There's no way we'll find a post where `wp_posts.ID = '0'` AND `wp_posts.ID = 6874`. 

I added some logic that determines if the parsed_url has query params, and if so, we check if the query_params are equal to the "pagename" and if so, we unset "pagename" because that means the path is _only_ query params and nothing else.

josephfusco:
I tested this locally, both `idType: SLUG` & `idType: URI` are working as expected.

Issue #1658, opened 2021-01-05
All the query now returns " cannot register duplicate Types to the Schema. The Type 'Template_' already exists in the Schema. Make sure to give new Types a unique name"


wcube-admin:
Ever since I've changed my WordPress site language from English to Traditional Chinese, All the GraphiQL calls start to return the following error / warning message

`"extensions": {
    "debug": [
      {
        "type": "DUPLICATE_TYPE",
        "message": "You cannot register duplicate Types to the Schema. The Type 'Template_' already exists in the Schema. Make sure to give new Types a unique name.",
        "type_name": "Template_",
        "stack": [
          "/home/public_html/wp-content/plugins/wp-graphql/src/Registry/TypeRegistry.php",
          "/home/public_html/wp-content/plugins/wp-graphql/access-functions.php",
          "/home/public_html/wp-includes/class-wp-hook.php",
          "/home/public_html/wp-includes/class-wp-hook.php",
          "/home/public_html/wp-includes/plugin.php",
          "/home/public_html/wp-content/plugins/wp-graphql/src/Registry/TypeRegistry.php",
          "/home/public_html/wp-includes/class-wp-hook.php",
          "/home/public_html/wp-includes/class-wp-hook.php",
          "/home/public_html/wp-includes/plugin.php",
          "/home/public_html/wp-content/plugins/wp-graphql/src/Registry/TypeRegistry.php",
          "/home/public_html/wp-content/plugins/wp-graphql/src/Registry/SchemaRegistry.php",
          "/home/farcent/web/farcent-wp.wstudio.io/public_html/wp-content/plugins/wp-graphql/wp-graphql.php",
          "/home/farcent/web/farcent-wp.wstudio.io/public_html/wp-content/plugins/wp-graphql/src/Request.php",
          "/home/farcent/web/farcent-wp.wstudio.io/public_html/wp-content/plugins/wp-graphql/src/Router.php",
          "/home/farcent/web/farcent-wp.wstudio.io/public_html/wp-content/plugins/wp-graphql/src/Router.php",
          "/home/farcent/web/farcent-wp.wstudio.io/public_html/wp-includes/class-wp-hook.php",
          "/home/farcent/web/farcent-wp.wstudio.io/public_html/wp-includes/class-wp-hook.php",
          "/home/farcent/web/farcent-wp.wstudio.io/public_html/wp-includes/plugin.php",
[...]

PR #2940, opened 2023-09-26
fix: Decode slug so it works with other languages

scottyzen:
# fix: Decode strings so it works with other languages

- Decodes the slug so it works with other languages. Like Hebrew, Arabic, etc.
- ~~Decodes the type name, so it works with different languages (e.g. Chinese, Hebrew, etc.)~~

##Example:
 - Before: "**%d7%90%d7%91%d7%99%d7%96%d7%a8%d7%99%d7%9d**"
 - After:  "**הטבה לרוכשים**"

Where has this been tested?
---------------------------

**WordPress Version:** 6.3.1

### DIFF
--- a/src/Model/Term.php
+++ b/src/Model/Term.php
@@ -145,7 +145,7 @@ class Term extends Model {
 					return ! empty( $this->data->name ) ? $this->html_entity_decode( $this->data->name, 'name', true ) : null;
 				},
 				'slug'                     => function () {
-					return ! empty( $this->data->slug ) ? $this->data->slug : null;
+					return ! empty( $this->data->slug ) ? urldecode( $this->data->slug ) : null;
 				},
 				'termGroupId'              => function () {
 					return ! empty( $this->data->term_group ) ? absint( $this->data->term_group ) : null;
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -663,11 +663,6 @@ class TypeRegistry {
 	public function register_type( string $type_name, $config ): void {
 
 		/**
-		* Decodes the type name, so it works with different languages (e.g. Chinese, Hebrew, etc.)
-		*/
-		$type_name = urldecode( $type_name );
-
-		/**
 		 * If the type should be excluded from the schema, skip it.
 		 */
 		if ( in_array( strtolower( $type_name ), $this->get_excluded_types(), true ) ) {
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -661,7 +661,6 @@ class TypeRegistry {
 	 * @return void
 	 */
 	public function register_type( string $type_name, $config ): void {
-
 		/**
 		 * If the type should be excluded from the schema, skip it.
 		 */
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -661,6 +661,12 @@ class TypeRegistry {
 	 * @return void
 	 */
 	public function register_type( string $type_name, $config ): void {
+
+		/**
+		* Decodes the type name, so it works with different languages (e.g. Chinese, Hebrew, etc.)
+		*/
+		$type_name = urldecode( $type_name );
+
 		/**
 		 * If the type should be excluded from the schema, skip it.
 		 */
--- a/src/Model/Menu.php
+++ b/src/Model/Menu.php
@@ -90,7 +90,7 @@ class Menu extends Model {
 					return ! empty( $this->data->name ) ? $this->data->name : null;
 				},
 				'slug'       => function () {
-					return ! empty( $this->data->slug ) ? $this->data->slug : null;
+					return ! empty( $this->data->slug ) ? urldecode( $this->data->slug ) : null;
 				},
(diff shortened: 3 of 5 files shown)


src/Registry/TypeRegistry.php
```diff
@@ -661,6 +661,12 @@ protected function register_connections_from_config( array $config ) {
 	 * @return void
 	 */
 	public function register_type( string $type_name, $config ): void {
+
+		/**
+		* Decodes the type name, so it works with different languages (e.g. Chinese, Hebrew, etc.)
+		*/
+		$type_name = urldecode( $type_name );
+
```
justlevine, line 669: @scottyzen can you explain the reason/case for this change?

GraphQL type names can only be English alphanumeric (http://spec.graphql.org/draft/#sec-Names), and I'm pretty sure that by this point in the lifecycle `graphql_format_type_name()` has already run, stripping out any non A-Z.0-9  characters (including the `%` markers which would break `urldecode()` ).

Is there a scenario where either the encoded or decoded string is going to be all (english) alpha-numeric characters?

(ps: @jasonbahl , maybe we should expand `is_valid_graphql_name()` to check for more than a leading `int`? I'm not seeing obvious negative side effects, and we might be able to catch some currently fatal errors. )

src/Model/Term.php
```diff
@@ -145,7 +145,7 @@ protected function init() {
 					return ! empty( $this->data->name ) ? $this->html_entity_decode( $this->data->name, 'name', true ) : null;
 				},
 				'slug'                     => function () {
-					return ! empty( $this->data->slug ) ? $this->data->slug : null;
+					return ! empty( $this->data->slug ) ? urldecode( $this->data->slug ) : null;
```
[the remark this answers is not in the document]
justlevine: Do we also need to `urldecode()` the [term slugs for a Menu](https://github.com/wp-graphql/wp-graphql/blob/2937585d4f73756535302b0ebe78592e51e23c94/src/Model/Menu.php#L93), or does WordPress handle their generation differently?

src/Registry/TypeRegistry.php
scottyzen: Hi @justlevine. There may be a better workaround to be done. But for some languages, the type doesn't get registered. For example, in Hebrew, the type name is "הטבה לרוכשים" and it doesn't work. So I encoded it and it worked. I'm not sure if it's the best solution though. Maybe the a **A-Z.0-9** needs to be expanded to handle other characters like Hebrew, Arabic, etc. It's just something I came across recently while working on a that had a lot of Hebrew content.

jasonbahl: > it's just something I came across
@scottyzen can you share more about the specific path that was registering a type with hebrew characters?

was it a direct call to `register_graphql_type()` or something like a page template that the Schema tries to register types from? 

Would be curious to learn more about what exactly was registering the type.

scottyzen: Hey guys, so I think I found the issue. I'll try to explain it as best I can. Say I go to create 2 new Product Attributes.
- First one named: test-אחד 
- Second one named: test-שני .

From the GraphQL IDE, I get the following response:

```js
{
  "productAttributes": {
    "nodes": [
      {
        "name": "test-אחד"
      },
      {
        "name": "test-שני"
      }
    ]
  }
}
```

That response is correct. But because the register_type() only allows A-Z.0-9, it strips out other characters ( in this case the Hebrew characters) and the type name becomes "test-". So it fails when I try to create the second attribute because the type name is the same.
```js
{
    "type": "DUPLICATE_TYPE",
    "message": "You cannot register duplicate Types to the Schema. The Type 'ContentTypesOfPaTestEnum' already exists in the Schema. Make sure to give new Types a unique name.",
    "type_name": "ContentTypesOfPaTestEnum"
}
```

I hope this helps.

justlevine: Thanks @scottyzen .

That tracks with what I said above. You can see [this issue](https://github.com/wp-graphql/wp-graphql/issues/1658) for the same problem in a different context. Tl;Dr WPGraphQL should probably include some sort of transliteration function (e.g. via `iconv`), and in the interim you can use the built-in type name filters to str_replace `'א'=>'a'` etc.

In the context of this PR, I think you should revert this line, so we can add the necessary unit tests and get this merged (I'm happy to try and find time this week to write them up if you're unable).

scottyzen: Great, thanks @justlevine.

src/Model/Term.php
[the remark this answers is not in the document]
scottyzen: Yea, I think we do 👍🏻

justlevine:
Tests added and ready for final review / merge. @jasonbahl @josephfusco

justlevine (approved):

jasonbahl (approved):

jasonbahl:
@zane-kanopi can you open a new issue with clear steps to reproduce and share any additional  information that might be helpful in debugging? Is it something that can be reproduced with WPGraphQL alone or do you have other plugins like Polylang or WPML that _might_ also be causing things to misbehave? The more info you can provide, the better.

PR #2945, opened 2023-09-30
fix: rename fields registered by connections when using `rename_graphql_field()`

### LINKED ISSUE #1843: rename_graphql_field doesn't work properly for connections
The new `rename_graphql_field` API doesn't work properly for connections. 

## Steps to Reproduce

Add the following code to functions.php or custom plugin code:

```php

  rename_graphql_field( 'RootQuery', 'users', 'wpUsers' );

});
```

### Expectation

The `RootQuery->Users` connections should be named `wpUsers` and I should be able to query: 

```graphql
{
  wpUsers {
    nodes {
       id
       name
    }
  }
}
```

### Actual Behavior

There is no `wpUsers` field added to the Schema. The `RootQuery.users` field remains in the Schema.

## More Info

This appears to be related largely to these filters: 

- https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/WPObjectType.php#L209
- https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/WPObjectType.php#L227
- https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/WPObjectType.php#L239

When the filter to rename fields runs, not all fields (connection fields, for example) have been registered to their connecting Type, so there's no field to rename yet. 

I believe we should formally deprecate the latter 2 filters in favor of the former. This way there's one central filter used to filter fields, instead of 3 different ones. 

This will allow us to better use priorities with filters to determine what's filtered early/later, vs using separate filters altogether.

We should be able to do this in a backward compatible way by having the latter 2 filters hook into the first one, instead of running _after_ the first one. 

There's likely more to the issue than these filters, but I think this is some good starting context.

justlevine:
This PR fixes `rename_graphql_field()` so fields registered by connections are also renamed (e.g. `RootQuery.users` => `RootQuery.wpUsers` ).

This is done by adding a `graphql_wp_connection_{$this->from_type}_from_field_name` filter to the `WPConnectionType::$from_field_name`. 

**Note: This filter is intended for internal use only, and marked as such**.

Additionally, fixes a PHP warning when the supplied field name doesnt exist.

Closes #1843 

Any other comments?
-------------------
See #1843 before closing.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + PHP 8.1.15)

**WordPress Version:** 6.3.1

### DIFF
--- a/src/Type/WPConnectionType.php
+++ b/src/Type/WPConnectionType.php
@@ -150,10 +150,17 @@ class WPConnectionType {
 
 		$this->validate_config( $config );
 
-		$this->config          = $config;
-		$this->from_type       = $config['fromType'];
-		$this->to_type         = $config['toType'];
-		$this->from_field_name = $config['fromFieldName'];
+		$this->config    = $config;
+		$this->from_type = $config['fromType'];
+		$this->to_type   = $config['toType'];
+
+		/**
+		 * Filter the connection field name.
+		 *
+		 * @internal This filter is internal and used by rename_graphql_field(). It is not intended for use by external code.
+		 */
+		$this->from_field_name = apply_filters( "graphql_wp_connection_{$this->from_type}_from_field_name", $config['fromFieldName'] );
+
 		$this->connection_name = ! empty( $config['connectionTypeName'] ) ? $config['connectionTypeName'] : $this->get_connection_name( $this->from_type, $this->to_type, $this->from_field_name );
 
 		/**
--- a/src/Type/WPConnectionType.php
+++ b/src/Type/WPConnectionType.php
@@ -158,6 +158,8 @@ class WPConnectionType {
 		 * Filter the connection field name.
 		 *
 		 * @internal This filter is internal and used by rename_graphql_field(). It is not intended for use by external code.
+		 *
+		 * @param string $from_field_name The name of the field the connection will be exposed as.
 		 */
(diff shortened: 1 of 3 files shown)


justlevine:
cc @TylerBarnes (since not sure if there's WPGatsby implications)

src/Type/WPConnectionType.php
```diff
@@ -150,10 +150,19 @@ public function __construct( array $config, TypeRegistry $type_registry ) {
...
+		$this->to_type   = $config['toType'];
+
+		/**
+		 * Filter the connection field name.
+		 *
+		 * @internal This filter is internal and used by rename_graphql_field(). It is not intended for use by external code.
+		 *
+		 * @param string $from_field_name The name of the field the connection will be exposed as.
+		 */
+		$this->from_field_name = apply_filters( "graphql_wp_connection_{$this->from_type}_from_field_name", $config['fromFieldName'] );
```
jasonbahl, line 164: @justlevine my first thought was that this should pass more context, as it would be difficult to use in just about any way other than the current use case. 

But I think marking it internal like you've done works well. 

If/when a use case arises where a 3rd party might need to filter this, we can consider passing more context to the filter at that time.

jasonbahl (approved):

PR #2947, opened 2023-10-01
Fix typos [DO NOT MERGE]

szepeviktor:
### This is a bug report, not a PR

Found some misspellings.

I let you add more commits.
This is not a PR just a bug report 🐛

PR #2950, opened 2023-10-02
fix: rename typo in component name - `AuthSwitchProvider`

justlevine:
This PR fixes the typo in the `AuthSwitchProvider` component name.

Cherrypicked from @szepeviktor 's work in https://github.com/wp-graphql/wp-graphql/pull/2947

Supercedes #2947

Any other comments?
-------------------
I am not sure whether this should be considered a breaking change since the component is exportable and exposed to a filter, and if it is what the necessary b/c steps would be. @jasonbahl ?

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + PHP 8.1.15 )

**WordPress Version:** 6.3.1

### DIFF
--- a/packages/graphiql-auth-switch/AuthSwitchContext.js
+++ b/packages/graphiql-auth-switch/AuthSwitchContext.js
@@ -7,7 +7,7 @@ export const useAuthSwitchContext = () => {
   return useContext(AuthSwitchContext);
 };
 
-export const AuthSwitchProvder = ({ children }) => {
+export const AuthSwitchProvider = ({ children }) => {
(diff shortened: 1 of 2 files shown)


jasonbahl:
@justlevine since this JS codebase isn't pushed up somewhere that anyone can import from (npm, etc) I don't believe we need to worry about this as a breaking change. 

If it turns out this does break someone's work, I would LOVE to learn how they went about importing this component to their project and what they were doing.

jasonbahl (approved):

PR #2953, opened 2023-10-02
fix: term uri, early return

jasonbahl:
this is a follow-up to #2341


justlevine:
The failing [`testQueryTermLinkCustomHostPortReplacement`](https://github.com/jasonbahl/wp-graphql/blob/c073d118cb67fb4acc817051130a624ff4a87015/tests/wpunit/TermNodeTest.php#L663) is due to the `site_url()` (and rly the `home_ur()`l which it what it probably should be using) being `localhost` on the CI.

(Also, the filtered URL really shouldn't be hardcoded so the suite works in envs where the test domain _isnt_ `localhost`, but thats not relevant here).

I still believe this test case is a [bug that should be addressed directly in faustwp](https://github.com/wp-graphql/wp-graphql/pull/2341#issuecomment-1102972545) ( Tl;Dr they should be filtering the `home_url()` conditionally with `WPGraphQL::is_graphql_request()` instead of most of their [string-replaces](https://github.com/wpengine/faustjs/blob/canary/plugins/faustwp/includes/replacement/callbacks.php)), but if we do need to accommodate this type of behavior, we can use the results of our `wp_parse_url()` call to manually build the `$schema . $host . $post` (and then I guess possibly use regex on the original link so the multisite subdirectory still gets the subdirectory ).

PR #2954, opened 2023-10-02
fix: regression to autoloader for bedrock sites

### LINKED ISSUE #2946: Regression: Missing dependencies notice when WPGraphQL is installed as a composer dependency
Ref: 
              @jasonbahl this change is breaking for me on bedrock (and my assumption is _any_ setup using composer + normal packagist), since the autoloader isnt in `wp-graphql/vendor`.

image

_Originally posted by @justlevine in https://github.com/wp-graphql/wp-graphql/pull/2935#discussion_r1340749589_

jasonbahl:
This refactors the logic around determining whether the plugin can load or not to play nicer with Bedrock.

closes #2946

### DIFF
--- a/wp-graphql.php
+++ b/wp-graphql.php
@@ -29,29 +29,60 @@ if ( ! defined( 'ABSPATH' ) ) {
 // If the codeception remote coverage file exists, require it.
 // This file should only exist locally or when CI bootstraps the environment for testing
 if ( file_exists( __DIR__ . '/c3.php' ) ) {
-	require_once __DIR__ . '/c3.php';
-}
-
-// Whether to autoload the files or not.
-// This must be defined here and not within the WPGraphQL.php because this constant
-// determines whether to autoload classes or not
-if ( ! defined( 'WPGRAPHQL_AUTOLOAD' ) ) {
-	define( 'WPGRAPHQL_AUTOLOAD', true );
-}
-
-// Run this function when WPGraphQL is de-activated
-register_deactivation_hook( __FILE__, 'graphql_deactivation_callback' );
-register_activation_hook( __FILE__, 'graphql_activation_callback' );
-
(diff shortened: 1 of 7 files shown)


wp-graphql.php
```diff
@@ -32,27 +32,59 @@
...
+
+	if ( isset( $composer_config->autoload->files ) ) {
+		foreach ( $composer_config->autoload->files as $file ) {
+			$file_path = __DIR__ . '/' . $file;
+			if ( file_exists( $file_path ) ) {
+				// @phpcs:ignore
+				require_once $file_path;
+			}
+		}
+	}
```
justlevine, line 59: It's only a few files, can we `require_once` them manually instead of pulling in + parsing a json file?

( would also solve `composer.json` including `src/WPGraphQL` both in `files` and in `psr-4` )

wp-graphql.php
```diff
@@ -68,38 +100,27 @@ function graphql_can_load_plugin(): bool {
...
 		// are loaded from the parent project.
-		if ( class_exists( '\WPGraphQL' ) ) {
-			return true;
-		}
+		return true;
 	}
 
-	// If the autoload file exists, load it
-	if ( file_exists( plugin_dir_path( __FILE__ ) . 'vendor/autoload.php' ) ) {
+	// @phpstan-ignore-next-line: this is ignored as the constant could be defined in wp-config, prior to being defined above
```
justlevine, line 106: Instead of ignoring the entire line, I think we can add `define( 'WPGRAPHQL_AUTOLOAD', true );` to `phpstan/constants.php`

wp-graphql.php
jasonbahl: ya, makes sense. I did have it this way in one commit, then reversed it. Seems sensible to re-reverse back to the manual require_once approach.

PR #2966, opened 2023-10-10
ci: Add CodeSpaces config 

moonmeister:
Implements CodeSpaces/Develompment containers config to allow easy dev environment for contributors. 

Don't believe so. 

Any other comments?
-------------------
We could add this to a Contributing doc if we wanted to help surface this for folks.

Where has this been tested?
---------------------------
I've tested it on codespaces. Someone who does local dev could test with development containers.


jasonbahl (changes requested):
see latest comment. 

Also, I think at minimum this should also probably include a README.md in the directory explaining a bit more about how folks can take advantage of this, etc

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+
+services: 
+  development:
+    build:
+      context: .
+      dockerfile: Dockerfile
+  app:
+    depends_on:
+      - app_db
+    image: ghcr.io/wp-graphql/wp-graphql:1.16.0-wp6.3-php8.2
```
justlevine, line 11: We'd ideally want this to be latest and not hardcoded

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    env_file:
+      - ../.env.dist
+    environment:
+      WP_URL: http://localhost:8080/
+      USING_XDEBUG: ${USING_XDEBUG:-}
+    ports:
+      - 8080:80
+
+  app_db:
+    image: mariadb:10.2
```
justlevine, line 24: I think just `mariadb:10` here

.nvmrc
```diff
@@ -1 +1 @@
-v14.19.0
+stable
```
justlevine, line 1: Will this affect our package.json scripts or is everything we're using already built on / compatible with whatever stable is currently?

.devcontainer/docker-compose.yml
moonmeister: I'll check, not sure if we have those tags.

.nvmrc
moonmeister: 20 is current stable. 14 is defunct. If anything they should be more comparable with 20. This won't affect anything in prod as that's all browser code. Usually changing this is very safe for these situations cause the package managers are relatively current. V18 would be a little safer.

justlevine: I'm a PHP dude, so my reference is to that and composer. _IF_ changing the node version also changes what package versions get installed, then we should fix the version of node to whatever and rebuild the plugin assets. 

(This would _not_ be a breaking change I believe, since its only used for building)

.devcontainer/docker-compose.yml
justlevine: It's not the end of the world if not, were already doing a manual string replace when we need to, but it would be nice to reduce the maintenance

.nvmrc
moonmeister: I don't believe it does. `package.json` determines those in node.

.devcontainer/docker-compose.yml
moonmeister: the closest we have is `[latest-wp6.3-php8.2](https://github.com/orgs/wp-graphql/packages/container/wp-graphql-testing/123900207?tag=latest-wp6.3-php8.2)` That atleast covers WPGraphQL which is the important bit, even if it doesn't include the latest WP or PHP. Should we create an issue to have a more general `latest` tag that covers all bits?

moonmeister:
@jasonbahl I blame WordPress absolute urls. What a terrible design decision, 😭 

The problem with Local VS Code access to codespaces gives you a `localhost` domain on all your port forwarding. Web editor access gives you the `app.github.dev` domain. There doesn't seem to be a way to detect this and the ENV vars they give you always provide the public domain. Providing different domains based on location would be helpful but we'd still have an issue where IF I were to switch from web to local the DB would have the wrong domain. I'll work on seeing if we can find a partial solution. Either way, we'll need good docs around this problem.

jasonbahl:
You can put the admin and front end on different domains. It's more complex than absolute vs relative. 

But, if you have a plugin that can solve it, it might be worth trying. . .but it might also cause other problems. 🤷‍♂️

jasonbahl:
@moonmeister this repo looks interesting: https://github.com/codeispoetry/wp-codespace

They take the domain that code spaces provides and write it to the wp-config file

jasonbahl:
Check the postCreateCommand.sh file

moonmeister:
> @moonmeister this repo looks interesting: https://github.com/codeispoetry/wp-codespace
> 
> They take the domain that code spaces provides and write it to the wp-config file
> 
> Check the postCreateCommand.sh file to see how they do it

good call, I've run into this setup before and got some info but I think this might be helpful. The biggest difference is that the script runs on the container you're given terminal access too. For me, that's not the same container I have WP running on. Maybe I need to change that but it made sense to me to have WP and the DB in separate containers and develop in a 3rd cause we don't care about changing WP itself.

moonmeister:
Okay. I just spent the morning going down another dead end. I've tried a couple of things, let me cover them, the issues I'm facing, and the path forward.

## What I've Tried:
1. As-is, this PR only works in a local VS Code instance. I'm really tempted to document this and move on with merging, but I want to try a few things. I have tested it, and easily made this work on the online VSCode instance @ `github.dev`. By using ENV vars to populate the domain. The hard part is knowing which domain to populate. The Codespaces provided vars don't change depending on how you're accessing it. Their built-in hooks for running scripts on startup that Jason's link uses only run on the container you specify for development, not secondary containers that run databases or WP in this case. I could easily run a script while creating the container but I'm not sure that'd fix the issue. Containers are cached and may not have access to domains at build time. Even if they do just cause I "opened" it in the online editor doesn't mean I'm not going to switch to a local VScode instance. Or from local to online. Either way, the domain is still hard coded. I'd have to rebuild and regenerate my DB to fix this.

2. This leads me to want to eliminate domains and make WP use root-relative links in the first place. I have this mostly working with a couple of outliers. I'm not sure if they are codespaces issues or WP issues. Either way, they are probably solvable. But, @justlevine pointed out this could mean serious bugs in WP. Given this functionality isn't native that could mean developing WPGraphQL on it could cause us to miss how things "should" work and introduce bugs when the MU is removed. 

## What's Next...
I used prior art from Local WP to influence my work to get relative links. See https://stackoverflow.com/questions/67502931/wordpress-styles-dont-load-on-ddev-ngrok-live-link/77023869#77023869. They are basically dynamically setting the domain based on headers. What I don't fully understand, is how Local WP accomplishes these domain changes for ngrok as this seemingly won't change the existing values in the DB when the domain changes . But, maybe they've solved that issue in some other way or just ignored it. At this point I need to research to understand how this works more and see if it's a viable solution. 

Any suggestions on other possibilities or insight into how that Local WP snippet works?

.devcontainer/docker-compose.yml
```diff
@@ -12,11 +12,13 @@ services:
     volumes:
       - '..:/var/www/html/wp-content/plugins/wp-graphql'
       - '../.log/app:/var/log/apache2'
+      - './mu-plugins:/var/www/html/wp-content/mu-plugins'
     env_file:
       - ../.env.dist
     environment:
-      WP_URL: http://localhost:8080/
+      WP_URL: 'https://glowing-garbanzo-ppp696r94c7v47-8080.app.github.dev/'
```
jasonbahl, line 19: ya, somehow this needs to be an .env var that codespaces provides. 

Right now I get this experience:

CleanShot 2023-11-14 at 14 27 05

jasonbahl:
@moonmeister I'm going to close this for now. If you end up getting it into a state where it's more generically useful, open up another PR and we can consider it again.

PR #2967, opened 2023-10-10
fix: use all roles for `UserRoleEnum` instead of the filtered `editible_roles`

### LINKED ISSUE #2537: `UserRoleEnum` should use all roles instead of `editible_roles`
### Description

Since WPGraphQL 1.11, trying to access the template property on a page or post throws an internal server error

### Steps to reproduce

Theses queries will return a 500 internal server error on WPGraphQL version 1.11 and 1.11.1. If I roll back to 1.10, the queries will work as expected.

```
query getPages {
  pages {  
    nodes  {
      template {
      __typename
      }
    }
  }
}

query getPage($id:ID!) {
  page(id:$id,idType:DATABASE_ID) {    
      template {
      __typename
      }
  }
}

````

### Additional context

_No response_

### WPGraphQL Version

1.11 and 1.11.1

### WordPress Version

6.0.2

### PHP Version

8.0

Wordpress setup based on Bedrock

- [X] Yes

- [X] Yes

justlevine:
This PR changes `UserRoleEnum` to use _all_ role values instead of just those filtered by [`editable_roles`](https://developer.wordpress.org/reference/hooks/editable_roles/).

The purpose of the `editable_roles` is to limit the roles that _the current user may assign to another user_. As such, it neither represents the roles a privileged user can query for, nor stays the same in the schema across all users.

Additionally, we're using `wp_roles()` instead of directly accessing the global, and early-returning to reduce code complexity.

fixes #2537 

Any other comments?
-------------------
- _Usage_ of `UserRoleEnum` is already validated in the codebase where it is used:
  - [Mutations](https://github.com/justlevine/wp-graphql/blob/32826dc3eb0afb500757a0cbb1ce6bf52d356b1f/src/Data/UserMutation.php#L253)
  - [Connections](https://github.com/justlevine/wp-graphql/blob/32826dc3eb0afb500757a0cbb1ce6bf52d356b1f/src/Data/Connection/UserConnectionResolver.php#L218-L230)
- Doesnt seem to be a need to write a testto cover #2537, since we're already testing the actual UserRoleEnum and no longer using `editable_roles` .

Where has this been tested?
---------------------------

**WordPress Version:** 6.3.1

### DIFF
--- a/src/Type/Enum/UserRoleEnum.php
+++ b/src/Type/Enum/UserRoleEnum.php
@@ -12,30 +12,29 @@ class UserRoleEnum {
 	 * @return void
 	 */
 	public static function register_type() {
-		global $wp_roles;
-		$all_roles      = $wp_roles->roles;
-		$editable_roles = apply_filters( 'editable_roles', $all_roles );
-		$roles          = [];
+		$all_roles = wp_roles()->roles;
+		$roles     = [];
 
-		if ( ! empty( $editable_roles ) && is_array( $editable_roles ) ) {
-			foreach ( $editable_roles as $key => $role ) {
-				$formatted_role = WPEnumType::get_safe_name( isset( $role['name'] ) ? $role['name'] : $key );
+		foreach ( $all_roles as $key => $role ) {
+			$formatted_role = WPEnumType::get_safe_name( isset( $role['name'] ) ? $role['name'] : $key );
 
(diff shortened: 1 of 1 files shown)


justlevine:
(accidentally used the wrong base 🤦repushed)

jasonbahl (approved):

PR #2976, opened 2023-11-01
chore: restore commenting sniffs to PHPCS ruleset

justlevine:
What does this implement/fix? Explain your changes.
---------------------------------------------------
This PR restores the `Squiz.Commenting` PHPCS ruleset, and addresses the resulting smells. 
More specifically: 

- `Squiz.Commenting.BlockComment`, `Squiz.Commenting.DocCommenting.Alignment`, `Squiz.FunctionComment.SpacingAfterParam`, and `Squiz.Commenting.InlineComment.NoSpaceBefore` were autofixed.
- added missing and fixed existing`@var` tags, `@params` and their types, `@throws` tags.
- fixed bad file comment formatting
- (smaller changes are listed as individual commits).

See additional comments for the remaining `Squiz.Commenting` smells that are now explicitly excluded.

**No production code has been changed in this PR**

Does this close any currently open issues?
------------------------------------------

Any relevant logs, error output, GraphiQL screenshots, etc?
-------------------------------------

Any other comments?
-------------------
- `Squiz.Functions.MultiLineFunctionDeclaration.SpaceAfterFunction` has been removed from the ruleset as it is already included by `WordPress`.
- `Generic.Commenting.DocComment.MissingShort` has been reclassified as "Should probably be added back".
- `Squiz.Commenting.FunctionComment.IncorrectTypeHint` throws false positives when used in conjunction with FCQN. which is significantly more important in terms of IDE hinting.
- `Squiz.Commenting.ClassComment.Missing`, `Squiz.Commenting.FileComment.Missing`. These don't affect IDE hinting, and a lot of effort to add.
- `Squiz.Commenting.FunctionComment.EmptyThrows`, `Squiz.Commenting.FunctionComment.MissingParamComment`. While adding param explanations would be nice, this would be a lot of work to remediate. Saving for a future PR.
- `Squiz.Commenting.InlineComment.InvalidEndChar`, would be a lot of work to remediate for only formatting.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + php8.1.15)

**WordPress Version:** 6.3.2

PR #2978, opened 2023-11-01
chore: implement stricter type hinting with `SlevomatCodingStandard.TypeHints` [phpcs]

justlevine:
This PR implements stricter phpdoc type hinting, and adds the `SlevomatCodingStandard.TypeHints`.

The primary change here is replacing the use of `array` with a traversable type (e.g. `string[]`, `array<int|string,string|array<mixed>>` etc). Additionally, more overloaded methods now use `{@inheritDoc}`, and there are some auto-fixable formatting patterns (e.g. using the short type name instead of the long syntax) to make the codebase easier to dedupe/refactor.

As a result of these changes, we were also able to change the phpstan ruleset to scan for valid traversable types.

**Note: No production code has been changed.**

Based on (and requires) #2976 . ([Filtered diff of this PR](https://github.com/wp-graphql/wp-graphql/pull/2978/files/8c8b0958ecc46817eadf12cb7008171043be41e7..f562b373de8e3b0990827416fea81772905c2af7))

Any other comments?
-------------------
- No real attempts were made to decrease type specificity from unions or `mixed` types. This sort of refactor work would be a huge code quality win but is a huge endeavor that will require code changes and should be broken up over several PRs.
- The fixed types will likely cause phpcs/phpstan smells in downstream projects, but _this is a good thing_ as it means they're now being notified that they're doing something that was _already_ wrong. Since only phpdocs have been altered these are _not_ breaking changes.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + php8.1.15)

**WordPress Version:** 6.3.2

### DIFF
--- a/access-functions.php
+++ b/access-functions.php
@@ -86,10 +86,10 @@ function graphql_format_type_name( $type_name ) {
 /**
  * Provides a simple way to run a GraphQL query without posting a request to the endpoint.
  *
- * @param array $request_data   The GraphQL request data (query, variables, operation_name).
- * @param bool  $return_request If true, return the Request object, else return the results of the request execution
+ * @param array<string,mixed> $request_data   The GraphQL request data (query, variables, operation_name).
+ * @param bool                $return_request If true, return the Request object, else return the results of the request execution
  *
- * @return array|\WPGraphQL\Request
+ * @return array<string,mixed>|\WPGraphQL\Request
  * @throws \Exception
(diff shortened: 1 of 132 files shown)


jasonbahl (approved):

justlevine:
sorry bout that, rebased to handle the merge conflict

PR #2988, opened 2023-11-10
feat: add missing extra fields to `EnqueuedAsset`, `EnqueuedScript` and `EnqueuedStylesheet`

### LINKED ISSUE #2938: Wrong data being returned from `EnqueuedAsset.extra` resolve function
### Description

If you are trying to query about the extra information of an enqueued asset whether it being a style or a script, you get the wrong results.

Here are the reasons for the error. 
* the resolve function for the `extra` field tries to return a string, meanwhile it's an supposed to be an object.
* the function is also set to return only one property `extra.data` and this property only exists on scripts, making it hard to extract other information.
Also, there are other fields other than extra.data which whiles on the object so by returning only `extra.data`, we are unable to access the other data.

Here's proof.

image
Over here, you can see it's returning null for the data.

image
By using GraphQL debug, we can verify that many properties exist on the extra field but it's being forced to return a string

### Steps to reproduce

1) Use this query on any random posts

```
{
  post(idType:DATABASE_ID, id:43){
    slug,
    enqueuedStylesheets {
      edges{
      	node{
          dependencies {
            id
          },
        	extra  
        }
      }
    }
  }
}
```

### Additional context

_No response_

### WPGraphQL Version

Version 1.16.0

### WordPress Version

6.3.1

### PHP Version

PHP version 8.2.4

_No response_

- [X] Yes

- [X] Yes

justlevine:
The PR registers GraphQL fields for script/style data stored in the `$extra` property on the instance. More specifically:

- `EnqueuedAsset`
  - added `after` and `before` fields, for code that is supposed to be inlined before/after the asset is called.
  - added `conditional` field, for HTML comment conditionals
  - deprecated `extra` in favor of `EnqueuedScript.extraData`.
  - deprecated `args` in favor of `EnqueuedStylesheet.media`.
  - fixed `dependency` to be an `EnqueuedAsset` instead of an `EnqueuedScipt`.
- `EnqueuedScript`
  - added `extraData` for inline js data passed to the script (previously `EnqueuedAsset.extra)
  - added `strategy`  and the corresponding `ScriptLoadingStrategyEnum`.
- `EnqueuedStylesheet`
  - Added `isRtl`, `media` `path`, `suffix`, and `title`
  - Added `rel`, which is derived from `$style->extra['alt']` since it seemed clunky to use an `isAlt` to force users to know they need to do `rel={isAlt ? 'alternative stylesheet' : 'stylesheet' }`

Fixes #2938 

Any other comments?
-------------------

Where has this been tested?
---------------------------
**Operating System:** Ubunut 20.04 (wsl2 + devilbox + php 8.1.15)

**WordPress Version:** 6.4.1


jasonbahl (dismissed):

src/Type/InterfaceType/EnqueuedAsset.php
```diff
@@ -41,38 +41,86 @@ public static function register_type( TypeRegistry $type_registry ) {
...
+						'resolve'     => static function ( \_WP_Dependency $asset ) {
+							if ( ! isset( $asset->extra['conditional'] ) || ! is_string( $asset->extra['conditional'] ) ) {
+								return null;
+							}
+
+							return $asset->extra['conditional'];
+						},
+					],
+					'dependencies' => [
+						'type'        => [ 'list_of' => 'EnqueuedScript' ],
```
jasonbahl, line 97: dependencies could be a `Script` or a `Stylesheet`, eh?

```suggestion
						'type'        => [ 'list_of' => 'EnqueuedAsset' ],
```

justlevine: I think you're right 😁,

That said, keep in mind that the `dependency` field wasn't changed (it's in the diff because I sorted the field list).

Also, not sure if this needs to be noted as a breaking change. It looks like all the previously-available `EnqueuedScript` are directly available from the `EnqueuedAsset, and I _think_ a `fragment MyFrag on EnqueuedScript` will work spread on an `EnqueuedAsset` field...

(Leaving this as is so you can review. Maintainer commits are enabled on this PR, so if you're good with the above, feel free to `commit suggestion` and merge no need to wait on me).

jasonbahl: Ya, this should be considered a bugfix. I'll callout in the release notes a filter should anyone feel like they need the previous behavior.

justlevine: updated the PR description to include this change ✔️

jasonbahl: Actually, I think we need to make one modification. The Interface says the dependency will be a list_of EnqueuedAsset, but each Type's dependencies field should explicitly return a list_of => EnqueuedScript or list_of => EnqueuedStylesheet. 

Will update.

justlevine: Wont that cause a schema conflict when querying the dependency field for both child types via the interface?

jasonbahl: @justlevine oh maybe. . . 🤔 

When registering a stylesheet, can it have a JS dependency? And when registering a Script can it have a CSS dependency? 

I feel like I should already know this (and am going to research the answer) but I _thought_ `register_script` dependencies were other scripts and `register_stylesheet` dependencies were always other stylesheets?? 👀

jasonbahl: CleanShot 2023-11-14 at 14 18 20

jasonbahl: CleanShot 2023-11-14 at 14 19 22

jasonbahl: @justlevine ya, looks like, at least according to docs, the dependencies of a script will always be other scripts, and the dependencies of a stylesheet will always be another stylesheet.

justlevine: Yeah, the deps IRL are always same-type, but because `EnqueuedAsset` is an interface not a union, fields (afaik) with the same name across children need to be defined as the same type and not just another type on a shared interface.

_If so_, then imo the more verbose `...on Enqueued{Script|Stylesheet}`, is probably preferred to a solution that breaks reusable fragments and leads to schema unpredictability by forcing devs to use their own field alias.

jasonbahl: @justlevine nah, we're good here. 

The Interface `EnqueuedAsset` defines a `dependencies` field that needs to return a `list_of => EnqueuedAsset`

Then the specific `EnqueuedScript.dependencies` returns a `list_of => EnqueuedScript` (which, since EnqueuedScript implements the EnqueuedAsset interface, it passes for properly returning `list_of => EnqueuedAsset`) and same for `EnqueuedStylesheet.dependencies` returning a `list_of => EnqueuedStylesheet`

This is how the `Connection` Interface works. 

Anything that is a `Connection` needs to have a field `nodes` that returns a `list_of => 'Node'` and so any connection that returns `Post` or `Page` or `User` is valid, because all of those things a are a `list_of => Node` too 🙌🏻

PR #3008, opened 2024-01-01
perf: Expose `graphql_should_analyze_queries` as setting

justlevine:
This PR adds an admin setting for enabling/disabling the Query Analyzer, and defaults the setting value to `off` on new installs, unless GraphQL Debugging is enabled. In other words **by default, the QueryAnalyzer will only be enabled when GraphQL Debugging is also enabled**.

The reason for this is the overhead caused by having QueryAnalyzer enabled. While this cost is justified during debugging, or when needed to cache future responses, on vanilla WPGraphQL setups there is no need to pay for that extra processing.

More specifically:
- Adds a new `query_analyzer_enabled` setting to the WPGraphQL Settings page. The setting control is _disabled_ if `GraphQL::debug()` is true.
- Adds a new `QueryAnalyzer::is_enabled()` _static_ method that can be used to check whether the QA is disabled site wide. The return value in order of priority is:
  - GraphQL Debugging enabled => `true`
  - The filtered value of the existing `graphql_should_analyze_queries` filter.
  - The stored setting value.
- Adds a new `QueryAnalyzer::is_enabled_for_query()` _instance_ method that uses a new `graphql_should_analyze_query` filter which passes the `\WPGraphQL\Request` object as a parameter, and can be used to enable/disable the QueryAnalyzer for specific requests.
- Sets the default `query_analyzer_enabled` value of _`true`_ for existing WPGraphQL installs on upgrade, to prevent back-compat issues.
- Outputting the data to `response.extensions` now requires _both_ `QueryAnalyzer::is_enabled_for_query()` and `\WPGraphQL::debug()` to be true.

Related #1873 

Per the below screenshots TTFB decreased from `165.68ms` => `125.93ms` on a simple query with very few types to analyze.

### Before (Query Analyzer enabled by default)
qa-enabled

### After (Query Analyzer disabled by default)
qa-disabled

### Settings toggle:
- GraphQL Debugging enabled
image

- GraphQL Debugging disabled
image

Any other comments?
-------------------
> [!IMPORTANT]
> Plugins that rely on the Query Analyzer should update their code with `add_action( 'graphql_should_analyze_queries', true );` or do a check around the `QueryAnalyzer::is_enabled()` method.

- I went with a Setting Label of `Enable GraphQL Type Tracking`, since "Query Analyzer" is the internal feature name, and not self-documenting for end users.

- Made the assumption that this will be part of the 1.20.0 release. If not the `activation.php` method should be updated.

- The future potential of `graphql_should_analyze_query` should make it easier to exclude or set custom rules around caching specific types/models/or even named queries. 

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2 + devilbox + 8.1.15)

**WordPress Version:** 6.4.2

### DIFF
--- a/activation.php
+++ b/activation.php
@@ -12,40 +12,6 @@ function graphql_activation_callback() {
 		return;
 	}
 
-	// Run any activation scripts before updating the version.
-	graphql_migrate_1_20_0();
-
 	// store the current version of WPGraphQL
 	update_option( 'wp_graphql_version', WPGRAPHQL_VERSION );
-}
-
-
-/**
- * Handles compatibility when updating from pre-v1.20.0 versions of WPGraphQL.
- *
- * @todo Remove this function in v2.0.0, when the default value for `query_analyzer_enabled` is set to `true`.
- */
-function graphql_migrate_1_20_0(): void {
-	// If the version is already set, we don't need to do anything.
-	$version = get_option( 'wp_graphql_version' );
-	if ( ! $version ) {
-		return;
-	}
-
-	// If the previous version is higher than 1.20.0, we don't need to do anything.
-	if ( version_compare( $version, '1.20.0', '>=' ) ) {
-		return;
-	}
-
-	/**
-	 * Set `query_analyzer_enabled` to `true` for preexisting installs.
-	 *
-	 * This is to prevent breaking changes in caching solutions that rely on the Query Analyzer, but aren't insuring that the `graphql_should_analyze_queries` filter is set to true.
-	 */
-
-	$graphql_settings = get_option( 'graphql_general_settings', [] );
-
-	$graphql_settings['query_analyzer_enabled'] = 'on';
-
-	update_option( 'graphql_general_settings', $graphql_settings );
 }
--- a/activation.php
+++ b/activation.php
@@ -12,6 +12,40 @@ function graphql_activation_callback() {
 		return;
 	}
 
+	// Run any activation scripts before updating the version.
+	graphql_migrate_1_20_0();
+
 	// store the current version of WPGraphQL
 	update_option( 'wp_graphql_version', WPGRAPHQL_VERSION );
+}
+
+
+/**
+ * Handles compatibility when updating from pre-v1.20.0 versions of WPGraphQL.
+ *
+ * @todo Remove this function in v2.0.0, when the default value for `query_analyzer_enabled` is set to `true`.
+ */
+function graphql_migrate_1_20_0(): void {
+	// If the version is already set, we don't need to do anything.
+	$version = get_option( 'wp_graphql_version' );
+	if ( ! $version ) {
+		return;
+	}
+
+	// If the previous version is higher than 1.20.0, we don't need to do anything.
+	if ( version_compare( $version, '1.20.0', '>=' ) ) {
+		return;
+	}
+
+	/**
+	 * Set `query_analyzer_enabled` to `true` for preexisting installs.
+	 *
+	 * This is to prevent breaking changes in caching solutions that rely on the Query Analyzer, but aren't insuring that the `graphql_should_analyze_queries` filter is set to true.
+	 */
+
+	$graphql_settings = get_option( 'graphql_general_settings' );
+
+	$graphql_settings['query_analyzer_enabled'] = 'on';
+
+	update_option( 'graphql_general_settings', $graphql_settings );
 }
--- a/activation.php
+++ b/activation.php
@@ -43,7 +43,7 @@ function graphql_migrate_1_20_0(): void {
 	 * This is to prevent breaking changes in caching solutions that rely on the Query Analyzer, but aren't insuring that the `graphql_should_analyze_queries` filter is set to true.
 	 */
 
-	$graphql_settings = get_option( 'graphql_general_settings' );
+	$graphql_settings = get_option( 'graphql_general_settings', [] );
 
 	$graphql_settings['query_analyzer_enabled'] = 'on';
 
--- a/src/Router.php
+++ b/src/Router.php
@@ -3,7 +3,6 @@
 namespace WPGraphQL;
 
 use GraphQL\Error\FormattedError;
-use WPGraphQL\Utils\QueryAnalyzer;
 use WP_User;
 
 /**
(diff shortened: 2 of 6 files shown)


renatonascalves:
I like this! I thought this was `off` by default.

jasonbahl:
Ok, so this will _technically_ be a breaking change as this feature is needed for folks that are querying via GET and have a caching client (Varnish, etc) or otherwise doing something with the headers that are returned by X-GraphQL-Keys. . .or when folks are using POST requests and have WPGraphQL Smart Cache object cache enabled. 

Before merging this, we need to make sure WPGraphQL Smart Cache ensures QueryAnalyzer is "ON" for GET requests or when GraphQL Object cache is available. If that's not already enforced by WPGraphQL Smart Cache, we need to make sure it is and communicate well that WPGraphQL Smart Cache users NEED to update to the version that does that before updating WPGraphQL to the version that switches the default to "OFF"

jasonbahl:
@justlevine lol, I just realized you already called this out: 

CleanShot 2024-01-08 at 10 46 22

🤦🏻‍♂️

justlevine:
> @justlevine lol, after writing my comment above, I realized you already called this out:

@jasonbahl sure, but I wasn't as detailed, and seeing it written out makes me worry about the compatibility for non-Smart Cache solutions (this would be a patch change in Smart Cache, so anyone with autoupdates enabled would get both updates and keep compatibility seamless).

Perhaps as a way to preserve back-compat, we can expose this as a setting and only default it to `false` on new installs. It'l be tech debt come v2.0, but it gets this into the codebase faster than waiting for me to finish the extension-compatibility update checker 🤔

jasonbahl:
> Perhaps as a way to preserve back-compat, we can expose this as a setting and only default it to false on new installs

@justlevine this seems sensible to me

justlevine:
@jasonbahl Changes made and the PR description updated with the specifics of the new implementation.

I still need to update/add the corresponding WPUnit tests, but in the interim please let me know if you like the new proposed solution.

activation.php
```diff
@@ -12,6 +12,40 @@ function graphql_activation_callback() {
...
+		return;
+	}
+
+	/**
+	 * Set `query_analyzer_enabled` to `true` for preexisting installs.
+	 *
+	 * This is to prevent breaking changes in caching solutions that rely on the Query Analyzer, but aren't insuring that the `graphql_should_analyze_queries` filter is set to true.
+	 */
+
+	$graphql_settings = get_option( 'graphql_general_settings' );
```
jasonbahl, line 46:
```suggestion
	$graphql_settings = get_option( 'graphql_general_settings', [] );
```

jasonbahl: I don't think there's a guarantee that `graphql_general_settings` will have any value stored yet, so get_option would return `false` by default and then we'd try to add a key/value to `false` below. 

This ensures that the default empty response, if there are no values returned, is an array.

jasonbahl (dismissed):

jasonbahl:
>@jasonbahl Changes made and the PR description updated with the specifics of the new implementation.

>I still need to update/add the corresponding WPUnit tests, but in the interim please let me know if you like the new proposed solution.

@justlevine ya, this looks good! Thanks 🙏🏻

src/Router.php
```diff
@@ -327,7 +326,7 @@ protected static function get_response_headers() {
 
 		// If the Query Analyzer was instantiated
 		// Get the headers determined from its Analysis
-		if ( self::get_request() instanceof Request && self::get_request()->get_query_analyzer() instanceof QueryAnalyzer ) {
+		if ( self::get_request() instanceof Request && self::get_request()->get_query_analyzer()->is_enabled() ) {
```
justlevine, line 329:
```suggestion
		if ( self::get_request() instanceof Request && self::get_request()->get_query_analyzer()->is_enabled_for_query() ) {
```

justlevine:
Overall coverage drop is caused by the increased settings config array, but the actual code-logic coveraged has gone up:

image

activation.php
```diff
@@ -12,6 +12,40 @@ function graphql_activation_callback() {
 		return;
 	}
 
+	// Run any activation scripts before updating the version.
+	graphql_migrate_1_20_0();
```
jasonbahl, line 16: Does this run for folks that update via composer too? 🤔

justlevine: Now that you mention it it might not even work on plugin _update_ and only on actual activation. 🤦

Not sure the best approach here:
- If it's non-composer, we could conditional it as a run_once on `admin_init`, but someone updating via composer would still need to log in. 
- Or we absorb the db version check on every `init`, the extra potential hit is annoying, but still less than with QA activated.
- Or we just keep the setting and filter logic, and wait until 2.0 to switch the the default behavior to `off`.


@jasonbahl thoughts?

jasonbahl: @justlevine perhaps in the `is_enabled_for_query` we can add some logic to handle folks that might not have triggered the `migrate` script. 

Like, when we determine if GRAPHQL_DEBUG is enabled, etc. . .

If we are able to determine no value has been set in the db for `query_analyzer_enabled` we can assume the migrate script hasn't run and/or the user hasn't manually saved the settings yet, so we can have some logic there to compensate for users that didn't trigger the migrate script?

justlevine:
Scope of this PR was reduced to just exposing the toggle/filters/method calls.

A mechanism for updating old DB values on migration will be evaluated/handled independently, to avoid the risks of tech debt by having the solution coupled [too deeply to this PR's use case](https://github.com/wp-graphql/wp-graphql/pull/3008#discussion_r1447726133).

@jasonbahl ready for your review again 🤞

jasonbahl (approved):

jasonbahl:
@justlevine looks good. 

I'll open a corresponding Issue and PR to handle enabling when WPGraphQL Smart Cache is active.

jasonbahl:
@justlevine fwiw https://github.com/wp-graphql/wp-graphql-smart-cache/issues/268

While the default will be "on" as of this PR, since it can easily be checked "off" I opened this issue for WPGraphQL Smart Cache to make it more difficult to turn of the analyzer as it would have negative impact for WPGraphQL Smart Cache users if it were disabled.

PR #3013, opened 2024-01-06
feat: output GRAPHQL_DEBUG message if requested amount is larger than connection limit

### LINKED ISSUE #3012: No error thrown when query value first exceeds max query amount
### Description

When a query is made exceeding the max query amount, no error is thrown to indicate that the value of `first` exceeds the amount. Instead, the query succeeds, but returns only the max query amount.

We expected some type of error, since the max query amount is not visible for a user of the API.

https://www.wpgraphql.com/filters/graphql_connection_max_query_amount

### Steps to reproduce

Reproducible in repl:

```
{
  posts(first:10000000){
    nodes {
      id
    }
  }
}
```

```
{
  "data": {
    "posts": {
      "nodes": [
        {
          "id": "cG9zdDox"
        }
      ]
    }
  },
  "extensions": {
    "debug": [
      {
        "type": "DEBUG_LOGS_INACTIVE",
        "message": "GraphQL Debug logging is not active. To see debug logs, GRAPHQL_DEBUG must be enabled."
      }
    ]
  }
}
```

### Additional context

This could be part of a larger conversation around query costs, which is how Shopify handles this issue with their GraphQL API.

https://shopify.engineering/rate-limiting-graphql-apis-calculating-query-complexity

```
{
  "errors": [
    {
      "message": "Query cost is 2002, which exceeds the single query max cost limit (1000).\n\nSee https://shopify.dev/concepts/about-apis/rate-limits for more information on how the\ncost of a query is calculated.\n\nTo query larger amounts of data with fewer limits, bulk operations should be used instead.\nSee https://shopify.dev/api/usage/bulk-operations/queries for usage details.\n",
      "extensions": {
        "code": "MAX_COST_EXCEEDED",
        "cost": 2002,
        "maxCost": 1000,
        "documentation": "https://shopify.dev/api/usage/rate-limits"
      }
    }
  ]
}
```

### WPGraphQL Version

1.19.0

### WordPress Version

N/A

### PHP Version

N/A

_No response_

- [X] Yes

- [X] Yes

justlevine:
This PR adds a debug message to the GraphQL response when the requested amount (`first`, `last`) is greater than the `graphql_connection_max_query_amount` filter value set on the server.

### Why 
This addition provides developers with an important indication that they're not actually getting the full results requested, which is particularly important to new users coming from traditional WP-PHP for whom setting `WP_Query( [ 'posts_per_page' => -1 ] )` is discouraged, but commonplace.

A debug message was chosen over throwing a `GraphQLError`, both to preserve back-compat, and because IMO this is the ideal GraphQL pattern: a too-large `first` isnt a schema error, it's something that gets filtered on the backend (ostensibly even based on user role). 

More importantly, it keeps the frontend code reusable and server-agnostic. E.g. a pagination pattern uses the last returned item to fetch the next page of results no matter what the server-limit is, whereas if a `GraphQLError` was thrown, users would need to manually update their frontend code and eschew component libraries / framework starters).

Closes #3012 

image

Any other comments?
-------------------
…

Where has this been tested?
---------------------------

**WordPress Version:** 6.4.2

### DIFF
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -415,8 +415,9 @@ abstract class AbstractConnectionResolver {
 	/**
 	 * Get_query_amount
 	 *
-	 * Returns the max between what was requested and what is defined as the $max_query_amount to
-	 * ensure that queries don't exceed unwanted limits when querying data.
+	 * Returns the max between what was requested and what is defined as the $max_query_amount to ensure that queries don't exceed unwanted limits when querying data.
+	 *
+	 * If the amount requested is greater than the max query amount, a debug message will be included in the GraphQL response.
 	 *
 	 * @return int
 	 * @throws \Exception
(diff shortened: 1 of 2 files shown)


justlevine:
(relabeled the PR as a `feat` because of the linter)

jasonbahl (approved):

Issue #2981, opened 2023-11-04
PHP deprecated


xbd63:
### Description

Hi, I’m getting this error when a GraphQL request is made. This is after updating from PHP 7.3 to PHP 8.2.

Error Level: E_DEPRECATED
Message: Creation of dynamic property WPGraphQL\AppContext::$connection_query_class is deprecated
/bitnami/wordpress/wp-content/plugins/wp-graphql/src/Type/WPConnectionType.php
Line: 528

This is the request: /graphql?query=query%20{%20posts(first:%2032,%20where:%20{mpid1:%20%22BBULRDSL000%22})%20{%20nodes%20{%20clientId%20databaseId%20title%20content%20featuredImage%20{%20node%20{%20id%20sourceUrl%20}%20}%20postFields%20{%20fieldGroupName%20galleries%20imageAltText%20isVideo%20matchedProducts%20{%20gbLink%20gbPrice%20gbTitle%20productId%20productName%20productPrice%20productUrl%20sku%20x%20y%20}%20socialNetwork%20sourceUrl%20}%20}%20}%20}

It's a request for some posts using ACF fields.

The correct JSON is returned, but this error persists.

WordPress version 6.3.2
Plugins:
Advanced Custom Fields PRO
Forminator Pro
Really Simple SSL
WooCommerce + related plugins
WP GraphQL
WPMU Dev Dashboard

Thanks

### Steps to reproduce

Execute a query with similar structure to this: /graphql?query=query%20{%20posts(first:%2032,%20where:%20{mpid1:%20%22BBULRDSL000%22})%20{%20nodes%20{%20clientId%20databaseId%20title%20content%20featuredImage%20{%20node%20{%20id%20sourceUrl%20}%20}%20postFields%20{%20fieldGroupName%20galleries%20imageAltText%20isVideo%20matchedProducts%20{%20gbLink%20gbPrice%20gbTitle%20productId%20productName%20productPrice%20productUrl%20sku%20x%20y%20}%20socialNetwork%20sourceUrl%20}%20}%20}%20}

(using ACF fields)

### WPGraphQL Version

1.18.0

### WordPress Version

6.3.2

### PHP Version

8.2

### Additional environment details

Bitnami on AWS

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

### Please confirm that you have disabled ALL plugins except for WPGraphQL.

- [X] Yes
[...]

PR #3027, opened 2024-01-29
feat: Add Admin Notice for WPGraphQL for ACF

josephfusco:

This introduces a new `register_graphql_admin_notice()` API that can be used to display an admin notice on the Plugins page, or on the WPGraphQL Settings or GraphiQL IDE page. 

This can be used by core WPGraphQL or for extensions to provide notices scoped to WPGraphQL pages.

It allows for notices to be permanent or dismissable and handles nonces for dismissing, storing a flag of dismissed notices in user_meta so that users won't be bugged again by the same notice after they've dismissed it.

no

Any other comments?
-------------------

Documentation: https://www.wpgraphql.com/functions/register_graphql_admin_notice

example usage:

```php
add_action( 'plugins_loaded', function() {

  if ( function_exists( 'register_graphql_admin_notice' ) ) {
    $config = [
      
      // required. string. The message to display in the notice. If no message, the notice won't display.
      'message' => __( 'Your notice message', 'your-textdomain' ), 
      
      // optional. boolean. whether the notice should be dismissable. False to be non-dismissable.
      'is_dismissable' => true, 
     
      // optional. string. The type of message. warning, success, error, info
      'type' => 'info', 
      
      // optional. callback. return true to display the notice (i.e. only display if another plugin is active)
      'conditions' => function() { 
         // if xx == true, return true
         // if xx == false, return false  
       }
    ];
    
    register_graphql_admin_notice( 'my-notice-slug', $config );
  }

} );
```
CleanShot 2024-02-02 at 16 58 50


justlevine:
> I tested this by replacing the Admin.php file in our live on our system yesterday, and the deprecation warnings still persist today.

1. This is a _draft_ PR, so regardless it shouldn't be expected to work in its current state.

3. The purpose this PR is to display an _admin notice_ on the WordPress dashboard. It has nothing to do with #2981 or any other PHP 8.3 compatibility work.

xbd63:
Ok thanks for explaining that.

On the WP GraphQL I do see a notice "Check out the new WPGraphQL for ACF".

Please let me know if I need to do anything else.

access-functions.php
```diff
@@ -953,3 +953,16 @@ function str_ends_with( string $haystack, string $needle ): bool {
 		return $needle_length <= strlen( $haystack ) && 0 === substr_compare( $haystack, $needle, -$needle_length );
 	}
 }
+
+/**
+ * @param string       $slug A unique slug to identify the admin notice by
```
justlevine, line 958:
```suggestion
 * Registers a configurable notice on the WordPress Dashboard for WPGraphQL-related notices.
 *
 * @param string       $slug A unique slug to identify the admin notice by.
```

access-functions.php
```diff
@@ -953,3 +953,16 @@ function str_ends_with( string $haystack, string $needle ): bool {
...
+ * @param array<mixed> $config The config for the admin notice. Determines visibility, context, etc.
```
justlevine, line 959:
```suggestion
 * @param array<string,mixed> $config The config for the admin notice. Determines visibility, context, etc.
```

src/Admin/Admin.php
```diff
@@ -43,6 +43,9 @@ public function init() {
 		$this->admin_enabled    = apply_filters( 'graphql_show_admin', true );
 		$this->graphiql_enabled = apply_filters( 'graphql_enable_graphiql', get_graphql_setting( 'graphiql_enabled', true ) );
 
+		$admin_notices = new AdminNotices();
```
justlevine, line 46: This should probably be wrapped in an `is_admin()` check. I don't think we're doing that earlier in the lifecycle but I might be wrong (reviewing via 📱)

src/Admin/AdminNotices.php
```diff
@@ -0,0 +1,299 @@
+<?php
+
+namespace WPGraphQL\Admin;
+
+/**
+ * This class isn't intended for direct extending or customizing.
```
justlevine, line 6: We should mark the class `@internal` and not just leave it in text.

src/Admin/AdminNotices.php
justlevine, line 1: Can we add a method to this for validation the config?

justlevine (commented):
Did a quick pass, will review more in depth tomorrow and also probably make those array types more specific

composer.json
```diff
@@ -47,7 +47,8 @@
 		"slevomat/coding-standard": "^8.9",
 		"szepeviktor/phpstan-wordpress": "~1.3.0",
 		"wp-graphql/wp-graphql-testcase": "~2.3",
-		"wp-cli/wp-cli-bundle": "^2.8"
+		"wp-cli/wp-cli-bundle": "^2.8",
+		"10up/wp_mock": "^0.5.0"
```
justlevine, line 51: Is this PR using this library somewhere?

jasonbahl: Ah! I was using it when drafting the tests but then decided not to.

src/Admin/AdminNotices.php
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	/**
+	 * Throw an error if the config is not valid.
+	 *
+	 * @since @TODO
+	 *
+	 * @param array<string,mixed> $config The config of the admin notice
+	 *
+	 * @return bool
```
jasonbahl, line 137:
```suggestion
	 * @param array<string,mixed> $config The config of the admin notice
```

Issue #3019, opened 2024-01-16
Let's Revamp Our IDE 🚀


josephfusco:
# Hey WPGraphQL Community! Let's Revamp Our IDE 🚀

## What's Cooking?

We're sprucing up our WPGraphQL IDE - think of it as giving it a fresh coat of paint and some cool new gadgets. Why? Because we want to make your development experience smoother, more intuitive, and totally in sync with the latest in the GraphQL world.

## Why This Rocks

- **Less Hassle, More Hustle**: We're switching to the official GraphiQL package. Less maintenance for us, more cool stuff for you.
- **Keeping Up with the GraphQL Joneses**: We're not just updating; we're aligning ourselves with the latest and greatest in the GraphQL universe.
- **Future-Proofing**: We're setting the stage for awesome new features down the line. Think custom scalars and beyond!
- **Everywhere, All the Time**: Want to use the IDE on the front-end, in wp-admin, or on the moon? (Okay, maybe not the moon). We've got you covered.
- **Test It Like You Mean It**: We know, the current E2E tests are a bit of a bummer. We're on it!

## Who Benefits?

All of us! Whether you're a WPGraphQL newbie or a seasoned pro, these updates are going to make your life easier.

## What's New and What's Going?

- **Goal**: We're not reinventing the wheel. We want to match what we've got, but with the new GraphiQL goodness.
- **Saying Goodbye**: No more `graphiql_explorer_operation_action_menu_items` and the old navy blue sidebar. It's time for something sleeker.
- **Routes and More**: 
    - Your usual GraphQL server settings? They're staying put.
    - The GraphiQL IDE page is getting a major facelift. Think minimalism, more power under the hood.
    - Help pages? Let's just link to the docs and keep things tidy.

## Join the Party 🎉

Got ideas, feedback, or just want to see what's happening? Jump into the [WPGraphQL IDE GitHub repo](https://github.com/wp-graphql/wpgraphql-ide). Your voice matters!

## The Endgame

We merge this baby into the core once we're all happy with it. The old repo gets a fond farewell, and we document any big changes to make your transition smooth.

---

We're super excited about this, and we hope you are too! Let's make WPGraphQL IDE better, together. 🚀💻🎉

jasonbahl:
Comments are locked on this. 

We'll be discussing technical details, feature requests, etc over on the WPGraphiQL IDE repo. 

👉🏻  https://github.com/wp-graphql/wpgraphiql-ide

[...]

PR #3031, opened 2024-01-31
fix(graphiql): Allow GraphiQL to run even if a valid schema cannot be returned.

linucks:
Currently GraphiQL won't load and just displays a spinner if there is an error generating a schema. An error is logged to the console, but it gives almost no indication of what the error was.

Even if a valid schema cannot be loaded, GraphiQL can still be used to make queries, and in the case of a broken schema, GraphiQL could then be used to determine which queries were failing and see why the schema was broken.

This fix allows GraphiQL to load, and displays the error message created in generating the schema in the console, so that a developer can see what caused the error.

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** Wordpress Docker image sha256:14425bb2eae98ef05853f78e2c8bc3d91cff3b465f2933c6f105ac697facb3ea

**WordPress Version:** 6.4.2


justlevine:
Thanks so much for this PR @linucks - this has personally been on my wishlist for a long time 🙌

cc: @josephfusco #3019

jasonbahl (approved):
@linucks I tested this change and I like it! 

We'll be sure to provide a smoother experience like this when the Schema fails to load in the new version of the WPGraphQL IDE we're building here: https://github.com/wp-graphql/wpgraphql-ide I'd love to take this a step fu

jasonbahl:
fwiw, a failed schema can currently be tested with the following: 

```php
<?php

	register_graphql_mutation( 'testMutation', [
		'inputFields' => [ 'test' => [ 'type' => 'String' ] ],
		'outputFields' => [ 'test' => [ 'type' => 'NonExistantType' ] ], // <-- this will cause the schema to fail to load
		'mutateAndGetPayload' => function( $input, $context, $info ) { return true; }
	] );

} );
```

jasonbahl:
I pushed up a commit that includes the built assets. Of course we ideally shouldn't be versioning the built assets of graphiql, but here we are in 2024 versioning them 🤷🏻‍♂️ 

As @josephfusco and I work on the new WPGraphQL IDE over here: https://github.com/wp-graphql/wpgraphql-ide we plan to have better automation/workflows.

PR #3058, opened 2024-02-22
fix: prevent wpml from redirecting in WPGraphQL requests

jasonbahl:

This adds a filter to prevent WPML from redirecting within WPGraphQL requests. 

Also, this tests the modified Github Workflow intended to run phpcbf to fix changes and commit them back to the PR. 

no

Any other comments?
-------------------
I've referred folks to this snippet many times. Adding it to core leads to a reduced support burden.

### DIFF
--- a/src/WPGraphQL.php
+++ b/src/WPGraphQL.php
@@ -341,14 +341,19 @@ final class WPGraphQL {
 		 * Prevent WPML from redirecting within WPGraphQL requests
 		 *
 		 * @see https://github.com/wp-graphql/wp-graphql/issues/1626#issue-769089073
-	 * @since @todo
+		 * @since @todo
 		 */
-add_filter( 'wpml_is_redirected', static function( bool $is_redirect ) {
-	if ( is_graphql_request() ) {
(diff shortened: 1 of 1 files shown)


justlevine:
> I've referred folks to this snippet many times. Adding it to core leads to a reduced support burden.

Even still, wouldn't the [official WPML GraphQL extension](https://wpml.org/documentation/related-projects/wpml-graphql/) be a better place for this?

jasonbahl:
@justlevine the problem happens when you have WPGraphQL active and WPML active even if you're not using WPML features in GraphQL. 

Just having WPML active has prevented a lot of users from accessing data from GraphQL until this is implemented. 

Some users have deactivated WPGraphQL thinking it just doesn't work at all because their first attempt at using it was a failure. 

I think intentional features of having WPML work with WPGraphQL belong in that extension but simply being able to coexist and WPGraphQL work normally feels like a small piece of tech debt we can take on with the trade off of reduced support and improved first time user experience. 🤷‍♂️

PR #3061, opened 2024-02-23
release: v1.22.0

jasonbahl:
# Release Notes

## New Features

- [#3044](https://github.com/wp-graphql/wp-graphql/pull/3044): feat: add `graphql_pre_resolve_menu_item_connected_node` filter
- [#3039](https://github.com/wp-graphql/wp-graphql/pull/3043): feat: add `UniformResourceIdentifiable` interface to `Comment` type
- [#3020](https://github.com/wp-graphql/wp-graphql/pull/3020): feat: introduce `graphql_query_analyzer_get_headers` filter

## Chores / Bugfixes

- [#3062](https://github.com/wp-graphql/wp-graphql/pull/3062): ci: pin wp-browser to "<3.5" to allow automated tests to run properly 
- [#3057](https://github.com/wp-graphql/wp-graphql/pull/3057): fix: `admin_enqueue_scripts` callback should expect a possible `null` value passed to it
- [#3048](https://github.com/wp-graphql/wp-graphql/pull/3048): fix: `isPostsPage` on content type
- [#3043](https://github.com/wp-graphql/wp-graphql/pull/3043): fix: return empty when filtering `menuItems` by a location with no assigned items
- [#3045](https://github.com/wp-graphql/wp-graphql/pull/3045): fix: `UsersConnectionSearchColumnEnum` values should be prefixed with `user_`

PR #3063, opened 2024-02-29
fix: taking care about the case of empty order by

mohjak:
This fix resolves the issue when the 'ORDER BY' is missing in cursor pagination.

#### Reproduce Link

I have installed WPGraphQL Version 1.21.0 in addition to this version of an offset pagination plugin <https://github.com/alaa-alshamy/wp-graphql-offset-pagination/releases/tag/v0.2.0.2>. 
I have the following categories for example as parent, child
image
When I run the following GraphQL query to count total of children of the targeted parent that id is `38`.
  ```graphql
  query terms {
    categories(where: {
      parent: 38
    }) {
      pageInfo {
        offsetPagination {
          total
        }
      }
      nodes {
        name
        databaseId
        slug      
      }
    }
  }
  ```
Then the total is returning `null` as you can see in the image below:
image

After Applying the fix of this pull request the issue resolves:
image

Any other comments?
-------------------
I believe the following testing environment should help you to reproduce the issue, this is hosted in local by flywheel:
<https://probable-tendency.localsite.io/wp-admin/admin.php?page=graphiql-ide>
**Basic Authentication:** 
**Username:** pattern
**Password:** unusual

**Admin Panel:**
**Username:** admin
**Password:** password

Where has this been tested?
---------------------------
**Operating System:** …
Ubuntu 22.04.4 LTS

**WordPress Version:** …
6.4.3

### DIFF
--- a/src/Data/Config.php
+++ b/src/Data/Config.php
@@ -205,7 +205,7 @@ class Config {
 		$key    = $cursor->get_cursor_id_key();
 
 		// If there is a cursor compare in the arguments, use it as the stablizer for cursors.
-		return "{$orderby}, {$key} {$order}";
+		return ( $orderby ? "{$orderby}, " : '' ) . "{$key} {$order}";
 	}
 
 	/**
@@ -307,7 +307,7 @@ class Config {
 		$cursor = new UserCursor( $query->query_vars );
 		$key    = $cursor->get_cursor_id_key();
 
-		return "{$orderby}, {$key} {$order}";
+		return ( $orderby ? "{$orderby}, " : '' ) . "{$key} {$order}";
 	}
 
 	/**
@@ -420,7 +420,7 @@ class Config {
 			$pieces['where'] = $pieces['where'] . $before_cursor->get_where();
 		}
 
-		// Check the cursor compare order
+		// Check the cursor compare order.
 		$order = '>' === $args['graphql_cursor_compare'] ? 'ASC' : 'DESC';
 
 		// Get Cursor ID key.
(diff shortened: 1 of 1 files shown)


justlevine:
Thanks so much for this commit, @mohjak 🙌

Can you provide an example of something that currently triggers this error - both for review purposes and to help us write the necessary unit tests?

(For example, I see that `graphql_wp_query_cursor_pagination_stability()` already enforces a string type, though `graphql_wp_user_query_cursor_pagination_stability()` only typehints it. Meanwhile, `graphql_wp_term_query_cursor_pagination_support()`'s `$pieces` has no typings so it definitely needs that guard. )

justlevine:
(tests are failing because this is based on `master` and not `develop` where the Codecption version is pinned. We can rebase now or wait for #3061 to ship)

mohjak:
You are welcome @justlevine, I have updated the pull request info. Please let me know if there are anything still missing.

mohjak:
@justlevine We are not enforcing string type, but we're trying to resolve if there is no value in `$orderby` parameter then a MySQL syntax error will appear.

in the old code:
```php
return "{$orderby}, {$key} {$order}";
```
So, If the `$orderby` is empty then the result will be for example ', date DESC'. Which is a MySQL syntax error.

jasonbahl:
@mohjak I was able to reproduce the issue you were seeing where `offsetPagination.total` returns 0 before this PR and returns `1` after the PR. 

Before merging, we'll need a test case that ensures this fix doesn't have a future regression.

I'm working on writing a test case now.

jasonbahl:
@mohjak I couldn't push up my test to your PR branch so I opened this PR: https://github.com/mohjak/wp-graphql/pull/1

But I think I'll just merge your PR then follow up with the test in a separate PR to core WPGraphQL

jasonbahl (approved):
Approving. I added a test here: https://github.com/mohjak/wp-graphql/pull/1

And will merge that test in a follow-up PR.

Issue #3036, opened 2024-02-06
set_query_arg is not properly respected when connections are resolved


jasonbahl:
### Description

When writing a resolver for a connection and using the `$connection->set_query_arg()` method, the args are not properly being respected by the resolving class. 

i.e. PostObjectConnectionResolver isn't properly consideration of any args set by `set_query_arg()`.

### Steps to reproduce

1. register a custom connection:

```php
add_action( 'graphql_register_types', function() {
	register_graphql_connection( [
		'fromType' => 'RootQuery',
		'toType' => 'Post',
		'fromFieldName' => 'curatedPosts',
		'resolve' => function( $root, $args, $context, $info ) {
			$resolver = new \WPGraphQL\Data\Connection\PostObjectConnectionResolver( $root, $args, $context, $info, 'any' );

			// REPLACE WITH A LIST OF VALID POST IDS.
			// ENOUGH TO PAGINATE AGAINST. IF NONE EXIST, GENERATE SOME.
			// ALSO MAKE SURE THEY ARE NOT STRICTLY ORDERED NUMERICALLY FOR SANITY CHECKING (i.e. 1, 5, 3, 2 instead of 1,2,3,4,5)
			$ids = [17, 105, 106, 15, 16, 18, 19, 30, 20, 21, 22, 23, 24, 7, 25, 26, 27, 28, 29, 31];
			return $resolver
				->set_query_arg( 'post_status', 'any' )
				->set_query_arg( 'post__in', $ids )
				->set_query_arg( 'orderby', 'post__in' )
				->get_connection();
		}
	] );
} );
```

2. Query for the posts: 

```graphql
{
  curatedPosts(
    first: 5
    after:null
) {
    nodes {
      databaseId
    }
    pageInfo {
      hasNextPage
      endCursor
    }
  }
}
```

3. ✅ Verify that the first 5 posts match the first 5 in the `set_query_arg( 'post__in' )`

$ids = [17, 105, 106, 15, 16, 18, 19, 30, 20, 21, 22, 23, 24, 7, 25, 26, 27, 28, 29, 31];

first 5: = 17, 105, 106, 15, 16

looks good!

![CleanShot 2024-02-06 at 10 55 54](https://github.com/wp-graphql/wp-graphql/assets/1260765/d453f802-2638-4d9b-9b1c-b1a55531d7c1)

4. Use the `endCursor` and query the next 5:

5. 👀 Expect to see the next 5 posts: 

[...]

PR #3066, opened 2024-03-04
fix: stop overwrite tax query filter if it exists

mohjak:
The change is aggregating the taxonomy filters. This is related to [wp-graphql-taxquery](https://github.com/wp-graphql/wp-graphql-tax-query) plugin.

Suppose the following scenario:

Given I installed a fresh copy of a WordPress v6.4.3 and WP GraphQL v1.22.0  and WPGraphQL Tax Query
 v0.2.0
And I have created a category with slug 'cat1'
And a tag with slug 'tag1'
And I have created a post that relates only to 'cat1'
When I run the following query:

```graphql
query categoriesSearch {
  categories(where: {slug: "cat1"}) {
    edges {
      node {
        name
        ... on Category {
          posts(
            first: 1
            where: {status: PUBLISH, taxQuery: {relation: AND, taxArray: [{field: NAME, operator: EXISTS, taxonomy: TAG}]}}
          ) {
            nodes {
              title
              status
              tags {
                nodes {
                  name
                }
              }
            }
          }
        }
      }
    }
  }
}
```

Then the following result returns:

```
{
  "data": {
    "categories": {
      "edges": [
        {
          "node": {
            "name": "Cat1",
            "posts": {
              "nodes": [
                {
                  "title": "Post Cat1",
                  "status": "publish",
                  "tags": {
                    "nodes": []
                  }
                }
              ]
            }
          }
        }
      ]
    }
  }
}
```

But, the expected result is to return nothing in posts.

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …
Ubuntu 22.04.4 LTS

**WordPress Version:** …
6.4.3


justlevine:
Hey @mohjak thanks for opening this PR!

This would be a breaking change to implement, and the same outcome can already be achieved with `get_query_arg()` and `array_merge()`.

I didn't really follow what issue you're trying to solve with this change, but perhaps it makes more sense to do in your custom code with one of the connection resolver hooks, or if it's a bug with the tax query extensions you referenced then a PR to that repo.

( ref: https://github.com/wp-graphql/wp-graphql/issues/3036#issuecomment-1933042912 )

mohjak:
Hi @justlevine,

What I am trying to resolve is the following `set_query_arg` function:

https://github.com/wp-graphql/wp-graphql/blob/9d021498122058854b7b206b406d3deeb5d7bd2c/src/Registry/Utils/TermObject.php#L234

Which is expecting that there is no `tax_query` in the query args that coming from graphql input, but actually it can have a tax_query like my above scenario. So, it overwrites the `tax_query` with the value mentioned in that referenced line of code.

justlevine:
> Hi @justlevine,
> 
> What I am trying to resolve is the following `set_query_arg` function:
> 
> https://github.com/wp-graphql/wp-graphql/blob/9d021498122058854b7b206b406d3deeb5d7bd2c/src/Registry/Utils/TermObject.php#L234
> 
> Which is expecting that there is no `tax_query` in the query args that coming from graphql input, but actually it can have a tax_query like my above scenario. So, it overwrites the `tax_query` with the value mentioned in that referenced line of code.

If anything, seems like it would be the `resolve` you referenced that needs to be "fixed", to check if `tax_query` is set, and if so merge it and pass the results to the `$resolver->set_query_arg()`.

@mohjak can you foresee any issues if we took that (nonbreaking) approach?

src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -832,7 +832,15 @@ public function get_page_info() {
 	 * @return static
 	 */
 	public function set_query_arg( $key, $value ) {
-		$this->query_args[ $key ] = $value;
+    if ( ! empty( $this->query_args[ $key ] ) && is_array( $this->query_args[ $key ] ) ) {
```
jasonbahl, line 835:
```suggestion
		if ( ! empty( $this->query_args[ $key ] ) && is_array( $this->query_args[ $key ] ) ) {
```

jasonbahl:
Apologies for the long-overdue review of this PR. 

I was able to walk through the scenario with the WPGraphQL Tax Query plugin active and reproduce the issue you outlined and I have confirmed that this PR does fix it. 

I wrote up a test based on the scenario that fails without the changes in this PR applied and passes with the changes applied.

jasonbahl (approved):

Issue #930, opened 2019-08-07
Handle Password Protected pages/posts


luchoster:
I'm re-doing a WP website with Gatsby, using the WPGraphQL plugin. We have a few pages that are password protected, and would like to know if there's an easy way to handle those pages in the front-end. So users can get to the page and enter the password to view the content.

Tagging @jasonbahl for help :)

jasonbahl:
Right now, WPGraphQL does not have a way for users to request password protected posts/pages/custom posts. 

At the moment, these objects are protected by the WPGraphQL model layer, so they will only be exposed to users with proper capabilities to see the content. 

In order for users to submit a password and have the objects returned, we need to add a mechanism that allows that input, and for resolving data to respect that. 

## Password Input Field?
An initial thought was to add an input field on the fields in the Schema that could return a password protected post, but this is problematic for many reasons. 

First, there are dozens of ways to access a post in the Graph, for example you can query a post individually, as part of a connection, as a parent/child of another post, in response to a mutation, etc. 

For all of those to have input fields to accept a post password would get very difficult to maintain and scale.

## Extensions!

The other thought would be to add support for some extension input that can be submitted alongside the query and variables of a request. 

The client can then persist the passwords however they want (localStorage, etc) and submit them with future requests. 

That would allow for all password protected posts to pass validation in the Model layer, and return content to the user, regardless of how they access the password protected node. 

```
wp_remote_post( $url, [
  'body' => wp_json_encode([
     'query' => '{ posts { nodes { id, title, content } } }',
     'variables' => null,
     'extensions' => [
        'postPasswords' => [
            [
              'id' => 123,
              'password' => 'my password'
           ],
           [
              'id' => 456,
              'password' => 'another password'
           ]
        ]
      ]
  ]);
] );
```

In order to support this, we will need to look for the extension input before execution, and set it as Context for the request. 

Then when the nodes are being resolved, they can look in the context and determine if the resolving node has a password in the context, and if the password matches, the node can be returned.

drewbaker:
Another idea:  Have the password be a required input in the query and used the same way FORMAT is used on a field?

```
query post($id: ID!, $password: String!) {
        id
        title(PASSWORD: $password)
        content(PASSWORD: $password)
    }
}
```

My thought came from the fact that [some fields aren't actually password protected](https://wordpress.org/support/article/using-password-protection/#password-protected-posts) in WP, for example titles will return but say `Protected:` before the title.

[...]

PR #3073, opened 2024-03-18
feat: expose `hasPassword` and `password` fields on Post objects

### LINKED ISSUE #3069: Expose hasPassword as a queryable field instead of just as an argument

On Faust-end, we would like the ability to know if a page/post has visibility set to password-protected. Currently there are no easy way to determine this. There are `hasPassword` and `password` query arguments but they are not helpful or direct in determining if a post is password protected. 

Expose a hasPassword field or some flag that can easily determine if a page/post is password protected in the GQL schema and readable on Faust end.

_No response_

### Additional Context

This request is based on this [Slack channel discussion](https://wp-graphql.slack.com/archives/C3NM1M291/p1710425784866659).

justlevine:
This PR exposes the `hasPassword` and `password` fields on Post objects.

It **does not** provide a mechanism to _provide_ a password, which will be handled in a separate PR.

closes #3069 

Any other comments?
-------------------
Until #930 is addressed, password-protected posts require a WP cookie or the `edit_posts` capability. You can see that ticket for workarounds.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (Wsl2 + devilbox + php 8.1.19)

**WordPress Version:** 6.4.3

### DIFF
--- a/src/Model/Post.php
+++ b/src/Model/Post.php
@@ -51,6 +51,7 @@ use WP_Post;
  * @property array   $editLock
  * @property string  $enclosure
  * @property string  $guid
+ * @property bool    $hasPassword
  * @property int     $menuOrder
  * @property string  $link
  * @property string  $uri
@@ -58,6 +59,7 @@ use WP_Post;
  * @property string  $featuredImageId
  * @property int     $featuredImageDatabaseId
  * @property string  $pageTemplate
+ * @property string  $password
  * @property int     $previewRevisionDatabaseId
  *
  * @property string  $captionRaw
@@ -152,6 +154,7 @@ class Post extends Model {
 			'isPostsPage',
 			'isFrontPage',
 			'isPrivacyPage',
+			'hasPassword',
 		];
 
 		if ( isset( $this->post_type_object->graphql_single_name ) ) {
--- a/src/Model/Post.php
+++ b/src/Model/Post.php
@@ -595,6 +595,12 @@ class Post extends Model {
 
 					return false;
 				},
+				'hasPassword'               => function () {
+					return ! empty( $this->data->post_password );
+				},
+				'password'                  => function () {
+					return ! empty( $this->data->post_password ) ? $this->data->post_password : null;
+				},
 				'toPing'                    => function () {
 					$to_ping = get_to_ping( $this->databaseId );
 
@@ -690,12 +696,6 @@ class Post extends Model {
 
 					return ! empty( $thumbnail_id ) ? absint( $thumbnail_id ) : null;
 				},
-				'password'                  => [
-					'callback'   => function () {
-						return ! empty( $this->data->post_password ) ? $this->data->post_password : null;
-					},
-					'capability' => isset( $this->post_type_object->cap->edit_others_posts ) ?: 'edit_others_posts',
-				],
 				'enqueuedScriptsQueue'      => static function () {
 					global $wp_scripts;
 					do_action( 'wp_enqueue_scripts' );
--- a/tests/wpunit/PostObjectQueriesTest.php
+++ b/tests/wpunit/PostObjectQueriesTest.php
@@ -2447,6 +2447,7 @@ class PostObjectQueriesTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase 
 				content
 				status
 				password
+				hasPassword
 			}
 		}
 		';
@@ -2464,6 +2465,7 @@ class PostObjectQueriesTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase 
 		$this->assertEquals( 'publish', $actual['data']['post']['status'], 'Status should be "publish" when unauthenticated' );
 		$this->assertNull( $actual['data']['post']['content'], 'Content should be null when unauthenticated' );
 		$this->assertNull( $actual['data']['post']['password'], 'Password should be null when unauthenticated' );
+		$this->assertTrue( $actual['data']['post']['hasPassword'], 'hasPassword should be true when unauthenticated' );
 
 		// Test password protected post as a subscriber.
 		wp_set_current_user( $subscriber );
@@ -2479,6 +2481,7 @@ class PostObjectQueriesTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase 
 		$this->assertEquals( 'publish', $actual['data']['post']['status'], 'Status should be "publish" when lacking permissions' );
 		$this->assertNull( $actual['data']['post']['content'], 'Content should be null when lacking permissions' );
 		$this->assertNull( $actual['data']['post']['password'], 'Password should be null when lacking permissions' );
+		$this->assertTrue( $actual['data']['post']['hasPassword'], 'hasPassword should be true when lacking permissions' );
 
 		// Test password protected post as different author.
 		wp_set_current_user( $post_author_two );
@@ -2494,6 +2497,7 @@ class PostObjectQueriesTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase 
 		$this->assertEquals( 'publish', $actual['data']['post']['status'], 'Status should be "publish" when lacking permissions' );
 		$this->assertNull( $actual['data']['post']['content'], 'Content should be null when lacking permissions' );
 		$this->assertNull( $actual['data']['post']['password'], 'Password should be null when lacking permissions' );
+		$this->assertTrue( $actual['data']['post']['hasPassword'], 'hasPassword should be true when lacking permissions' );
 
 		// Test password protected post as current author.
 		wp_set_current_user( $post_author_one );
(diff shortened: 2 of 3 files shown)


src/Model/Post.php
```diff
@@ -595,6 +595,12 @@ protected function init() {
 
 					return false;
 				},
+				'hasPassword'               => function () {
+					return ! empty( $this->data->post_password );
+				},
+				'password'                  => function () {
+					return ! empty( $this->data->post_password ) ? $this->data->post_password : null;
+				},
```
jasonbahl, line 603: why did the capability check get removed from this field? Is it already protected by other logic in the model? Want to make sure we're not exposing passwords on posts to folks that shouldn't be able to access the password 🤔

tests/wpunit/PostObjectQueriesTest.php
```diff
@@ -2406,4 +2409,84 @@ public function testQueryPostBySlugWithNonAsciiSlug() {
...
+		] );
+
+		$this->assertArrayNotHasKey( 'errors', $actual );
+		$this->assertEquals( $post->post_title, $actual['data']['post']['title'], 'Title should be returned when unauthenticated' );
+		$this->assertEquals( 'publish', $actual['data']['post']['status'], 'Status should be "publish" when unauthenticated' );
+		$this->assertNull( $actual['data']['post']['content'], 'Content should be null when unauthenticated' );
+		$this->assertNull( $actual['data']['post']['password'], 'Password should be null when unauthenticated' );
+
+		// Test password protected post as admin.
+		wp_set_current_user( $this->admin );
```
jasonbahl, line 2470: I think we need to also test this with other roles. 

The code used to check `$this->post_type_object->cap->edit_others_posts` before returning the password. 

So I believe this means the password will be exposed to admins but not other roles like subscriber. (similar to the wp-admin, where an admin could see the password of a post, but a subscriber could not)

jasonbahl: The logic of just checking the capability _might_ not even be enough though, because I think if you don't have that capability BUT you are the author of the post itself you should be able to see the password?

jasonbahl (changes requested):

src/Model/Post.php
justlevine: Yup, it's protected top-level by [`Model::has_restricted_cap()`](https://github.com/wp-graphql/wp-graphql/blob/3b326a5887c90e7e697618046a144a4701decaef/src/Model/Post.php#L252) and tested [here](https://github.com/wp-graphql/wp-graphql/blob/3b326a5887c90e7e697618046a144a4701decaef/tests/wpunit/PostObjectQueriesTest.php#L2453).

tests/wpunit/PostObjectQueriesTest.php
justlevine: What other caps do you want to test against?

This PR intentionally doesn't change the existing logic for handling password-protected posts (which should be addressed in #930 ), and the test currently checks:
- an unauthenticated user
- an authenticated user that doesn't have the correct cap (a regular subscriber)
- an authenticated user that does have the correct cap (admin)

Once we tackle #930, we _may_ want to change the cap check to `edit_post` (as handled in [WP_REST_Posts_Controller::can_access_password_content()](https://developer.wordpress.org/reference/classes/wp_rest_posts_controller/can_access_password_content/) versus the even more restrictive than currently frontend check by [post_password_required()](https://developer.wordpress.org/reference/functions/post_password_required/)), but that security loosening should probably be considered a breaking change.

jasonbahl: @justlevine ya I mean the password field itself. 

In normal WordPress, not _everyone_ is able to see the password that has been set for a post, even if they can see other fields of that post. 

For example, if I'm an admin, I can set the password on a post and therefore also see the value of the existing password that is set for a post. 

However, if I am authenticated in WordPress as an "author", I can see the "password protected post" and fields like the title, author, date, etc. . .BUT, I cannot see the "password" field. 

CleanShot 2024-03-27 at 15 55 16


That field is restricted only to users with the `edit_others_posts` capability. 

Since WordPress itself does not expose the "password" value to users that do not have proper permission to see/update the password, WPGraphQL also should not "leak" the password to users that should not be able to see the password.

So, being able to resolve the value of `password` field is more nuanced than the Post itself. The `has_restricted_cap()` covers the Post node as a whole, but this field has more granular requirements. Not only do you have to be able to see the Node itself, BUT you also have to be able to see the password, which not every user will have permission to do so. 

That's what [this code](https://github.com/wp-graphql/wp-graphql/pull/3073/files#diff-8346b33bd0059a34c486fab6a109aaab8180a6a15d3049eccc473249ec0cc0bbL697) does. . .it further restricts this field on the model to resolve only if the user has proper capabilities to see the password field. 

I think the current logic is a bit naiive and should be more robust to not JUST check if the current user has `edit_other_post` cap, but also other conditions such as "is the current user the author of the post"

For example, an Author can actually publish their own post and set a password on it, and so the author of a post should be able to see the password for their own post:

CleanShot 2024-03-27 at 15 57 22

But _also_ an admin that can `edit_others_posts` should _also_ be able to see the password field.

Regardless, at minimum we should probably add a test to test against the `author` role and ensure that they can get the title and date of a pw protected post they did not author, but cannot get the password .

justlevine: @jasonbahl in b9e25bf13a226dc8e7b6be122effafee9cc1de7e, I added asserts against the current post author (password/content shown) and a different post author (password/content hidden), and fixed the subscriber assert (forgot to call `set_current_user()` 🤦). So _hopefully_ that covers your concerns, but tbh I'm still not sure I fully understand:

> For example, if I'm an admin, I can set the password on a post and therefore also see the value of the existing password that is set for a post.
> 
> However, if I am authenticated in WordPress as an "author", I can see the "password protected post" and fields like the title, author, date, etc. . .BUT, I cannot see the "password" field.
> [image]
>That field is restricted only to users with the edit_others_posts capability.

To clarify, on the _frontend_, WordPress uses [post_password_required()](https://developer.wordpress.org/reference/functions/post_password_required/), and requires everyone - including admin or the post author to provide a password, in order to see the `post_content` and `password`, while everything else remains visible.

On the dashboard list views/ edit screen, WP uses [WP_REST_Posts_Controller::can_access_password_content()](https://developer.wordpress.org/reference/classes/wp_rest_posts_controller/can_access_password_content/), allowing the current post author and any users with the `edit_others_posts` capability to open the Edit screen and view the password.

In **both** scenarios, other Post Object data remains visible (such as custom post meta).

> So, being able to resolve the value of password field is more nuanced than the Post itself. The has_restricted_cap() covers the Post node as a whole, but this field has more granular requirements. Not only do you have to be able to see the Node itself, BUT you also have to be able to see the password, which not every user will have permission to do so.

Agreed, but to the _opposite_ intent. Currently `get_restricted_cap()`'s [check](https://github.com/justlevine/wp-graphql/blob/b9e25bf13a226dc8e7b6be122effafee9cc1de7e/src/Model/Post.php#L252) for `post_password` results in a _stricter_ permissions model than either WP's frontend or REST API behavior, and makes the existing `$field_config['capability']` check redundant.

In an ideal world (i.e. after or as part of #930 ) we would find what the happy medium should be (e.g. should post meta be hidden by default), but it's not going to be doable with a simple `$field_config['capability']` check.

Where are we losing each other?

jasonbahl: > Where are we losing each other?

I think I was lost in the history of the project. 

We had 2 different APIs for handling the same thing. 

- The capability check on individual Model fields
- The `allowed_restricted_fields` / `get_restricted_cap()` on the Model itself

You are indeed correct that this was already being covered by the `allowed_restricted_fields` / `get_restricted_cap()`.

To verify manually I executed the following query as 3 different users (`public`, `author`, `admin`)

```graphql
query GetPostsAndCurrentViewer {
  posts {
    nodes {
      id
      title
      date
      password
    }
  }
  viewer {
    name
    roles {
      nodes {
        name
      }
    }
  }
}
```

### Public User

As a public user, I can see that Password Protected posts exist, but I cannot see the password nor the password protected content. This is good. However, I cannot see whether the post `hasPassword: true/false` either, so we need to add `hasPassword` as an allowed restricted field. Public users should be able to identify whether a post has a password or not, but they should not be able to access the password itself.

CleanShot 2024-04-01 at 11 15 15

### Author

As an Author, I can see that password protected posts exist, and for a PW protected post that I am the author of, I can see both the content and the password. But for a PW protected post authored by another user, I can see that it exists, but I cannot see the content nor the password. However, just like a public user, I cannot see the value of the `hasPassword` field. 

CleanShot 2024-04-01 at 11 15 39

### Admin

As an admin user I can see all the posts, their content, their passwords, etc whether I am the author of the post or not. This is as expected.

CleanShot 2024-04-01 at 11 14 34

justlevine: Thanks for clarifying - and great catch with `hasPassword`!

Will push that fix + tests and ping you for the re-review when it's ready

jasonbahl: @justlevine I just pushed changes 😉

PR #3091, opened 2024-04-03
feat: introduce actions and filters for GraphQL Admin Notices

### LINKED ISSUE #3090: Introduce actions/filters for the `register_graphql_admin_notice` API

As a WordPress plugin developer, I would like to modify the default behaviors of the `register_graphql_admin_notice`. For example, I would like to be able to filter [the list of allowed pages](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Admin/AdminNotices.php#L295-L300) the notices should show on so that they could show on additional pages for plugins that extend WPGraphQL.

Additionally, I would like to be able to hook into the rendering of the notices to be able to add custom styles, etc.

I propose that some new actions and filters be introduced in the `/src/Admin/AdminNotices.php` class.

Getting a little hacky with 3rd party extension code.

### Additional Context

I'm working on the new WPGraphQL IDE and want to be able to support `register_graphql_admin_notices()` and how they render on the new admin page that renders the new WPGraphQL IDE.

jasonbahl:
This PR introduces the following:

- `graphql_admin_notices_render_notices` action
- `graphql_admin_notices_render_notice` action
- `graphql_admin_notices_is_plugin_scoped_page` filter

These actions and filters enable 3rd party codebases to hook into the rendering of admin notices registered via `register_graphql_admin_notice` and modify things like: 

- which page(s) the notices can be displayed on
- styles applied to the notices

closes #3090 

Any other comments?
-------------------
While working on the new WPGraphQL IDE (https://github.com/wp-graphql/wpgraphql-ide) I wanted to add support for the admin notices registered via `register_graphql_admin_notice()` but because of the lack of actions/filters I was not able to add proper support without modifying core WPGraphQL. 

By introducing these actions/filters I can now add support for GraphQL Admin Notices to additional pages, such as the new IDE.

ex: 

CleanShot 2024-04-03 at 12 16 24

### DIFF
--- a/src/Admin/AdminNotices.php
+++ b/src/Admin/AdminNotices.php
@@ -249,6 +249,16 @@ class AdminNotices {
 		</style>
 		<?php
 		$count = 0;
+
+		/**
+		 * Fires before the admin notices are rendered.
+		 *
+		 * @param array<mixed> $notices The notices to be rendered
+		 *
+		 * @since todo
+		 */
+		do_action( 'graphql_admin_notices_render_notices', $notices );
+
 		foreach ( $notices as $notice_slug => $notice ) {
 			$type = $notice['type'] ?? 'info';
 			?>
@@ -260,7 +270,7 @@ class AdminNotices {
 			<div id="wpgraphql-admin-notice-<?php echo esc_attr( $notice_slug ); ?>" class="wpgraphql-admin-notice notice notice-<?php echo esc_attr( $type ); ?> <?php echo $this->is_notice_dismissable( $notice ) ? 'is-dismissable' : ''; ?>">
 				<p><?php echo ! empty( $notice['message'] ) ? wp_kses_post( $notice['message'] ) : ''; ?></p>
 				<?php
-				if ( $this->is_notice_dismissable( $notice ) ) {
+				if ( $is_dismissable = $this->is_notice_dismissable( $notice ) ) {
 					$dismiss_acf_nonce = wp_create_nonce( 'wpgraphql_disable_notice_nonce' );
 					$dismiss_url       = add_query_arg(
 						[
@@ -275,6 +285,17 @@ class AdminNotices {
 				<?php } ?>
 			</div>
 			<?php
+			/**
+			 * Fires for each admin notice that is rendered.
+			 *
+			 * @param string $notice_slug The slug of the notice
+			 * @param array<mixed> $notice The notice to be rendered
+			 * @param bool $is_dismissable Whether the notice is dismissable or not
+			 * @param int $count The count of the notice
+			 *
+			 * @since todo
+			 */
+			do_action( 'graphql_admin_notices_render_notice', $notice_slug, $notice, $is_dismissable, $count );
 			++$count;
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/Admin/AdminNotices.php
```diff
@@ -249,6 +249,16 @@ protected function render_notices(): void {
 		</style>
 		<?php
 		$count = 0;
+
+		/**
+		 * Fires before the admin notices are rendered.
+		 *
+		 * @param array<mixed> $notices The notices to be rendered
```
justlevine, line 256: It's probably overkill to fully-type these, but we can can at least make them `array<string,mixed>` to hint that there's actual string keys that need to be used.

src/Admin/AdminNotices.php
```diff
@@ -301,7 +322,19 @@ protected function is_plugin_scoped_page(): bool {
...
+		 * Filter to determine if the current admin page is within the scope of the plugin's own pages.
+		 * This filter can be used to add additional pages to the list of allowed pages.
+		 *
+		 * The filter receives the following arguments:
+		 *
+		 * @param bool $is_plugin_scoped_page True if the current page is within scope of the plugin's pages.
+		 * @param string $current_page_id The ID of the current admin page.
+		 * @param array<string> $allowed_pages The list of allowed pages.
+		 */
+		return apply_filters( 'graphql_admin_notices_is_plugin_scoped_page', $is_plugin_scoped_page, $current_page_id, $allowed_pages );
```
justlevine, line 337: Don't ask me what it _should_ be, but can we think on a more self-documenting name for this? I'm not a fan of `graphql_admin_notices_is_allowed_admin_page`, but at least it doesn't require a definition for what a "plugin-scoped page" means.

src/Admin/AdminNotices.php
jasonbahl, line 256:
```suggestion
		 * @param array<string,mixed> $notices The notices to be rendered
```

src/Admin/AdminNotices.php
jasonbahl: 2 hardest things in programming: 

- naming things
- exactly once delivery
- exactly once delivery

jasonbahl: Based on the fact that I don't have a better name, we'll run with `graphql_admin_notices_is_allowed_admin_page`

src/Admin/AdminNotices.php
```diff
@@ -301,7 +323,19 @@ protected function is_plugin_scoped_page(): bool {
...
```
jasonbahl, line 338:
```suggestion
		$is_allowed_admin_page = in_array( $current_page_id, $allowed_pages, true );

		/**
		 * Filter to determine if the current admin page is within the scope of the plugin's own pages.
		 * This filter can be used to add additional pages to the list of allowed pages.
		 *
		 * The filter receives the following arguments:
		 *
		 * @param bool $is_plugin_scoped_page True if the current page is within scope of the plugin's pages.
		 * @param string $current_page_id The ID of the current admin page.
		 * @param array<string> $allowed_pages The list of allowed pages.
		 */
		return apply_filters( 'graphql_admin_notices_is_allowed_admin_page', $is_allowed_admin_page, $current_page_id, $allowed_pages );
```

PR #3093, opened 2024-04-07
ci: Update actions in GitHub workflows and cleanup

justlevine:
This PR:
- Updates the various GitHub actions used by our workflows to their latest versions.
- Removes `matrix` testing from our workflows when only one value (e.g. `php-version`, `node-version`) used.
- Replaces the use of `styfle/cancel-workflow-action` with the built-in `concurrency` property.
- Bumps the default PHP version used for workflows to PHP 8.2
- Updates the WordPress/PHP matrixes to make them more readable/manageable.

This PR **does not**
- Bump the active Node version used for workflows to Node 20. 

Any other comments?
-------------------
Once this is merged, we'll need to update the repo settings for required actions.

Where has this been tested?
---------------------------
**Operating System:** n/a (See action history for this PR)

**WordPress Version:** 6.5.0

### DIFF
--- a/.github/workflows/build-graphiql.yml
+++ b/.github/workflows/build-graphiql.yml
@@ -18,9 +18,10 @@ jobs:
   build-graphiql:
     name: Build GraphiQL
     runs-on: ubuntu-latest
+
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           ref: ${{ github.head_ref }}
           token: ${{ secrets.PAT }}
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 8.1
+          php-version: 8.2
           tools: composer:v2
           coverage: none
       
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -10,33 +10,31 @@ on:
       - develop
       - master
 
+# Cancel previous workflow run groups that have not completed.
+concurrency:
+  # Group workflow runs by workflow name, along with the head branch ref of the pull request
+  # or otherwise the branch or tag ref.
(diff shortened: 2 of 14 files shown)


justlevine:
WP 5.2 on PHP 7.4 is failing for [`NodeByUriTest:testDefaultTaxTermByUri`](https://github.com/wp-graphql/wp-graphql/blob/33c6e97387479fd5b2428736b0722a2d91109455/tests/wpunit/NodeByUriTest.php#L1420), but I am unsure why `get_term_by()` doesn't return the valid default term 🤔

@jasonbahl @josephfusco any ideas (branch edits open, so feel free to fix directly on this branch if you have a solve)

justlevine:
> @justlevine I'm not sure we need to prioritize testing 5.2.
> 
> We haven't been testing against it before and according to the telemetry data we collect, we have 0 users on 5.2.

It's less about 5.2 specifically, and more about testing our minimum supported version - 5.2 is just the lowest version who's docker still builds. It gives us a nice spread as we work to 2.0, and lets us drop middle versions in the future without worrying too much about coverage.

If it's not worth fixing the Codeception test, I can keep bumping until we find a minimum version that just works.

jasonbahl:
@justlevine ya, I think the effort is better spent ensuring things work on more recent versions and coming up with a plan / policy for supported versions that we can put into place and follow over time.  

i.e. "we will officially support the latest xx versions of WordPress and the last xx versions of PHP" or something to that tune. 

Definitely needs more thought than that, but I think there's more value in encouraging folks to update vs putting effort into testing versions we wouldn't actively recommend be used in the first place 🤷🏻‍♂️

justlevine:
>Definitely needs more thought than that, but I think there's more value in encouraging folks to update vs putting effort into testing versions we wouldn't actively recommend be used in the first place 🤷🏻‍♂️

💯

The goal of the matrix changes in this PR is _not_ philosophical, only a cleaner restructure of the matrix to make it easier to add/change versions without losing coverage.

Will find a docker build that works as-is and ping you for re-review.

justlevine:
@jasonbahl WP 5.5 is the magic number. 
Ready for your review.

jasonbahl (approved):

PR #3104, opened 2024-04-24
refactor: make `AbstractConnectionResolver::should_execute()` non-abstract and add `::pre_should_execute()`

justlevine:
This PR refactors the DX around `AbstractConnectionResolver::should_execute()`.

More specifically:
- `AbstractConnectionResolver::$should_execute` is no longer instantiated with a value. Instead `::should_execute()` is no longer abstract and returns true by default, so only Resolvers that require custom `should_execute` logic need to overload the method.
- Resolver-specific `pre_should_execute` logic has been abstracted out of AbstractConnectionResolver's constructor and into a new `::pre_should_execute()` method. This overloadable method is then wrapped in a new `::get_pre_should_execute()` method which uniformly filters the value with the new `graphql_connection_pre_should_execute` hook.
- `::get_should_execute()` has been modified to call `::should_execute()` if the class property has yet to be set. This ensure backwards compatibility and paves the way for future enhancements, such as improving the evaluation logic and filtering in `::execute_and_get_ids()`.

_There are no breaking changes in this PR_.

Part of #2749 

Any other comments?
-------------------
- Due to the way these changes are implemented, this PR is fully backwards compatible. Existing Resolvers will continue to work as before, and developers will be able to update their classes at their leisure to "opt in" to the new features like moving their constructor pre-checks into `get_pre_should_execute()`, separating the pre/should_execute logic, and removing unnecessary parameter instantiation, method overloading.

Where has this been tested?
---------------------------

**WordPress Version:** 6.5.2

### DIFF
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -64,9 +64,13 @@ abstract class AbstractConnectionResolver {
 	/**
 	 * Whether the connection resolver should execute.
 	 *
-	 * @var bool
+	 * If `false`, the connection resolve will short-circuit and return an empty array.
+	 *
+	 * Filterable by `graphql_connection_pre_should_execute` and `graphql_connection_should_execute`.
+	 *
+	 * @var ?bool
 	 */
-	protected $should_execute = true;
+	protected $should_execute;
 
 	/**
 	 * The loader name.
@@ -172,8 +176,8 @@ abstract class AbstractConnectionResolver {
 		 */
 		$this->args = $args;
 
-		// Bail if the Post->ID is empty, as that indicates a private post.
-		if ( $source instanceof Post && empty( $source->ID ) ) {
+		// Pre-check if the connection should execute so we can skip expensive logic if we already know it shouldn't execute.
+		if ( ! $this->get_pre_should_execute( $this->source, $this->unfiltered_args, $this->context, $this->info ) ) {
 			$this->should_execute = false;
 		}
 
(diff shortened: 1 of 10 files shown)


src/Data/Connection/AbstractConnectionResolver.php
```diff
@@ -257,22 +261,29 @@ abstract public function get_query_args();
...
+	 * @param \GraphQL\Type\Definition\ResolveInfo $info    Info about fields passed down the resolve tree.
 	 */
-	abstract public function should_execute();
+	protected function pre_should_execute( $source, array $args, AppContext $context, ResolveInfo $info ): bool {
+		$should_execute = true;
+
+		/**
+		 * If the source is a Post and the ID is empty, we should not execute the query.
+		 *
+		 * @todo this can probably be moved to the PostObjectConnectionResolver when we don't care about b/c.
```
jasonbahl, line 279: I don't think this can move to the PostObjectConnectionResolver because the `$source` is the node the connection resolver is coming _from_ not going to. 

So a `Post->User` connection which would use the `UserConnectionResolver` needs this, etc. 

We could probably generalize it better to validate that the `$source` is valid in some way to include this specific scenario and others. I can't recall the intent of this and what edge case we faced where we were passing `Post` without an `ID` 🤔. . .but I assume we could generalize this to better validate any source 🤔

jasonbahl (changes requested):
Left a comment regarding a `@todo` comment

src/Data/Connection/AbstractConnectionResolver.php
justlevine: Updated in https://github.com/wp-graphql/wp-graphql/pull/3104/commits/8014e2c93d9a612a79998fd09b0a52786f57e5d4

jasonbahl (approved):

Issue #3108, opened 2024-04-26
Arguments defined on Interface Fields are not inherited by the fields on the Implementing Object Type


jasonbahl:
### Description

When registering an Interface Type with fields that have arguments, then registering an Object Type that implements that Interface, the fields of the Interface Type are inherited by the ObjectType, but the argument(s) on the fields are not inherited, causing invalid schema issues.

### Steps to reproduce

1. register an interface type that contains a field with an argument:

```php
register_graphql_interface_type( 'InterfaceWithArgs', [
	'fields' => [
		'fieldWithArgs' => [
			'type' => 'String',
			'args' => [
				'interfaceArg' => [
					'type' => 'String',
				],
			],
			'resolve' => function( $source, $args ) {
				return $args['arg'];
			}
		],
	]
] );
```

2. register an object type that contains a field with the same name, but with no arguments: 

```php
register_graphql_object_type( 'ObjectTypeImplementingInterfaceWithArgs', [
	'interfaces' => [ 'InterfaceWithArgs' ],
	'fields' => [
		'fieldWithArgs' => [
			'type' => 'String',
			'resolve' => function() {
				return 'object value';
			}
		],
	],
] );
```

3. Register a field to add the Types to the Schema: 

```php
register_graphql_field( 'RootQuery', 'interfaceArgsTest', [
	'type' => 'ObjectTypeImplementingInterfaceWithArgs',
	'resolve' => function() {
		return true;
	},
]);
```

4. Open the GraphQL IDE and see the following error: 

```js
Interface field argument InterfaceWithArgs.fieldWithArgs(interfaceArg:) expected but 
ObjectTypeImplementingInterfaceWithArgs.fieldWithArgs does not provide it.
```

### Additional context

This appears to be a regression of v1.24.0 release (specifically #3100). 

### WPGraphQL Version

1.24.0

### WordPress Version

6.5.2

### PHP Version

8.2

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

### Please confirm that you have disabled ALL plugins except for WPGraphQL.

- [X] Yes
- [ ] My issue is with compatibility with a specific WordPress plugin, and I have listed all my installed plugins (and version info) above.

jasonbahl:
related: https://github.com/wp-graphql/wp-graphql/issues/3115

PR #3109, opened 2024-04-26
fix: arguments on Interface Fields were not properly applied to interfaces and object types implementing the Interface

jasonbahl:
This fixes a regression to #3102 / #3100 

- [x] Failing Test: https://github.com/wp-graphql/wp-graphql/actions/runs/8852128733
- [x] Passing Test: https://github.com/wp-graphql/wp-graphql/actions/runs/8852138235

closes: #3108 

Any other comments?
-------------------

Given the following code: 

```php

	register_graphql_interface_type( 'InterfaceWithArgs', [
		'fields' => [
			'fieldWithArgs' => [
				'type' => 'String',
				'args' => [
					'interfaceArg' => [
						'type' => 'String',
					],
				],
				'resolve' => function( $source, $args ) {
					return $args['arg'];
				}
			],
		]
	] );

	register_graphql_object_type( 'ObjectTypeImplementingInterfaceWithArgs', [
		'interfaces' => [ 'InterfaceWithArgs' ],
		'fields' => [
			'fieldWithArgs' => [
				'type' => 'String',
				'resolve' => function() {
					return 'object value';
				}
			],
		],
	] );

	register_graphql_field( 'RootQuery', 'interfaceArgsTest', [
		'type' => 'ObjectTypeImplementingInterfaceWithArgs',
		'resolve' => function() {
			return true;
		},
	]);

});
```

### Before

The schema would be invalid:

CleanShot 2024-04-26 at 11 33 56

And this query would be invalid:

CleanShot 2024-04-26 at 11 35 38

### After

The schema loads as valid and the test field can be queried

CleanShot 2024-04-26 at 11 35 05

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Type/WPInterfaceTrait.php
+++ b/src/Type/WPInterfaceTrait.php
@@ -160,10 +160,6 @@ trait WPInterfaceTrait {
 		foreach ( $fields as $field_name => $field ) {
 			$new_field = $field;
 
-			if ( empty( $new_field['description'] ) && ! empty( $interface_fields[ $field_name ]['description'] ) ) {
-				$new_field['description'] = $interface_fields[ $field_name ]['description'];
-			}
-
 			if ( ! isset( $new_field['type'] ) ) {
 				if ( isset( $interface_fields[ $field_name ]['type'] ) ) {
 					$new_field['type'] = $interface_fields[ $field_name ]['type'];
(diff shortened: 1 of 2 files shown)


justlevine (approved):
Code lgtm. 

With those `description` lines moved,  could probably simplify the complexity by `continue`ing early if `! isset( $interface_fields[ $field_name ]['type'] )`. But that can also be addressed by adding a few inline comments.

*Note*: I'm not by my computer to check, but I've run into previous issues using `unset()` on the looping array and then checking `isset()` on the modified array, vaguely related to scoping. If we don't early return instead, I'd say this might benefit from another unit test to ensure that conditional doesn't pass when it shouldn't, but also probably not necessary.

justlevine:
> Code lgtm.
> 
> With those `description` lines moved, could probably simplify the complexity by `continue`ing early if `! isset( $interface_fields[ $field_name ]['type'] )`. But that can also be addressed by adding a few inline comments.
> 
> _Note_: I'm not by my computer to check, but I've run into previous issues using `unset()` on the looping array and then checking `isset()` on the modified array, vaguely related to scoping. If we don't early return instead, I'd say this might benefit from another unit test to ensure that conditional doesn't pass when it shouldn't, but also probably not necessary.

See https://github.com/jasonbahl/wp-graphql/pull/11 for an example of what I mean.

jasonbahl:
closing in favor of #3112 which handles some additional cases not covered by this PR

Tests from this PR are handled in #3112:

- https://github.com/wp-graphql/wp-graphql/pull/3112/files#diff-b9b0ab687b2170fedfb7568ae8b1ae59ca53beec92447cba4fff22a2dbd25c09R457
- https://github.com/wp-graphql/wp-graphql/pull/3112/files#diff-b9b0ab687b2170fedfb7568ae8b1ae59ca53beec92447cba4fff22a2dbd25c09R546

PR #3116, opened 2024-05-02
chore: WPGraphQLTestcase upgraded to v3

kidunot89:
- Upgrades WPGraphQL Testcase to v3.
- Updates tests using `assertIsResponseValid` assertion to use the new `assertResponseIsValid` assertion.
- Fixes issues in preexisting wpunit tests surfaced by the new versions.

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/tests/wpunit/AccessFunctionsTest.php
+++ b/tests/wpunit/AccessFunctionsTest.php
@@ -2240,7 +2240,7 @@ class AccessFunctionsTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase {
 				'toType'                => 'Post',
 				'fromFieldName'         => 'test_field_with_underscores',
 				'allowFieldUnderscores' => true,
-				'connectionTypeName'    => 'TestConnectionFieldWithUnderscores',
+				'connectionTypeName'    => 'Test_Connection_With_Underscores',
 				'resolve' 			 => static function () {
 					return [
 						'nodes' => [],
--- a/tests/wpunit/AccessFunctionsTest.php
+++ b/tests/wpunit/AccessFunctionsTest.php
@@ -19,10 +19,10 @@ class AccessFunctionsTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase {
 
 	public function tearDown(): void {
 		// your tear down methods here
+		$this->clearSchema();
 
 		// then
 		parent::tearDown();
-		$this->clearSchema();
 	}
 
 	public function testGraphQLPhpVersion() {
@@ -1306,7 +1306,7 @@ class AccessFunctionsTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase {
 
 		$actual = graphql( compact( 'query' ) );
 
-		$this->assertIsValidQueryResponse( $actual );
+		$this->assertResponseIsValid( $actual );
 		$this->assertArrayNotHasKey( 'errors', $actual );
 
 		// Ensure Post-related types have been removed.
@@ -1372,7 +1372,7 @@ class AccessFunctionsTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase {
 
 		$actual = $this->graphql( compact( 'query' ) );
 
-		$this->assertIsValidQueryResponse( $actual );
+		$this->assertResponseIsValid( $actual );
 		$this->assertArrayHasKey( 'errors', $actual );
 
 		// Ensure Post is removed from interfaces.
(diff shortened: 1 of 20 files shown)


tests/wpunit/ConnectionRegistrationTest.php
```diff
@@ -317,7 +317,7 @@ static function () {
 		$this->assertQueryError( $response, $expected );
 
 		\wp_set_current_user( 1 );
-		$this->clearSchema();
+
 		$response = $this->graphql( compact( 'query' ) );
 		$expected = [
 			$this->expectedField( 'failingAuthConnection.nodes.0', self::NOT_NULL ),
```
justlevine, line 323: Per slack, I don't understand this entire test case. Since the test was previously passing, we should verify that we're testing the behavior we actually want.

tests/wpunit/AccessFunctionsTest.php
```diff
@@ -2232,15 +2232,21 @@ public function testRegisterConnectionWithUnderscores() {
 	}
 
 	public function testDeRegisterConnectionWithUnderscores() {
-
-		deregister_graphql_connection( 'test_field_with_underscores' );
+		deregister_graphql_connection( 'TestConnectionFieldWithUnderscores' );
```
jasonbahl, line 2235: This change defeats the purpose of the test 🤔

tests/wpunit/ConnectionRegistrationTest.php
```diff
@@ -243,7 +243,7 @@ static function () {
 						],
 						'fromFieldName' => 'failingAuthConnection',
 						'resolve'       => static function () {
-							return [ 'nodes' => [ null, false, 0 ] ];
+							return [ 'nodes' => [ '', false, 0 ] ];
```
jasonbahl, line 246: why did this change? Is there a regression somewhere we need to document?

jasonbahl (changes requested):
Left a couple comments / questions

tests/wpunit/ConnectionRegistrationTest.php
kidunot89: If you return a `null` for an object's root you GraphQL does not resolve that types fields. Which was making this test fail.

tests/wpunit/AccessFunctionsTest.php
justlevine: Intent of the test as I understood is that `deregister_graphql_connection()` also removes the connecting field even if the _field_ is in underscores.`deregister_graphql_field()` is tested separately for underscores.

justlevine: Or maybe it is a _type_ with underscores, and the `connectionTypeName` needs to be changed to include underscores?

tests/wpunit/ConnectionRegistrationTest.php
justlevine: Specifically, the Connection Interfaces release made `nodes` non nullable. The error was just getting suppressed before.

justlevine: Verbose pre-change error log in this workflow attempt: https://github.com/wp-graphql/wp-graphql/actions/runs/8918725858/job/24495559399

kidunot89: Okay, I think I figure out the way the test was suppose to be in the first place and fixed it.

tests/wpunit/AccessFunctionsTest.php
jasonbahl: @justlevine I bet the intent was for the Type Name to have underscores. Because it is valid to have a Type like `My_Custom_Type` and I believe the intent of the test was to ensure that any normalization on the strings under the hood isn't stripping underscores, etc 🤷🏻‍♂️ 

Like I know the registry normalizes types to lowercase strings in some places, so this should be ensuring that a connection registered with underscores is also removed when deregistered

tests/wpunit/ConnectionRegistrationTest.php
kidunot89: > Specifically, the Connection Interfaces release made `nodes` non nullable. The error was just getting suppressed before.

That first node was never supposed to be a falsy value in the first place. I change it to `test`.

tests/wpunit/AccessFunctionsTest.php
```diff
@@ -2232,15 +2232,21 @@ public function testRegisterConnectionWithUnderscores() {
...
 
 		register_graphql_connection(
 			[
 				'fromType'              => 'RootQuery',
 				'toType'                => 'Post',
 				'fromFieldName'         => 'test_field_with_underscores',
 				'allowFieldUnderscores' => true,
+				'connectionTypeName'    => 'TestConnectionFieldWithUnderscores',
```
jasonbahl, line 2243:
```suggestion
				'connectionTypeName'    => 'Test_Connection_With_Underscores',
```

tests/wpunit/AccessFunctionsTest.php
jasonbahl, line 2235:
```suggestion
		deregister_graphql_connection( 'Test_Connection_With_Underscores' );
```

jasonbahl (approved):

PR #3133, opened 2024-05-16
build: Clean up dist

### LINKED ISSUE #3126: Remove development-related files from SVN on release

Some of those files are not necessary to be [packaged](https://plugins.trac.wordpress.org/browser/wp-graphql/#trunk/build).

CleanShot 2024-05-06 at 10 56 23@2x

We do not send to SVN development-related files.

_No response_

### Additional Context

_No response_

josephfusco:
Updates the `.distignore` to prevent unnecessary files from being included in the build. 

Fixes #3126


josephfusco:
@justlevine good call!

Given that `.gitattributes` is only being used for export-ignore, it seems the only benefit would be for when a zip is downloaded from GitHub.

download zip

If vendor dependencies aren't available in this zip, it doesn't feel like it's adding any value right now.

PR #3139, opened 2024-05-21
fix: `$settings_fields` param on "graphql_get_setting_section_field_value" filter not passing the correct type

### LINKED ISSUE #3138: Bug: graphql_get_setting_section_field_value filter params return different types than documented
### Description

The filter `graphql_get_setting_section_field_value` has the following `$params`:

```php
* @param mixed               $value          The value of the field
* @param mixed               $default_value  The default value if there is no value set
* @param string              $option_name    The name of the option
* @param array<string,mixed> $section_fields The setting values within the section
* @param string              $section_name   The name of the section the setting belongs to
```

However, the 4th param `$section_fields` doesn't always return an array. Sometimes it returns `null` because of the `get_option()` call above the filter (see: https://github.com/wp-graphql/wp-graphql/blob/develop/access-functions.php#L816)

### Steps to reproduce

1. Add a filter with Typing on the callback like so: 

```php
add_filter( 'graphql_get_setting_section_field_value', function( $value, $default_value, string $option_name, array $section_fields, string $section_name ) { 
  // do something custom here
  return $value;
}, 10, 5 );
```

2. see the following error: 

```php
Fatal error: Uncaught TypeError: WPGraphQLIDE\ensure_graphiql_link_is_unchecked(): Argument #4 ($section_fields) must be of type array, string given
```

### Additional context

_No response_

### WPGraphQL Version

1.26.0

### WordPress Version

6.6-alpha-58181

### PHP Version

8.2.19

_No response_

- [X] Yes

- [X] Yes

jasonbahl:
This ensures that the `$settings_fields` param passed to the `graphql_get_setting_section_field_value` filter is an array as documented in the filter params. Previously the value was documented to be an array, but was not always returning an array. 

closes #3138

### DIFF
--- a/access-functions.php
+++ b/access-functions.php
@@ -813,21 +813,24 @@ function register_graphql_settings_fields( string $group, array $fields ) {
  * @since 0.13.0
  */
 function get_graphql_setting( string $option_name, $default_value = '', $section_name = 'graphql_general_settings' ) {
-	$section_fields = get_option( $section_name );
+	$section_fields = get_option( $section_name, [] );
 
 	/**
 	 * Filter the section fields
-
+	 *
 	 * @param array<string,mixed> $section_fields The values of the fields stored for the section
 	 * @param string              $section_name   The name of the section
 	 * @param mixed               $default_value  The default value for the option being retrieved
 	 */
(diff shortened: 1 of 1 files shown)


josephfusco (approved):
Looks good! Possibly add tests to cover this scenario from popping up again

PR #3143, opened 2024-05-29
feat: Enhance tab state management with query arguments and localStorage fallback

### LINKED ISSUE #3142: Deep Linking for WPGraphQL Settings pages

The WPGraphQL Settings page allows for settings to be separated into tabs. (Plugins like WPGraphQL for WooCommerce and others leverage this feature). 

I would like to be able to provide links directly to a specific tab. i.e. `/wp-admin/admin.php?page=graphql-settings&group=${name_of_setting_group}` that would allow the settings page to be linked to with a specific tab already in focus.

I propose that the tabs push state to the URL and load state from the URL using a query param such as "group" or similar.

This would allow for "deep linking" to the tabbed pages of the WPGraphQL Settings page.

_No response_

### Additional Context

The WPGraphQL IDE that is being re-architected over here: https://github.com/wp-graphql/wpgraphql-ide will be merged into WPGraphQL core in the future and we want to be able to guide folks through setting up the new IDE by providing a link directly to the new IDE settings page.

josephfusco:
This pull request enhances the tab state management for the WPGraphQL settings page by integrating query arguments with a fallback to localStorage. This ensures that tab states are sharable via URLs while maintaining backward compatibility with users' existing localStorage preferences.

closes #3142

Testing
1. Navigate to a tab using a URL with the tab query argument and verify the correct tab opens.
2. Clear the query argument, navigate to different tabs, and ensure the last active tab is remembered across page loads.
3. Click on different tabs and verify that both the URL and localStorage are updated correctly.

[2024-05-29 at 17.09.37 - Yellow Finch.webm](https://github.com/wp-graphql/wp-graphql/assets/6676674/5ab35e61-054c-4b82-8018-67a1b23c51d8)

### DIFF
--- a/src/Admin/Settings/SettingsRegistry.php
+++ b/src/Admin/Settings/SettingsRegistry.php
@@ -678,7 +678,7 @@ class SettingsRegistry {
 	/**
 	 * Tabbable JavaScript codes & Initiate Color Picker
 	 *
-	 * This code uses localstorage for displaying active tabs
+	 * This code uses query arguments and localstorage for displaying active tabs
 	 *
 	 * @return void
 	 */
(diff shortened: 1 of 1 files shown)


jasonbahl:
@josephfusco we should add an e2e test for this. 

The Codeception tests don't run JavaScript in the client, so we'd need to add a PlayWright test, similar to this one: https://github.com/wp-graphql/wp-graphql/pull/3137/files#diff-9d15a2c2ea09662e39dfeb06f50ed190a3c6319ba2d0f3a637527a98fa6f7fb3

jasonbahl (changes requested):
add test(s)

src/Admin/Settings/SettingsRegistry.php
```diff
@@ -678,37 +678,38 @@ public function show_forms() {
...
 				$('.group').hide();
 				var activetab = '';
-				if (typeof (localStorage) != 'undefined') {
-					activetab = localStorage.getItem("activetab");
-				}
+				var urlParams = new URLSearchParams(window.location.search);
+				var queryTab = urlParams.get('tab');
 
-				//if url has section id as hash then set it as active or override the current local storage value
-				if (window.location.hash) {
```
jasonbahl, line 700: @josephfusco perhaps we should maintain this mechanism of using the hash? 

the deeplinking is technically already working. If I hover the tabs I can see the `#name-of-tab` in the url and if I copy the link and visit it, the link works 🤔 

I do think the improvments to store in localStorage and push to the url are good, but maybe instead of using `tab` query param, we can continue to use the hash in case anyone already is making use of linking with the hash

jasonbahl (changes requested):

src/Admin/Settings/SettingsRegistry.php
josephfusco: Yes, great call out. I've updated this PR to use the existing hash functionality

PR #3144, opened 2024-05-30
feat: add WPGraphQL IDE as an "Experiment"

jasonbahl:
This uses the beta "Experiments API" to register the new WPGraphQL IDE as an "Experiment"

WIP

- note: the actual raw JS files are not in here, just the built IDE files. We could maintain the JS in here or an external repo that then publishes the build files here or whatever 🤷🏻


src/Admin/Ide/Ide.php
```diff
@@ -0,0 +1,581 @@
...
+
+		if ( $this->user_lacks_capability() ) {
+			return;
+		}
+
+		// Don't enqueue new styles/scripts on the legacy IDE page
+		if ( function_exists( 'get_current_screen' ) ) {
+			$screen = get_current_screen();
+			if ( 'toplevel_page_graphiql-ide' === $screen->id ) {
+				return;
```
justlevine, line 247:
```suggestion
			if ( null !== $screen && 'toplevel_page_graphiql-ide' === $screen->id ) {
				return;
```

src/Admin/Ide/Ide.php
```diff
@@ -0,0 +1,581 @@
...
+	 *
+	 * @param mixed                $value          The value of the field.
+	 * @param mixed                $default_value  The default value if there is no value set.
+	 * @param string               $option_name    The name of the option.
+	 * @param array<string, mixed> $section_fields The setting values within the section.
+	 * @param string               $section_name   The name of the section the setting belongs to.
+	 * @return mixed The modified value of the field.
+	 */
+	public function ensure_graphiql_link_is_unchecked( $value, $default_value, string $option_name, array $section_fields, string $section_name ) {
+		if ( 'show_graphiql_link_in_admin_bar' === $option_name && 'graphql_general_settings' === $section_name ) {
```
justlevine, line 481: Not sure _why_ `$section_fields` can be a `string`, but since we dont need it, easy enough to ignore the type.
```suggestion
	public function ensure_graphiql_link_is_unchecked( $value, $default_value, string $option_name, $section_fields, string $section_name ) {
```

src/Admin/Ide/Ide.php
```diff
@@ -0,0 +1,581 @@
...
+		}
+
+		if ( empty( $key ) ) {
+			return null;
+		}
+
+		$plugin_data = get_plugin_data( __FILE__ );
+
+		return $plugin_data[ $key ] ?? null;
+	}
```
justlevine, line 334: PHPStan: lets enforce the string return type.

```suggestion
		return isset( $plugin_data[ $key ] ) ? (string) $plugin_data[ $key ] : null;
	}
```

justlevine (commented):
Only implementation suggestion (until we add more Admin-related stuff in the API) is to colocate `Ide` in `Experimental\Experiment\WPGraphQLIDE`.

This would theoretically allow us to split up that 600-line class, but more importantly it makes it easy at a glance to determine what code is related to an experiment vs what code is mainlined into core.

justlevine (commented):
> Not sure _why_ `$section_fields` can be a `string`, but since we dont need it, easy enough to ignore the type.
```suggestion
	public function ensure_graphiql_link_is_unchecked( $value, $default_value, string $option_name, $section_fields, string $section_name ) {
``` 

This suggestion should be on the previous line 🤦

jasonbahl:
I'm closing this PR as I don't believe the IDE makes sense as an Experiment. 

The IDE is a full product on its own that deserves its own release lifecycle. 

The Experiments API is still something we want to add to core WPGraphQL and will benefit features such as: 

- Custom Scalars
- Refactored "Where" args (i.e. sort/filter arguments on connections)

These features are smaller in scope and are intended to directly relate to core WPGraphQL, but will require _some_ iteration and feedback before being merged to the core codebase. Once merged, there's high likelihood there would be little iteration of these features, where the IDE will likely have continuous iteration long-term.

justlevine:
Note to future selves:

If a feature is in a _very active state of development_ - defined as "likely requires more frequent public releases than WPGraphQL core's release cadence of `~2-4 weeks`" - it would be better served as a separate plugin than an `Experiment`.

PR #3161, opened 2024-07-02
feat: introduce `$connection_resolver->override_query_arg` and `$connection_resolver->add_query_arg`

jasonbahl:
Deprecates `$connection_resolver->set_query_arg` in favor of newly introduced: 

- `$connection_resolver->override_query_arg`: used to override query args at the resolver level. Good for resolvers that want to enforce a final argument without impact of internal filters or user input. 
- `$connection_resolver->add_query_arg` used to add a query arg which appends / merges with existing query args. Good for args that have arrays, like tax_query. 

n/a

- related to: #3066 

Any other comments?
-------------------
The changes merged in #3066 could have caused a regression, especially for plugin authors that used `$connection_resolver->set_query_arg()` intentionally to override other arguments. 

This now separates the method into 2 methods with (hopefully) better naming: 

- override_query_arg (overrides existing args)
- add_query_arg (appends/merges args when possible)

### DIFF
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -824,22 +824,6 @@ abstract class AbstractConnectionResolver {
 	 */
 
 	/**
-	 * Given a key and value, this sets a query_arg which will modify the query_args used by ::get_query();
-	 *
-	 * @param string $key   The key of the query arg to set
-	 * @param mixed  $value The value of the query arg to set
-	 *
-	 * @deprecated Deprecated since v1.28.0 in favor of $this->get_args();
-	 *
-	 * @return static
-	 */
-	public function set_query_arg( string $key, $value ) {
-		_deprecated_function( __METHOD__, '1.28.0', static::class . '::override_query_arg()' );
-		$this->query_args[ $key ] = $value;
-		return $this;
-	}
-
-	/**
 	 * Given a key and value, this adds a query_arg which will modify the query_args used by ::get_query();
 	 *
 	 * Unlike ::override_query_arg(), this method will append the value to the query arg if the query arg is an array.
@@ -1599,6 +1583,23 @@ abstract class AbstractConnectionResolver {
 		_deprecated_function( __METHOD__, '0.3.0', static::class . '::set_query_arg()' );
 
 		return $this->set_query_arg( $key, $value );
+	}
+
+	/**
+	 * Given a key and value, this sets a query_arg which will modify the query_args used by ::get_query();
+	 *
+	 * @param string $key   The key of the query arg to set
+	 * @param mixed  $value The value of the query arg to set
+	 *
+	 * @deprecated 1.28.0
+	 *
+	 * @return static
+	 *
+	 * @codeCoverageIgnore
+	 */
+	public function set_query_arg( string $key, $value ) {
+		_deprecated_function( __METHOD__, '1.28.0', static::class . '::override_query_arg()' );
+		return $this->override_query_arg( $key, $value );
 	}
 
 	/**
--- a/src/Data/Connection/AbstractConnectionResolver.php
+++ b/src/Data/Connection/AbstractConnectionResolver.php
@@ -829,9 +829,27 @@ abstract class AbstractConnectionResolver {
 	 * @param string $key   The key of the query arg to set
 	 * @param mixed  $value The value of the query arg to set
 	 *
+	 * @deprecated Deprecated since v1.28.0 in favor of $this->get_args();
+	 *
 	 * @return static
 	 */
-	public function set_query_arg( $key, $value ) {
+	public function set_query_arg( string $key, $value ) {
+		_deprecated_function( __METHOD__, '1.28.0', static::class . '::override_query_arg()' );
+		$this->query_args[ $key ] = $value;
+		return $this;
+	}
+
+	/**
+	 * Given a key and value, this adds a query_arg which will modify the query_args used by ::get_query();
+	 *
+	 * Unlike ::override_query_arg(), this method will append the value to the query arg if the query arg is an array.
+	 *
+	 * @param string $key   The key of the query arg to set
+	 * @param mixed  $value The value of the query arg to set
+	 *
+	 * @return static
+	 */
+	public function add_query_arg( string $key, $value ) {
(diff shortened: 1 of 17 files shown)


justlevine:
Code LGTM, but I'm not entirely sold we want to deprecate/add these in 1.x, when they'll prob need to be changed again after a breaking release, and a safer/more robust pattern is for extenders to overload the GraphQL args over the WPQuery args.

Will respond/elaborate in more detail hopefully by EOD tomorrow

justlevine:
Okay breaking the argument down:

## Background

1. _As of v1.27.1_, [`AbstractConnectionResolver::set_query_arg()`](https://github.com/wp-graphql/wp-graphql/blob/0bfa7b368790692159f6ddc52afca7789a766a30/src/Data/Connection/AbstractConnectionResolver.php#L834) exists as a way to "manually" set an underlying `$query_args` key-value, used by [`AbstractConnectionResolver::$query_class`](https://github.com/wp-graphql/wp-graphql/blob/0bfa7b368790692159f6ddc52afca7789a766a30/src/Data/Connection/AbstractConnectionResolver.php#L113) (e.g. `WP_Query`, `WP_Term_Query`, etc.
2. The **justifiable use case** for `::set_query_arg()` is when extenders need to quickly overload resolution behavior for a `$query_args` key-value that _is not already exposed as/mapped from a GraphQL connection arg. (e.g. `RootQuery.myPostsFilteredByMyCustomField` is a WP_Query for posts matching `my_custom_field` meta value).
3. However, **if there is an GraphQL connection arg, we should almost _always_ use that** instead of manipulating the `$query_args` key-value directly. For example when getting [comments on a specific Post](https://github.com/wp-graphql/wp-graphql/blob/3a96572155ebcbaf47a0dad5e609138aaf131058/src/Registry/Utils/PostObject.php#L118-L128), we should:

   ```diff
   'resolve' => static function ( Post $post, $args, $context, $info ) {
     if ( $post->isRevision ) {
       $id = $post->parentDatabaseId;
     } else {
       $id = $post->ID;
     }

   +  // Set the Post ID on the GraphQL args.
   +  $args['contentId'] = absint( $id );

     $resolver = new CommentConnectionResolver( $post, $args, $context, $info );

   -  return $resolver->set_query_arg( 'post_id', absint( $id ) )->get_connection();
   +  return $resolver->get_connection();
   ```

   - This provides lifecycle conformity and reliability when reusing and extending Connection Resolvers and is a huge DX benefit. Extenders can target the same behavior regardless of where the resolver class is consumed; the args are the same at each hook (i.e. a "lifecycle breakpoint"), and extending child connection resolvers remains DRY.
   - The reason we currently _arent_ doing this approach was the lifecycle flakiness of the `AbstractConnectionResolver` class. It was safer to simply have a cudgel on our `AbstractConnectionResolver` class than it was to target the implementation inconsistences (different/missing hooks, inconsistent mapping patterns, retriggering logic (e.g. rerunning the query) when trying to interact with class parameters). However, due the series of improvements to `AbstractConnectionResolver`, these are no longer concerns.
   - Once we are able to introduce breaking changes into `AbstractConnectionResolver`, there will be even _more_ benefits to overloading the connection `$args` instead of using `set_query_arg()`, include performance bumps (e.g. by initializing class variables on demand instead of in the constructor).

## In the context of #3066 
4. The issue the author was _attempting_ to address was **our use of `::set_query_arg()`** leading to the buggy behavior described in point 3 (above)
https://github.com/wp-graphql/wp-graphql/blob/release/v1.27.1/src/Registry/Utils/TermObject.php#L233C7-L243C9
Specifically, by using `::set_query_arg()` for `tax_query` we are _overloading_ any values set for `tax_query` in the `PostObjectConnectionResolver` lifecycle. That means any plugin that wants to also set `tax_query` (e.g. wp-graphql-tax-query)` needs to overload _our_ overload.
5. The author addressed this by changing the behavior of `::set_query_arg()` so instead of _setting_ the arg, it _adds_ the arg to any previously set args. This is bad for several reasons:
   - **It's a breaking change**. If I was currently using `::set_query_arg()` as intended to overload an arg (for legacy reasons described in 3), this is now broken. E.g. a `trashedPosts` field using `$post_object_connection_resolver->set_query_arg( 'post_status', 'trashed'` ), now returns all published _and_ trashed posts.
   - **It's an unwanted change**. For the above reason, this removes the ability to overload query vars altogether, without resorting to a an `add_filter()`, breaking the `resolve` callback pattern.

## In the context of _this_ PR
6. **We should never encourage the use of overloading `$query_args` directly vs passing it as GraphQL `$args` to the constructor** for the reasons mentioned in point 3 (above). There should always be an explicit and specific reason for using `::set_query_arg()` (or whatever it's called - see below).
7. As such, it seems counterproductive to add a special `::add_query_arg()`, when we can accomplish that by using `::get_query_args()` and then `array_merge()`ing what we want back into `::set_qeury_arg()`. It's only 2-3 lines of code, and in this case **the friction is good**.
8. Worse, since the `array_merge()` pattern is implicit to overloading the _GraphQL `$args` in the constructor, we now have two paths to accomplish what _seems_ to be the same thing, but actually has major edge cases that need to be accounted for (since regardless of whether`$query_args` are appended or overloaded, it happens only _after_ the connection resolver has been instantiated).
9. Additionally, as efforts to migrate away from `::set_query_arg()` take place (either due to post-breaking change improvements mentioned above and in the Proposed Steps below), developers who adopt these functions will likely need to almost immediately begin migrating away from them in 2.x).

## Proposed Steps
10. What's remaining is whether to rename `::set_query_arg()` to `::overwrite_query_arg()`, or really the question "why do people keep thinking a function called `**set**_query_arg()" would _append_ a value?
11. I believe that's due to our own use of `::set_query_arg()` whether due to the legacy reasons noted in point 3 above or as a cudgel when it would be assumed that our intent is to add (e.g. the bug with wp-graphql-tax-query).
12. **Taken together**: I think the best step to address the confusion is to simply fix/replace _all_ our uses of `::set_query_arg()` which we need to do anyway (and is the last thing on my list before writing up the "migration guide" for the connecton resolver lifecycle improvements).
   - Whenever possible, overload the GraphQL `$args` passed to the constructor instead of directly manipulating the underlying `$query_args` key-value pair.
   - Review our patterns, and see where it makes more sense to `array_merge()` to a possible existing value instead of overloading it.

 I'm on no sleep, so hope this is coherent 🤞

jasonbahl:
closing in favor of #3162

PR #3169, opened 2024-07-08
feat: enqueued asset changes

kidunot89:
Modifies the `enqueuedScripts` field on the `ContentNode` to better simulate the rendering of a content node a WordPress theme, as well as sorting the results in the correct order for rendering.

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …


src/Type/InterfaceType/ContentNode.php
```diff
@@ -60,6 +132,20 @@ public static function register_type( TypeRegistry $type_registry ) {
 					'enqueuedStylesheets' => [
 						'toType'  => 'EnqueuedStylesheet',
 						'resolve' => static function ( $source, $args, $context, $info ) {
+							// Simulate WP template rendering
+							ob_start();
+							wp_head();
```
justlevine, line 137: This will be incompatible with eventual FSE support, and break several existing WPGraphQL extensions (e.g. seo). I haven't reviewed the PR for real, so without knowing the full goal I gotta hope there's a different way to accomplish this.

kidunot89: No, it shouldn't. I've tested quite extensively at this point.

src/Type/InterfaceType/EnqueuedAsset.php
```diff
@@ -94,8 +94,8 @@ static function ( $before ) {
 						},
 					],
 					'dependencies' => [
-						'type'        => [ 'list_of' => 'EnqueuedAsset' ],
-						'description' => __( 'Dependencies needed to use this asset', 'wp-graphql' ),
+						'type'        => [ 'list_of' => 'String' ],
```
justlevine, line 97: Why drop the entire object? The handles are available via `dependencies.nodes.handle`, and since there's a dataloader + we parse from the global and not a db query there's no performance loss for the extra context.

src/Type/ObjectType/EnqueuedScript.php
```diff
@@ -56,6 +59,17 @@ public static function register_type() {
 							return $script->extra['strategy'];
 						},
 					],
+					'group'     => [
+						'type'        => 'ScriptLoadingGroupEnum',
+						'description' => __( 'The loading strategy to use on the script tag', 'wp-graphql' ),
+						'resolve'     => static function ( \_WP_Dependency $script ) {
+							if ( ! isset( $script->extra['group'] ) ) {
+								return 0;
```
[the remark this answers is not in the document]
justlevine: Should this be on the [interface](https://github.com/WordPress/wordpress-develop/blob/6.5/src/wp-includes/class-wp-dependencies.php#L482)?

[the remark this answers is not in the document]
justlevine: Also, defaulting to `HEADER` only seems true for [scripts](https://github.com/WordPress/wordpress-develop/blob/f1489fbc81fa9e053982a9a8df5ebf1839fca728/src/wp-includes/class-wp-scripts.php#L627), but per the above link, a `false` value is `NO_GROUP`.  Not sure if that's a helpful demarcation for non-script assets, but it does respect WP as source-of-truth

src/Type/InterfaceType/EnqueuedAsset.php
kidunot89: True, and it's breaking CI too. I honestly don't remember if there was major reason for remember, but I'll restore and see if it breaks my solutions. Hopefully, I can just put it back.

src/Type/ObjectType/EnqueuedScript.php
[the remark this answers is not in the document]
kidunot89: Gotcha, I've copied the `group` field to the `EnqueuedAsset` type with a return type to `Integer`. Then renamed the `group` field to `location` in the `EnqueuedScript` file.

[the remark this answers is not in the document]
justlevine: Not sure I follow the change. What is the difference between `group` and `location`?

src/Type/InterfaceType/ContentNode.php
justlevine: Also it runs on the resolvers for both `enqueuedStyles` and `enqueuedScripts`, which means these `do_action()` and `apply_filters()` are getting triggered multiple times too... What's the performance hit with this change?

src/Type/ObjectType/EnqueuedScript.php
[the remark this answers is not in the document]
kidunot89: `group` is the group level that can be applied to all _WP_Dependencies, and `location` is the enum representation of this value exclusive to scripts being either `HEADER` for 0 and  `FOOTER` for 1.

src/Type/InterfaceType/ContentNode.php
kidunot89: Nothing that I've noticed, but I'm just using WC and some WC premium plugins, and while WC emits a lot of localize scripts for their pages I've haven't seen any slow downs in the query speed due to this change.

PR #3170, opened 2024-07-09
feat: Ability to Sort Variation Terms by Menu order

alexookah:

This PR updates the TermObjectsConnectionOrderbyEnum.php file to include the MENU_ORDER option. This change enables GraphQL queries to sort terms by menu order (MENU_ORDER) when fetching data related to attributes. It enhances the flexibility of sorting options available to users, improving how terms associated with attributes are organized and presented.

```
  terms(where: {orderby: MENU_ORDER, order: ASC}) {
    nodes {
      name
      slug
      taxonomyName
      databaseId
    }
 }
```

https://github.com/wp-graphql/wp-graphql-woocommerce/issues/475
https://github.com/wp-graphql/wp-graphql-woocommerce/issues/389

Any other comments?
-------------------
...

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Type/Enum/TermObjectsConnectionOrderbyEnum.php
+++ b/src/Type/Enum/TermObjectsConnectionOrderbyEnum.php
@@ -38,6 +38,10 @@ class TermObjectsConnectionOrderbyEnum {
 						'value'       => 'description',
 						'description' => __( 'Order the connection by description.', 'wp-graphql' ),
 					],
+					'MENU_ORDER'    => [
+						'value'       => 'menu_order',
+						'description' => __( 'Order by the menu order value', 'wp-graphql' ),
+					],
 					'COUNT'       => [
 						'value'       => 'count',
 						'description' => __( 'Order the connection by item count.', 'wp-graphql' ),


josephfusco:
Hey @alexookah, thanks for opening this pull request!

WP_Term_Query doesn't use menu_order as an orderby parameter. https://github.com/WordPress/wordpress-develop/blob/6.5/src/wp-includes/class-wp-term-query.php#L19-L1182

Can you clarify what you expect this to do and provide some examples?

alexookah:
@josephfusco 
Thanks for checking my PR out.

You cen see these two issues in woocomerce:
https://github.com/wp-graphql/wp-graphql-woocommerce/issues/475
https://github.com/wp-graphql/wp-graphql-woocommerce/issues/389

And I was trying to use my custom order for my variations
Similar to this screenshot:
100875255-10712c80-34a6-11eb-85b0-50acefeda20f

The only way to make it possible and also resolve these 2 issues from wp-graphql-woocomerce is to introduce and add the MENU_ORDER in the terms enum.

I am not sure why menu order is not already used in the order_query as you mention. 

But I tested it out and added the MENU_ORDER in the terms_type_enum (like from the code changes) and the custom order defined from user was returned in the response as expected.

alexookah:
Adding some screenshots showing how it works:

Screenshot 2024-07-09 at 19 48 20
Screenshot 2024-07-09 at 19 48 32

justlevine:
If it's WooCommerce that's exposing a `menu_order` for taxonomies, then wp-graphql-woocommerce is where a filter should be added on the enum to add the possible value (cc @kidunot89 ).

If you're using a custom/different 3rd party plugin that exposes `menu_order` you should filter the enum (and possibly the resolver mapping) based on that implementation locally.

Marking this as invalid (not because your desire to filter Woo Terms by `menu_order` is invalid, but because the core WPGraphQL plugin focuses on WordPress _core functionality_ so we're limited in our ability to address this).

I believe those WooGraphQL issues contain examples of the custom code you need, but if there's something missing, please feel free to ask for help in the WPGraphQL Discord or follow up here).

kidunot89:
@justlevine is correct. @alexookah could you open this PR on the wp-graphql-woocommerce repo?

alexookah:
Thanks @justlevine  for the detailed feedback
@kidunot89 I will try to make a PR in WoographQL repo instead.

I beleive this has to be extended reading existing values and appending the MENU_ORDER type as an additional type when woographQL is installed? and registered on the same type.

Something like this?
Screenshot 2024-07-09 at 22 49 23

jasonbahl:
I'm going to transfer this issue to the WooGraphQL repo.

PR #3172, opened 2024-07-09
fix: only `eagerlyLoadType` on introspection requests

### LINKED ISSUE #3159: Eagerly Load Types should only be true for introspection requests

When a type is defined as `eagerlyLoadType => true` there is a run-time performance cost for each query and this is typically only needed for Introspection queries. 

We should consider "forcing" eagerlyLoadType to be false for non introspection queries and true for introspection. 

_No response_

### Additional Context

see: https://github.com/wpengine/wp-graphql-content-blocks/issues/134#issuecomment-1648475383

jasonbahl:
Any type that is registered to `eagerlyLoadType` will now only be eagerly loaded for Introspection requests (typically used for tooling such as GraphiQL IDE, Codegen, etc)

For standard execution of queries the eagerly loaded types will not be eagerly loaded which dramatically reduces the number of Object Types that are required to be generated per request.

This introduces: 

- WPGraphQL::set_is_introspection_query
- WPGraphQL::is_introspection_query

Closes #3159 

Any other comments?
-------------------
This _might_ be a breaking change if shipped as-is, as the Schema is now generated later in the lifecycle. 

### Performance Tests

I ran some (super basic) performance tests using k6. 

<details>
  <summary>(performance-test.js)</summary>
    
    ```js
    import http from 'k6/http'
    import { check, sleep } from 'k6'
    
    export const options = {
        vus: 10,
        duration: '10s',
    };
    
    export default function() {
        const url = 'http://uri-debugging.local/graphql'
        const query = `
        query {
            posts(first: 10) {
                nodes {
                    __typename
                    title
                    date
                }
            }
        }
        `;
    
    
        // format the query as a query param and append to the url
        // const res = http.get(`${url}?query=${encodeURIComponent(query)}`)
    
        const headers = { 'Content-Type': 'application/json' }
        const res = http.post(url, JSON.stringify({ query: query }), { headers: headers })
        sleep(1)
    
        check(res, {
            'status was 200': r => r.status === 200,
            'transaction time OK (less than 200ms)': r => r.timings.duration < 200
        });
    
    }
    ```
</details>

The test simulates 10 concurrent users executing the following query as a POST request and continues to execute as many times as it can for 10 seconds.

```graphql
    query {
        posts(first: 10) {
            nodes {
                __typename
                title
                date
            }
        }
    }
```

#### BEFORE 

```bash
http_req_duration..............: avg=156.21ms min=52.89ms med=109.72ms max=585.73ms p(90)=331.21ms p(95)=371.64ms
```

#### AFTER

```bash
http_req_duration..............: avg=67.21ms  min=44.3ms  med=56.27ms max=279.04ms p(90)=70.38ms p(95)=164.29ms
```

### DIFF
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -714,7 +714,7 @@ class TypeRegistry {
 			return $this->prepare_type( $type_name, $config );
 		};
 
-		if ( is_array( $config ) && isset( $config['eagerlyLoadType'] ) && true === $config['eagerlyLoadType'] && ! isset( $this->eager_type_map[ $this->format_key( $type_name ) ] ) ) {
+		if ( \WPGraphQL::is_introspection_query() && is_array( $config ) && isset( $config['eagerlyLoadType'] ) && true === $config['eagerlyLoadType'] && ! isset( $this->eager_type_map[ $this->format_key( $type_name ) ] ) ) {
 			$this->eager_type_map[ $this->format_key( $type_name ) ] = $this->format_key( $type_name );
 		}
 	}
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -4,6 +4,7 @@ namespace WPGraphQL\Registry;
 
 use GraphQL\Error\Error;
 use GraphQL\Type\Definition\Type;
+use WPGraphQL;
 use WPGraphQL\Data\DataSource;
 use WPGraphQL\Mutation\CommentCreate;
 use WPGraphQL\Mutation\CommentDelete;
@@ -407,10 +408,10 @@ class TypeRegistry {
 		 *
 		 * @var \WP_Post_Type[] $allowed_post_types
 		 */
-		$allowed_post_types = \WPGraphQL::get_allowed_post_types( 'objects' );
+		$allowed_post_types = WPGraphQL::get_allowed_post_types( 'objects' );
 
 		/** @var \WP_Taxonomy[] $allowed_taxonomies */
-		$allowed_taxonomies = \WPGraphQL::get_allowed_taxonomies( 'objects' );
+		$allowed_taxonomies = WPGraphQL::get_allowed_taxonomies( 'objects' );
 
 		foreach ( $allowed_post_types as $post_type_object ) {
 			PostObject::register_types( $post_type_object );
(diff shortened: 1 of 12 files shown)


src/Registry/TypeRegistry.php
```diff
@@ -714,7 +714,7 @@ public function register_type( string $type_name, $config ): void {
 			return $this->prepare_type( $type_name, $config );
 		};
 
-		if ( is_array( $config ) && isset( $config['eagerlyLoadType'] ) && true === $config['eagerlyLoadType'] && ! isset( $this->eager_type_map[ $this->format_key( $type_name ) ] ) ) {
+		if ( \WPGraphQL::is_introspection_query() && is_array( $config ) && isset( $config['eagerlyLoadType'] ) && true === $config['eagerlyLoadType'] && ! isset( $this->eager_type_map[ $this->format_key( $type_name ) ] ) ) {
```
justlevine, line 717: Can we `use` the class instead of backlashing it inline?doesn't matter _too_ much in TypeRegistry, but generally:
1. Helps prevent IDEs from getting confused.
2. Makes it easy at a glance at the headers to see the class dependencies.

Not saying we should go through the codebase and update old references, but at least we can prevent new ones from slipping in.

src/Utils/QueryAnalyzer.php
```diff
@@ -133,15 +144,18 @@ public static function is_enabled(): bool {
 	}
 
 	/**
-	 * @param \WPGraphQL\Request $request The GraphQL request being executed
+	 * Get the GraphQL Schema.
+	 * If the schema is not set, it will be set.
+	 *
+	 * @return \GraphQL\Type\Schema|\WPGraphQL\WPSchema|null
```
justlevine, line 150: I think the property type is wrong, and it/this function can return just `?\WPGraphQL\WPSchema`

src/WPGraphQL.php
```diff
@@ -140,12 +145,30 @@ public static function set_is_graphql_request( $is_graphql_request = false ) {
...
 	}
 
+	/**
+	 * Set whether the request is an introspection query or not
+	 *
+	 * @param bool $is_introspection_query
+	 *
+	 * @return void
+	 */
+	public static function set_is_introspection_query( bool $is_introspection_query = false ) {
```
justlevine, line 161: New function, can we inline the return type into the method signature?

justlevine:
Tests passing against this branch for:
- wp-graphql-gravity-forms
- wp-graphql-facetwp
- wp-graphql-headless-login
- wp-graphql-rank-math

Of them, Rank Math is strangely experiencing a significant slowdown on interface queries, but that could also be something with my local machine. I'll try to run some k6/xhprof benchmarks over the weekend and see if I can spot a real discrepancy between this branch and `develop`

jasonbahl:
@justlevine sounds good! Ya as much as I want to get this out ASAP I want to get it "right" more than get it out fast

justlevine:
@jasonbahl not sure if its because I'm stuck trying to run these tests on LocalWP + Windows, but while i'm seeing marginal->modest improvements on object queries, queries that rely on interfaces seem to be slower. Would recommend you do some performance comparisons locally as well.

DMing you the test conditions + data I used.

src/Utils/QueryAnalyzer.php
jasonbahl: phpstan disagrees: 

```
Method WPGraphQL\Utils\QueryAnalyzer::get_schema() should return WPGraphQL\WPSchema but returns GraphQL\Type\Schema.
```

jasonbahl: Ah, found out why. Ya, I can get this fixed.

PR #3188, opened 2024-08-12
feat: Add WPGraphQL Extensions Page to the WordPress admin

josephfusco:
This is the 1st pass at #304

## Features

This page is registered as a submenu of WPGraphQL: `/wp-admin/admin.php?page=wpgraphql-extensions`

## UI Scenarios

| State       | Button Label          | Description                     |
|-------------|-----------------------|---------------------------------|
| Uninstalled | Install & Activate    | Indicates the plugin is not installed and can be installed and activated. |
| Installed   | Activate              | Indicates the plugin is installed but not active. |
| Active      | Active (subdued)      | Indicates the plugin is currently active. |

## Demo

https://github.com/user-attachments/assets/85252646-0c0d-4297-a449-1d23988c437d

## Key Points

- Extensions/plugins cannot be deactivated or deleted on this screen by design.
- Plugin data is hardcoded now with the intent to fetch from WordPress.org

## Possible Additions

- Add plugin thumbnail to card (similar to WordPress.org)
- Allow new platforms to be registered (Bitbucket, Gitlab, etc)

## TODO

- [ ] Setup @wordpress/data in order to use @wordpress/notice store


src/Admin/Extensions/Extensions.php
```diff
@@ -310,18 +197,52 @@ private function get_installed_plugins() {
...
+	 */
+	private function load_extensions_from_file(): ?array {
+		// Check if the file exists and is readable
+		if ( ! file_exists( $this->extensions_file_path ) || ! is_readable( $this->extensions_file_path ) ) {
+			return null;
+		}
+
+		// phpcs:ignore WordPressVIPMinimum.Performance.FetchingRemoteData.FileGetContentsUnknown
+		$contents = file_get_contents( $this->extensions_file_path );
+
```
justlevine, line 213: This is a valid smell.

If we really need to use .json over PHP, then we should store the parsed json in a transient (that expires when WPGraphQL updates - so long as it's served locally ), and only `file_get_contents()` when the transient is expired/unset.

jasonbahl: Ya, I want to use JSON as the source of truth as it will ultimately also be used to source the wpgraphql.com/extensions page as well.

jasonbahl: @justlevine I just pushed up a new commit. 

This includes the suggested change of caching the contents of the file and then invalidating the cache whenever WPGraphQL is updated. . .which led to the introduction of a new `update routine` functionality (which I've been wanting to introduce for quite a while). 

I'm still not sold that we _need_ to cache/invalidate cache on a static JSON file as it's on the local file system. We're not doing a remote HTTP request. 

Gutenberg uses file_get_contents, without any additional caching / invalidation: https://github.com/WordPress/gutenberg/blob/6ffb6720e1edbfa7ccf8603ea31c66a1e7521423/lib/blocks.php#L151

jasonbahl: Other places in core use file_get_contents without additional caching as well: 
- https://github.com/WordPress/WordPress/blob/7b9673722e9f9198ae02a17465e2c562c08d5815/wp-includes/embed.php#L1081-L1090
- https://github.com/WordPress/WordPress/blob/7b9673722e9f9198ae02a17465e2c562c08d5815/wp-includes/embed.php#L520-L522

justlevine: So I would actually use a transient instead of `wp_cache_get()`. The goal is to prevent reading a file and parsing json on page visits, not to avoid re-reading within the same request.

jasonbahl: @justlevine ya, I'm not sure a transient is really needed? The sniff is coming from the WordPress VIP ruleset and suggests using `wpcom_vip_file_get_contents` which doesn't use transients directly, it uses `wp_cache_*` methods (see: https://github.com/Automattic/vip-go-mu-plugins/blob/87ffe8c418151b0fa041395be1b967cbcd0681d6/vip-helpers/vip-utils.php#L532-L673).

There's a discussion about a similar sniff here that suggests ignoring the sniff for local files: https://github.com/WordPress/WordPress-Coding-Standards/issues/943#issuecomment-341592431

justlevine: @jasonbahl that thread is referring to an old smell that was recommending using `wp_remote_get` _instead_ of `file_get_contents()`

Putting aside any specific PHPCS error messages, the reasons to _transient_ is that:
- Reading/decoding json is slower than PHP
- All file_get_* are extremely slow on VIP (since they dont have a real file system)
- **The "definitions" change _rarely_ between individual requests** and therefore it's unnecessary overhead to do this _every single time_.

The last point is the most important one, and IMO just as valid for a local json file as when we eventually start serving it from .com (where then too, we should use a transient over caching).

jasonbahl: @justlevine do you have examples of this pattern at play (and even better if there's a corresponding issue explaining _why_)? Searching various codebases and it seems that file_get_contents for local files is pretty common.

jasonbahl: more examples even in core WordPress: 

- https://github.com/WordPress/WordPress/blob/7b9673722e9f9198ae02a17465e2c562c08d5815/wp-includes/block-editor.php#L803
- https://github.com/WordPress/WordPress/blob/7b9673722e9f9198ae02a17465e2c562c08d5815/wp-includes/block-editor.php#L204
- https://github.com/WordPress/WordPress/blob/7b9673722e9f9198ae02a17465e2c562c08d5815/wp-includes/functions.php#L3352
- https://github.com/WordPress/WordPress/blob/7b9673722e9f9198ae02a17465e2c562c08d5815/wp-includes/functions.php#L6893
- https://github.com/WordPress/WordPress/blob/7b9673722e9f9198ae02a17465e2c562c08d5815/wp-includes/formatting.php#L5947

jasonbahl: I'm not really seeing the suggested pattern anywhere else: https://github.com/search?q=org%3AWordPress+file_get_contents&type=code

justlevine: Didn't have time to go through all the links, but most of the GB stuff _is_ cached/transiented or in process. https://github.com/WordPress/gutenberg/issues/45171

From VIP(https://docs.wpvip.com/vip-file-system/media-uploads/)
> Communication with the VIP File System occurs over HTTP. At a minimum, every filesystem-related operation takes 200 milliseconds to process. As a result, file operations such as file_exists() and file_get_contents() should be used at a minimum to prevent a negative impact to page load time. 

Jetpack caches their json using a weird static class pattern https://github.com/Automattic/jetpack/blob/d1ebbe9fd49e3dfe9f01290f3993df8a086f7292/projects/plugins/jetpack/class.jetpack-gutenberg.php#L327

---
Tbh I'm assuming that the switch from a local file to the website will be a relatively short transition (and I'm assuming you will agree that transienting a remote JSON file is the correct approach). Between that and the fact that currently the JSON is small and I don't want to turn on xhprof to get some actual timings, if you want to call this premature optimization I can let it slide and wait until `wp_remote_get()` is used. 🤷

jasonbahl: > I'm assuming you will agree that transienting a remote JSON file is the correct approach
@justlevine yes, agreed here. We're not doing a remote file, however. This JSON file is included in the plugin and will be a local file to each install of the plugin, not a remote call to JSON on wpgraphql.com. 

I will be using the same local.json to populate the extensions page on wpgraphql.com, sourcing from the latest released version on Github, just like how docs source .md files from Github as well.

PR #3191, opened 2024-08-19
devops: WPGraphQLTestCase constant usage updated for extendability

kidunot89:
Replaces all calls to `WPGraphQLTestcase` using `self` to use `static`. e.g. `self::IS_NULL` became `static::IS_NULL`.

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/tests/wpunit/AccessFunctionsTest.php
+++ b/tests/wpunit/AccessFunctionsTest.php
@@ -943,7 +943,7 @@ class AccessFunctionsTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase {
 		$this->assertQuerySuccessful(
 			$actual,
 			[
-				$this->expectedField( 'graphqlInResolver', self::NOT_NULL ),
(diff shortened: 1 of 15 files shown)


jasonbahl (approved):

justlevine:
While this is compatible with existing tests, I would love clarification on why this change is necessary. Afaik our tests aren't meant to be extended, so shouldn't `self::` be the correct syntax here?

(It's a slevomat violation, but thata not reason enough to block, there's lots of code in our tests that aren't compliant with our CS)

kidunot89:
It's because the values that you `self` are being defined by a child class and not the root WPGraphQLTestCommon class.

justlevine:
>are defined by a child class 

Thats what I'm trying to understand: _what_  child class? What would be extending `\tests\wpunit\{SomeSpecific}Test.php`

PR #3196, opened 2024-08-24
feat: expose `EnqueuedAsset.group` and `EnqueuedScript.location` to schema

justlevine:
This PR adds support for exposing the `WP_Dependency->extra['group']` field to the Schema. This field is useful for developers trying to imitate WordPress's frontend, and load their assets at the correct point in the rendered document.

Cherrypicked from #3169 (with some changes noted below - @kidunot89 see bd1627b9c51257b9c6610c7d9b3ec123eeca06e2)

### What's Included
- Exposes the `EnqueuedAsset.group` field, which returns the `Int` value of `$asset->extra['group']`
  - Change from #3169: If unset, this will return `null` for fidelity with WordPress, instead of cast to a default `0`
- Exposes the `EnqueuedScript.groupLocation` field, which returns a `ScriptLoadingGroupEnum` corresponding to header/footer scripts. This semantic meaning is _for DX simplicity only_ (WordPress internally uses `int`s).
  - Change from #3169: renamed from `location` to `groupLocation` to make it clearer that this is a derived value.

Tests (missing from #3169) have been added as well.

Any other comments?
-------------------
This is parceled off from #3169 so it can be reviewed/merged while more investigation continues on the best ways to fix missing assets/dependencies.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 20.04 (wsl2+ devilbox + PHP 8.1.14)

**WordPress Version:** 6.6.1

### DIFF
--- a/src/Type/Enum/ScriptLoadingGroupEnum.php
+++ b/src/Type/Enum/ScriptLoadingGroupEnum.php
@@ -1,0 +1,39 @@
+<?php
+/**
+ * Register the ScriptLoadingGroupEnum Type to the Schema
+ *
+ * @package WPGraphQL\Type\Enum
+ * @since TBD
+ */
+
+namespace WPGraphQL\Type\Enum;
+
+/**
+ * Class ScriptLoadingGroupEnum
+ */
+class ScriptLoadingGroupEnum {
+
+	/**
+	 * Register the ScriptLoadingStrategy Enum Type to the Schema
+	 *
+	 * @return void
+	 */
+	public static function register_type() {
+		register_graphql_enum_type(
+			'ScriptLoadingGroupEnum',
+			[
+				'description' => __( 'Locations for script to be loaded', 'wp-graphql' ),
+				'values'      => [
+					'HEADER' => [
+						'value'       => 0,
+						'description' => __( 'Script to be loaded in document `<head>` tags', 'wp-graphql' ),
+					],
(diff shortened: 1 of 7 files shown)


src/Type/Enum/ScriptLoadingGroupEnum.php
```diff
@@ -0,0 +1,39 @@
...
+class ScriptLoadingGroupEnum {
+
+	/**
+	 * Register the ScriptLoadingStrategy Enum Type to the Schema
+	 *
+	 * @return void
+	 */
+	public static function register_type() {
+		register_graphql_enum_type(
+			'ScriptLoadingGroupEnum',
```
kidunot89, line 23: Change the name to `ScriptLoadingLocationEnum`

kidunot89 (changes requested):

src/Type/Enum/ScriptLoadingGroupEnum.php
justlevine: Fixed in https://github.com/wp-graphql/wp-graphql/pull/3196/commits/cd6c302b74dacfa6d7175d96091b3740d71cea77

justlevine:
@jasonbahl putting this on your post 🦃 radar for review 🙏

jasonbahl (approved):

PR #3208, opened 2024-09-10
feat: expose commenter edge fields

### LINKED ISSUE #3197: CommentAuthor no longer returning the name set in comment_author for comments by non-anonymous users
### Description

Querying the `name` field of a comment author used to return the name set in the `comment_author` field in the `wp_comments` database, it no longer does if the comment was created by a logged in user from the WordPress Admin-Panel. Instead, it always returns the name of the user.

I understand that the node returned by `CommentToCommenterConnectionEdge` can either be a `User` *or* a `CommentAuthor`, but I think if there is a name set in `comment_author` querying the `name` field should always return that value.

This seems to be how it used to work, as we’ve been adding a generic team name to our replies to anonymous comments and until recently (probably v1.27.0) got the team name back in the query, not the name of the WordPress user.

### Steps to reproduce

1. Add a comment in the WordPress Admin Panel
2. Change the `comment_author` field to `'Generic Team'` by editing the comment in the Admin Panel
3. Note the Comment’s ID
4. In the GraphQL IDE execute the following query:

```graphql
query {
  comment(id: [the ID of the comment], idType: DATABASE_ID) {
    author {
      node {
        name
      }
    }
  }
}
```

The expected result should be the following:

```json
{
  "data": {
    "comment": {
      "author": {
        "node": {
          "name": "Generic Team"
        }
      }
    }
  }
}
```

Instead, the result is:

```json
{
  "data": {
    "comment": {
      "author": {
        "node": {
          "name": [Your WordPress User First Name + Last Name]
        }
      }
    }
  }
}
```

### Additional context

I’m quite new to WP GraphQL, so it might very well be that I’m missing something, but the code worked before and only recently stopped working.

Interestingly enough, comments created *before* the update to `v1.27.0` still return the content of `comment_author` despite the comment having a `user_id` that isn’t `0`.

### WPGraphQL Version

1.28.1

### WordPress Version

6.6.1

### PHP Version

8.3.6

Frontend is running via Next.js, but the issue can be reproduced in the GraphQL IDE. WordPress is run through Roots Bedrock.

- [X] Yes

- [X] Yes

jasonbahl:

This introduces new fields on the "Edge" on the `CommentToCommenter` connection:

- name: The display name of the comment author for this particular comment
- url: The url entered for the comment author on this particular comment
- email: The email address representing the author for this particular comment
- ipAddress: IP address of the author at the time of making this comment. 

This allows for the relational data between a Commenter and the Comment to be queried for, in addition to the Commenter node itself. 

### Why?

Comment Authors can be represented by both their User object data _and_ conditional data related to a specific comment.

By exposing this edge data, the client application can combine data from the `User` node and the `Comment.author` edge fields to provide a UI matching similar UIs as the WordPress Dashboard. 

For example:

In the WordPress Dashboard, when moderating comments, the Author is represented as a mix of a User _and_ comment-specific data. 

With these edge fields, we can now query the User that authored the comment as well as the conditional fields specific to the comment:

CleanShot 2024-09-10 at 13 20 53

With this data queried at the edge between the Comment and the Author, we could create a UI similar to what is displayed in the WP Admin, where we see some information (avatar, etc) from the User node, and some information (commenter's name) in place of the User's information:

CleanShot 2024-09-10 at 13 19 25

closes #3197

### DIFF
--- a/access-functions.php
+++ b/access-functions.php
@@ -974,11 +974,6 @@ function register_graphql_admin_notice( string $slug, array $config ): void {
  * @return array|mixed[][] An array of admin notices
  */
 function get_graphql_admin_notices() {
-
-	if ( ! is_admin() ) {
-		return [];
-	}
-
 	$admin_notices = \WPGraphQL\Admin\AdminNotices::get_instance();
 	return $admin_notices->get_admin_notices();
 }
--- a/access-functions.php
+++ b/access-functions.php
@@ -967,3 +967,18 @@ function register_graphql_admin_notice( string $slug, array $config ): void {
 		}
 	);
 }
+
+/**
+ * Get the admin notices registered for the WPGraphQL plugin screens
+ *
+ * @return array|mixed[][] An array of admin notices
+ */
+function get_graphql_admin_notices() {
+
+	if ( ! is_admin() ) {
+		return [];
+	}
+
+	$admin_notices = \WPGraphQL\Admin\AdminNotices::get_instance();
+	return $admin_notices->get_admin_notices();
+}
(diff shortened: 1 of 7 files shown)


josephfusco (approved):
CleanShot 2024-09-12 at 13 38 49@2x
I'm able too see the commenter fields using the same query provided in the PR 🎉

PR #3214, opened 2024-09-12
fix: lazy-resolve `Post.sourceUrl` and deprecate `Post.sourceUrlsBySize`

justlevine:
This PR fixes an issue where trying to get a `sourceUrl` by `size` would cause the image URLs for _all_ the sizes to resolve. This inevitably leads to slowdowns on sites with lots of registered image sizes (e.g. WooCommerce sites)

The fix works by introducing a new `Post::get_source_url_by_size()` public method used instead of the `sourceUrlsBySize` model field, which has been deprecated.

Discord: https://discord.com/channels/838854618406453300/1282730629305208832/1283796889615073321

Any other comments?
-------------------
- Alternatives considered:
  - Initially, I tried turning `Post::sourceUrlsBySize` into a callback, but that would be a breaking change.
  - Moreover, trying to turn _any_ Model field into a callback won't work, since the first time the field is called, the callback + args will be cached and reused. 
- There's some questionable logic around arg coersion in the `sourceUrl` and `fileSize` resolvers in `Registry\Utils\PostObject::register_attachment_fields()`, but that's unaffected by this particular bug/fix.
- ~Not entirely sure if/how to test this, since this is an under-the-hood fix.~

Where has this been tested?
---------------------------

**WordPress Version:** 6.6.2

### DIFF
--- a/src/Model/Post.php
+++ b/src/Model/Post.php
@@ -35,12 +35,12 @@ use WP_Post;
  * @property string  $pingStatus
  * @property string  $slug
  * @property array   $template
- * @property bool $isFrontPage
- * @property bool $isPrivacyPage
- * @property bool $isPostsPage
- * @property bool $isPreview
- * @property bool $isRevision
- * @property bool $isSticky
+ * @property bool    $isFrontPage
+ * @property bool    $isPrivacyPage
+ * @property bool    $isPostsPage
+ * @property bool    $isPreview
+ * @property bool    $isRevision
+ * @property bool    $isSticky
  * @property string  $toPing
  * @property string  $pinged
  * @property string  $modified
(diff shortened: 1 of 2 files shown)


jasonbahl:
> Not entirely sure if/how to test this, since this is an under-the-hood fix.

Do we have tests that query for the image sizes already and can at least confirm this is not a regression?

We could also add specific unit test(s) that call the function directly but not necessarily  execute a GraphQL query like most of our wpunit  tests do.

justlevine:
> > Not entirely sure if/how to test this, since this is an under-the-hood fix.
> 
> Do we have tests that query for the image sizes already and can at least confirm this is not a regression?
> 

Haha yes, that's why the test is failing now.

Got pulled away but basically returning a generator would be a breaking change for classes extending the Post model, while surfacing a model-level function doesn't fit with current conventions. Probably just needs to be wrapped better. 🤔

justlevine:
Callback approach won't work due to how fields are "cached" on the model. Instead I've made the `Post::get_source_url_by_size()` method public, which also means we have a way to test it explicitly (incoming).

justlevine:
PS: this is ready once tests are added.

justlevine:
@jasonbahl this needed tests 🫣

jasonbahl:
@justlevine I just opened this PR with some tests: https://github.com/wp-graphql/wp-graphql/pull/3219

Let me know if you think there are other tests we should consider adding.

PR #3219, opened 2024-10-29
test: add tests for querying different sizes of media items

jasonbahl:
This adds tests for the PR #3214 

n/a


justlevine (approved):
From the coveralls and the test contents, this is probably good enough.

Two potential cases to add - your call if they're worth it.

- That our default size is the same one as the explicit one
- That a custom registered size works as expected (e.g. a product_thumb or a size slug woo uses IRL)

Issue #3179, opened 2024-08-01
Hebrew filename causes composer install error


heggemsnes:
### Description

The hebrew filename in `/tests/_data/templates/` causes `composer require wp-graphql/wp-graphql` to fail on my system:

```
Failed to extract wp-graphql/wp-graphql: (50) '/usr/bin/unzip' -qq '/Users/sigurdheggemsnes/Documents/GitHub/Local/odt/app/public/wp-content/vendor/composer/tmp-a58962bb4a0560b633defac71bbf8065.zip' -d '/Users/sigurdheggemsnes/Documents/GitHub/Local/odt/app/public/wp-content/vendor/composer/83f5e8d8'

error:  cannot create /Users/sigurdheggemsnes/Documents/GitHub/Local/odt/app/public/wp-content/vendor/composer/83f5e8d8/wp-graphql-wp-graphql-a44eef4/tests/_data/templates/ά??????ά-ήΣ??.php
        Illegal byte sequence
        
```

Because of this I receive this error:

```
WPGraphQL appears to have been installed without it's dependencies. It will not work properly until dependencies are installed. This likely means you have cloned WPGraphQL from Github and need to run the command `composer install`.
```

### Steps to reproduce

Run `composer require wp-graphql/wp-graphql` in the terminal 

### Additional context

Using WP local

### WPGraphQL Version

1.27.2

### WordPress Version

6.6.1

### PHP Version

8.3.8

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

### Please confirm that you have disabled ALL plugins except for WPGraphQL.

- [X] Yes
- [ ] My issue is with compatibility with a specific WordPress plugin, and I have listed all my installed plugins (and version info) above.

justlevine:
Thanks for reporting this @heggemsnes 

I'm unable to replicate this (Ubuntu 20.04 + PHP8.2 ) - can you share what Composer version you are using?

Separately, in the majority of use cases you want to use the distribution package from wpackagist `compoaer require wpackagist-plugin/wp-graphql` and not source repo from GitHub. That version doesn't include the test files or dev-dependencies

heggemsnes:
Thanks! This seemed to do the trick! 

Would be great if the docs perhaps reflected this: https://www.wpgraphql.com/docs/quick-start#install-with-composer

PR #3234, opened 2024-11-15
chore: update quick-start.md to provide more clarity around using wpackagist

jasonbahl:

- use wpackagist-plugin/wp-graphql in the composer example

closes: #3179


justlevine (commented):
We should probably link to wpackagist, or even better, include the other composer.json changes necessary

PR #3241, opened 2024-11-29
feat: add option to get filesize externally

SrPatinhas:
This allows the developer to get the filesize of an attachment that can be offloaded or the size be already in the DB, that way, it will not return 0 if the file is not local

No

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …
Linux 5.15.167.4-microsoft-standard-WSL2
**WordPress Version:** …
6.6.2

### DIFF
--- a/src/Registry/Utils/PostObject.php
+++ b/src/Registry/Utils/PostObject.php
@@ -613,6 +613,11 @@ class PostObject {
 						if ( empty( $source_url ) ) {
 							return null;
 						}
+						// allow to get filesize from other sources, like database or offloaded file
+						$filesize = apply_filter('graphql_get_filesize', $image->databaseId);
+						if(!empty($filesize)) {
+							return $filesize;
+						}
 
 						$path_parts    = pathinfo( $source_url );
 						$original_file = get_attached_file( absint( $image->databaseId ) );


justlevine:
Hey @SrPatinhas - `get_attached_file()` is already filterable by WordPress core: https://developer.wordpress.org/reference/hooks/get_attached_file/

Is there some limitation with that hook which is addressed by the `graphql_get_filesize` you're adding here?

SrPatinhas:
Yes, so, in this case the file is on the S3 bucket, and because of the file being external, to get the information, we would need to download it and then get the filesize of the local file, this way, we can simple use the information that is already updated on the DB.

If I just change the `get_attached_file()` the function will continue even though the file does not exits. This is to give an option for the ones that don't keep the files locally.

SrPatinhas:
To give an example:

1. user uploads file to wordpress
2. file is uploaded as a temp file
3. temp file is uploaded to external server (eg: S3, google drive, etc)
4. information about file is saved on database
5. local temp file is deleted

Then to get the information as it is right now,

1. function gets DB file path
2. File doesn't exist locally
3.  returns error / 0
- OR
2. File is download to a local temp file
3. System functions are called to get the file information
- OR 
2. custom action gets information from database
3. Information is returned

justlevine:
Thanks for clarifying @SrPatinhas 🙏

What about the existing `graphql_pre_resolve_field` [filter](https://www.wpgraphql.com/filters/graphql_pre_resolve_field)?

You would
1.  use the `$type_name` and `$field_key` to target the field, 
2. Check if the file is offloaded, and if so return the proxied filesize from your custom meta field.
3. Otherwise, return null so the field uses it's default resolver for native attachments.

If not that filter, there are dozens of others that similarly allow you to granularly override fields and their resolvers at different stages of the lifecycle.

@jasonbahl is the decision maker here, but to me this whole scenario makes this PR feel like tech-debt that deviates from WP core for an edge case.

PR #3236, opened 2024-11-19
feat: submit an extension criteria and validation

jasonbahl:
What does this implement/fix? Explain your changes.
---------------------------------------------------

This PR adds the following: 

- [x] Documentation for how to [Submit a WPGraphQL Extension](https://github.com/wp-graphql/wp-graphql/compare/feat/3049-extensions-page-fix...feat/3205-extensions-criteria#diff-1430e990fa80806f404d3bdfbf61c5eeb9eb3638212c452b13282f5c5c7f3f1e)
- [x] Adds [new Schemas](https://github.com/wp-graphql/wp-graphql/compare/feat/3049-extensions-page-fix...feat/3205-extensions-criteria#diff-b0566199e40770510d3d5ab842ca8936fbbfc552998e25174060f97e45a4a234) for [extensions](https://github.com/wp-graphql/wp-graphql/compare/feat/3049-extensions-page-fix...feat/3205-extensions-criteria#diff-781c952724797bf3d20654e024be4839bc015d7c6b025d73627940ac967e787d) and [single-extension](https://github.com/wp-graphql/wp-graphql/compare/feat/3049-extensions-page-fix...feat/3205-extensions-criteria#diff-deef9690990447d55a995ead176e127f1d5a835515578947003c0e0a32cf59c8) to validate changes to the extensions.json
- [x] [Github Workflow that validates the extensions.json](https://github.com/wp-graphql/wp-graphql/compare/feat/3049-extensions-page-fix...feat/3205-extensions-criteria#diff-da44d06437b7b430c5181ef82a61156ec1ab6b24d9c8c21b66bd244acdfa1e0d) file against `schemas/extensions.json` to ensure that any changes to the extensions.json file (PRs to the community, etc) in the future are valid. 
- [x] [Validate extensions](https://github.com/wp-graphql/wp-graphql/compare/feat/3049-extensions-page-fix...feat/3205-extensions-criteria#diff-11ce1dda3598611f59138dd383821ad207277fac9142acde2c2ab6c6fdefce55R254) before localizing them for the react app
- [x] Adds filters for plugin developers to be able to customize the extensions list
  - [x] `graphql_get_extensions`: Filter the list of extensions
  - [x] `graphql_pre_get_extensions`: Prevent the JSON file from being loaded in the first place (i.e. override it completely)
[...]

PR #3246, opened 2024-11-30
refactor: validate extensions with PHP

justlevine:
Alternative approach to #3236 

Rationale:

- No additional dependencies
- No messing around with JSON
- Built-in validation using PHPStan (no additional CI)
- Closer mimics to how we would handle a REST response from .org with extension data.

(Additionally some internals were cleaned up, but I can always leave that feedback on the original PR if this gets rejected.

Any other comments?
-------------------
Leaving comments on the diff itself.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 24.04 (WSL + ddev + php 8.2.24)

**WordPress Version:** 6.7.1


src/Admin/Extensions/Extensions.php
```diff
@@ -205,139 +204,114 @@ private function get_installed_plugins() {
...
 	/**
 	 * Validate an extension based on the schemas/extensions.json schema.
 	 *
-	 * @param array<string, array<string, mixed>> $extension The extension to validate.
+	 * Sanitization ensures that the values are correctly types, so we just need to check if the required fields are present.
 	 *
-	 * @return mixed|bool|string True if the extension is valid, otherwise an error message.
+	 * @param array<string,mixed> $extension The extension to validate.
+	 *
+	 * @return true|\WP_Error True if the extension is valid, otherwise an error.
```
justlevine, line 234: `WP_Error` makes it more obvious that it's an error.

src/Admin/Extensions/Registry.php
```diff
@@ -0,0 +1,65 @@
...
+ *  description: non-empty-string,
+ *  plugin_url: non-empty-string,
+ *  support_url: non-empty-string,
+ *  documentation_url: non-empty-string,
+ *  repo_url?: string,
+ *  author: array{
+ *   name: non-empty-string,
+ *   homepage?: string,
+ *  },
+ * }
```
justlevine, line 27: This is all we need to validate submissions programmatically with PHPStan.

And the entire file can be dropped once we have a server payload of data with minimal changes to `Extensions.php`

src/Admin/Extensions/Extensions.php
```diff
@@ -205,139 +204,114 @@ private function get_installed_plugins() {
...
 			if ( isset( $installed_plugins[ $slug ] ) ) {
 				$extension['installed'] = true;
 				$extension['active']    = $installed_plugins[ $slug ]['is_active'];
 				$extension['author']    = $installed_plugins[ $slug ]['author'];
-			} else {
-				$extension['installed'] = false;
-				$extension['active']    = false;
 			}
 
+			// @todo Where does this come from?
```
justlevine, line 298: I'm not seeing `extension['setting_path']` populated anywhere or noted in any of the docs.

src/Admin/Extensions/Extensions.php
```diff
@@ -355,6 +329,45 @@ static function ( $a, $b ) {
...
+			$extensions = Registry::get_extenions();
+
+			/**
+			 * Filter the list of extensions, allowing other plugins to add or remove extensions.
+			 *
+			 * @see Admin\Extensions\Registry::get_extensions() for the correct format of the extensions.
+			 *
+			 * @param array<string,mixed> $extensions The list of extensions.
+			 */
+			$extensions = apply_filters( 'graphql_get_extensions', $extensions );
```
justlevine, line 352: Since we're not doing anything expensive there's no need for the tech-debt of a second `pre_*` filter to short-circuit anything.

src/Admin/Extensions/Extensions.php
```diff
@@ -355,6 +329,45 @@ static function ( $a, $b ) {
...
+
+			$valid_extensions = [];
+			foreach ( $extensions as $extension ) {
+				$sanitized = $this->sanitize_extension( $extension );
```
justlevine, line 356: Doing this in a separate step from validation, to reduce the tech debt of an eventual migration to a server payload.

src/Admin/Extensions/Registry.php
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Class Registry
+ *
+ * @phpstan-type Extension array{
+ *  name: non-empty-string,
+ *  description: non-empty-string,
+ *  plugin_url: non-empty-string,
+ *  support_url: non-empty-string,
+ *  documentation_url: non-empty-string,
+ *  repo_url?: string,
```
justlevine, line 22: These were in the documentation, but not used in the JSON schema.

Single source of truth means removing/updating them here is all we need for PHPStan to make sure that they're correctly consumed everywhere else.

src/Admin/Extensions/README.md
justlevine, line 1: 1. We no longer need any special instructions about looking up schemas or validation.
2. I did not like the idea of having to maintain two separate sets of docs with the same info. Did my best to dedupe what's in `docs/submit-extension.md`, but we can probably prune even more.

src/Admin/Extensions/Extensions.php
jasonbahl: @justlevine ah yes, this was one that we had at one point, but removed it and didn't clean up 🤦🏻‍♂️

justlevine: Another reasons to keep the validation coupled with the code in PHP and not in an external schema ;-)

jasonbahl:
@justlevine so, one of the reasons I went with .json was that the WPGraphQL.com website would also source the .json to populate wpgraphql.com/extensions. 

Since the site is built in Next it would be easy to source the .json files and use them there as well as source them in the plugin's extensions page. 

The .JSON could then be validated in Github to ensure extensions being added by users is valid as a merged extension will show on wpgraphql.com as well as the extensions page. 

The .json would also be validated at run-time in the plugin to ensure that it's valid when localizing to wp_localize_script.

src/Admin/Extensions/Extensions.php
jasonbahl: ha! For sure. 

Here was my thoughts on why I was using .json files: 

https://github.com/wp-graphql/wp-graphql/pull/3246#issuecomment-2515620055

src/Admin/Extensions/README.md
jasonbahl: Ok, so with proper PHP Stan types, submitting an extension will now either pass/fail the phpstan checks which nulls the Github workflow, eh?

Then as far as using the data in WPGraphQL.com, I suppose it could be as easy as localizing the json (or similar) for consumption in the NextJS site as well

src/Admin/Extensions/Extensions.php
jasonbahl: > eventual migration to a server payload

What is this server payload?

src/Admin/Extensions/Registry.php
jasonbahl: Ah yes. Good point. I did make a decision on the fly to leave them out for now as the UI doesn't use them. Figured we could make a decision on these fields later.

src/Admin/Extensions/Extensions.php
```diff
@@ -205,139 +204,114 @@ private function get_installed_plugins() {
...
+			'repo_url'    => ! empty( $extension['repo_url'] ) ? esc_url_raw( $extension['repo_url'] ) : null,
+			'author'      => [
+				'name'     => ! empty( $extension['author']['name'] ) ? sanitize_text_field( $extension['author']['name'] ) : null,
+				'homepage' => ! empty( $extension['author']['homepage'] ) ? esc_url_raw( $extension['author']['homepage'] ) : null,
+			],
+		];
 	}
 
 	/**
 	 * Validate an extension based on the schemas/extensions.json schema.
```
jasonbahl, line 228:
```suggestion
	 * Validate an extension.
```

src/Admin/Extensions/Registry.php
```diff
@@ -0,0 +1,65 @@
...
+	 * Array keys are solely used to sort the array and prevent merge conflicts when diffing. They should be unique.
+	 *
+	 * @return array<string,Extension>
+	 */
+	public static function get_extenions(): array {
+		return [
+			'wp-graphql/wp-graphql-smart-cache' => [
+				'name'        => 'WPGraphQL Smart Cache',
+				'description' => 'A smart cache for WPGraphQL that caches only the data you need.',
+				'documentation_url' => 'https://docs.wpgraphql.com/extensions/smart-cache/', // @todo.
```
jasonbahl, line 55:
```suggestion
				'documentation_url' => 'https://github.com/wp-graphql/wp-graphql-smart-cache',
```

justlevine:
@jasonbahl Understood, but I do think that approach is short-sighted. Responding inline:

> @justlevine so, one of the reasons I went with .json was that the WPGraphQL.com website would also source the .json to populate wpgraphql.com/extensions.
> 
> Since the site is built in Next it would be easy to source the .json files and use them there as well as source them in the plugin's extensions page.

We can just as easily expose the extension data to GraphQL in a .com mu-plugin. Or even as a REST endpoint (even pending using .com as the source of truth), since btw the WP REST API also generates schemas and that's the pattern

> The .JSON could then be validated in Github to ensure extensions being added by users is valid as a merged extension will show on wpgraphql.com as well as the extensions page.

PHPStan handles that validation automatically, be it publicly in WPGraphQL or in an mu-plugin sourced endpoint. _And_ extension submitters get it in their IDE automatically without having to do any additional configs.

The manual .json approach doesn't just require a new GH workflow but additional dependencies for what is a short-term solution in technical terms (from a time POV, it might take a while to serve this from .com, but I think we can agree that it doesn't make sense for the extensions source of truth to be in the plugin itself. And even if you don't, I think you can agree that creating a PR to manually update a registry - whether PHP or external .json file - is _definitely_ not an ideal pattern and one of the first things we'd want to abstract away if the Extensions page takes off ).

> 
> The .json would also be validated at run-time in the plugin to ensure that it's valid when localizing to wp_localize_script.

This is worded like a feature, but to me it reads like a workaround for a bug: "we aren't using PHP to control the schema so we need a json validation library instead of just validating-in-code."

src/Admin/Extensions/Extensions.php
jasonbahl: Ok, ya, the more I think about this, the more I like the simplicity. I was way over-complicating, because of my "need" to use the same list of extensions as localized data for the react extensions page in the plugin but also within the WPGraphQL.com site. 

In my mind we needed something that could be read by both JS and PHP and .json seemed to be the right approach. 

This simplifies things, and now I will just need to get the extensions data for display on wpgraphql.com, but since it already sources Blog Posts, etc from a WordPress server, it can easily source this list from a WordPress server as well, it doesn't need to source it from the Github repo, which was my initial plan.

jasonbahl:
> We can just as easily expose the extension data to GraphQL in a .com mu-plugin. Or even as a REST endpoint (even pending using .com as the source of truth), since btw the WP REST API also generates schemas and that's the pattern

Or dogfood GraphQL 🤦🏻 

Easy enough to register a WPGraphQLExtension Type and WPGraphQLExtensionAuthor Type and expose the lists to be used on wpgraphql.com Next.js (or Astro or whatever) front end.

> I think we can agree that it doesn't make sense for the extensions source of truth to be in the plugin itself

🤔 Like you're thinking that this will "phone home" and get a list of extensions from outside of the plugin? 

That wasn't my intent. I'm not 💯 opposed to that, but I wasn't looking for something to phone home to an "extensions" server or whatever. 

On one hand, it could make things easier to maintain a list of extensions if it is maintained outside of the codebase, but on the other hand maintaining it in the codebase is much more transparent.

src/Admin/Extensions/Extensions.php
justlevine: E.g. `wp_remote_get( 'https://wpgraphql.com/{wp-json-backend-proxy}/extensions' );`

If the payload of extensions is coming from outside the plugin, we need to sanitize/validate it. So I'm trying to follow conventions that would require minimal changes to that pattern switch.

justlevine:
> That wasn't my intent. I'm not 💯 opposed to that, but I wasn't looking for something to phone home to an "extensions" server or whatever.
> 
> On one hand, it could make things easier to maintain a list of extensions if it is maintained outside of the codebase, but on the other hand maintaining it in the codebase is much more transparent.

_Long term_ I don't think anything else is really feasible or maintainable. And I think transparency is more about the guidelines, the process, and access to the list than where the list is physically stored/served from.

Either way that's not a decision that needs to be made now, nor does it take away the value of doing the local sourcing in PHP vs .json, but it was definitely a factor I took into account with design.

PR #3247, opened 2024-12-03
release: v2.0.0-beta.1

jasonbahl:

This PR updates graphql-php to 15.18.1 and increases the minimum version of PHP to 7.4 (used to be 7.1+). 

The goal here was to make only required updates needed to be compatible with graphql-php 15.18.1. I did update phpstan from 1.11 to 1.12 as well, to help us prepare for updating to phpstan 2 (which is php7.4+ as well). 

I'd be interested to hear from folks how these changes impact extension plugins, etc.


justlevine:
PS: PHPStan is already at 1.12.x in https://github.com/wp-graphql/wp-graphql/pull/3242

readme.txt
justlevine, line 1: `Requires PHP` tag needs a bump.

readme.txt
```diff
@@ -4,7 +4,7 @@ Tags: GraphQL, Headless, REST API, Decoupled, React
 Requires at least: 5.0
 Tested up to: 6.6
 Requires PHP: 7.1
```
jasonbahl, line 6:
```suggestion
Requires PHP: 7.4
```

justlevine:
Just did a smoke test against `wp-graphql-headless-login`, and things seem mostly fine. Two takeaways:

1. The change from `errors.$.debugMessage` to `errors.$.extensions.debugMessage` is innocuous enough for users, but means lots of time on plugin devs making tests compatible, especially if they want to continue to support WPGraphQL 1.x . Perhaps it's worth creating an assert in `wp-graphql-test-case`?

2. Noticed this PR is _already_ out of sync w. bugfixes `develop`, and there's no good way to iron those out against a single, evolving PR. Perhaps we can cut an actual `v2.0` branch, and then give a real prerelease tag for a `v2.0.0-beta.1`?

Will continue to test my other extensions and flag anything else that comes up.

justlevine:
> 2\. Noticed this PR is _already_ out of sync w. bugfixes `develop`, and there's no good way to iron those out against a single, evolving PR. Perhaps we can cut an actual `v2.0` branch, and then give a real prerelease tag for a `v2.0.0-beta.1`?

https://github.com/wp-graphql/wp-graphql/pull/3268

src/Type/WPInterfaceType.php
```diff
@@ -17,24 +17,41 @@ class WPInterfaceType extends InterfaceType {
 	public $type_registry;
 
 	/**
-	 * @var array<string,mixed>
+	 * @var array{
+	 *     name?: string|null,
+	 *     description?: string|null,
+	 *     fields: (callable(): iterable<string, mixed>)|iterable<string, mixed>,
+	 *     interfaces?: (callable(): iterable<callable(): \GraphQL\Type\Definition\InterfaceType|\GraphQL\Type\Definition\InterfaceType>)|iterable<callable(): \GraphQL\Type\Definition\InterfaceType|\GraphQL\Type\Definition\InterfaceType>,
+	 *     resolveType?: (callable(mixed, mixed, \GraphQL\Type\Definition\ResolveInfo): (callable(): (\GraphQL\Type\Definition\ObjectType|string|null)|\GraphQL\Deferred|\GraphQL\Type\Definition\ObjectType|string|null))|null,
```
justlevine, line 25: Where did these type definitions come from? I'm trying to understand whether  it's the types that are incorrect or if our code is out of date with what is now expected with WPGraphQL v15

E.g 
image

readme.txt
```diff
@@ -3,8 +3,8 @@ Contributors: jasonbahl, tylerbarnes1, ryankanner, chopinbach, kidunot89, justle
 Tags: GraphQL, Headless, REST API, Decoupled, React
 Requires at least: 5.0
```
justlevine, line 4: Are we not bumping the minimum version of wordpress?

src/Registry/TypeRegistry.php
```diff
@@ -815,7 +810,10 @@ public function prepare_type( string $type_name, $config ) {
 					$prepared_type = new WPEnumType( $config );
 					break;
 				case 'input':
-					$prepared_type = new WPInputObjectType( $config, $this );
+					// Cast config to exact type expected
+					/** @var array{name?: string|null, description?: string|null, fields: (callable(): iterable<string, mixed>)|iterable<string, mixed>, parseValue?: callable(array<string, mixed>): mixed, astNode?: \GraphQL\Language\AST\InputObjectTypeDefinitionNode|null, extensionASTNodes?: array<\GraphQL\Language\AST\InputObjectTypeExtensionNode>|null} $input_config */
```
justlevine, line 814: Why are all these type coercions necessary (everywhere in this PR).

At bare minimum we should `@phpstan-import-type`, but if we need coercion to begin with its usually a sign of a bigger issue.

This makes me very nervous in a breaking refactor.

I'll be taking a pass on this after #3288

src/Type/WPInterfaceType.php
justlevine: https://github.com/wp-graphql/wp-graphql/actions/runs/13102963390/job/36553507641

src/Request.php
```diff
@@ -201,7 +201,7 @@ protected function get_validation_rules(): array {
 		$validation_rules = GraphQL::getStandardValidationRules();
 
 		$validation_rules['require_authentication'] = new RequireAuthentication();
-		$validation_rules['disable_introspection']  = new DisableIntrospection();
+		$validation_rules['disable_introspection']  = new DisableIntrospection( 1 );
```
justlevine, line 204: Not sure about this, as we're overloading the `DisableIntrospection::isEnabled()` function which is the only thing that uses the constructor param.

readme.txt
justlevine: Per `wp-graphql.php`

```suggestion
Requires at least: 6.0
```

.github/workflows/testing-integration.yml
```diff
@@ -60,11 +60,6 @@ jobs:
             php: '8.1'
           - wordpress: '6.0'
             php: '7.4'
-          - wordpress: '5.9'
```
justlevine, line 63: Also need to remove versions from `deploy-docker-image.yml`

phpcs.xml.dist
```diff
@@ -39,7 +39,7 @@
 		Tests for PHP version compatibility.
 		https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards#Recomended-additional-rulesets
 	-->
-	<config name="testVersion" value="7.1-"/>
+	<config name="testVersion" value="7.4-"/>
```
justlevine, line 42: Also need to bump `minimum_wp_version` to 6.0

src/Type/WPInputObjectType.php
```diff
@@ -112,4 +148,36 @@ public function prepare_fields( array $fields, string $type_name, array $config,
...
+
+		/**
+		 * @todo: This is a temporary fix to prevent the InvariantViolation from being thrown
+		 *    when the fields are not iterable. This is a temporary fix until the issue is resolved.
+		 */
+		if ( ! \is_iterable( $fields ) && ! $fields instanceof InputObjectField ) {
+			$invalidFields = Utils::printSafe( $fields );
+
+			// translators: %1$s is the name of the type and %2$s is the invalid fields
+			throw new InvariantViolation( sprintf( esc_html__( '%1$s fields must be an iterable or a callable which returns an iterable, got: %2$s.', 'wp-graphql' ), esc_html( $this->name ), esc_html( $invalidFields ) ) );
```
justlevine, line 174: The error message doesn't match the todo (which I also am not entirely sure I get, but I'm assuming we can stop this in the constructor instead of in `assertValid()`.

jasonbahl: @justlevine I agree that the todo doesn't make sense. I probably left it as a note to self thinking I had a different way to handle this, but clearly didn't explain well enough 🤦🏻‍♂️ 

I think we can remove the todo comment.

I don't know if we can / should move it to the constructor because of lazy loading types, so I think the assertValid is the right place as it is part of graphql-php's validation phase, which happens after types are registered but before the schema is used. This should ensure that all cross-references between types can be properly validated.

jasonbahl: I pushed up a change, updating the comment and error message.

src/Request.php
jasonbahl: The `\GraphQL\Validator\Rules\DisableIntrospection` class in graphql-php now requires an argument. 

Without this we get phpstan error:

```
 204    Class WPGraphQL\Server\ValidationRules\DisableIntrospection constructor invoked with 0 parameters, 1 required.  
         🪪  arguments.count  
```

I suppose we could update the extending class as well, but this was an easy change to satisfy the need and not change behavior

src/Registry/TypeRegistry.php
jasonbahl: Ya, this could be some of my ignorance with phpstan. There was a lot of phpstan warnings I don't think I fully had a grasp on and was working through things to satisy it. Open to other ways of handling it

justlevine: Mostly resolved by #3289

src/Request.php
jasonbahl: @justlevine I opened this PR to refactor how we're handling the DisableIntrospection instantiation: https://github.com/wp-graphql/wp-graphql/pull/3290

PR #3250, opened 2024-12-06
fix: receiving post for Incorrect uri

### LINKED ISSUE #3240: URI Query - Receiving Post for Incorrect Slug Path
### Description

Is there a way to query by URI?
I have attempted this, but I encountered an issue in my React frontend. When I use $slug with /test/, I receive the post as expected. The same happens when I use /news/test/.
However, I only want to receive the post for /news/test/.

### Steps to reproduce

1. Set backend permalink to `/news/%postname%/`
2. Try query
```
query GetPage($slug: String!) {
  nodeByUri(uri: $slug) {
    __typename
    ... on Page {
      id
      title
      slug
    }
    ... on Post {
      id
      title
      slug
    }
    uri
  }
}
```
First 
```
{
  "slug": "test"
}
```

Second

```
{
  "slug": "/news/test"
}
```

### Additional context

grafik
grafik
grafik

### WPGraphQL Version

4.3.1 

### WordPress Version

6.6.2.

### PHP Version

8.2.x

Im at the moment testing it local.

- [X] Yes

- [X] Yes

jasonbahl:
This fixes a bug where entering a uri for an invalid node might return a node with a different uri. 

- [x] failing test: (didn't run in Github properly, but you can confirm locally with this commit: https://github.com/wp-graphql/wp-graphql/pull/3250/commits/b5e25aaddd1c10e1302665679ff129ee1a284ab8)
- [x] passing test: https://github.com/wp-graphql/wp-graphql/actions/runs/12190547800

Closes #3240 

With this PR in place, following the same steps to reproduce (as documented in #3240):

## Before

Querying for a node by uri might return a node that has a _different_ uri. 

CleanShot 2024-12-05 at 16 43 44

## After

Now, querying for a node using a uri that would 404 in traditional WP will return a `null` instead of returning a node that doesn't match the uri input. 

CleanShot 2024-12-05 at 16 42 24

And querying with the correct uri continues to return the expected node: 

CleanShot 2024-12-05 at 16 47 42

### DIFF
--- a/src/Data/NodeResolver.php
+++ b/src/Data/NodeResolver.php
@@ -213,7 +213,6 @@ class NodeResolver {
 		// Resolve Post Objects.
 		if ( $queried_object instanceof WP_Post ) {
 
-
 			// If Page for Posts is set, we need to return the Page archive, not the page.
 			if ( $query->is_posts_page ) {
 				// If were intentionally querying for a something other than a ContentType, we need to return null instead of the archive.
@@ -501,7 +500,6 @@ class NodeResolver {
 			}
 		}
 
-
 		/**
 		 * Filters the query variables allowed before processing.
 		 *
@@ -604,8 +602,6 @@ class NodeResolver {
 
 		// We don't need the GraphQL args anymore.
 		unset( $this->wp->query_vars['graphql'] );
-
-
 
 		do_action_ref_array( 'parse_request', [ &$this->wp ] );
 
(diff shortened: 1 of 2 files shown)


justlevine (commented):
Instead of extra class props and deviating further from core in methods like `::parse_request()`, i'd recommend we implement a `register_globals()` and `handle_404()` method.

That's significantly more work though.

PR #3251, opened 2024-12-07
feat: implement SemVer-compliant update checker

### LINKED ISSUE #2543: Idea: SemVer auto-update enforcement

justlevine:
This PR implements Semantic Version checking when updating WPGraphQL to a newer version.

### How it works:

- A new `Admin\Updates\UpdateChecker` class has been introduced to check whether an update is likely "non-breaking" (per Semantic Versioning). This class is used both to prevent auto-updates and to display update information before a manual update is applied.

- The `UpdateChecker` class also checks if there are any _explicitly incompatible_ or _possibly untested_ WPGraphQL extensions that should be checked/updated before (auto)updating.

  - New Plugin headers for extensions:
  
     - `Requires WPGraphQL: {x.y.z}` sets a minimum plugin version.
     - `WPGraphQL tested up to: {x.y.z}` sets the last _known_ compatible version.

    Plugins that use the `Requires Plugins: wp-graphql` or use `WPGraphQL` in the plugin name or description are detected with an `Unknown` version compatibility.

- The `PluginScreenLoader` and `UpdateScreenLoader` classes inject the notices/modals into the admin screens to warn and confirm about untested extensions that are potentially breaking.

  - A new `packages/updates` package contains the `.scss` to style our notices/modals. Additionally, `webpack-remove-empty-scripts` has been added to our NPM deps to prevent the generation of an empty `.js` file.

### New filters:

-  `wpgraphql_enable_major_autoupdates` - allows for major versions to be auto-updateable. Defaults `false`.
-  `wpgraphql_untested_release_type` - determines which SemVer release type should be considered potentially "breaking". Defaults `major`.
-  `graphql_get_dependents` and `graphql_get_possible_dependents` filter the list of (possible) WPGraphQL extensions.

### Todo
- [x] Implement APIs
- [x] Implement Autoupdate Prevention
- [x] Prevent activation ( + notice ) of plugins that don't meet the `Requires WPGraphQL` version
- [x] Inject Upgrade notice on Plugin Screen
- [x] Implement warning modals on Manual Update.
- [x] Attribute prior art.
- [ ] (Bonus) Show Upgrade Notices from `readme.txt`
- [ ] (Bonus) Explore reuse in semver-compatible extensions.

Closes #2543

Testing Instructions
-------------------
1. Install a WPGraphQL Extension. (you can copy the mock extensions from `test/data/plugins` into `wp-content/plugins` )
2. Mock an available update:

   ```php
   add_action( 'init', static function() {
	$update_plugins = get_site_transient( 'update_plugins' );

	if ( ! $update_plugins ) {
		return;
	}

	// Prime.
	if ( empty( $update_plugins->response ) ) {
		$update_plugins->response = [];
	}

	if ( isset( $update_plugins->no_update['wp-graphql/wp-graphql.php'] ) ) {
		$update_plugins->response['wp-graphql/wp-graphql.php'] = $update_plugins->no_update['wp-graphql/wp-graphql.php'];

		unset( $update_plugins->no_update['wp-graphql/wp-graphql.php'] );
	}

	// Stub new version.
	if ( isset( $update_plugins->response['wp-graphql/wp-graphql.php'] ) ) {
		// Ensure array.
		$update_plugins->response['wp-graphql/wp-graphql.php']->new_version = '4.0.0';
	}

	set_site_transient( 'update_plugins', $update_plugins );
   } );
   ```
   
2. **Updates Screen**:
  a. Visit `/wp-admin/update-core.php` and refresh the screen a few times* until the update is available.
  b. Select `WPGraphQL` from the list of plugins to view the modal.
  c. Click `Cancel` to see WPGraphQL get unchecked.
  
3. **Plugins Screen**
  a. Visit `wp-admin/plugins.php` and refresh the screen a few times* until the update is available.
  b. Click the inline `Update now` link to view the modal.

> [!WARNING]
> **DO NOT** try to update WPGraphQL to the fictitious version. (e.g. clicking the `Update now` button in the *modal*).
> It will delete the existing `wp-graphql` plugin, including the sym-linked reference to the repo.

### Plugin Screen notice:
image

### Plugin Screen modal:
image

### Update Screen modal:
image

### Deactivated Plugin notice
image

Any other comments?
-------------------

- Displaying Upgrade Notes:
   Possibly beyond the scope, but if as part of our Upgrade Policy (#3218) we include Upgrade Notes in the dist version of the plugin somewhere, we can display them in our notices. Prior art: https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/admin/plugin-updates/class-wc-plugins-screen-updates.php#L80
   
- SemVer-compliant Extension Updates
  A personal goal is to make the API reusable enough that any WPGraphQL extension could easily tap into it to make their own updates SemVer compliant. While that's _definitely_ beyond the scope, I will be auditing the API/method signatures so we can support this with non-breaking iteration.

Where has this been tested?
---------------------------
**Operating System:** Ubuntu 24.04 ( wsl2 + ddev + php8.2.26)

**WordPress Version:** 6.7.1


jasonbahl:
@justlevine very cool! 

Will take a look 👀

justlevine:
PR is working and ready for feedback.

Of the remaining todos in the PR description:

> -  (Bonus) Explore reuse in semver-compatible extensions.

I'll be testing against `wp-graphql-headless-login` and `wp-graphql-content-blocks` to see how easy it is for them to reuse the classes for their own SemVer auto-updates. This might result in some tweaks to the hooks and signature visibility, but the logic should remain intact.

> - (Bonus) Show Upgrade Notices from readme.txt

@jasonbahl this is your decision as to whether we do this in this PR or even at all.

justlevine:
> (Bonus) Explore reuse in semver-compatible extensions.

Did a basic review against `wp-graphql-headless-login`:

- Using the new `SemVer` class to disable major extension autoupdates took ~15 LOC ✔️
- There is _no good way_ to reuse the `*ScreenLoader` classes and the underlying `UpdateChecker` class. That said, there's no _real_ reason for reuse vs extension-native code, since we're currently only displaying issues regarding possibly incompatible versions, and not a general `Upgrade Notice` or similar. Further, we can make the necessary extensibility changes iteratively, so I see no reason to hold up review/merge for this.

Marking ready for review.

---

Disable autoupdates example:
```php
add_filter(
	'auto_update_plugin', 
	static function( $should_update, $plugin ) {
		// Bail early if the plugin is not the one we're looking for.
		if ( ! isset( $plugin->plugin ) || ! isset( $plugin->new_version) ||'wp-graphql-headless-login/wp-graphql-headless-login.php' !== $plugin->plugin ) {
			return $should_update;
		}

		// If autoupdates are already disabled we don't need to check further.
		if ( false === $should_update ) {
			return $should_update;
		}

		$new_version = sanitize_text_field( $plugin->new_version );
		if ( '' === $new_version ) {
			return $should_update;
		}

		// Check if it's a major release.
		$release_type = \WPGraphQL\Admin\Updates\SemVer::get_release_type( WPGRAPHQL_LOGIN_VERSION, $new_version );
		if ( 'major' === $release_type ) {
			return false;
		}

		return $should_update;
	},
	10,
	2
);
```

src/Admin/Updates/PluginsScreenLoader.php
```diff
@@ -0,0 +1,156 @@
+<?php
+/**
+ * Handles plugin update checks and notifications on the plugins screen.
+ *
+ * Code is inspired by and adapted from WooCommerce's WC_Plugin_Screen_Updates class.
```
jasonbahl, line 5:
```suggestion
 * Code is inspired by and adapted from WooCommerce's WC_Plugins_Screen_Updates class.
```

src/Admin/Updates/Updates.php
```diff
@@ -0,0 +1,191 @@
...
+				'version' => $plugin[ UpdateChecker::VERSION_HEADER ],
+			];
+			deactivate_plugins( $file );
+		}
+
+		// Display a notice to the user.
+		if ( ! empty( $notice_data ) ) {
+			set_transient( 'wpgraphql_incompatible_plugins', $notice_data );
+		}
+	}
```
jasonbahl, line 159: Upon quick scan (haven't tested or thoroughly read all the code yet), this stands out as something a bit concerning. 

Mostly leaving a note for myself here, but I need to fully understand what situations will cause this. My initial reaction is that I don't want WPGraphQL to be responsible for de-activating other plugins as it could cause a lot of confusion / distrust. 

That said, if this is VERY CLEARLY documented and communicated as a feature and is limited very specifically to plugins that require WPGraphQL (which I assume is already the case - need to look through and test more), then we're probably ok with it.

jasonbahl (commented):
I'm still looking through and testing things but wanted to leave a few comments.

jasonbahl:
> Plugins that use the Requires Plugins: wp-graphql or use WPGraphQL in the plugin name or description are detected with an Unknown version compatibility.

Correct me if I'm mistaken, but I just want to clarify that the recommendation for extension authors now would be to use _all 3_ headers. (i.e. we're not replacing "Requires Plugins: wp-graphql", but enhancing things with _additional_ headers)

i.e. a current WPGraphQL Extension plugin that had the following header: 

```php
 * Requires Plugins: wp-graphql // This header is recognized by WordPress core, and used on some existing extensions
 ```
 
 would adjust like so:
 
 ```php
  * Requires Plugins: wp-graphql
  * Requires WPGraphQL: 1.30 // Adjust as needed
  * WPGraphQL tested up to: 1.30 // Adjust version as testing is confirmed
 ```

jasonbahl:
@justlevine I opened a PR to your branch here with some updates to the messaging: https://github.com/justlevine/wp-graphql/pull/3

jasonbahl:
> Requires Plugins: wp-graphql - prevents an extension from being activated if WPGraphQL is not active, and prevents WPGraphQL from being deactivated if the extension is active.
This would e.g. make sense for WPGraphQL IDE (that can't run at all without WPGraphQL) but you wouldnt want it for let's say CPT UI (or arguably for WPGraphQL for ACF) since you want to still be able to configure your types even if WPGraphQL is (temporarily) disabled.

Ok, I get where you're going with this. WPGraphQL for ACF specifically does use some functions from WPGraphQL in the admin UI screens for configuring ACF Field Groups and Fields to show in the schema, so WPGraphQL really is required. There's likely some defensive programming to ensure the functions exist before calling them (I should probably double check 👀, but WPGraphQL really is required for it). . .but I get the spirit of the idea. Some plugins might be _compatible with_ WPGraphQL but might not TRULY require it. 

>Requires WPGraphQL: 1.30 this sets a minimum version constraint, which must be matched for the extension to be activated.
If WPGraphQL isn't active, the extension can still be activated (e.g. CPT UI above).


👌🏻 

> WPGraphQL tested up to 1.30: This offers an extra level of reliability when upgrading, by allowing autoupdates for major WPGraphQL releases (e.g. requires 1.30, tested up to 3.5 ).
If an extension is missing this header/it falls outside the incoming WPGraphQL version, autoupdates will be prevented for major releases, and the confirmation modal is shown before manually updating.

Great summary! This will help document this feature for release. 🙌🏻

jasonbahl (dismissed):

jasonbahl (approved):

PR #3261, opened 2024-12-13
Update term connection resolver in PostObject.php

### LINKED ISSUE #3238: terms return null if asPreview is set to true
### Description

When I try to make a preview of my single post type (post or custom), the terms node is returned as null, even if I trigger a "draft" revision from WordPress

### Steps to reproduce

1. Create a post with at least one terms (custom or category)
2. Go to GraphQL and try to request the post with the `asPreview` to `true`
```graphql
query ($postId: ID!, $asPreview: Boolean){
  post(id: $postId, idType: DATABASE_ID, asPreview: $asPreview) {
    terms {
      nodes {
        name
      }
    }
  }
}
```

### Additional context

image

### WPGraphQL Version

1.29.2

### WordPress Version

6.7

### PHP Version

8.1.23

- Add WPGraphQL SEO 4.22.5
- JSON Basic Authentication 0.1
- WPGraphQL Content Blocks 4.0.1
- WPGraphQL for ACF 2.4.1
- WPGraphQL Offset Pagination 0.2.0

- [X] Yes

- [X] Yes

jasonbahl:

This updates the resolver to use the parentDatabaseId if the connection is coming from a preview post.

NOTE: This might require an update to TermObjectConnectionResolver.php to allow `$taxonomy` to be an array instead of just a string, to satisfy phpstan. Unconfirmed yet, but highly likely going to see some phpstan errors I believe. (I just made these changes directly in Github to get a quick PR together).

closes #3238 

Any other comments?
-------------------

## Before

Querying a preview post returns `null` for terms instead of returning the parent (published post) terms

CleanShot 2024-12-13 at 09 35 00

## After

Querying a preview post returns the parent (published post) terms

CleanShot 2024-12-13 at 09 34 10

### DIFF
--- a/src/Registry/Utils/PostObject.php
+++ b/src/Registry/Utils/PostObject.php
@@ -198,14 +198,15 @@ class PostObject {
 					),
 					'resolve'        => static function ( Post $post, $args, AppContext $context, ResolveInfo $info ) {
 						$taxonomies = \WPGraphQL::get_allowed_taxonomies();
-						$terms      = wp_get_post_terms( $post->ID, $taxonomies, [ 'fields' => 'ids' ] );
+						$object_id = true === $post->isPreview && ! empty( $post->parentDatabaseId ) ? $post->parentDatabaseId : $post->ID;
 
-						if ( empty( $terms ) || is_wp_error( $terms ) ) {
+						if ( empty( $object_id ) || ! absint( $object_id ) ) {
 							return null;
 						}
-						$resolver = new TermObjectConnectionResolver( $post, $args, $context, $info, $taxonomies );
-						$resolver->set_query_arg( 'include', $terms );
 
+						$resolver = new TermObjectConnectionResolver( $post, $args, $context, $info, $taxonomies );
+						$resolver->set_query_arg( 'object_ids', absint( $object_id ) );
+						$resolver->set_query_arg( 'taxonomies', $taxonomies );
 						return $resolver->get_connection();
 					},
 				];


src/Registry/Utils/PostObject.php
```diff
@@ -198,14 +198,15 @@ protected static function get_connections( WP_Post_Type $post_type_object ) {
...
 						$taxonomies = \WPGraphQL::get_allowed_taxonomies();
-						$terms      = wp_get_post_terms( $post->ID, $taxonomies, [ 'fields' => 'ids' ] );
+						$object_id = true === $post->isPreview && ! empty( $post->parentDatabaseId ) ? $post->parentDatabaseId : $post->ID;
 
-						if ( empty( $terms ) || is_wp_error( $terms ) ) {
+						if ( empty( $object_id ) || ! absint( $object_id ) ) {
 							return null;
 						}
-						$resolver = new TermObjectConnectionResolver( $post, $args, $context, $info, $taxonomies );
-						$resolver->set_query_arg( 'include', $terms );
```
justlevine, line 206: `empty( '0' ) === true` and both `->parentDatabaseId` and `->ID` are already integers


```suggestion

						if ( empty( $object_id ) ) {
							return null;
						}

```

src/Registry/Utils/PostObject.php
```diff
@@ -198,14 +198,15 @@ protected static function get_connections( WP_Post_Type $post_type_object ) {
...
 
+						$resolver = new TermObjectConnectionResolver( $post, $args, $context, $info, $taxonomies );
+						$resolver->set_query_arg( 'object_ids', absint( $object_id ) );
+						$resolver->set_query_arg( 'taxonomies', $taxonomies );
```
justlevine, line 209: 1. If we're already updating this callback, better to insert these via `$args['where]` than to overload it after the fact with `->set_query_arg()`
2. Not by my desk to confirm, but `set_query_arg( 'taxonomies' )` seems redundant when it's also passed to the constructor.

jasonbahl: ah yes! I think I added the set_query_arg first, changed my mind, but didn't remove it 🤦🏻 

I made this PR from the Github UI mostly so I had a placeholder. Will update the PR locally now.

Issue #3259, opened 2024-12-12
Filter comments by date


jasonbahl:
### What problem does this address?

Currently there is no way to query for comments and filter them by date. 

I would like to be able to query for comments and filter them by date / date range. 

### What is your proposed solution?

I propose introducing a CommentDateQuery input type similar to what is in place for filtering posts by date, then introducing a `$where` arg field on comment connections that allow for filtering comments by date or date range. 

I think this ultimately needs more thought, but being able to query something like the following would be nice: 

```graphql
query GetCommentsByDate($after: String, $before: String) {
	comments(
		where: {
			dateQuery: {
				after: $after
				before: $before
			}
		}
	) {
		nodes {
			id
			content
			date
		}
	}
}
```

### Additional Context

Custom Scalars and `@oneOf` directive will likely change how this works. Doesn't mean we should wait for those features, but they will overlap.

jasonbahl:
Marking this blocked with some reasoning: we are preparing an RFC for newly designed Connection `filter` and `sort` args that would replace the current `where` args, and date filtering on comments is a use case that design needs to cover. Adding a `dateQuery` where arg now would mean shipping the shape twice, and a custom DateTime scalar is likely relevant to how the date input gets typed, so this should be addressed after those conventions are figured out. In the meantime this is achievable today with `register_graphql_connection_where_arg()` plus the `graphql_map_input_fields_to_wp_comment_query` filter, since `WP_Comment_Query` supports `date_query` natively.

Issue #3049, opened 2024-02-20
WPGraphQL Extensions Page


jasonbahl:
### What problem does this address?

As a WPGraphQL maintainer I often have to manually redirect folks to extensions that can help solve problems they're running into. 

As a WPGraphQL maintainer I would like to maintain different functionality (i.e. the GraphiQL IDE) as a different plugin that can be updated independent from the core WPGraphQL plugin, and want to easily surface the plugin for users (possibly even auto install it or have UIs that highly recommend installing it for new installs?)

As a WPGraphQL user I have to search for existing plugins that might bridge WPGraphQL with another plugin I'm using (i.e. WPGraphQL for ACF) and vet the solutions (i.e. compare WPGraphQL Content Blocks vs WPGraphQL for Gutenberg vs other plugins in the block editor domain). 

As a WPGraphQL Extension Author it's hard for me to surface my plugin for WPGraphQL users and indicate what version(s) of WPGraphQL it's compatible with, etc. 

### What is your proposed solution?

I propose that WPGraphQL adds an "Extensions" page in the WordPress dashboard (similar to WooCommerce, etc) that would allow WPGraphQL extensions to be surfaced in the WordPress admin in a uniform way.

Extensions could be installed and activated from this screen, update notifications, etc could be centralized, and common things like logic around auto-update semver checks could be centralized.

I think there should be some sort of formal "registry" within WPGraphQL that lists all the extension plugins and meta such as "tags" / "documentation urls", etc.

### Acceptance Criteria

- [x] An "Extensions" menu item is added under the WPGraphQL menu in the WordPress admin dashboard.
- [x] The Extensions page displays a list of available WPGraphQL extensions.
- [x] Each extension listing includes the following details, displayed in a card layout (similar to the Performance Lab plugin)
  - [x] Extension/Plugin Name
  - [x] Short description
  - [x] ~Version (maybe)~ would require remote request, possible add later
  - [x] Extension specific support link (even if it opens a modal)
  - [x] Plugins from wp.org can be inline activated
  - [x] Support Link CTA
  - [x] Link to settings page tab that shows when plugin is active
  - [x] ~Dynamic repository host, e.g. "View on GitHub/Bitbucket/GitLab"~ this can come later

### Follow Up

  - [x] #3202

justlevine:
TL;DR from yesterday's Open Hours:

[...]

PR #3262, opened 2024-12-13
chore: update docs & improve comment functionality

jasonbahl:
This improves some documentation: 

### Authentication and Authorization

Re-organize the article and clarify and expand on some points. 

### Build your First WPGraphQL Extension

- Re-organize the article and expand on some topics to give a bit more depth and clarity. 
- Added pre-reqs. 
- Explained some best practices.
- Provided conceptual and practical example. 

### Categories and Tags

- re-organized the article
- fixed some typos
- provided more examples
- called out performance considerations

### Comments

- Updated content on hierarchical queries / mapping
- Provided more example queries
- Added more clarity around comment visibility / user capabilities and comment statuses
- Discussed comment meta and provided examples for using comment meta
- Called out performance considerations and using pagination
- Added more examples of comment mutations

### Performance

The previously empty performance guide now has content!

----

## Filtering Comments by status

In addition to the documentation updates, this PR also includes a change to Comment connections to better support querying comments by status. 

We've deprecated the `status` where arg on comments which accepted a string, and introduced a new `statusIn` argument which accepts a `[CommentStatusEnum]` (one or more than one CommentStatusEnum value).

This allows for more flexible filtering by status to allow one or more and ensure the statuses are valid by using an Enum instead of a string.

In addition to this functionality change to comments, I also added several more tests around different comment use cases. 

I left a placeholder test re: #3259 (and have it intentionally marked incomplete) which we can visit in the future.

### DIFF
--- a/docs/authentication-and-authorization.md
+++ b/docs/authentication-and-authorization.md
@@ -3,66 +3,76 @@ uri: "/docs/authentication-and-authorization/"
 title: "Authentication and Authorization"
 ---
 
-## A quick word about GraphQL Mutations vs Queries
+## Understanding GraphQL Operations
 
-From a technical perspective, the only differences between GraphQL Queries and Mutations is the `mutation` keyword, and the GraphQL spec requires mutations to be processed synchronously, where queries can be processed Async (in environments that support it).
+Before diving into authentication and authorization, it's important to understand how GraphQL operations work in WPGraphQL.
 
-Other than that, Queries and Mutations are the same, they’re both just strings that map to functions.
+GraphQL has two main operation types:
+- **Queries**: Used for fetching data
+- **Mutations**: Used for modifying data 
 
-Now that we’re clear on Queries vs. Mutations (both are just maps to functions), authentication & authorization is left up to the application layer, not the GraphQL API layer, although some mechanisms in GraphQL can help facilitate these processes.
+While mutations use the `mutation` keyword and must be processed synchronously (one after another), both queries and mutations are fundamentally similar - they map input to resolver functions that interact with WordPress.
 
-## Authentication & Authorization
 
-- **Authentication**: the process of verifying who you are (logging in)
-- **Authorization**: the process of verifying that you have access to something – (the ability to view/change private data)
+## Authentication & Authorization Concepts
 
-## Authentication with WPGraphQL
+- **Authentication**: The process of verifying a user's identity (logging in, validating credentials)
+- **Authorization**: The process of verifying what a user can access or modify (permissions)
 
-Since WPGraphQL is a WordPress plugin that adheres largely to common WordPress practices, there are many ways to make authenticated WPGraphQL requests.
+In WPGraphQL, these processes build on WordPress's existing user and capability systems.
 
-For remote HTTP requests to the `/graphql` endpoint, existing authentication plugins *should* work fine. These plugins make use of sending data in the Headers of requests and validating the credentials and setting the user before execution of the API request is returned:
+## Authentication Methods
 
-- https://github.com/wp-graphql/wp-graphql-jwt-authentication
-- https://github.com/WP-API/Basic-Auth (even though it’s labeled for the REST API, it works well with WPGraphQL – but not recommended for non-SSL connections)
-- https://github.com/WP-API/OAuth1 (labeled for use with the WP REST API, but works well with WPGraphQL)
-- https://make.wordpress.org/core/2020/11/05/application-passwords-integration-guide/ (For WordPress 5.6 and above)
+WPGraphQL supports multiple authentication approaches depending on your use case:
 
-If the remote request is within the WordPress admin, such as the WPGraphiQL plugin, you can use the existing Auth nonce as seen in action [here](https://github.com/wp-graphql/wp-graphiql/blob/82518eafa5f383c5929111431e4a641caace3b57/assets/app/src/App.js#L58-L75).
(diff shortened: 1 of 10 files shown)


src/Data/Connection/CommentConnectionResolver.php
```diff
@@ -48,6 +48,7 @@ protected function prepare_query_args( array $args ): array {
 		 */
 		$query_args['orderby'] = 'comment_date';
 
+
```
justlevine, line 51: Flagging the extra new-lines here, but I'm gonna `update phpcs.xml.dist` so these get fixed automatically by `phpcbf`

tests/wpunit/CommentObjectQueriesTest.php
```diff
@@ -914,4 +914,290 @@ static function ( $text, $comment ) use ( $filtered ) {
...
+					id
+					content
+					status
+				}
+			}
+		}
+		';
+	
+		// Test APPROVE status
+		$variables = ['status' => [ 'APPROVE' ] ];
```
justlevine, line 1004: We probably want a test case with multiple status too.

docs/performance.md
```diff
@@ -3,4 +3,136 @@ uri: "/docs/performance/"
...
+
+## Performance Monitoring
+
+Monitor your GraphQL performance using:
+- Browser Developer Tools Network tab
+- WPGraphQL Query Logging and Trace data (enabled under WPGraphQL > Settings)
+- Server-side logging
+- Application Performance Monitoring (APM) tools
+
+## Additional Tips
```
justlevine, line 130: _Maybe_ also suggest adding a global `id` fields to your queries/fragments, since most query libraries use it for query deduplication/caching.

docs/performance.md
```diff
@@ -3,4 +3,136 @@ uri: "/docs/performance/"
...
 ---
 
-Add some interesting performance details here.
+# Performance Best Practices
+
+WPGraphQL is designed to be performant, but there are several best practices you can follow to ensure optimal performance in your implementation.
+
+## Query Optimization
+
+### Request Only What You Need
```
justlevine, line 12: Wondering if a suggestion to "avoid unnecessary connection nesting" to limit db joins (or similar) works here or is beyond the scope. Practical advice would be to:
- When possible, grab data from the parent object instead of from the connection.
- If you only _sometimes_ need data from the connected object, consider using `@include`/`@defer` directives or splitting your queries up.

docs/authentication-and-authorization.md
```diff
@@ -3,66 +3,76 @@ uri: "/docs/authentication-and-authorization/"
...
+## Authentication Methods
 
-- https://github.com/wp-graphql/wp-graphql-jwt-authentication
-- https://github.com/WP-API/Basic-Auth (even though it’s labeled for the REST API, it works well with WPGraphQL – but not recommended for non-SSL connections)
-- https://github.com/WP-API/OAuth1 (labeled for use with the WP REST API, but works well with WPGraphQL)
-- https://make.wordpress.org/core/2020/11/05/application-passwords-integration-guide/ (For WordPress 5.6 and above)
+WPGraphQL supports multiple authentication approaches depending on your use case:
 
-If the remote request is within the WordPress admin, such as the WPGraphiQL plugin, you can use the existing Auth nonce as seen in action [here](https://github.com/wp-graphql/wp-graphiql/blob/82518eafa5f383c5929111431e4a641caace3b57/assets/app/src/App.js#L58-L75).
+### 1. Remote HTTP Requests
```
justlevine, line 28: Shameless self-promotion for WPGraphQL Headless Login, which supports:
- Basic Auth
- JWT
- OAuth2,
- + the ability register custom authentication providers with PHP.

**But** it's not under the org namespace, so probably better to exclude - I just want it to be an intentional decision.

docs/performance.md
jasonbahl: Ya, I think that makes sense. First goal is to have something on this page rather than the current nothing 😆 

CleanShot 2024-12-13 at 13 33 43

docs/authentication-and-authorization.md
jasonbahl: @justlevine I'm down to include it. Do you want to write up a snippet and PR it?

tests/wpunit/CommentObjectQueriesTest.php
jasonbahl: done!

docs/authentication-and-authorization.md
justlevine: Lol was actually hoping you'd agree about excluding it - at minimum until #3049

docs/performance.md
jasonbahl: > When possible, grab data from the parent object instead of from the connection.

@justlevine do you have a good example of this that you use? I can for sure come up with some examples, just curious if you have any common ones.

jasonbahl: @justlevine we don't have support for the `@defer` directive, unfortunately.

docs/authentication-and-authorization.md
jasonbahl: I'm happy to exclude for now as well. Either way is fine with me. lol

jasonbahl: Been excluded this long. . .no harm in excluding longer 😆

justlevine: haha perfect. Marking resolved as-is

PR #3271, opened 2025-01-15
fix: default cat should not be added when other categories are added

### LINKED ISSUE #3269: Uncategorized is added by default when adding a post
### Description

image

image

When adding a post, it adds the 'Uncategorized' category by default. Is there a way to fix this?

### Steps to reproduce

1. Create a create post mutation as in the picture and use it (only two categories are defined)
2. "Uncategorized" is added

### Additional context

_No response_

### WPGraphQL Version

1.29.3

### WordPress Version

 6.7.1

### PHP Version

7.4.33

_No response_

- [X] Yes

- [ ] Yes

jasonbahl:
This fixes a bug where the `createPost` mutation would assign the "default category" to newly created posts even when other categories were explicitly assigned during the mutation. 

The default category should only be assigned if other categories are not assigned during the creation.

- [x] **FAILING TEST:** https://github.com/wp-graphql/wp-graphql/actions/runs/12798916700/job/35683922462?pr=3271
- [x] **PASSING TEST:** https://github.com/wp-graphql/wp-graphql/actions/runs/12798966725

closes #3269 

Any other comments?
-------------------

## Before

Submitting the `createPost` mutation with categories would create the post with the submitted categories, but _also_ include the default category. 

CleanShot 2025-01-15 at 16 32 04

## After

Submitting the `createPost` mutation with categories would create the post with the submitted categories, and **not include** the default category. 

CleanShot 2025-01-15 at 16 32 30

### DIFF
--- a/src/Data/PostObjectMutation.php
+++ b/src/Data/PostObjectMutation.php
@@ -355,9 +355,9 @@ class PostObjectMutation {
 						return;
 					}
 
-					if ( 'category' === $tax_object->name  ) {
+					if ( 'category' === $tax_object->name ) {
 						$default_category_id = absint( get_option( 'default_category' ) );
-						if ( ! in_array( $default_category_id, $terms_to_connect ) ) {
+						if ( ! in_array( $default_category_id, $terms_to_connect, true ) ) {
 							wp_remove_object_terms( $post_id, $default_category_id, 'category' );
 						}
 					}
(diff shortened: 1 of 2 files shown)


src/Data/PostObjectMutation.php
```diff
@@ -355,6 +355,13 @@ protected static function set_object_terms( int $post_id, array $input, WP_Post_
 						return;
 					}
 
+					if ( 'category' === $tax_object->name  ) {
```
justlevine: Does the bug happen also when `$append` is false? If not, it would be great to gate against and avoid the unnecessary extra db write.

PR #3283, opened 2025-01-27
fix: Error in update checker when WPGraphQL is active as an mu-plugin

### LINKED ISSUE #3282: Bug with get_plugins()
### Description

There is a bug within the Updates Class where is not correctly retrieving plugins from the mu-plugins directory when calling `get_plugins()`. This causes issues with plugin functionality and prevents some plugins from being recognized or loaded.
I think it is necessary to get also the list of mu-plugins with `get_mu_plugins()` and merge with the list of plugins.
`get_plugins` function only gets the plugins that are on /plugins path (Reference: https://developer.wordpress.org/reference/functions/get_plugins/), or don't even check for updates if the plugin is into Must Use.

### Steps to reproduce

1. Set up a Bedrock-based WordPress project:
Follow the Bedrock installation steps if you haven't already, which involves using Composer to manage WordPress setup.
`composer create-project roots/bedrock my-project
cd my-project
`

2. Put the `"wp-graphql/wp-graphql": "^1.29.3"` inside required dependencies on composer.json
3. The plugin needs to be passed into mu-plugins this way:
  ```
"extra": {
    "installer-paths": {
      "web/app/mu-plugins/{$name}/": [
        "type:wordpress-muplugin",
        "wp-graphql/wp-graphql"
      ],
    }
  }
```
4. Run `composer install`

### Additional context

_No response_

### WPGraphQL Version

^1.29.3

### WordPress Version

6.7.1

### PHP Version

>=8.1

_No response_

- [x] Yes

- [x] Yes
- [x] My issue is with compatibility with a specific WordPress plugin, and I have listed all my installed plugins (and version info) above.

jasonbahl:
This fixes a bug where wp-graphql can't be found in the array of plugins when WPGraphQL is active as an mu-plugin.

Closes #3282 

## BEFORE

Error shown when WPGraphQL is active as an MU plugin

CleanShot 2025-01-27 at 15 51 28

## AFTER

No error:

CleanShot 2025-01-27 at 15 53 05

Any other comments?
-------------------

I haven't done thorough testing yet to ensure this change doesn't accidentally cause unpredicted side-effects. The get_plugins function that we were using does use get_plugin_data under the hood, so I think this more direct call to get_plugin_data should do what we need with the added flexibility of working when WPGraphQL is a "regular" or "mu" plugin. 

Would love @justlevine to confirm this change doesn't cause any unintended side effects.

### DIFF
--- a/src/Admin/Updates/Updates.php
+++ b/src/Admin/Updates/Updates.php
@@ -136,8 +136,12 @@ final class Updates {
 	 * Disables plugins that don't meet the minimum `Requires WPGraphQL` version.
 	 */
 	public function disable_incompatible_plugins(): void {
+
+		// Get the plugin data.
+		$plugin_data = get_plugin_data( WPGRAPHQL_PLUGIN_DIR . '/wp-graphql.php' );
+
 		// Initialize the Update Checker.
-		$update_checker = new UpdateChecker( (object) get_plugins()['wp-graphql/wp-graphql.php'] );
+		$update_checker = new UpdateChecker( (object) $plugin_data );
 
 		// Get the incompatible plugins.
 		$incompatible_plugins = $update_checker->get_incompatible_plugins( WPGRAPHQL_VERSION, true );


justlevine:
>Would love @justlevine to confirm this change doesn't cause any unintended side effects.

I should have time to look this over this weekend, but probably won't have time before

jasonbahl:
I've done some tests and everything seems to work as expected with this change and all of the wpunit tests pass with this change as well. 

One thing to note is that much of the admin notice UI work won't show anywhere when WPGraphQL is an mu-plugin due to the fact that: 

> Plugins in the must-use directory will not appear in the update notifications nor show their update status on the plugins page, so you are responsible for learning about and performing updates on your own.

(source: https://developer.wordpress.org/advanced-administration/plugins/mu-plugins/#caveats)

justlevine:
> One thing to note is that much of the admin notice UI work won't show anywhere when WPGraphQL is an mu-plugin due to the fact that:
> 
> > Plugins in the must-use directory will not appear in the update notifications nor show their update status on the plugins page, so you are responsible for learning about and performing updates on your own.
> 
> (source: https://developer.wordpress.org/advanced-administration/plugins/mu-plugins/#caveats)

Curious how woo handles it 🧐

PR #3290, opened 2025-02-04
fix: update DisableIntrospection rule to pass the enabled/disabled state to class instead of being re-defined within the class

jasonbahl:
This determines whether the DisableIntrospection rule is enabled before instantiating the rule. 

related: https://github.com/wp-graphql/wp-graphql/pull/3247#discussion_r1941685001

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -201,15 +201,8 @@ class Request {
 		$validation_rules = GraphQL::getStandardValidationRules();
 
 		$validation_rules['require_authentication'] = new RequireAuthentication();
-
-		// By default, introspection is disabled for non-logged-in users and when debug is off
-		$disable_introspection_rule_enabled = 0;
-		if ( ! get_current_user_id() && ! \WPGraphQL::debug() && 'off' === get_graphql_setting( 'public_introspection_enabled', 'off' ) ) {
-			$disable_introspection_rule_enabled = 1;
-		}
-
-		$validation_rules['disable_introspection'] = new DisableIntrospection( $disable_introspection_rule_enabled );
-		$validation_rules['query_depth']           = new QueryDepth();
+		$validation_rules['disable_introspection']  = new DisableIntrospection();
+		$validation_rules['query_depth']            = new QueryDepth();
 
 		/**
 		 * Return the validation rules to use in the request
--- a/src/Request.php
+++ b/src/Request.php
@@ -201,8 +201,15 @@ class Request {
 		$validation_rules = GraphQL::getStandardValidationRules();
 
 		$validation_rules['require_authentication'] = new RequireAuthentication();
-		$validation_rules['disable_introspection']  = new DisableIntrospection( 1 );
-		$validation_rules['query_depth']            = new QueryDepth();
+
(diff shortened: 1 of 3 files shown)


src/Request.php
```diff
@@ -201,8 +201,15 @@ protected function get_validation_rules(): array {
...
 		$validation_rules['require_authentication'] = new RequireAuthentication();
-		$validation_rules['disable_introspection']  = new DisableIntrospection( 1 );
-		$validation_rules['query_depth']            = new QueryDepth();
+
+		// By default, introspection is disabled for non-logged-in users and when debug is off
+		$disable_introspection_rule_enabled = 0;
+		if ( ! get_current_user_id() && ! \WPGraphQL::debug() && 'off' === get_graphql_setting( 'public_introspection_enabled', 'off' ) ) {
+			$disable_introspection_rule_enabled = 1;
+		}
+
```
justlevine, line 210: Can we do this in the `DisableIntrospection` constructor instead of dirtying up request?

(Ref: I need a way to overload with [this logic](https://github.com/rtCamp/snapwp-helper/blob/be2a9f137aeed70d5ecad82ad31e354713862e1b/src/Modules/GraphQL/Server/DisableIntrospectionRule.php#L24)

src/Server/ValidationRules/DisableIntrospection.php
```diff
@@ -10,15 +10,9 @@
...
-	public function isEnabled(): bool {
-		$enabled = false;
-
-		if ( ! get_current_user_id() && ! \WPGraphQL::debug() && 'off' === get_graphql_setting( 'public_introspection_enabled', 'off' ) ) {
-			$enabled = true;
-		}
-
-		return $enabled;
+	public static function introspectionDisabledMessage(): string {
+		return 'The query contained __schema or __type, however GraphQL introspection is not allowed for public requests by default. Public introspection can be enabled under the WPGraphQL Settings.';
```
justlevine, line 16:
```suggestion
		return __( 'The query contained __schema or __type, however GraphQL introspection is not allowed for public requests by default. Public introspection can be enabled under the WPGraphQL Settings.', 'wp-graphql' );
```

src/Request.php
jasonbahl: @justlevine pushed up some changes that handle it in the constructor and should allow for overloading like your example.

src/Server/ValidationRules/DisableIntrospection.php
justlevine, line 1: This is my recommendation:

Note how it keeps _our_ logic separate from graphql-php's , and so the parent's `isEnabled()` (and other lifecycle) remains intact.

image

src/Request.php
justlevine: Sorry my github is acting wonky: https://github.com/wp-graphql/wp-graphql/pull/3290#discussion_r1943456875

src/Server/ValidationRules/DisableIntrospection.php
jasonbahl: Pushed up another variation. See what you think

justlevine (approved):
😍

PR #3291, opened 2025-02-05
fix: return 415 status code for missing or invalid content type header

### LINKED ISSUE #1692: Missing Content-Type request header results in 500 response code
The `Content-Type` header is rightfully required to properly interpret the POST body of a GraphQL request. When this header is not specified, the response includes a 500 response code indicating a server error. Since this error is caused by a misbehaving client, I would expect a response in the 400 range indicating a client error. Specifically, I expect either a 400 directly (Bad Request) or a 415 (Unsupported Media Type). 

I discovered this using a testing tool that didn't send the `Content-Type` header automatically, but this can be easily reproduced from a shell with curl by suppressing the default header:

`$ curl --request POST --data '{"query": "query QP { posts(first:10) { edges { node { id } } } }"}' -H 'Content-Type:' -o /dev/null -w "%{http_code}" $GRAPHQL_ENDPOINT_URL`

WP GraphQL Version: 1.1.2
Platform: WP Engine; PHP 7.4

jasonbahl:
This updates the response to return a 415 status code if the POST request to the endpoint is missing a "content-type" header or has an invalid value (something other than `application/json`)

- [x] Failing Test: https://github.com/wp-graphql/wp-graphql/actions/runs/13168466543/job/36753997929#step:6:592
- [x] Passing Test: https://github.com/wp-graphql/wp-graphql/actions/runs/13187533719/job/36813105163?pr=3291

This closes #1692 

Any other comments?
-------------------
**This is considered a breaking change** as clients will now have different behavior. 

While it's already documented that the content-type needs to be application/json and that errors were returned, this changes the status code from 500 to 415, which might require some clients to make adjustments. 

We've introduced the `graphql_invalid_content_type_status_code` filter, which would allow for the status code to be set back to `500` (or any other status code) should users want to maintain backward compatibility with v1.* where the status code was 500 when the content type was invalid. 

This PR is to the v2.0 release branch and should be documented in the v2.0 release notes.

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -722,7 +722,7 @@ class Request {
 			return false;
 		}
 
-		return 'application/json' === trim( strtolower( $content_type ) );
+		return 0 === stripos( $content_type, 'application/json' );
 	}
 
 	/**
--- a/src/Request.php
+++ b/src/Request.php
@@ -678,7 +678,7 @@ class Request {
 	 * @throws \Exception
 	 */
 	public function execute_http() {
-		if ( ! $this->validate_http_content_type() ) {
+		if ( ! $this->is_valid_http_content_type() ) {
 			return $this->get_invalid_content_type_response();
 		}
 
@@ -712,13 +712,17 @@ class Request {
 	/**
 	 * Validates the content type for HTTP POST requests
 	 */
-	private function validate_http_content_type(): bool {
+	private function is_valid_http_content_type(): bool {
 		if ( ! isset( $_SERVER['REQUEST_METHOD'] ) || 'POST' !== $_SERVER['REQUEST_METHOD'] ) {
 			return true;
 		}
 
 		$content_type = $this->get_content_type();
-		return ! empty( $content_type ) && 0 === strpos( $content_type, 'application/json' );
+		if ( empty( $content_type ) ) {
+			return false;
+		}
+
+		return 'application/json' === trim( strtolower( $content_type ) );
 	}
 
 	/**
--- a/src/Request.php
+++ b/src/Request.php
@@ -694,6 +694,7 @@ class Request {
 				return [
 					'errors' => [
 						[
+							// translators: %s is the content type header value that was received 
 							'message' => sprintf( esc_html__( 'HTTP POST requests must have Content-Type: application/json header. Received: %s', 'wp-graphql' ), $content_type ),
 						],
 					],
--- a/src/Request.php
+++ b/src/Request.php
@@ -678,42 +678,8 @@ class Request {
 	 * @throws \Exception
 	 */
 	public function execute_http() {
-		// Validate content-type header for POST requests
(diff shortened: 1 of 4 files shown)


moonmeister (approved):

moonmeister:
I was thinking https://graphql.github.io/graphql-over-http/draft/#sec-Media-Types might apply but it looks like that' only for the response header. So looks good to me.

src/Router.php
```diff
@@ -282,10 +282,19 @@ public static function send_header( $key, $value ) {
 	/**
 	 * Sends an HTTP status code.
 	 *
+	 * @param int|null $status_code The status code to send.
+	 *
 	 * @return void
 	 */
-	protected static function set_status() {
-		status_header( self::$http_status_code );
+	protected static function set_status( $status_code = null ) {
```
justlevine, line 289: I believe this is supported by 7.4

```suggestion
	protected static function set_status( ?int $status_code = null ) {
```

src/Request.php
```diff
@@ -678,6 +678,30 @@ function ( $data ) {
 	 * @throws \Exception
 	 */
 	public function execute_http() {
+		// Validate content-type header for POST requests
```
justlevine, line 681: I'd SOLID and move this into a `protected function is_valid_content_type_header(): bool`

src/Request.php
```diff
@@ -678,6 +678,30 @@ function ( $data ) {
...
+		if ( isset( $_SERVER['REQUEST_METHOD'] ) && 'POST' === $_SERVER['REQUEST_METHOD'] ) {
+			// Check both possible content-type header locations
+			$content_type = '';
+			if ( isset( $_SERVER['CONTENT_TYPE'] ) ) {
+				$content_type = sanitize_text_field( $_SERVER['CONTENT_TYPE'] );
+			} elseif ( isset( $_SERVER['HTTP_CONTENT_TYPE'] ) ) {
+				$content_type = sanitize_text_field( $_SERVER['HTTP_CONTENT_TYPE'] );
+			}
+
+			if ( empty( $content_type ) || 0 !== strpos( $content_type, 'application/json' ) ) {
```
justlevine, line 691:
```suggestion
			if ( 'application/json' !== trim( strtolower( $content_type ) ) ) {
```

src/Request.php
jasonbahl: Thanks for the feedback! 🙏🏻 

Pushed up a refactor with better abstraction of logic into more single use methods

src/Request.php
jasonbahl: Updated! 🙌🏻

jasonbahl: @justlevine had to revert this change as it causes failing test: https://github.com/wp-graphql/wp-graphql/actions/runs/13168466543/job/36753997929#step:6:584

some clients (I've seen postman send these headers, for example) send headers like: 

`Content-type: application/json; charset=utf-8` which we should not return a 415 for

Issue #3296, opened 2025-02-08
Unable to change from mutation updateCategory to top level


kinnn2:
### Description

trying to modify a subcategory element to be a top-level category, but it's not working.
tried parentId value of null, -1, and 0.
Moving from a subcategory to another subcategory works fine
Moving to the top level is not working.
I'm confused if it's a bug or tried that doesn't work by default in wp-graphql.

### Steps to reproduce

```
mutation updateCaegory($id: ID!, $parentId: ID) {
  updateCategory(input: {id: $id, parentId: $parentId, clientMutationId: "34343"}) {
    category {
      id
      databaseId
      parentDatabaseId
      name
      parentId
    }
    clientMutationId
  }
}
``` 

```
{
"id":37,
"parentId":0
}
``` 

```
{
"id":37,
"parentId": null
}
```

```
{
"id":37,
"parentId": -1
}
```

### WPGraphQL Version

1.31.1

### WordPress Version

6.7.1

### PHP Version

8.1.29

### Please confirm that you have searched existing issues in the repo.

- [x] Yes

### Please confirm that you have disabled ALL plugins except for WPGraphQL.

- [x] Yes
- [ ] My issue is with compatibility with a specific WordPress plugin, and I have listed all my installed plugins (and version info) above.

jasonbahl:
@kinnn2 thanks for reporting this. My hunch is that this is a legit bug. I will try to reproduce and confirm.

jasonbahl:
@kinnn2 I was able to reproduce. 

The problem is due to some checks within the mutation to see if `parentId` is empty and then bailing if it is. But a `0` id is empty, but should be valid for cases when we want to set the parent to be `0`

I believe I might have a solution for this. Hope to have a PR shortly.

jasonbahl:
@kinnn2 I opened a PR here: https://github.com/wp-graphql/wp-graphql/pull/3308. I believe this should resolve the issue but would love it if you were able to confirm as well. Thanks! 🙏🏻

kinnn2:
@jasonbahl That's perfect, thank you very much

PR #3308, opened 2025-02-11
fix: update term mutation was preventing terms from removing the parentId 

jasonbahl:
This fixes a bug where updating child terms to have no parent was being prevented. 

The bug was that we were using some `if ( empty() ) ` checks vs `if ( ! isset() )` checks.

So, when the ID `0` was passed for the parentId, it was bailing because it was "empty", even though it was set intentionally as 0 and 0 is a valid ID to set as the parent of a term.

- [x] failing tests: https://github.com/wp-graphql/wp-graphql/actions/runs/13274749264/job/37062012106?pr=3308
- [x] passing tests: https://github.com/wp-graphql/wp-graphql/actions/runs/13274806513/job/37062179962?pr=3308

closes: #3296 

Any other comments?
-------------------

### Before:

Executing the updateCategory mutation on a child term, setting the parentId to `"0"` would not properly set the term's parent to `0`, leaving the child category a child category instead of a top level category. 

CleanShot 2025-02-11 at 20 24 53

Also, if we use an encoded id of `term:0` (`dGVybTow`) that also does not work:

CleanShot 2025-02-11 at 20 26 24

### After

Now, setting the `parentId` to `"0"` (or the encoded id for `term:0` (`dGVybTow`)) properly updates the term to be a top-level term with no parent.

CleanShot 2025-02-11 at 20 28 23

CleanShot 2025-02-11 at 20 28 14

Additionally, we've also added the `parentDatabaseId` input as an option. So mutations can accept encoded Ids for `parentId` _or_ the `parentDatabaseId` using the integer value. The docs for each field have been updated to indicate one or the other should be used but not both, and an error is returned if both are used.

### DIFF
--- a/src/Data/TermObjectMutation.php
+++ b/src/Data/TermObjectMutation.php
@@ -77,7 +77,6 @@ class TermObjectMutation {
 			}
 
 			$insert_args['parent'] = 0 !== $parent_id ? $parent_term->term_id : 0;
-
 		}
 
 		/**
--- a/src/Data/TermObjectMutation.php
+++ b/src/Data/TermObjectMutation.php
@@ -48,17 +48,22 @@ class TermObjectMutation {
 		}
 
 		/**
-		 * If the parentId argument was entered, we need to validate that it's actually a legit term that can
-		 * be set as a parent
+		 * If both parentId and parentDatabaseId are provided, throw an error
 		 */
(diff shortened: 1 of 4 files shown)


jasonbahl:
@kinnn2 thanks for confirming! Will get it released shortly 🙏🏻

PR #3338, opened 2025-03-10
fix: Avoid the deprecation warning when sending null header values

marcwieland95:
This fixes the following error appearing on the GraphQL endpoint: `Deprecated: preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated`

…

## Any other comments?

…

## Where has this been tested?

**Operating System:** …

**WordPress Version:** …

### DIFF
--- a/src/Router.php
+++ b/src/Router.php
@@ -260,8 +260,8 @@ class Router {
 	/**
 	 * Sends an HTTP header.
 	 *
-	 * @param string      $key   Header key.
-	 * @param string|null $value Header value. Null values will be converted to empty strings.
+	 * @param string $key   Header key.
+	 * @param string $value Header value.
 	 *
 	 * @return void
 	 * @since  0.0.5
--- a/src/Router.php
+++ b/src/Router.php
@@ -260,8 +260,8 @@ class Router {
 	/**
 	 * Sends an HTTP header.
 	 *
-	 * @param string $key   Header key.
-	 * @param string $value Header value.
+	 * @param string      $key   Header key.
+	 * @param string|null $value Header value. Null values will be converted to empty strings.
 	 *
 	 * @return void
 	 * @since  0.0.5
--- a/src/Router.php
+++ b/src/Router.php
@@ -275,7 +275,7 @@ class Router {
 		 * single SP before interpreting the field value or forwarding the
 		 * message downstream.
 		 */
-		$value = preg_replace( '/\s+/', ' ', $value );
+		$value = ! empty( $value ) ? preg_replace( '/\s+/', ' ', $value ) : '';
 		header( apply_filters( 'graphql_send_header', sprintf( '%s: %s', $key, $value ), $key, $value ) );
 	}
 


src/Router.php
```diff
@@ -275,7 +275,7 @@ public static function send_header( $key, $value ) {
 		 * single SP before interpreting the field value or forwarding the
 		 * message downstream.
 		 */
-		$value = preg_replace( '/\s+/', ' ', $value );
+		$value = preg_replace( '/\s+/', ' ', $value ?? '' );
```
justlevine, line 278: Instead of preg-replacing an empty string, let's check if the string is _empty_ before preg-replacing.

```suggestion
		$value = ! empty( $value ) ? preg_replace( '/\s+/', ' ', $value ): '';
```

justlevine (changes requested):
@marcwieland95 thanks so much for this PR 🙌

Any chance you can provide a trace for the error you're trying to address?

While not enforced by the method signature (luckily for you, since it would fatal), `$value` is typed to accept a `string` not a `string | null`.

I left some feedback to harden this check in a way that isn't negated by passing the correct parameter type, but we should also address the problem at the source before we start obfuscating the real issue.

jasonbahl:
@justlevine the send_header method is typed in the docblock for phpstan, but is not typed in the function signature, and given that it's a public function that can be used by plugins to send headers along with the response, we need to move forward assuming that `null` values are being passed. 

I'll update the docblock to reflect as much and this solution will convert null values to an empty string to play nice.

jasonbahl:
@justlevine I updated the docblock to reflect the reality of how the function is already being used. https://github.com/wp-graphql/wp-graphql/pull/3338/commits/b932ecf961920422ca8536bb9009672559eaa59f

justlevine:
> given that it's a public function that can be used by plugins to send headers along with the response, we need to move forward assuming that null values are being passed.

@jasonbahl there's a difference between assuming and encouraging though.

https://github.com/wp-graphql/wp-graphql/commit/8151e945a8775674fd08827d33b7c7ae543c19a6 was already enough to defensively prevent people from passing `null` IRL. 

Changing the doc-block, however _encourages_ people to pass null.

Question is what's the better path we want to encourage: to pass a header with an empty value (`?string`) or to only send a header if it has a value.

I strongly lean towards the latter.

jasonbahl:
@justlevine do you think it would make sense to add a deprecation notice within the send_header method to discourage null inputs?

i.e. something like:

```php
if ( null === $value ) {
    _deprecated_argument( __METHOD__, '@since next-version', 'Passing null values to send_header is deprecated. Please pass a string value.' );
}
```

This way the docblock can accurately represent how the function currently works (regardless of ideals) and actively discourage users from using it in a way we don't encourage?

Or we can leave the docblock string as-is (even though null is technically valid) and let the defensive fix do it's thing in the mean time.

justlevine:
> @justlevine do you think it would make sense to add a deprecation notice within the send_header method to discourage null inputs?
> 

Yes, I'd want to mark this as deprecated for signature change in a future release, _but_ i'd rather wait until we know when that release might happen, and not prematurely force it for 3.0.  There's no reason to cause a break over this, so we can slate it strategically to occur alongside related changes.

> Or we can leave the docblock string as-is (even though null is technically valid) and let the defensive fix do it's thing in the mean time.

In lieu of a method signature, the docs define what is "technically valid" (regardless if PHP throws an error), so yeah even if there wasn't a good reason to temporarily hold off on event logging, I personally wouldn't feel too bad about the fact that passing `null` doesn't cause an issue despite the function wanting a string 🤷

iandrewt:
Looks like this PR got reverted in this commit: https://github.com/wp-graphql/wp-graphql/commit/835321e6a573165b1f0662f9e01e7103be6e0f01

Issue #3354, opened 2025-04-11
Enhancement: Allow Lazy Type Configuration via Callable Configs


jasonbahl:
### What problem does this address?

Currently, WPGraphQL's type registration requires the entire configuration array upfront, which causes several performance issues:

1. **Translation Function Overhead**: Every type description using `__()` is processed during registration, even if the type is never used:
```php
register_graphql_object_type( 'Hero', [
    'description' => __( 'A hero section with title and image', 'textdomain' ),
    'fields' => [
        'title' => [
            'description' => __( 'The title of the hero section', 'textdomain' ),
            'type' => 'String',
        ],
    ],
] );
```

2. **Eager Configuration Processing**: All configuration values, including:
   - Type descriptions
   - Field descriptions
   - Interface implementations
   - Type resolvers

3. **Memory Usage**: Large configurations with many translated strings and complex field definitions consume memory even when types aren't used in queries.

### What is your proposed solution?

Allow the entire type configuration to be provided as a callable:

```php
register_graphql_object_type( 'Hero', function() {
    return [
        'description' => __( 'A hero section with title and image', 'textdomain' ),
        'fields' => [
            'title' => [
                'description' => __( 'The title of the hero section', 'textdomain' ),
                'type' => 'String',
            ],
        ],
    ];
} );
```

Benefits:
1. **Deferred Translation Processing**: `__()` calls only happen when the type is actually used
2. **Lazy Configuration**: All configuration processing is deferred until needed
3. **Context Awareness**: Configuration can be generated with runtime context
4. **Memory Efficiency**: Configuration arrays only exist when types are instantiated

Example implementation:
```php
class TypeRegistry {
    protected function register_type( string $type_name, $config ) {
        if ( is_callable( $config ) ) {
            // Store the callable for later
            $this->type_configs[$type_name] = $config;
        } else {
            // Existing implementation for array configs
            $this->types[$type_name] = $this->prepare_type( $type_name, $config );
        }
    }

    public function get_type( string $type_name ) {
        if ( isset( $this->type_configs[$type_name] ) && is_callable( $this->type_configs[$type_name] ) ) {
            // Only execute the callable when the type is actually needed
            $config = $this->type_configs[$type_name]();
            $this->types[$type_name] = $this->prepare_type( $type_name, $config );
            // Clear the callable to free memory
            unset( $this->type_configs[$type_name] );
        }
        return $this->types[$type_name] ?? null;
    }
}
```

[...]

PR #3356, opened 2025-04-11
feat: lazy loading fields for Object Types and Interface Types

jasonbahl:

This PR introduces lazy evaluation of GraphQL field definitions through callable functions in the core type system. The key changes focus on:

1. Added support in `WPInterfaceTrait` to handle callable field definitions:
```php
if ( is_callable( $config['fields'] ) ) {
    $config['fields'] = $config['fields']();
}
```

2. Modified `WPInterfaceType` and `WPObjectType` to use callable functions for field definitions:
```php
$config['fields'] = function () use ( $config ) {
    return ! empty( $this->fields ) ? $this->fields : $this->get_fields( $config, $this->type_registry );
};
```

This architectural change defers the execution of field definitions until they are actually needed, rather than defining all fields upfront during schema initialization.

### Before/After Example

Here's an example of how the `User` type fields were refactored:

**Before:**
```php
'fields' => [
    'id' => [
        'description' => __( 'The globally unique identifier for the user object.', 'wp-graphql' ),
    ],
    'name' => [
        'type' => 'String',
        'description' => __( 'Display name of the user.', 'wp-graphql' ),
    ],
    // ... more fields
]
```

**After:**
```php
'fields' => static function () {
    return [
        'id' => [
            'description' => __( 'The globally unique identifier for the user object.', 'wp-graphql' ),
        ],
        'name' => [
            'type' => 'String',
            'description' => __( 'Display name of the user.', 'wp-graphql' ),
        ],
        // ... more fields
    ];
}
```

To measure the impact of this change, we added a `TestGetText` class that tracks all `__()` function calls during schema generation and query execution. This logging shows that translations are now only executed when fields are actually queried, rather than during schema initialization.

```php
class TestGetText {

	/**
	 * @var array
	 */
	protected $gettext_log = [];

	/**
	 * Constructor
	 */
	public function __construct() {
		// Reset at the very start of schema generation
		add_action( 'graphql_process_http_request', [ $this, 'init_log' ], 1 );

		// Track all translations
		add_filter(
			'gettext',
			function ( $text ) {
				$this->gettext_log[] = $text;
				return $text;
			},
			10,
			3
		);

		// Add to extensions
		add_filter( 'graphql_request_results', [ $this, 'add_to_extensions' ], 10, 1 );
	}

	/**
	 * Initialize the log
	 */
	public function init_log(): void {
		$this->gettext_log = [];
	}

	/**
	 * Add translation data to GraphQL extensions
	 *
	 * @param \GraphQL\Executor\ExecutionResult $response The GraphQL response
	 */
	public function add_to_extensions( \GraphQL\Executor\ExecutionResult $response ): \GraphQL\Executor\ExecutionResult {
		$extensions                 = $response->extensions ?? [];
		$extensions['translations'] = [
			'count' => count( $this->gettext_log ),
			// 'log'   => $this->gettext_log,
		];
		$response->extensions       = $extensions;
		return $response;
	}
}

new TestGetText();
```

Working on #3354 (more to come)

## Any other comments?

This change provides significant performance benefits by:

1. **Deferred Translation**: The `__()` function calls for field descriptions are now only executed when the fields are actually used in a query, rather than during schema initialization. This is particularly important for multilingual sites where translation lookups can be expensive.

2. **Reduced Memory Usage**: Field definitions that are rarely used in queries no longer need to be stored in memory during the entire request lifecycle.

3. **Faster Schema Initialization**: The schema can be initialized more quickly since field definitions are not processed until they are actually needed.

4. **Better Resource Management**: Resources like database connections or complex calculations needed for field definitions are only allocated when the fields are actually queried.

The changes maintain full backward compatibility while providing these performance improvements. All existing functionality remains the same, but the implementation is now more efficient, especially for large WordPress installations with complex schemas.


src/Registry/TypeRegistry.php
```diff
@@ -811,6 +811,20 @@ protected function prepare_type( string $type_name, $config ) {
 			$kind           = isset( $config['kind'] ) ? $config['kind'] : null;
 			$config['name'] = ucfirst( $type_name );
 
+			// Allow the description to be a callable
+			if ( WPGraphQL::is_introspection_query() && isset( $config['description'] ) && is_callable( $config['description'] ) ) {
+				$config['description'] = $config['description']();
+			} else {
+				$config['description'] = is_string( $config['description'] ) ? $config['description'] : '';
+			}
+
```
justlevine, line 820: Is there a way we can do this lower-down, I.e only when `$config['description']` or `$config['deprecationReason']` are accessed?

If not, and we need to keep these conditionals then the `else {}` logic here needs to be fixed to avoid the PHP error if `$config['deprecationReason]` isn't set.

justlevine: Regardless of _where_ we do it, I think the how is better like this:

```php
/**
	 * Remove or resolve config args, depending on whether the type is being used for introspection.
	 *
	 * @param array<string,mixed> $config The config array to sanitize
	 * @return array<string,mixed>
	 */
	private function prepare_introspection_config_args( array $config ): array {
		$introspection_keys = [
			'description',
			'deprecationReason',
		];

		foreach ( $introspection_keys as $key ) {
			// If the key isn't callable, we can skip it.
			if ( ! isset( $config[ $key ] ) || ! is_callable( $config[ $key ] ) ) {
				continue;
			}

			// If we're not introspecting, we should remove the key instead of resolving it.
			if ( ! WPGraphQL::is_introspection_query() ) {
				unset( $config[ $key ] );
				continue;
			}

			// If the key is callable, we should resolve it.
			$config[ $key ] = $config[ $key ]();
		}

		return $config;
	}
```

jasonbahl:
Here's a comparison of the current master branch vs this PR:

As we can see comparing cachegrind files (before on the left, after on the right) there's significant reduction to the number of calls to:

- apply_filters
- get_translations_for_domain
- translate
- NOOP_Translations->translate

And the total cost is reduced from: `63,558,472` to `55,086,122`

CleanShot 2025-04-15 at 21 39 17@2x

The cache grind files were generated using:

- LocalWP
  - the following modification to the php.ini.hbs:
```bash
[xdebug]
{{#if os.windows}}
zend_extension = php_xdebug.dll
{{else}}
zend_extension = {{extensionsDir}}/xdebug.so
{{/if}}

{{#if xdebugEnabled}}
xdebug.mode=profile
xdebug.output_dir="/Users/jasonbahl/Local Sites/wpgraphql/logs/xdebug"
xdebug.profiler_output_name="cachegrind.out.%t.%p"
xdebug.use_compression=0
xdebug.profiler_enable=1
xdebug.profiler_enable_trigger=0
xdebug.client_port=9003
xdebug.start_with_request=yes
xdebug.discover_client_host=yes
xdebug.log_level=0
xdebug.start_with_request=yes
xdebug.show_local_vars=1
xdebug.remote_connect_back=1
{{else}}
xdebug.mode=off
{{/if}}
```

Then I enabled xdebug and executed the following query on both the `master` branch and the `feat/3354-lazy-loading-configs` branch. 

```graphql
{
  posts {
    nodes {
      id
    }
  }
}
```

Attached are the 2 files:

- [cachegrind-files.zip](https://github.com/user-attachments/files/19770262/cachegrind-files.zip)

jasonbahl:
Ok, I just pushed up a few more changes and I'm seeing even more improvement. 

I updated my test class:

```php
class TestGetText {

	/**
	 * @var array
	 */
	protected $gettext_log = [];

	/**
	 * Constructor
	 */
	public function __construct() {
		// Reset at the very start of schema generation
		add_action( 'graphql_process_http_request', [ $this, 'init_log' ], 1 );

		// Track all translations
		add_filter(
			'gettext',
			function ( $translation, $text, $domain ) {
				if ( 'wp-graphql' === $domain ) {
					$this->gettext_log[] = $text;
				}
				return $translation;
			},
			10,
			3
		);

		// Add to extensions
		add_filter( 'graphql_request_results', [ $this, 'add_to_extensions' ], 10, 1 );
	}

	/**
	 * Initialize the log
	 */
	public function init_log(): void {
		$this->gettext_log = [];
	}

	/**
	 * Add translation data to GraphQL extensions
	 *
	 * @param \GraphQL\Executor\ExecutionResult $response The GraphQL response
	 */
	public function add_to_extensions( \GraphQL\Executor\ExecutionResult $response ): \GraphQL\Executor\ExecutionResult {
		$extensions                 = $response->extensions ?? [];
		$extensions['translations'] = [
			'count' => count( $this->gettext_log ),
			// 'log'   => $this->gettext_log,
		];
		$response->extensions       = $extensions;
		return $response;
	}
}

new TestGetText();
```

Now I'm only logging translations with the `wp-graphql` text domain so that we can narrow our focus on what we can improve within the plugin. 

- Master branch: `4485`
- Latest commit of this branch (sha: cf73e9780a52d7b100fef3f999f6f8413fa18f20): `63`

CleanShot 2025-04-16 at 10 43 46

And here's the latest cachegrind files:

[cachegrind-cf73e9780a52d7b100fef3f999f6f8413fa18f20.zip](https://github.com/user-attachments/files/19781199/cachegrind-cf73e9780a52d7b100fef3f999f6f8413fa18f20.zip)

src/Type/Enum/PluginStatusEnum.php
```diff
@@ -4,6 +4,7 @@
 
 class PluginStatusEnum {
 
+
```
justlevine, line 7:
```suggestion
class PluginStatusEnum {

```

src/Type/WPEnumType.php
```diff
@@ -96,6 +97,17 @@ private static function prepare_values( $values, $type_name ) {
 		$values = apply_filters( 'graphql_' . lcfirst( $type_name ) . '_values', $values, $type_name );
 		$values = apply_filters( 'graphql_' . $type_name . '_values', $values, $type_name );
 
+		// map over the values and if the description is a callable, call it
+		$values = array_map(
+			static function ( $value ) {
+				if ( ! is_array( $value ) ) {
+					$value = [ 'value' => $value ];
+				}
+				return TypeRegistry::prepare_config_for_introspection( $value );
```
justlevine, line 106: I hate that we're doing this here, and that `prepare_config_for_introspection()` is public), but I don't have a better solution since we don't have an Abstract type, and utils are even worse.

Just sharing in care you have a better idea, if not just hit resolve.

src/Registry/TypeRegistry.php
```diff
@@ -792,6 +819,56 @@ public function register_union_type( string $type_name, array $config ): void {
...
+		return self::$introspection_keys;
+	}
+
+	/**
+	 * Prepare the config for introspection. This is used to resolve callable values for description and deprecationReason for
+	 * introspection queries.
+	 *
+	 * @param array<string,mixed> $config The config to prepare.
+	 *
+	 * @return array<string,mixed> The prepared config.
```
justlevine, line 848: Can we mark this as `@internal`?

(https://github.com/wp-graphql/wp-graphql/pull/3356#discussion_r2051293680)

src/Registry/TypeRegistry.php
```diff
@@ -150,6 +150,13 @@ class TypeRegistry {
 	 */
 	protected $types;
 
+	/**
+	 * The keys that are prepared for introspection.
+	 *
+	 * @var array<string>|null
+	 */
+	protected static $introspection_keys = null;
```
justlevine, line 158: We're 7.4+ we can do cool stuff like this now 🚀🚀🚀

```suggestion
	protected static ?array $introspection_keys = null;
```

src/Registry/TypeRegistry.php
```diff
@@ -920,6 +999,8 @@ protected function prepare_field( string $field_name, array $field_config, strin
 			$field_config['name'] = lcfirst( $field_name );
 		}
 
+		$field_config = self::prepare_config_for_introspection( $field_config );
```
justlevine, line 1002: We should be able to do this _at least_ as late after the `if ( is_array( $field_config['type'] ) ) {}` block (L1067) in this commit)

src/Type/WPEnumType.php
```diff
@@ -64,9 +65,9 @@ private static function prepare_values( $values, $type_name ) {
 			if ( ! empty( $value['description'] ) && is_callable( $value['description'] ) ) {
 				$description = $value['description']();
 			} else {
-				$description = is_string( $value['description'] ) ? $value['description'] : '';
+				$description = isset( $value['description'] ) && is_string( $value['description'] ) ? $value['description'] : '';
 			}
-			$values[ $key ]['description'] = $description;
+			$values[ $key ]['description'] = $description ?? '';
```
justlevine, line 70: Do we still need this foreach loop if we're preparing_for_introspection later in this method? (or vice versa)

jasonbahl: @justlevine this is to prepare the values of the enum, the other is for the Enum Type description

PR #3358, opened 2025-04-18
chore: Update Composer deps

justlevine:

This PR updates the Composer deps to their latest (semver) compatible versions.

Of note:
- `webonyx/graphql-php` v15.20.0 : https://github.com/webonyx/graphql-php/releases/tag/v15.20.0
- `slevomat/coding-standards` deprecated `SlevomatCodingStandard.TypeHints.UnionTypeHintFormat` which we weren't explicitly using in favor of `SlevomatCodingStandard.TypeHints.DNFTypeHintFormat` which we're already implicitly including, so the former is now excluded. Ref: https://github.com/slevomat/coding-standard/blob/34325312cbd9465a222564597e9f56e42750939c/SlevomatCodingStandard/Sniffs/TypeHints/UnionTypeHintFormatSniff.php#L74

-

Nope

## Any other comments?

(Are changesets a thing we need to be adding now?)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
 	],
 	"require": {
 		"php": "^7.4 || ^8.0",
-		"webonyx/graphql-php": "15.19.1",
+		"webonyx/graphql-php": "15.20.0",
 		"ivome/graphql-relay-php": "0.7.0",
 		"appsero/client": "2.0.4"
 	},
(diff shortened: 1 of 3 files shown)


justlevine:
Noticed this didnt trigger Codeception so I made #3359

jasonbahl (approved):

jasonbahl:
> (Are changesets a thing we need to be adding now?)

@justlevine no need to add a changeset. They should be automated when a PR is merged. 😄

PR #3362, opened 2025-04-19
chore: inline (non-breaking) native return types

justlevine:

This PHP inlines return types for number functions where this change has been confirmed to be nonbreaking i.e.:
 - cant cause an error
 - wont cause an error
 - would already cause an error
 - are private and cant affect extenders/users

Native return types provide:
- better IDE support (dont need an extra doc-block parser)
- better LLM support (less likely to get confused or lose context)
- better DX (since the error happens where it's caused and not where the correct type is expected ).
- better refactorability (since we know what parts of the codebase are "safe".

No but I Should make more for all the 3.0 breaking type changes I wanna make 🧑‍🍳

### DIFF
--- a/access-functions.php
+++ b/access-functions.php
@@ -39,7 +39,7 @@ function graphql_format_name( string $name, string $replacement = '_', string $r
  *
  * @todo refactor to use Utils::format_field_name()
  */
-function graphql_format_field_name( $field_name ) {
+function graphql_format_field_name( $field_name ): string {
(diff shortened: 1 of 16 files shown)


jasonbahl:
> No but I Should make more for all the 3.0 breaking type changes I wanna make 🧑‍🍳

@justlevine yes please! 🙏🏻

jasonbahl (approved):

Issue #3370, opened 2025-04-26
Remove deprecations and improve type safety for v3.0


justlevine:
### What problem does this address?

There is a lot of tech debt in the existing codebase including
- Deprecations that need to be removed
- Type safety and DX issues that could be a breaking change to implement, but can be choreographed in advance via deprecations and other messages.

During 2.0 we made a decision to limit the number of breaking changes. With actual planning and a target window we can do this seamlessly in time for v3.0 .

### What is your proposed solution?

- [ ] Update all _existing_ codebase deprecations to log a deprecation message about 3.0 removal
- [ ] Choose a cut-off date after which additional deprecations will only be slated for _after_ 3.0
- [ ] Audit codebase for other 3.0 deprecation candidates for deprecation + fix.
- [ ] Audit codebase for _4.0_ early deprecation candidates (so devs can preemptively handle these same time they're  handling the _required_ ones).
- [ ] Write an announcement post.
- [ ] (Bonus) Relocate whatever possible deprecations to their own namespace.

### What alternatives have you considered?

Letting the tech debt _continue_ to accumulate for another major release.

### Additional Context

- We could theoretically leverage #3292 for some kind of "upgrade compatibility report". 
- Part of the tech debt to deprecate is the overly-large access. We may want to take this opportunity to (deprecate and eventually) mark some classes as final, change method/variable access levels, and consolidate different hooks that do the same purpose.

justlevine:
@jasonbahl to make this happen, I need

1. Your greenlight to update existing deprecation messages.
2. A _tentative_ target for when the cutoff should be for deprecating things for v3.0

Everything else can be figured out as progress is made here or on specific PRs.

jasonbahl:
Designating **#3902** as the master v3.0 tracking issue. All items from this issue have been carried over there so they're tracked in one place:

- ✅ **Approval to remove existing deprecations in v3.0** (granted above) — recorded in #3902.
- 🗓️ **Cutoff date** — the tentative "end of 2025" target has lapsed, so there's no hard-deadline pressure. The #3902 inventory is the working plan and is still open to input/refinement (e.g. from @justlevine); net-new deprecations will generally be slated for after 3.0.
- **Remove existing deprecations** — inventoried in #3902 (Groups 1–4); first removal in progress (#3899).
- **Type safety & access** (final classes, visibility, hook consolidation, namespace relocation) — #3902 Group 5.
- **4.0 early-deprecation candidates** — #3902 Group 6.
[...]

PR #3368, opened 2025-04-26
chore: narrow/fix php types on WPGraphQL, Server, Utils namespaces

justlevine:
## What does this implement/fix? Explain your changes.

This PR audits and fixes the PHPDoc types in the `WPGraphQL`, `WPGraphQL\Server`, `WPGraphQL\Utils` namespaces, in the following ways:

- Replace use of `mixed` with a more specific type whenever possible.
   **Note:** Due to cross-class dependencies, numerous uses of `mixed` were left as is, specifically regarding the `Router` class.
- Used PHPStan conditional return types to narrow types when helpful.
- Fixed various small bugs and type errors in those and other classes as a result of the above changes.

There are _no breaking changing_ in this PR. I've left a comment on the (limited) code-related changes.

> [!IMPORTANT]
> ~This PR is based on #3367 which should be merged first.~
> 
> ~Relevant diff: https://github.com/wp-graphql/wp-graphql/pull/3368/commits/f663eba17aed38747b28b001ca6b038b5e9045ab~

## Any other comments?

This is a "first pass" because I really dont understand the `Router` class, and its reuse of the same parameters at different points in the lifecycle. Will take a second pass (in a follow up PR) ~once I get clarity from @jasonbahl~ ✔️

PR #3371, opened 2025-04-26
fix!: make `Tracing::init_trace()` void and call on `add_action()`

justlevine:

This PR fixes a code quality bug where the `do_graphql_request` was incorrectly being called via `add_filter()` instead of `add_action()`. 

As a result of this change, the `Tracing::init_trace()` has been changed from returning a `float` unused by the application to returning `void`.

This is _technically_ a breaking change, as that method is public, and the Tracing class is not marked `final`. However, there doesn't seem to be a way _in_ WPGraphQL to replace this version with an extended version (unlike some other filterable classes).

> [!IMPORTANT]
> ~This PR is based on #3368 which should be merged first.~
>
> ~Relevant diff for this PR is: https://github.com/wp-graphql/wp-graphql/commit/3f0de2398b75dcf63f7ff98ea0c677f3fa5d54a9~
> Rebased ✔️

Not directly, but if we consider this a breaking change than it's part of #3370 . Converting the entire class to `final` would be part of that ticket for sure.

### DIFF
--- a/src/Utils/Tracing.php
+++ b/src/Utils/Tracing.php
@@ -125,7 +125,7 @@ class Tracing {
 			return;
 		}
 
-		add_filter( 'do_graphql_request', [ $this, 'init_trace' ] );
+		add_action( 'do_graphql_request', [ $this, 'init_trace' ] );
 		add_action( 'graphql_execute', [ $this, 'end_trace' ], 99, 0 );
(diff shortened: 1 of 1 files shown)


justlevine:
(Rebased on #3368 . Relevant diff is now: https://github.com/wp-graphql/wp-graphql/commit/3f0de2398b75dcf63f7ff98ea0c677f3fa5d54a9 )

PR #3372, opened 2025-04-28
fix: update skipped since tags

jasonbahl:

This updates missed `@since next-version` tags that weren't properly updated in the previous release. 

It also updates the release workflow to ensure we properly update since tags during release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -154,6 +154,7 @@ jobs:
           # Check if this is a PR from develop to master
           if [[ "${{ github.event_name }}" == "pull_request_target" && "${{ github.event.pull_request.head.ref }}" == "develop" ]]; then
             # Create a temporary file for the PR body
+            # Use printf for safer handling of special characters and newlines
             printf "%s" '${{ github.event.pull_request.body }}' > /tmp/release-notes/pr_body.md
 
             # Process the PR body using sed instead of awk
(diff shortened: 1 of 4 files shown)


justlevine:
Did we also intentionally skip v2.2 in favor of v2.3?

jasonbahl:
@justlevine I think the skip must be something going on in my release tooling. It should not have skipped but something in my workflow triggered it to that I need to debug. 😬

jasonbahl:
@justlevine it looks like maybe I had a 2.2.0 release that didn't properly run and I wasn't paying enough attention to the details because the workflows report they succeeded, but really they didn't. (see: https://github.com/wp-graphql/wp-graphql/actions/runs/13957891456/job/39073292846#step:9:864)

🤔 

- https://github.com/wp-graphql/wp-graphql/actions/runs/14184340363
- https://github.com/wp-graphql/wp-graphql/actions/runs/13957891456

Got some fixing to do in my automations. 🤦🏻 😬

jasonbahl:
2.2.0 and 2.3.0 both are not properly deployed to WordPress.org, it seems 🤔 

CleanShot 2025-05-02 at 14 52 52

jasonbahl:
@justlevine looks like I had an error in the workflow that was passing the `v` prefix of our release versions `v2.2.0` and `v2.3.0` which were causing some funky issues. 

I have it strip the v before deploying to WordPress.org now. 

Both 2.2.0 and 2.3.0 are on WordPress.org now and I manually published the 2.2.0 release to Github as well. 

Hopefully this settles the kinks with the workflows. We might find some other issues as we go. 

Thanks for pointing this out. Gave me a good chance to step back and look at things I was overlooking.

Issue #2517, opened 2022-09-12
Standardize Action tracking / event emitting


jasonbahl:
### What problem does this address?

There are currently dozens of plugins ([WPGatsby](https://wordpress.org/plugins/wp-gatsby/), [WP Headless Trigger](https://github.com/nicoandrade/wp-headless-trigger), [Deploy Webhook Button](https://wordpress.org/plugins/webhook-netlify-deploy/), [Vercel Deploy Hooks](https://wordpress.org/plugins/vercel-deploy-hooks/), just to name a few),  in the ecosystem that listen for events in WordPress, then send out webhooks to their headless host (or elsewhere), to do _something_ in reaction to said event. 

A common problem that I've seen in these plugins, is that they don't always know what events to hook into, and what conditions to apply within the events they've hooked to, etc. 

For example, if a plugin were to hook into `wp_insert_post` and send a webhook to their build service each time that were executed, that would generate A LOT of unnecessary webhooks. 

`wp_insert_post` is called when auto-drafts are saved, or drafts, or other non-public data. If not guarded against, this will cause an influx of unnecessary webhooks to be sent to the external service and potentially create a bottleneck for when a webhook that _should_ have been sent needs to be processed. 

### What is your proposed solution?

I believe WPGraphQL should standardize some conventions around listening to relevant events in WordPress (publish_post, etc), standardize the "rules" or "conditions" that should be met, to determine whether the event is a "qualifying event" that should indeed dispatch a notification to an external service. 

I believe that centralizing things in core WPGraphQL will empower developers to be more successful using WPGraphQL and let them focus on project-specific problems. 

### What alternatives have you considered?

Let external plugins handle this.

I believe that WPGraphQL can greatly reduce (but not necessarily eliminate) the amount of custom code/custom plugins needed to connect WPGraphQL powered sites with remote services. 

### Additional Context

I built WPGatsby's action monitor, and while working on WPGraphQL Smart Cache and experimenting with WPGraphQL Subscriptions, I'm noticing a LOT of identical situations that I believe can/should be handled more centrally.

justlevine:
+1

As someone working to bring WP plugin support to graphql, I can see immediate benefit for a centralizing something like this. Currently WPGraphQL for GF only supports WPGatsby from the above list and the now-defunct [WP Jamstack Deployments](https://github.com/crgeary/wp-jamstack-deployments) - and the quirks of working with _both of em_ have dissuaded me from adding support for others, for the time being.

[...]

PR #3376, opened 2025-05-07
feat: introduce GraphQL event registry system and event registration hook

theodesp:

This PR introduces a new event registry system for WPGraphQL to centralize the registration and handling of WordPress hooks as GraphQL events. It includes:

- A `EventRegistry` class to manage event configurations and attach them to WordPress actions.
- A `register_graphql_event()` helper function for external plugins or themes to declaratively register events.
- An `EventMonitor` utility to track dispatched events via hooks like `wpgraphql_event_tracked_{event}`.

Draft PR for https://github.com/wp-graphql/wp-graphql/issues/2517

## Any other comments?

Looking forward to feedback on the pattern and implementation approach and other suggestions around testing and documentation

### DIFF
--- a/src/Registry/EventRegistry.php
+++ b/src/Registry/EventRegistry.php
@@ -1,0 +1,96 @@
+<?php
+
+namespace WPGraphQL\Registry;
+
+use WPGraphQL\Utils\EventMonitor;
+
+/**
+ * Class EventRegistry
+ *
+ * @package WPGraphQL\Registry
+ */
+class EventRegistry {
+
+    /**
+     * Array of registered events.
+     *
+     * @var array
+     */
+    protected static $registered_events = [];
+
+    /**
+     * EventRegistry constructor.
+     *
+     * @param array $default_events Optional. Array of event configurations to initialize with.
+     */
+    public function __construct( array $default_events = [] ) {
+        $this->registered_events = $default_events;
+    }
+
+    /**
+     * Initialize the Event Registry.
+     *
+     * @return void
+     */
+    public function init(): void {
+        do_action( 'graphql_register_events', $this );
+        $this->registered_events = apply_filters( 'graphql_registered_events', $this->registered_events, $this );
+        $this->attach_events();
+    }
+
+    /**
+     * Register a new event.
+     *
+     * Adds an event configuration to the static list of registered events.
+     *
+     * @param string   $name      Unique identifier for the event.
+     * @param string   $hook_name The WordPress action hook to attach the event to.
+     * @param callable $callback  Callback function to execute when the action fires.
+     * @param int      $priority  Optional. Priority for the callback. Default 10.
+     * @param int      $arg_count Optional. Number of arguments accepted by the callback. Default 1.
+     *
+     * @return void
+     */
+    public static function register_event( $name, $hook_name, $callback, $priority = 10, $arg_count = 1 ) {
+        self::$registered_events[] = [
+            'name'      => $name,
+            'hook_name' => $hook_name,
+            'callback'  => $callback,
+            'priority'  => $priority,
+            'arg_count' => $arg_count,
+        ];
+    }
+
+    /**
+     * Attach all registered events to their corresponding WordPress actions.
+     *
+     * @return void
+     */
+    public static function attach_events() {
+        foreach ( self::$registered_events as $config ) {
+            add_action( $config['hook_name'], function ( ...$hook_args ) use ( $config ) {
+                // Skip event escape hatch.
+                $maybe_skip = apply_filters( "graphql_event_should_handle_{$config['name']}", null, ...$hook_args );
+                if ( null !== $maybe_skip && false === $maybe_skip ) {
+                    return;
+                }
+
+                // Execute event callback.
+                $payload = call_user_func( $config['callback'], ...$hook_args );
+
+                // Handle errors.
+                if ( is_wp_error( $payload ) ) {
+                    do_action( "graphql_event_error_{$config['name']}", $payload, $hook_args );
+                    return;
+                }
+
+                // Filter the payload before tracking.
+                $payload = apply_filters( "graphql_event_payload_{$config['name']}", $payload, $hook_args );
+
+                // Track event in a custom event monitor or DB table.
+                EventMonitor::track( $config['name'], $payload );
+
+            }, $config['priority'], $config['arg_count'] );
+        }
+    }
+}
--- a/access-functions.php
+++ b/access-functions.php
@@ -7,6 +7,7 @@
  */
 
 use GraphQL\Type\Definition\Type;
+use WPGraphQL\Registry\EventRegistry;
 use WPGraphQL\Registry\TypeRegistry;
 use WPGraphQL\Request;
 use WPGraphQL\Router;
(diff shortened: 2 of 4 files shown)


src/Registry/EventRegistry.php
justlevine, line 1: Pattern-wise, I think it might be better if we co-locate everything in `WPGraphQL\EventMonitor` (or `\Events`), instead of splitting things across `/Registry` and `/Utils`. Especially considering we still need to add things like the db layer for recording

theodesp: Thanks @justlevine I will refactor

theodesp:
Hey @justlevine and @jasonbahl,  just wanted to follow up on this PR and check whether it’s being actively considered, especially regarding the proposed public functions like `register_graphql_event()`.

The reason I ask is that we’re currently experimenting with a [headless webhooks plugin](https://github.com/wpengine/hwptoolkit/tree/main/plugins/wp-graphql-headless-webhooks) built around this PR (POC stage for now) and would love to hear your thoughts or suggestions on how we might help move this PR forward.



Appreciate any feedback or direction you can share!

justlevine:
> Hey @justlevine and @jasonbahl, just wanted to follow up on this PR and check whether it’s being actively considered, especially regarding the proposed public functions like `register_graphql_event()`.
> 
> The reason I ask is that we’re currently experimenting with a [headless webhooks plugin](https://github.com/wpengine/hwptoolkit/tree/main/plugins/wp-graphql-headless-webhooks) built around this PR (POC stage for now) and would love to hear your thoughts or suggestions on how we might help move this PR forward.
> 
> Appreciate any feedback or direction you can share!

Hey @theodesp I actually wanted to ask you about this on last night's community call, since I wasn't sure what the scope of this PR is/should be.

As of now this PR adds an event store, but doesn't actually _use_ it for anything. Personally when there's no scope or RFP, I prefer we minimally dogfood our APIs in core, otherwise they risk turning into tech debt immediately. 

Once this scope is clarified, next step would be wpunit tests I assume.

theodesp:
Hey David I gueess I agree that without an immediate consumer in core, there’s a risk of adding speculative APIs.

We’re happy to port this functionality into our headless webhooks plugin for now, as it doesn’t have any deep dependencies on WPGraphQL itself beyond hooking into the WPGraphQL class actions. This way we can continue experimenting, get real-world usage behind it without the pressure of it needing to live in core immediately.

If and when a clearer scope or use case emerges for core, we’d be happy to revisit it or upstream improvements based on what we learn. In the meantime, we'll keep it cleanly contained in our plugin.

Thanks for the heads-up, and  really appreciate the context.

jasonbahl:
@theodesp I have shoulder surgery scheduled for June 5 and likely will not get to approving this or super deeply discussing further until after that. In short, yes, it is seriously being considered, but unfortunately I'm going to be a bottleneck for a bit longer and delay the momentum. Hope to get momentum picked up as I begin my recovery from surgery.

justlevine:
> Possibly one thing that would help would be to come up with some real-world use cases to solve (i.e. how could we accomplish the same thing that WPGraphQL Smart Cache does in a more centralized event-emitter API like this)

Here's prior art of WPGatsby and the even older Jamstack Deployments integrations for WPGraphQL for GF:

- https://github.com/AxeWP/wp-graphql-gravity-forms/blob/develop/src/Extensions/WPGatsby/GravityFormsMonitor.php

- https://github.com/AxeWP/wp-graphql-gravity-forms/blob/develop/src/Extensions/WPJamstackDeployments/WPJamstackDeployments.php

jasonbahl:
Ok, so after thinking a bit more about this, I think what we ultimately want to start working on here is the underlying api for [GraphQL Subscriptions](https://graphql.org/learn/subscriptions/).

GraphQL Subscriptions are a 3rd root type (alongside Query and Mutation), and the intent is to allow for real-time (or near-real-time) updates from a GraphQL Server. 

With this in mind, we should be able to start building the underlying event-tracking and emitting system, with the ultimate goal of exposing subscriptions in the schema when we feel like we have valid options for the subscriptions transport, but we can still enable the event tracking and emitting even before we expose Subscriptions to the schema, to solve for existing use cases such as cache invalidation (WPGraphQL Smart Cache).

I'll spend some more time working up a more formal RFC, but below is my initial thought on what an API like this could look like that would allow us to register "event listeners" in a central wpgraphql-esque way, with the goal of solving current situations like cache invalidation and the ultimate goal of bringing proper GraphQL Subscriptions to WPGraphQL.

## register_graphql_subscription

Here's what I'm thinking for a `register_graphql_subscription()` API:

```php
// hook in to plugins_loaded or later to make sure WPGraphQL is loaded 
// and `register_graphql_subscription` exists. 
// we don't hook into a graphql specific hook like `graphql_register_types` 
// because we will be tracking events outside of graphql and want 
// as wide of a scope as possible to track a wide variety of actions in 
// WordPress and we shouldn't require the GraphQL schema, etc to bootstrap.
add_action( 'plugins_loaded', function() { 

  // register a subscription. This not only will add a field to the root Subscription type, 
  // but will also track events in WordPress and emit messages that will trigger subscription operations. 
  register_graphql_subscription( 
     
     // $name: The name of the registered subscription
    'postUpdated',
    
     // $config: The config array for the subscription to be registered
    [
      'description' => __( 'Describe the subscription. This will map to the description in the schema', 'wp-graphql' ),
    
      // This is an array of events to "subscribe" to in WordPress. 
      // Each event in this array will be mounted as `add_action( $event_name, $callback, $args, $priority )`
      'subscribe' => [ 
        'wp_insert_post' => function( $dispatcher, $event_args ) { 
        
          // early return if the event doesn't qualify
          if ( 'xxx' !== $event_args['yyy'] ) {
             return;
          }
          
          // Build the payload. 
          // Potentially we use a class here instead of an arbitrary array to ensure some consistency in payloads
          // i.e. `$payload = new \WPGraphQL\SubscriptionPayload( ... );` or similar
          $payload = [
            'nodeId' => deriveNodeIdFromEventArgs( $event_args ),
            'nodeType' => deriveTypeFromEventArgs( $event_args ),
            'whateverElseIsRelevant' => deriveOtherRelevantInfoFromEventArgs( $event_args ), 
          ];
          
          // We dispatch an event that will get added to a queue, and ideally processed async
          // The queue should be de-duped then processed, by sending events out to whatever
          // transport mechanisms are configred.
          // The transport mechanisms would be configured separately.
          // i.e. websocket, webhook, server-sent-events
          // for webhook and SSE subscriptions, WordPress probably will
          // need to store the subscription data (custom post type + custom taxonomy)
          // so that the queue can be processed by WordPress.
          // for websocket subscriptions, the websocket server would need to 
          // persist the subscription data and handle connections/disconnections, etc.
          $dispatcher->dispatch( 'postUpdated', $payload );
        },
      ]
      
      // The fields that will be queryable on the subscription
      'fields' => [
      
        'post' => [
          'type' => 'Post',
          'description' => __( 'The post that was updated', 'wp-graphql' ),
        ],
        
      ],
      
      // The args that allow a user to refine/filter the subscription
      'args' => [
      
        // allow the subscription to be filtered by status
        'status' => [
          'type' => 'PostStatusEnum',
        ],
         
      ],
      
      // The resolver to execute when the subscription operation is executed.
      // The $payload is the root object that will be set by the payload that was emitted
      // The $args will be the args defined by the user in the document
      'resolve' => function( $payload, $args, $context, $info ) {
      
        // The $payload would be the context that was sent out with the dispatch
        $post = derivePostFromPayload( $payload );
        
        if ( ! $post ) {
          return null;
        }
        
        // Bail if the args on the subscriptions don't match the node we subscribed to
        if ( isset( $args['status'] ) && $args['status'] !== $post->post_status ) {
          return;
        }
      
        // If the event is relevant still, we resolve the subscription
        return new \WPGraphQL\Model\Post( $post_id );
      
      }
    ]
  );
}
```

PR #3382, opened 2025-05-20
chore: improve type safety [2nd pass]

justlevine:

### DIFF
--- a/src/Registry/TypeRegistry.php
+++ b/src/Registry/TypeRegistry.php
@@ -141,6 +141,8 @@ use WPGraphQL\Utils\Utils;
  * @phpstan-import-type WPEnumTypeConfig from \WPGraphQL\Type\WPEnumType
  * @phpstan-import-type WPScalarConfig from \WPGraphQL\Type\WPScalar
  *
+ * @phpstan-type TypeDef \GraphQL\Type\Definition\Type&\GraphQL\Type\Definition\NamedType
+ *
  * @package WPGraphQL\Registry
  */
 class TypeRegistry {
@@ -148,7 +150,7 @@ class TypeRegistry {
 	/**
 	 * The registered Types
 	 *
-	 * @var array<string,mixed>
+	 * @var array<string,?TypeDef>
 	 */
 	protected $types;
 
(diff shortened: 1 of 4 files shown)


src/Registry/TypeRegistry.php
```diff
@@ -933,13 +941,19 @@ protected function prepare_type( string $type_name, $config ) {
...
 	public function get_type( string $type_name ) {
 		$key = $this->format_key( $type_name );
 
 		if ( isset( $this->type_loaders[ $key ] ) ) {
-			$type                = $this->type_loaders[ $key ]();
+			$type = $this->type_loaders[ $key ]();
+			/**
+			 * Filter the type before it is loaded into the registry.
+			 *
+			 * @param ?TypeDef $type The type to load.
```
justlevine, line 954: This is what's letting us do all this type-narrowing without it being a real breaking change, since `InstrumentSchema::instrument_resolvers() would fatal error if anything other than a `Type` is passed here.
- https://github.com/wp-graphql/wp-graphql/blob/161565b87bd4e3d5bc675ee1f5fbd86801db6493/src/Utils/InstrumentSchema.php#L23

Which has been in the codebase since 2021: https://github.com/wp-graphql/wp-graphql/commit/6c7cc4228e0b6705f5aedf2fd5cf0919d08bcb81

jasonbahl: @justlevine just to clarify, "type-narrowing" in this case is referring to PHP/PHPStan types and _not_ GraphQL Types? The intent of this PR is all about PHP Types, so that's my assumption but wanted to make sure I'm not missing something in regards to GraphQL Types here.

jasonbahl: Ok, following your linked code in InstrumentSchema, I'm following what you're saying. We were already doing type checks during execution so the Type declaration is safe to add as anything that would break now would have already been broken. 👌

jasonbahl (approved):

src/Registry/TypeRegistry.php
justlevine: Haha yes _PHP_ type namrrowing.

PR #3420, opened 2025-09-25
fix: git updater compatibility

### LINKED ISSUE #3418: Version on WP Plugin Repository Contains a Git-Updator Header which breaks it if Git-Updator updates it.
### Description

GitHub Plugin URI: https://github.com/wp-graphql/wp-graphql

Git-Updater automatically picks up this comment entry, so when it runs it tries to update it which breaks the plugin as the dependencies are not installed.  the plugin warns you to install them via composer which might not be available in a shared hosting environment.

That line should be excluded from the version release to the WP Plugin Repository.

### Steps to reproduce

Install Git-Updater plugin and WP-GraphQL Plugin through WP Plugin Repository.  Next update it will break WP-GraphQL if a new version is available as the dependencies do not get installed with the update from GitHUb.

### Additional context

_No response_

### WPGraphQL Version

Any

### WordPress Version

Any

### PHP Version

Any

_No response_

- [x] Yes

- [ ] Yes

jasonbahl:

This updates the plugin to be compatible with Git Updater by setting `Release Asset: true` header and updating the release workflow to upload an asset named `$repo-$tag.zip` following the docs here: https://git-updater.com/knowledge-base/required-headers/

closes #3418

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -547,11 +547,23 @@ jobs:
         run: |
           composer run-script zip
 
+      - name: Create GitHub Updater compatible artifact
+        run: |
+          # Create a copy of the zip file with GitHub Updater naming convention ($repo-$tag.zip)
+          cp plugin-build/wp-graphql.zip plugin-build/wp-graphql-${{ steps.get_version_info.outputs.tag_name }}.zip
+          echo "Created GitHub Updater compatible artifact: wp-graphql-${{ steps.get_version_info.outputs.tag_name }}.zip"
+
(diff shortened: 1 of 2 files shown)


afragen:
I have to thank you @jasonbahl as I haven't really encountered anyone who has more than one release asset per release. I was previously picking up the first release asset, which in you case would be `schema.graphql`. I've created a fix in my `develop` branch that will pick the release asset whose name starts with the repo's slug. Therefore it is not necessary to include the version number.

I should merge this and release by early next week. I will also update my KB doc. Thanks.

PR #3422, opened 2025-10-01
fix: post meta array on preview

PatelUtkarsh:
When meta is stored as serialize array in post meta table, during preview instead of returning array it returns first value from array.

When `$single` is true, WordPress core's `get_post_metadata` filter returns the first value from an array. This fix ensures proper array wrapping to return array back when stored value is array.

Ref: https://github.com/WordPress/wordpress-develop/blob/2fe26ceb7a1f3fb57ec8726fc5f425d00a12ace9/src/wp-includes/meta.php#L666

### DIFF
--- a/tests/wpunit/PreviewTest.php
+++ b/tests/wpunit/PreviewTest.php
@@ -1191,4 +1191,113 @@ class PreviewTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase {
 		// value
 		$this->assertEmpty( get_post_meta( $this->preview, $published_meta_key, true ) );
 	}
+
+	/**
+	 * @see https://github.com/wp-graphql/wp-graphql/pull/3422
+	 */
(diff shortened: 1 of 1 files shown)


jasonbahl:
@PatelUtkarsh thanks for this PR!

Would you have any interest in contributing a test for this? An ideal test would be one that makes a preview query that includes meta and fails without this fix in place but passes with this fix in place.

jasonbahl (approved):

jasonbahl:
Thanks @PatelUtkarsh! 🙌🏻

PatelUtkarsh:
Hello @jasonbahl, I couldn't figure out how to run the specific file / method in the setup, I tried setting up SUITS env. Do you have any advice on that? 

```
SUITES="wpunit tests/wpunit/PreviewTest.php:testPreviewMetaWithArrayValueAndSingleTrue" composer run-test
```

PR #3448, opened 2025-12-15
fix: align cookie authentication with WordPress REST API pattern

### LINKED ISSUE #3447: Align Cookie Authentication with WordPress REST API Pattern

WPGraphQL's cookie authentication handling should be more aligned with the WordPress REST API's 
implementation to ensure consistent behavior across WordPress APIs.

1. **Add cookie authentication detection** - Implement equivalent of 
   `rest_cookie_collect_status()` to properly track when cookie-based auth is used

2. **Move auth check timing** - Run authentication validation before query execution 
   rather than after, matching the REST API's `rest_authentication_errors` filter pattern

3. **Introduce WPGraphQL-specific nonce** - Add `wp_graphql` nonce action with 
   `wp_rest` fallback for backward compatibility (i.e. https://github.com/wp-graphql/wpgraphql-ide/blob/47d7f672092ffbf3ff403339c80400e8f267d43a/wpgraphql-ide.php#L440)

4. **Add helper function** - Provide `graphql_get_nonce()` for extensions and 
   integrations to obtain valid nonces

_No response_

## Testing Plan

- Ensure existing test suites pass
- Write tests before implementation (TDD approach)
- Verify compatibility with common auth plugins
  - **WPGraphQL IDE**: Already sends nonces - no changes needed
  - **WPGraphQL JWT Authentication**: Uses non-cookie auth path - verify unaffected  
  - **Application Passwords**: Uses non-cookie auth path - verify unaffected
  - **WPGraphQL Headless Login**: Uses non-cookie auth path - verify unaffected

## Related

- WordPress core: `rest_cookie_check_errors()` 
  https://developer.wordpress.org/reference/functions/rest_cookie_check_errors/
- WordPress core: `rest_cookie_collect_status()`
  https://developer.wordpress.org/reference/functions/rest_cookie_collect_status/

jasonbahl:

This PR aligns WPGraphQL's cookie authentication handling with the WordPress REST API's implementation pattern.

### The Issue:

WPGraphQL's cookie authentication was not fully aligned with WordPress REST API conventions. Specifically:
- Authentication checks were mistakenly running _after_ query execution, rather than _before_
- Nonce validation for cookie-authenticated requests wasn't _fully_ following the REST API pattern
- There was no dedicated WPGraphQL nonce action

### The Fix:

- Authentication checks now run in `Router::validate_http_request_authentication()` BEFORE `graphql_process_http_request` fires, ensuring plugins cannot access unvalidated user context
- Cookie-authenticated requests properly validate nonces via `X-WP-Nonce` header or `_wpnonce` parameter
- Added `wp_graphql` nonce action (with `wp_rest` fallback, for backward compatibility)
- Added `graphql_get_nonce()` helper function for extensions and integrations
- Added `graphql_cookie_auth_require_nonce` filter for development environments

### ⚠️ Behavior Changes:

| Scenario | Behavior |
|----------|----------|
| No nonce provided | Request downgrades to guest (i.e. `{ viewer: null }`) |
| "Falsy" nonce (empty, `null`, `undefined`) | Treated as no nonce, downgrades to guest |
| Invalid nonce | Returns 403 error: "Cookie nonce is invalid" (status filterable via `graphql_authentication_error_status_code`) |
| Valid nonce | Full authenticated access |
| Authorization header (JWT, App Passwords) | Nonce not required |

## Why This Is a Patch, Not a Major Version

This change is classified as a **bug fix** rather than a breaking change for the following reasons:

1. **Existing Code Intent**: WPGraphQL already had code in place to verify nonces for cookie-authenticated requests (via `has_authentication_errors()`). However, the implementation was incomplete. Nonce checks were mistakenly running _after_ query execution instead of _before_. This PR corrects the timing to match the original intent.

2. **Defense in Depth**: WPGraphQL's CORS headers (`Access-Control-Allow-Origin`) were already configured to block cross-origin credentialed requests. This nonce validation is a defense-in-depth measure that provides _additional_ protection, particularly for same-origin scenarios.

3. **REST API Alignment**: The WordPress REST API has always required nonces for cookie authentication. WPGraphQL was designed (initial code added in 2016) to follow this pattern but the implementation had gaps. This PR closes those gaps.

4. **Minimal Real-World Impact**: _Most_ implementations are unaffected:
   - Decoupled/headless apps use JWT, Application Passwords, or OAuth (not affected)
   - The built-in GraphiQL IDE already sends nonces (not affected)
   - Any custom WordPress admin integrations that fetch the WPGraphQL endpoint already have access to nonces and can pass them via `wp_localize_script()` or `wp_inline_script()`. If not implemented you can add them following the [built-in GraphiQL IDE pattern](https://github.com/wp-graphql/wp-graphql/blob/d854b76dfe3a901fdfb1b67bfd6b1b4b8a2b3b34/packages/wpgraphiql/utils/fetcher.js#L15)

If any implementations were relying on the incomplete behavior, and cannot upgrade immediately, or want to bypass the nonce checks for whatever reason to maintain previous behavior (i.e. during development in local/staging environments), they can use the new `graphql_cookie_auth_require_nonce` filter to restore the previous behavior while they update their client code: 

```php
<?php
add_filter('graphql_cookie_auth_require_nonce', function($require) {
    return wp_get_environment_type() !== 'local';
});
```

5. **Developer Convenience Note**: If you're logged into wp-admin in one browser tab and execute a GraphQL GET request in another tab (e.g., typing `yoursite.com/graphql?query={viewer{name}}` in the URL bar), the request will now be treated as unauthenticated unless you also include a `_wpnonce` query parameter.

Closes #3447

### Test Results

- [x] **Failing test added**: Tests demonstrate expected behavior for various authentication scenarios
- [x] **Fix implemented**: Authentication checks moved to `before_execute()` with proper nonce validation
- [x] **All tests passing**: All existing tests continue to pass
- [x] **Regression tests**: Comprehensive acceptance, functional, and WPUnit tests added

### Tests Added

**Acceptance Tests (13 tests):**
- Unauthenticated request returns only public data
- Valid nonce header authenticates user
- `wp_rest` nonce backward compatibility
- Logged-in user without nonce treated as guest
- Logged-in user without nonce cannot see draft posts
- Logged-in user with nonce CAN see draft posts
- Invalid nonce returns error
- Application Password auth works without nonce
- Application Password can access private data
- Application Password can execute mutations
- Invalid Application Password is rejected
- Application Password works with GET requests
- Auth error returns 403 by default (filterable)

**Functional & WPUnit Tests:**
- Cookie auth flow validation
- Nonce verification scenarios
- Authentication state management

### Before:
```graphql
# Cookie-authenticated request without nonce (most commonly embedded js apps in wp-admin) 
# Would execute with full user context
query {
  viewer {
    databaseId
    email
  }
}
# Result: Full authenticated data returned
```

### After:
```graphql
# Cookie-authenticated request without nonce
# Now properly downgrades to guest
query {
  viewer {
    databaseId
    email
  }
}
# Result: viewer is null (guest access)
```

### Authenticated Request (with nonce):
```javascript
fetch('/graphql', {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
    'X-WP-Nonce': wpApiSettings.nonce // or graphql_get_nonce()
  },
  body: JSON.stringify({ query: '{ viewer { name } }' })
});
// Result: Full authenticated access
```

### Compatibility Notes

- **WPGraphQL IDE (built-in)**: ✅ Unaffected - already sends nonces
- **JWT Authentication**: ✅ Unaffected - uses Authorization header, bypasses nonce check
- **Application Passwords**: ✅ Unaffected - uses Authorization header, bypasses nonce check
- **WPGraphQL Headless Login**: ✅ Unaffected - uses non-cookie auth path
- **WPGraphQL IDE (extension)**: ⚠️ Needs to update to [send the nonce with the fetcher](https://github.com/wp-graphql/wpgraphql-ide/blob/47d7f672092ffbf3ff403339c80400e8f267d43a/src/components/App.jsx#L74-L90) when state is authenticated 

⚠️ Some extension plugins _may_ need updates to tests using `amOnPage()` for authenticated requests to include a nonce.

### New Features

1. **`graphql_get_nonce()`** - Helper function to generate nonces:
   
   ```php
   wp_localize_script('my-app', 'mySettings', [
       'nonce' => graphql_get_nonce()
   ]);
   ```

2. **`graphql_cookie_auth_require_nonce` filter** - For development environments or migration:
   
   ```php
   // Only use in local development!
   add_filter('graphql_cookie_auth_require_nonce', function($require) {
       return wp_get_environment_type() !== 'local';
   });
   ```

3. **`graphql_authentication_error_status_code` filter** - Customize the HTTP status code for auth errors:
   
   ```php
   // For legacy clients expecting 200 with GraphQL error response
   add_filter('graphql_authentication_error_status_code', function($status, $error) {
       return 200;
   }, 10, 2);
   ```

### Documentation Updates

- Updated `docs/authentication-and-authorization.md` with nonce requirements
- Updated `docs/security.md` with CSRF protection details
- Updated `CHANGELOG.md`
- Added upgrade notice to `readme.txt`

---

> [!NOTE]
> Aligns cookie-based auth with REST API by validating nonces before execution, adds a nonce helper and filters, updates docs, and introduces comprehensive tests and dev tooling to support the flow.
> 
> - **Auth/Core**
>   - Move CSRF/nonce handling to `Router::validate_http_request_authentication()` (runs before hooks); downgrade to guest when no nonce; return 403 for invalid nonce; accept `X-WP-Nonce`/`_wpnonce` and both `wp_graphql`/`wp_rest` actions.
>   - Update `Request` to capture auth errors before execution and refresh `AppContext->viewer` accordingly; remove post-exec auth checks.
>   - Add `graphql_get_nonce()` helper; new filters: `graphql_cookie_auth_require_nonce`, `graphql_authentication_error_status_code`.
> - **Docs**
>   - Update `docs/authentication-and-authorization.md` and `docs/security.md` with nonce/CSRF guidance; add upgrade notice in `readme.txt`.
> - **Tests**
>   - Add acceptance (`tests/acceptance/CookieAuthenticationCest.php`), functional, and wpunit suites covering nonce flows, guest downgrade, 403 errors, and Application Passwords.
> - **Dev/Infra**
>   - Enable WebDriver in `codeception.dist.yml`; add Selenium Chrome service in `docker-compose.yml`.
>   - Ensure `Authorization` header passthrough in Apache and `.htaccess`; set `WP_ENVIRONMENT_TYPE=local` in setup.
>   - Minor type/doc fixes in `DataSource`; bump dev Composer deps.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a89deafc2a44f9f8b1eaa0afcbf616ee9fe629fa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tests/functional/CookieAuthenticationCept.php
```diff
@@ -0,0 +1,215 @@
...
+
+$I->seeResponseCodeIs( 200 );
+$I->seeResponseIsJson();
+
+$response       = $I->grabResponse();
+$response_array = json_decode( $response, true );
+
+codecept_debug( $response_array );
+
+$I->assertArrayNotHasKey( 'errors', $response_array, 'Missing nonce should not produce error, just downgrade' );
```
kidunot89, line 158: You should try using the [GraphQL E2E assertions](https://github.com/wp-graphql/wp-graphql-testcase/blob/develop/tests/codeception/functional/QueryAssertsModuleCest.php#L42-L51)

jasonbahl: Nice! I will follow up with updating acceptance/functional tests to use those assertions in another PR.

src/Request.php
```diff
@@ -352,18 +396,90 @@ protected function has_authentication_errors() {
...
+		$require_nonce = apply_filters( 'graphql_cookie_auth_require_nonce', true, $this );
+
+		/**
+		 * No nonce provided - downgrade to unauthenticated (unless nonce requirement is disabled).
+		 *
+		 * This is not an error, we just treat the request as if it came from
+		 * an unauthenticated user (CSRF protection).
+		 */
+		if ( null === $nonce && $require_nonce ) {
 			wp_set_current_user( 0 );
```
chriszarate: It still seems too late to call `wp_set_current_user`. Can this request-level authentication be done immediately _before_ `graphql_init`?

If _any_ WPGraphQL hooks run before this check, it feels like additional opportunity for mismatched expectations and potential leaks.

For example, `graphql_process_http_request` runs before this check, which plugins might hook into to enable sensitive debugging output.

jasonbahl: @chriszarate `graphql_init` is fired within the `after_setup_theme` callback, so adding something like this there would be overhead on _every_ WordPress request, even ones that are not making a GraphQL request. 

I believe we could move it to the Router and run it early there, as this is only protecting HTTP requests to the GraphQL endpoint. Having the code in Request meant that it was central to _all_ kinds of requests, including direct calls to the `graphql()` method which instantiates a `new \WPGraphQL\Request()` and fires it off. However, we're not protecting against direct calls to the graphql() method, just HTTP requests to the endpoint, so I think we can safely move this code early in the Router. 

I'll work on that refactor.

PR #3454, opened 2026-01-04
chore: update node, npm, and composer deps

justlevine:

This PR updates the minimum `node` version used by nvmrc and CI to v22, and Composer + NPM deps to their latest SemVer-compatible versions*.

Additionally, the `npm run lint:php:fix` command was fixed to correctly call `composer fix-cs`

This PR **does not** attempt to update any deps that require breaking changes, or FE dependencies related to Graphiql IDE and other parallel efforts, nor does it fix the preexisting (failing) E2E test.

No

### Test Results

- [ ] **Failing test added**: Commit that demonstrates the bug with a failing test
- [ ] **Fix implemented**: Commit that resolves the bug and makes tests pass
- [x] **All tests passing**: All existing tests continue to pass
- [ ] **Regression tests**: Added tests to prevent the bug from returning

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,8 +61,8 @@
 			"phpstan/extension-installer": true
 		}
 	},
-  "minimum-stability": "dev",
-  "prefer-stable": true,
+	"minimum-stability": "dev",
+	"prefer-stable": true,
 	"autoload": {
 		"files": [],
 		"psr-4": {


.github/workflows/graphiql-e2e-tests.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: '20'
+          node-version: '22'
```
justlevine, line 40: Future PR to consolidate/cleanup our GH workflows will update `setup-node` to use nvmrc (and better caching)

composer.json
```diff
@@ -39,14 +39,13 @@
 		"codeception/module-webdriver": "^1.0",
 		"codeception/util-universalframework": "^1.0",
 		"lucatume/wp-browser": "<3.5",
-		"phpcompatibility/php-compatibility": "^9.3",
-		"phpcompatibility/phpcompatibility-wp": "^2.1",
+		"phpcompatibility/phpcompatibility-wp": "^3.0.0-alpha",
```
justlevine, line 42: This contains `phpcompatibility/php-compatibility^3.0.0-alpha` so we no longer need to manually require + update both.

composer.json
```diff
@@ -39,14 +39,13 @@
...
 		"phpstan/extension-installer": "^1.1",
 		"phpstan/phpstan-deprecation-rules": "^2.0.1",
 		"phpstan/phpstan": "~2.1.2",
 		"phpunit/phpunit": "^9.5",
 		"slevomat/coding-standard": "^8.9",
-		"szepeviktor/phpstan-wordpress": "~2.0.1",
+		"szepeviktor/phpstan-wordpress": "^2.0.1",
```
justlevine, line 48: (It's locked by `phpstan/phpstan` )

tests/wpunit/AccessFunctionsTest.php
```diff
@@ -25,12 +25,6 @@ public function tearDown(): void {
 		parent::tearDown();
 	}
 
-	public function testGraphQLPhpVersion() {
```
justlevine, line 28: The changelog is no longer bundled, but this test is pointless so I just removed it entirely. We'll get a fatal error using the library if it doesnt exist.

composer.json
```diff
@@ -62,6 +61,8 @@
 			"phpstan/extension-installer": true
 		}
 	},
+  "minimum-stability": "dev",
```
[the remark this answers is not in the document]
justlevine: You idiotic nonce. You have a lockfile to prevent yourself from making such dumb comments. Do better or ELON MUSK will kill my family.

Issue #3477, opened 2026-01-16
RFC: Centralize plugin configuration using release-please manifest for CI orchestration


jasonbahl:
## Problem

As we bring more plugins into the monorepo, each plugin needs:
- phpcs (coding standards)
- phpstan (static analysis)
- Unit tests (codeception)
- E2E tests (playwright)
- Build steps

Currently, adding a new plugin requires updating **5+ configuration files**:

| File | Purpose |
|------|---------|
| `package.json` → `workspaces` | npm workspace commands |
| `.release-please-manifest.json` | Version tracking |
| `release-please-config.json` | Release settings |
| `.wp-env.json` | WordPress environment |
| GitHub workflows (8 files) | CI pipelines |

This is error-prone and violates DRY principles.

## Proposed Solution

Use `.release-please-manifest.json` as the **single source of truth** for plugin discovery:

```json
{
  "plugins/wp-graphql": "2.6.0",
  "plugins/wp-graphql-acf": "1.0.0"
}
```

### Why this file?
- ✅ Already exists and **must** be maintained (release-please requires it)
- ✅ Keys are the exact plugin paths needed for CI
- ✅ Simple to parse: `jq -r 'keys[]' .release-please-manifest.json`
- ✅ No additional config file to maintain

### For richer metadata
`release-please-config.json` can provide additional details when needed:
```json
"packages": {
  "plugins/wp-graphql": {
    "component": "wp-graphql",
    "package-name": "wp-graphql"
  }
}
```

### 1. CI Workflows - Change Detection Matrix

Update workflows to dynamically detect which plugins changed and only run jobs for those:

```yaml
jobs:
  detect-changes:
    runs-on: ubuntu-latest
    outputs:
      plugins: ${{ steps.filter.outputs.changes }}
    steps:
      - uses: actions/checkout@v4
      
      - name: Build paths filter from manifest
        id: build-filter
        run: |
          # Generate filter config from manifest
          jq -r 'keys[] | "\(.):
            - \(.)/**"' .release-please-manifest.json > filters.yml
          
      - uses: dorny/paths-filter@v3
        id: filter
        with:
          filters: filters.yml
          
  phpcs:
    needs: detect-changes
    if: needs.detect-changes.outputs.plugins != '[]'
    strategy:
      matrix:
        plugin: ${{ fromJson(needs.detect-changes.outputs.plugins) }}
    steps:
      - name: Run PHPCS
        working-directory: ${{ matrix.plugin }}
[...]

PR #3478, opened 2026-01-16
fix: simplify PHP lint scripts to not require wp-env

jasonbahl:
## Description

This PR simplifies the PHP linting scripts in the wp-graphql plugin to run composer directly instead of requiring wp-env to be running.

### What Changed

1. **Root `package.json`**: Added `lint:php`, `lint:php:fix`, `lint:php:stan` scripts that use turbo
2. **Plugin `package.json`**: Simplified lint scripts to use composer directly (no wp-env dependency)
3. **`turbo.json`**: Added `lint:php:fix` and `lint:php:stan` tasks

### Before vs After

**Before (required wp-env running):**
```json
"lint:php": "npm run wp-env:cli -- composer run check-cs -- -d memory_limit=256M"
```

**After (runs directly):**
```json
"lint:php": "composer run check-cs"
```

### Usage

```bash
# From root - runs across all workspaces via turbo
npm run lint:php

# From plugin directory
cd plugins/wp-graphql && npm run lint:php
```

### Benefits

- **Simpler local dev** - Run linting without starting wp-env first
- **Turbo caching** - Repeated runs are instant (cache hit)
- **Consistent** - Same approach for all lint:php variants (fix, stan)

---

Related: #3477 (RFC for centralizing plugin configuration)


jasonbahl:
I'll close this and only revisit if I run into similar problems.

PR #3493, opened 2026-01-21
refactor: deprecate `WPGraphQL_CLI_Command` in favor of `WPGraphQL\CLI\Commands`

justlevine:

This PR deprecates the `WPGraphQL_CLI_Command` class (located in `cli/wp-cli.php`) for a `\WPGraphQL\CLI\Commands` class _inside_ the namespace.

This allows the class to be properly linted, stanned, and checked for test coverage. (and it's where LLMs assume it will be)

Then the new file was linted and stanned in 6c653542dc6b1e5459ad33e48cce835c771acd55 (so you could have a clear diff of the changes).

## Any other comments?

Did not feel the need to add tests to this, since it's not like there were tests before. We can backfill them in a future PR, but at least we're now aware it's there.

### DIFF
--- a/.github/workflows/integration-tests-reusable.yml
+++ b/.github/workflows/integration-tests-reusable.yml
@@ -117,7 +117,7 @@ jobs:
           timeout_minutes: 10
           max_attempts: 3
           command: |
-            npm run wp-env start -- ${{ inputs.coverage && '--xdebug=coverage' || '' }} ${{ secrets.ACTIONS_STEP_DEBUG && '--debug' || '' }}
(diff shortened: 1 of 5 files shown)


justlevine:
@jasonbahl that test matrix is the gift that keeps on giving 😂😭. Even the smoke tests and schema linter are passing, so I'm assuming it's not the actual changes here...

---

Leaving the copilot feedback for you to review. My goal wasnt to improve the class, just to relocate it, but if the feedback is already there and you agree with some of it 🤷

PR #3494, opened 2026-01-21
ci: refactor integration tests into reusable workflow

jasonbahl:
## Summary

Refactors the integration tests workflow into a reusable workflow pattern, making it easier to add tests for additional plugins as they're migrated to the monorepo.

Inspired by @justlevine's work on the reusable linter workflow and his constructive feedback. 🙏 (and his sass 😆)

## Changes

- **Renamed** `testing-integration.yml` → `integration-tests.yml`
- **Created** `integration-tests-reusable.yml` - reusable workflow with all test execution logic
- **Simplified matrix** - removed the complicated bash-generated conditional matrix in favor of a straightforward `include` matrix that's easier to read and maintain
- **Updated** `docs/ARCHITECTURE.md` - documented how to add tests for new plugins

## How it works

1. `integration-tests.yml` defines plugin-specific jobs with test matrices
2. Each job calls the reusable workflow with plugin path/name
3. The reusable workflow's `pre_run` job checks if that specific plugin has changes
4. Tests only run for plugins that actually changed

## Adding a new plugin

Copy the `wp-graphql` job block and update the plugin name/path. The reusable workflow handles the rest.

## Test Matrix

Uses "boundary testing" approach with a simple `include` matrix (no more bash generation):
- 4 coverage jobs: Latest WP/PHP with theme × multisite combinations
- 3 boundary jobs: Current stable, mid-range, oldest supported
- 1 experimental: WordPress trunk (allowed to fail)


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,183 @@
...
+            multisite: false
+            coverage: false
+          # Experimental: WordPress trunk (allowed to fail)
+          - name: "WP trunk / PHP 8.4"
+            php: "8.4"
+            wp: trunk
+            theme: twentytwentyfive
+            multisite: false
+            coverage: false
+            experimental: true
```
justlevine, line 171: Just giving you fair warning that `theme` and `experimental` are both on my sass list for a some-unspecified-point-in-the-future PR.

jasonbahl: @justlevine the experimental I can understand. Want to flag some way to indicate that we're testing against wp nightly and allow it to fail. so it's signifying 2 things, not just a wp version, but open to changing the name or whatever. 

For the theme, I think it's important to test against a block theme and a non-block theme, especially as we add more "modern" wp support and will have different areas of code that will need different tests for each case

justlevine: I agree we need to test it, but that it (theme switching or activating the theme feature that impacts results) should happen as a test implementation detail and not as a matrix branch flags. Most themes aren't classic vs fse but "hybrid" and classic also != legacy

justlevine: > Want to flag some way to indicate that we're testing against wp nightly and allow it to fail.

not finding it via my phone, but this specifically I think we can do by just adding `nightly` to the matrix and then a continue-if . Either way beyond the scope of this PR

PR #3498, opened 2026-01-22
perf(ci): faster wp-env config

justlevine:

This PR speeds up CI and our wp-env setup in the following ways

1. ~Runs commands _inside_ the containers in parallel (via `bin/setup.sh`)~
2. Used PCOV instead of xdebug for coverage.
3. Removes unnecessary directly CI calls to the script.

Additionally it:

4. Isolates the commands into reusable functions for human clarity, and better LLM handling for future contribs (and evaluation for removing if not actually necessary)
5. ~Renames `setup-wp-env.sh` to `after-start.sh` to make the differentiation between it and the internally run `setup.sh` clearer.~ 

## Any other comments?

- Does _not_ move the PHP spaghetti into a `.php` file since the root wp-env doesn't make it obvious where it would live cross-repo. If we are keeping our wp-envs top-level for the future, then this will go in the resulting `tools/wp-env` or whatever (by having the other repos migrated in the location will become more obvious).
- For the same reason I reverted the attempt to isolate a setup.sh that could be run from inside the container. we need a good place for our `wp-env.json.mappings`

Prior art:
- https://github.com/AxeWP/wp-graphql-gravity-forms/pull/467

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -56,7 +56,7 @@ jobs:
 
       - name: Detect changed files
         id: filter
-        uses: tj-actions/changed-files@v44
+        uses: tj-actions/changed-files@v46
         with:
           files_yaml: |
             wp-graphql:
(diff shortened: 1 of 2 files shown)


justlevine:
@jasonbahl ready for your review. Cutting down a good 5 minutes even after reverting some of the changes I want to make because I didn't have time to deal with the monorepo

PR #3499, opened 2026-01-22
fix(ci): add WP version to child job name for better sidebar visibility

jasonbahl:
## Description

Follow-up to #3497. Adds WordPress version to the beginning of child job names so it's visible in GitHub's Actions sidebar.

## Problem

GitHub's sidebar for reusable workflows only shows the child job name, not the full parent job context. This meant the sidebar was showing just theme names like `twentytwentyfive 📊` without any version information.

## Solution

Add WP version to the start of the child job name:
- Before: `twentytwentyfive 📊`
- After: `WP 6.9 / twentytwentyfive 📊`

## Trade-off

The WP version will appear twice in the full job header (`Integration: wp-graphql / WP 6.9 / PHP 8.4 / WP 6.9 / twentytwentyfive 📊`), but this ensures version info is visible in the sidebar where it matters most for quick scanning.

### DIFF
--- a/.github/workflows/integration-tests-reusable.yml
+++ b/.github/workflows/integration-tests-reusable.yml
@@ -44,9 +44,9 @@ on:
 jobs:
   # Job naming strategy:
   # The parent workflow sets name to "Integration: plugin / WP X.X / PHP X.X".
-  # The child job name shows theme + modifiers (no PHP to avoid duplication).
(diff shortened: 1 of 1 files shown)


jasonbahl:
Meh. Still some weird GitHub UI issues.

CleanShot 2026-01-22 at 09 53 09
CleanShot 2026-01-22 at 09 53 16

 Going to just leave it as it was before this PR and close this PR without merging 🤷‍♂️

PR #3506, opened 2026-01-24
fix: reorder arguments in graphql_map_input_fields_to_wp_query filter

abhishekmanegit:
### Summary
This PR reorders the arguments passed to the `graphql_map_input_fields_to_wp_query` filter to place `$where_args` (user input) before `$query_args`, improving consistency and developer expectations across WPGraphQL filters.

### Before:
```php
apply_filters( 'graphql_map_input_fields_to_wp_query', $query_args, $where_args, ... );

### After: 
```php
apply_filters( 'graphql_map_input_fields_to_wp_query', $where_args, $query_args, ... );


### Why?

In most WPGraphQL filters, user input ($where_args) is the first argument. This change aligns this filter with that convention and simplifies predictability for developers extending WPGraphQL behavior.


jasonbahl:
Thanks again for working through this, @abhishekmanegit. After looking at it more carefully, I don't think we should land this change.

The PR's premise was that `$where_args` should come first for consistency. But the WordPress filter convention is that the first argument is the value being filtered and returned — and this filter returns `$query_args`. All three sibling filters (`graphql_map_input_fields_to_wp_user_query`, `graphql_map_input_fields_to_get_terms`, `graphql_map_input_fields_to_wp_comment_query`) follow that convention with `$query_args` first.

Once we keep the original signature (which we must for backward compatibility anyway), what's left is a rename from `…_wp_query` to `…_wp_query_args`. That isn't a win either: the `_args` suffix doesn't match any of the sibling filters, and the rename creates migration work for every existing handler with no behavioral benefit.

Going to close this one. Really do appreciate the effort, and please don't be discouraged from opening more PRs — there are plenty of areas where this kind of consistency thinking pays off.

PR #3518, opened 2026-01-29
fix(deps): configure dependabot commit prefixes to trigger releases for plugin dependencies

jasonbahl:
## What's Changing

This PR updates the Dependabot configuration to use different commit message prefixes based on where dependencies are located and whether they're production or development dependencies. The goal is to ensure that plugin production dependency updates trigger releases via release-please, while dev dependencies and root/tooling dependency updates don't.

## The Problem

I noticed in #3516 that Dependabot was creating PRs with `chore(deps-dev):` prefixes for plugin dependency updates. Since release-please only triggers releases for `feat:`, `fix:`, and `perf:` commit types, these dependency updates wouldn't generate releases even though they affect the distributable plugins.

However, after thinking about it more, dev dependencies (like phpunit in #3516) don't actually ship with the plugin, so they probably shouldn't trigger releases anyway. Only production dependencies that end users actually get should trigger releases.

## The Solution

The Dependabot configuration now uses different commit prefixes based on the directory and dependency type:

- **Plugin Production Composer dependencies** (`/plugins/*`): Use `fix(deps):` → triggers **PATCH** releases
- **Plugin Dev Composer dependencies** (`/plugins/*`): Use `chore(deps-dev):` → no releases (testing/build tooling only)
- **Plugin Production NPM dependencies** (`/plugins/*`): Use `feat(deps):` → triggers **MINOR** releases  
- **Plugin Dev NPM dependencies** (`/plugins/*`): Use `chore(deps-dev):` → no releases (testing/build tooling only)
- **Root NPM dependencies** (`/`): Use `chore(deps):` or `chore(deps-dev):` → no releases (tooling only)
- **GitHub Actions** (`/`): Use `chore(deps):` → no releases (CI only)

## How It Works

The nice thing about this approach is that release-please uses path-based detection to determine which components should get releases. So even if we used `fix(deps):` for all dependency updates, release-please would only create a release for a plugin if files within that plugin's directory actually changed. This means:

- A `fix(deps):` commit that only touches `/plugins/wp-graphql-smart-cache/composer.json` will trigger a release for that specific plugin
- A `fix(deps):` commit that only touches root `/package.json` won't trigger any plugin releases

But using `chore:` for dev dependencies and root/tooling dependencies is still the right approach because it's more semantically correct and makes the intent clear in the git history.

## Impact

Going forward:
- ✅ Plugin production dependency updates will trigger releases when merged (as they should have been all along)
- ✅ Plugin dev dependency updates won't trigger releases (they don't ship with the plugin)
- ✅ Root/tooling dependency updates won't trigger releases (as intended)
- ✅ Each plugin gets its own release when its production dependencies are updated
- ✅ The commit history will be clearer about what type of change each dependency update represents

This ensures that users get updates when we bump plugin production dependencies, which is especially important for security patches and bug fixes in dependencies that actually ship with the plugin.

### DIFF
--- a/plugins/wp-graphql/composer.lock
+++ b/plugins/wp-graphql/composer.lock
@@ -7092,16 +7092,16 @@
         },
         {
             "name": "symfony/process",
-            "version": "v5.4.47",
+            "version": "v5.4.51",
             "source": {
                 "type": "git",
                 "url": "https://github.com/symfony/process.git",
-                "reference": "5d1662fb32ebc94f17ddb8d635454a776066733d"
+                "reference": "467bfc56f18f5ef6d5ccb09324d7e988c1c0a98f"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/symfony/process/zipball/5d1662fb32ebc94f17ddb8d635454a776066733d",
-                "reference": "5d1662fb32ebc94f17ddb8d635454a776066733d",
+                "url": "https://api.github.com/repos/symfony/process/zipball/467bfc56f18f5ef6d5ccb09324d7e988c1c0a98f",
+                "reference": "467bfc56f18f5ef6d5ccb09324d7e988c1c0a98f",
                 "shasum": ""
             },
             "require": {
@@ -7134,7 +7134,7 @@
             "description": "Executes commands in sub-processes",
             "homepage": "https://symfony.com",
             "support": {
-                "source": "https://github.com/symfony/process/tree/v5.4.47"
+                "source": "https://github.com/symfony/process/tree/v5.4.51"
             },
             "funding": [
                 {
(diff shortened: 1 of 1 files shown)


justlevine:
> However, after thinking about it more, dev dependencies (like phpunit in https://github.com/wp-graphql/wp-graphql/pull/3516) don't actually ship with the plugin, so they probably shouldn't trigger releases anyway. Only production dependencies that end users actually get should trigger releases.

Not 100% accurate. "Dev dependencies" are also our _build_ dependencies, so they can very much change behavior and should imo trigger a release.

.github/dependabot.yml
```diff
@@ -26,29 +32,60 @@ updates:
...
+    commit-message:
+      prefix: 'feat'
+      prefix-development: 'chore'
+      include: 'scope'
+    groups:
+      npm-plugin-minor-patch:
+        update-types: [minor, patch]
+      npm-plugin-dev-major:
+        dependency-type: 'development'
+        update-types: [major]
```
justlevine: Per https://github.com/wp-graphql/wp-graphql/pull/3509#issuecomment-3819382290 I think we should get rid of this group entirely and handle majors individually to increase the chances of them passing and us being able to save some manual dev time.

PR #3539, opened 2026-02-03
fix: Allow compatible interface field override with `register_graphql_field()`

### LINKED ISSUE #3096: DX: Can't narrow down interface subtypes on an implementing object with `register_graphql_field()`
### Description

When using `register_graphql_field()`, a `DUPLICATE_FIELD` error is thrown even if the field being registered is compatible with the existing one.

For example, let's say a `NodeWithSeo` interface registers the `seo: SeoInterface` field. I should be able to overwrite `PostObject.seo` to be type `PostObjectSeo implements SeoInterface`. However, because at this point in the lifecycle, the interface fields are already merged with the fields in `WPObjectType`, the blind key-check sees that `seo` already exists and bails.

Ideally, if the field already exists, the new object type would be compared to see if it is compatible with the interface instead of throwing an error. (However I'm unaware of any internal methods for getting a list of implementing interfaces for a GraphQL object name).

### Steps to reproduce

1. Register your custom interface, e.g. `SeoInterface`
2. Register your custom object that implements that interface. e.g. `PostObjectSeo`.
3. Create a new interface to add a `seo: SeoInterface` to all `ContentNodes`.
4. Register a custom field on `PostObject` to override the type of your custom field: `seo: PostObjectSeo`.
5. Enable GraphQL Debugging and run the query.

image

### Additional context

- The workaround is manually using the 'graphql_{$type_name}_fields' filter, but that's bad DX.
- This affects the schema usability for both WPGraphQL for Gravity Forms and WPGraphQL for RankMath Seo.

### WPGraphQL Version

1.22.1

### WordPress Version

6.5.0

### PHP Version

8.1.15

Ubuntu 20.04 ( wsl2 + devilbox).

- [X] Yes

- [X] Yes

jasonbahl:
Fixes #3096

## Problem

When using `register_graphql_field()`, WPGraphQL could throw `DUPLICATE_FIELD` even when an override was compatible with an inherited interface field.

Example: if an interface defines `seo: SeoInterface`, overriding `Post.seo` to `PostSeo implements SeoInterface` should be valid, but duplicate checks previously happened before compatibility could be confirmed.

## Solution

`TypeRegistry::register_field()` now performs an explicit compatibility check for duplicate overrides that appear to come from interface inheritance:

1. Resolve the existing duplicate field's underlying type (including wrapped types such as `list_of` / `non_null`)
2. Confirm the existing field is interface-derived
3. Confirm the overriding object type implements that interface
4. Allow only verified compatible overrides; otherwise keep `DUPLICATE_FIELD`

This includes recursion-safe handling and registered-but-lazy-loaded type scenarios.

## Correctness Fix Note (Behavioral Change)

This PR intentionally tightens duplicate-field override validation.

- **Before:** Some edge cases could be permissive/inconsistent based on type-loading timing.
- **After:** Duplicate overrides are only allowed when compatibility is verified.

Downstream plugins that accidentally relied on those permissive edge cases may now see `DUPLICATE_FIELD` where they previously did not.

We are documenting this as a **correctness fix**, not a formal breaking change, because it aligns behavior with valid schema/interface rules and preserves intended extension behavior for compatible overrides.

## Additional Schema Safety Adjustment

To avoid exposing an invalid root field in environments with no exposed settings:

- `Settings` type is only registered when real settings fields exist
- `RootQuery.allSettings` is conditionally exposed only when settings fields are available

This avoids synthetic placeholder fields while keeping schema validity.

### Interface override compatibility
- `testIncompatibleInterfaceFieldOverrideStillErrors`
- `testCompatibleInterfaceFieldOverrideWithWrappedListInterfaceType`
- `testCompatibleInterfaceFieldOverrideWithListOfInterface` (based on downstream real-world plugin scenario)
- Existing duplicate/connection behavior tests preserved and passing

### Interface field inheritance cleanup
- Removed ambiguous TODO/debug-noise in interface inheritance tests and kept assertions definitive

### Settings schema behavior
- `testAllSettingsFieldIsHiddenWhenNoSettingsAreExposed`

## Validation

- `tests/wpunit/TypesTest.php`
- `tests/wpunit/InterfaceTest.php`
- `tests/wpunit/SettingsQueriesTest.php`
- PHPStan (`composer run phpstan -- --memory-limit=2G`) for `@wpgraphql/wp-graphql`

All passing locally in wp-env.

## Impact

- **Breaking Changes:** None (documented correctness fix)
- **Backward Compatibility:** Compatible interface overrides continue to work; invalid/unverifiable duplicate overrides are now consistently rejected
- **Performance:** Minimal; added checks run only in duplicate-field registration paths

## Related

- Closes #3096
- Addresses downstream reports from WPGraphQL extension maintainers (including list modifier/interface narrowing cases)


plugins/wp-graphql/src/Type/ObjectType/Settings.php
```diff
@@ -24,8 +24,22 @@ class Settings {
...
-			return;
+			$fields = [
+				'_empty' => [
+					'type'        => 'String',
+					'description' => __( 'Placeholder field when no settings are registered', 'wp-graphql' ),
+					'resolve'     => static function () {
+						return null;
+					},
+				],
+			];
```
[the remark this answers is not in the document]
justlevine: I was going to ask the same question. Is this related or just in here?

justlevine:
Sadly this isnt working for me downstream.


image

First thing to note, is the schema docs still thinks the field type is from the interface (`GfFieldChoice`, even though `__typename` resolves to the correct type (`RadioFieldChoice`).

When I try to access fields specific to downstream, you can see that validation fails, assumedly because the Duplicate Field check is preventing the application of the narrowed field.

image

---
You can see the relevant change in this commit: https://github.com/AxeWP/wp-graphql-gravity-forms/pull/487/commits/a4fd6bccc3a574796c7ff3b64530ed3cd087cc0a

Hoping it's just because the function doesn't account for list_of/non_null, but it's very noisy and hard for me to tell.

To test on that branch (so you don't need to manually test gf.
```
nvm use && npm install
npm run wp-env start
npm run wp-env run -- --env-cwd=wp-content/plugins/wp-graphql-gravity-forms cli composer install
npm run wp-env run -- --env-cwd=wp-content/plugins/wp-graphql cli bash

{ then I applied the branch }

npm run tests:codecept -- run tests/wpunit/AddressFieldTest:testGetField -vvv
```

plugins/wp-graphql/src/Type/WPInterfaceTrait.php
```diff
@@ -234,7 +234,30 @@ private function get_fields_from_implemented_interfaces( TypeRegistry $registry
...
+						// This will merge new args into existing, keeping all unique args
+						// @phpstan-ignore-next-line - Args are already in correct format from interface field configs
+						$new_field['args'] = $this->merge_field_args( $interface_field_name, $new_args, $existing_args );
+					}
+
+					// Merge other config properties, with new field taking precedence
+					$interface_fields[ $interface_field_name ] = array_merge( $existing_field, $new_field );
+				} else {
+					$interface_fields[ $interface_field_name ] = $interface_field->config;
+				}
```
justlevine, line 258: Havn't found the bug, but reduced the nesting anyway:

```suggestion
				if ( ! isset( $interface_fields[ $interface_field_name ] ) ) {
					$interface_fields[ $interface_field_name ] = $interface_field->config;
					continue;
				}

				// If this field already exists from another interface, merge the arguments
				// instead of overwriting. This allows arguments from parent interfaces
				// to be merged with child interfaces.
				$existing_field = $interface_fields[ $interface_field_name ];
				$new_field      = $interface_field->config;

				// Merge arguments from both interfaces
				// We want to combine all unique args from both fields
				if ( ! empty( $existing_field['args'] ) || ! empty( $new_field['args'] ) ) {
					$existing_args = $existing_field['args'] ?? [];
					$new_args      = $new_field['args'] ?? [];
					// Use merge_field_args which handles type compatibility checking
					// Pass new_args as field args and existing_args as interface args
					// This will merge new args into existing, keeping all unique args
					// @phpstan-ignore-next-line - Args are already in correct format from interface field configs
					$new_field['args'] = $this->merge_field_args( $interface_field_name, $new_args, $existing_args );
				}

				// Merge other config properties, with new field taking precedence
				$interface_fields[ $interface_field_name ] = array_merge( $existing_field, $new_field );
```

plugins/wp-graphql/src/Registry/TypeRegistry.php
```diff
@@ -1282,6 +1291,131 @@ function ( $fields ) use ( $type_name, $field_name, $config ) {
 						return $fields;
 					}
```
justlevine, line 1302: For starters, can we move this into a private function? maybe having the bounaries clearer will make the bug more obvious

plugins/wp-graphql/tests/wpunit/InterfaceTest.php
```diff
@@ -1420,37 +1441,40 @@ public function testInterfaceFieldInheritanceAndMerging() {
...
 		}';
 
-		$actual = $this->graphql([ 'query' => $query ]);
+		$actual = $this->graphql( [ 'query' => $query ] );
 
 		// Debug the response
-		codecept_debug('Type query response:');
-		codecept_debug($actual);
+		codecept_debug( 'Type query response:' );
+		codecept_debug( $actual );
```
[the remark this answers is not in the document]
justlevine: These were already here.

plugins/wp-graphql/src/Registry/TypeRegistry.php
```diff
@@ -1282,6 +1291,131 @@ function ( $fields ) use ( $type_name, $field_name, $config ) {
...
+							$this->checking_compatibility = true;
+
+							try {
+								// Resolve the existing field type (should be an interface)
+								$resolved_existing_type = is_callable( $existing_field_type ) ? $existing_field_type() : null;
+
+								// If existing field type is a callable, resolve it
+								if ( $resolved_existing_type instanceof \GraphQL\Type\Definition\InterfaceType ) {
+									$interface_name = $resolved_existing_type->name;
+								} elseif ( is_string( $existing_field_type ) ) {
```
justlevine: I think it's here....  we first need to check if its's a listof/nonnull, and if so unwrap it...

plugins/wp-graphql/tests/wpunit/TypesTest.php
```diff
@@ -469,4 +469,1587 @@ public function testRegisterDuplicateTypesOutputsDebugMessage() {
...
+		$response = $this->graphql( compact( 'query' ) );
+
+		// Should have DUPLICATE_FIELD error since the override is incompatible
+		// (not because of an error, but because TestErrorType doesn't implement the interface)
+		$debug_types = wp_list_pluck( $response['extensions']['debug'] ?? [], 'type' );
+		$this->assertContains( 'DUPLICATE_FIELD', $debug_types, 'Should show DUPLICATE_FIELD error when override is incompatible' );
+		
+		// The error handling code path (try/catch/finally) ensures that even if errors occur
+		// during type resolution, the schema still builds and appropriate errors are shown
+	}
```
justlevine, line 2994: Here's a failing test for you:
```suggestion
	}

	/**
	 * Test that interface field override works with list_of type modifiers.
	 */
	public function testCompatibleInterfaceFieldOverrideWithListOfInterface() {
		add_action(
			'graphql_register_types',
			static function () {
				// Register a base interface for choices with a resolver that returns a default value.
				// This resolver should be INHERITED by types that don't override it.
				register_graphql_interface_type(
					'ChoiceWithText',
					[
						'fields' => [
							'text'  => [
								'type'        => 'String',
								'description' => '[Interface] The display text for this choice',
								'resolve'     => static fn ( $source ) => $source['text'] ?? '[Interface Default] No text provided',
							],
						],
					]
				);

				// Register a secondary interface with its own field and resolver.
				register_graphql_interface_type(
					'ChoiceWithValue',
					[
						'fields' => [
							'value'  => [
								'type'        => 'String',
								'description' => '[Interface] The display text for this choice',
								'resolve'     => static fn ( $source ) => '[Interface Default]',
							],
						],
					]
				);

				// Register a concrete type that implements both interfaces.
				// Note: We DON'T override the 'text' resolver - it should inherit from TestChoiceInterface.
				// We DO override 'isSelected' to demonstrate resolver override behavior.
				register_graphql_object_type(
					'TestRadioChoice',
					[
						'interfaces' => [ 'ChoiceWithText', 'ChoiceWithValue' ],
						'fields'     => [
							'value' => [
								'description' => '[Override] Override value field from ChoiceWithValue interface',
								'resolve'     => static fn ( $source ) => '[Override Resolver] Value for ' . ( $source['text'] ?? 'unknown choice' )
							]
						],
					]
				);

				// Create an interface that defines a choices field with a default resolver.
				// This resolver should be REPLACED by the override on Post.
				register_graphql_interface_type(
					'TestNodeWithChoices',
					[
						'interfaces' => [ 'Node' ],
						'fields'     => [
							'choices' => [
								'type'        => [ 'list_of' => 'TestChoiceInterface' ],
								'description' => '[Interface] Default choices from TestNodeWithChoices interface',
								'resolve'     => static fn() => [
									[ 'text' => '[Interface Resolver] Default from TestNodeWithChoices' ],
								],
							],
						],
					]
				);

				// Make the interface available on ContentNode
				register_graphql_interfaces_to_types( 'TestNodeWithChoices', [ 'ContentNode' ] );

				// Override Post.choices with narrowed type and new description/resolver.
				register_graphql_field(
					'Post',
					'choices',
					[
						'type'        => [ 'list_of' => 'TestRadioChoice' ],
						'description' => '[Override] Post-specific radio choices with narrowed type [TestRadioChoice]',
						'resolve'     => static function () {
							return [
								[
									'text'       => '[Override Resolver] Choice #1 from Post.choices',
									'value'      => 'choice1',
								],
								[
									'text'       => '[Override Resolver] Choice #2 from Post.choices',
								],
							];
						},
					]
				);
			}
		);

		// Create a post to query
		$this->factory()->post->create(
			[
				'post_title' => 'Test Post',
			]
		);

		$query = '
			query {
				posts {
					nodes {
						choices {
							__typename
							text
							value
						}
					}
				}
			}
		';

		$response = $this->graphql( compact( 'query' ) );

		// Should NOT have any errors
		$this->assertArrayNotHasKey( 'errors', $response, 'Should not have errors when overriding list_of interface with list_of implementing type' );

		// Should NOT have DUPLICATE_FIELD debug message
		$debug_types = wp_list_pluck( $response['extensions']['debug'] ?? [], 'type' );
		$this->assertNotContains( 'DUPLICATE_FIELD', $debug_types, 'Should not have DUPLICATE_FIELD error when overriding list_of interface with compatible type' );

		// Verify the override resolver is used (not the interface resolver)
		$this->assertNotEmpty( $response['data']['posts']['nodes'] );
		$choices = $response['data']['posts']['nodes'][0]['choices'];
		$this->assertCount( 2, $choices );

		$this->assertEquals( 'TestRadioChoice', $choices[0]['__typename'], 'Field type should be narrowed to TestRadioChoice' );
		$this->assertEquals( '[Override Resolver] Choice #1 from Post.choices', $choices[0]['text'], 'Should use override resolver for text field' );
		$this->assertEquals( '[Override Resolver] Value for [Override Resolver] Choice #1 from Post.choices', $choices[0]['value'], 'Should use override resolver for value field (using text as source)' );

		$this->assertEquals( 'TestRadioChoice', $choices[1]['__typename'], 'Field type should be narrowed to TestRadioChoice' );
		$this->assertEquals( '[Override Resolver] Choice #2 from Post.choices', $choices[1]['text'], 'Should use override resolver for text field' );
		$this->assertEquals( '[Override Resolver] Value for [Override Resolver] Choice #2 from Post.choices', $choices[1]['value'], 'Should use override resolver for value field (using text as source)' );


		// Introspection query to verify field type and description

		$query = '
			query {
				__type(name: "Post") {
					fields(includeDeprecated:false) {
						name
						description
						type {
							kind
							name
							ofType {
								kind
								name
								ofType {
									kind
									name
								}
							}
						}
					}
				}
			}
		';

		$response = $this->graphql( compact( 'query' ) );
		$fields               = $response['data']['__type']['fields'];
		$choices_field        = null;
		foreach ( $fields as $field ) {
			if ( 'choices' === $field['name'] ) {
				$choices_field = $field;
				break;
			}
		}

		$this->assertNotNull( $choices_field, 'Post.choices field should exist' );

		$this->assertEquals( 'LIST', $choices_field['type']['kind'], 'Field type should be a List' );
		$this->assertEquals( 'TestRadioChoice', $choices_field['type']['ofType']['name'], 'List item type should be TestRadioChoice (narrowed from TestChoiceInterface)' );
		$this->assertEquals( 'OBJECT', $choices_field['type']['ofType']['kind'], 'List item should be an OBJECT type (not INTERFACE)' );

		$this->assertStringContainsString( '[Override]', $choices_field['description'], 'Description should be from override, not interface' );
		$this->assertStringContainsString( 'TestRadioChoice', $choices_field['description'], 'Description should mention the narrowed type TestRadioChoice' );
	}
```

PR #3551, opened 2026-02-05
feat: add core user admin preferences fields to User type

izzygld:
Addresses Issue #479

### DIFF
--- a/src/Model/User.php
+++ b/src/Model/User.php
@@ -239,26 +239,44 @@ class User extends Model {
 					return 'true' === $toolbar_preference_meta;
 				},
 				'adminColor'                   => function () {
-					$admin_color = get_user_meta( $this->data->ID, 'admin_color', true );
+					$admin_color = get_user_option( 'admin_color', $this->data->ID );
 
 					return ! empty( $admin_color ) ? $admin_color : 'fresh';
 				},
 				'hasRichEditingEnabled'        => function () {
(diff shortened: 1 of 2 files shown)


jasonbahl (approved):
Thanks for the PR!

It addresses the core issue described in #479 and adds the core user admin preference fields to the User type.

Similar to our discussion about options, I'd love at some point to have a more formal way to map registered user options to the schema more programmatically (i.e. when a plugin adds fields to the user profile, I'd love for those fields to automatically map to the schema), but until we spend the time to come up with a centralized registry/map, this is a good step.

I like that you followed the existing patterns using the Model layer and added tests! 🙌

I made a few changes, but wanted to see what you think:

- Switched from `get_user_meta()` to `get_user_option()` for all four fields:
Better alignment with WordPress core (especially multisite)
`get_user_option()` handles site-specific user options correctly

- Updated default handling since `get_user_option()` returns false when not set

- Added edge case tests:
  - `testUserAdminPreferencesExplicitFalseValues()`: Verifies explicit 'false' string values are handled correctly
  - `testUserAdminPreferencesFieldVisibility()`: Tests visibility rules across different user roles (admin, subscriber, public)

- Updated existing tests to use `update_user_option()` for consistency with the new implementation

**Note**: Mutation support for these fields (mentioned in the original issue) can be tracked in a follow-up issue.

PR #3778, opened 2026-04-16
feat: allow additional valid hosts

### LINKED ISSUE #3777: NodeResolver parse_request() ignores WPML language domains

When querying a post by URI, if the URI is fully qualified with a hostname that doesn't match the host as returned by site_url() or home_url() the resolver returns null. 

For example, on a site running on `https://domain.com` using WPML set-up with 'domain per language', the query `post(id:"https://de.domain.com/hallo-welt", idType: URI)` will fail.

Allow plugins to add additional allowed hosts via a filter. A good place for this is in NodeResolver around line 364.
A merge request will follow. 

NOTE: It would also be great if the original URI was saved somewhere so that other plugins can access it. When calling parse_request, the return value of this call replaces the $uri used by NodeResolver, making it irretrievable by plugins that hook late to the resolve hooks.

_No response_

### Additional Context

_No response_

d-packs:
### Update NodeResolver.php to allow additional valid hosts

Adds a new filter called 'graphql_allowed_hosts' which allows additional domains to be registered by other plugins. For example a translation plugin will want to add language speciffic domains running on the same install.

closes #3777

### DIFF
--- a/plugins/wp-graphql/src/Data/NodeResolver.php
+++ b/plugins/wp-graphql/src/Data/NodeResolver.php
@@ -362,21 +362,11 @@ class NodeResolver {
 
 		// Bail if external URI.
 		if ( isset( $parsed_url['host'] ) ) {
-			$site_url = wp_parse_url( site_url() );
-			$home_url = wp_parse_url( home_url() );
+			$site_url      = wp_parse_url( site_url() );
+			$home_url      = wp_parse_url( home_url() );
+			$allowed_hosts = apply_filters( 'graphql_allowed_hosts', [ $site_url['host'], $home_url['host'] ] );
 
-			/**
-			 * @var array<string,mixed> $home_url
-			 * @var array<string,mixed> $site_url
-			 */
-			if ( ! in_array(
-				$parsed_url['host'],
-				[
-					$site_url['host'],
-					$home_url['host'],
-				],
-				true
-			) ) {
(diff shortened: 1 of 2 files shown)


jasonbahl (approved):

jasonbahl (dismissed):

jasonbahl:
@d-packs thanks for the PR and taking the time to share the use case!

plugins/wp-graphql/src/Data/NodeResolver.php
```diff
@@ -362,21 +362,11 @@ public function parse_request( string $uri, $extra_query_vars = '' ) {
 
 		// Bail if external URI.
 		if ( isset( $parsed_url['host'] ) ) {
-			$site_url = wp_parse_url( site_url() );
-			$home_url = wp_parse_url( home_url() );
+			$site_url      = wp_parse_url( site_url() );
+			$home_url      = wp_parse_url( home_url() );
+			$allowed_hosts = apply_filters( 'graphql_allowed_hosts', [ $site_url['host'], $home_url['host'] ] );
```
justlevine: Can we add a doc-block here, ~and pass them in as two separate args instead of as an `$args[0]` and `$args[1]`~?

jasonbahl (dismissed):

jasonbahl (approved):

PR #3834, opened 2026-05-19
fix(security): drop file:// support from createMediaItem mutation

jasonbahl:
## Summary

- Removes `'file'` from the default `$allowed_protocols` list in `createMediaItem`, so the mutation no longer advertises local-file sideload as supported.
- Deletes the now-unreachable `file_get_contents()` + `wp_upload_bits` branch that consumed `file://` paths — including the `file_get_contents()` call named as the sink in a recent external security report.
- Adds a regression test (`testFileProtocolIsRejected`) asserting `file://` filePaths are rejected and no attachment URL is returned.

## Why

`'file'` was added to the default `$allowed_protocols` in v1.14.6 (#2840) alongside a `wp_http_validate_url()` check pushed in the same PR. `wp_http_validate_url()` strips non-`http`/`https` schemes and requires a non-empty host, so `file://...` URLs have been rejected by core validation before the protocol allowlist is ever consulted. That made:

- the `'file'` allowlist entry effectively a no-op,
- the `wp_upload_bits` branch (lines 236–239 prior) unreachable in stock WordPress,
- the `graphql_media_item_create_allowed_protocols` filter unable to serve as a real escape hatch for `file://`, since `wp_http_validate_url()` runs first and has no filter inside it.

A public GitHub code search for `graphql_media_item_create_allowed_protocols` returns only WPGraphQL's own source (vendored copies), no consumer code calling `add_filter()`. So removing `file` from the default has no known compatibility impact.

An external security researcher (@gokul965) reported the design intent (allowing `file` + calling `file_get_contents()` on user input) as a Critical LFI vulnerability. The exploit as written is not reachable on stock WordPress for the reasons above, but the design smell is real defense-in-depth weakness: anything that loosened `wp_http_validate_url()` would flip this to exploitable. This PR removes both the smell and the sink.

## What changed

1. `plugins/wp-graphql/src/Mutation/MediaItemCreate.php`
   - Line 199: `$allowed_protocols = [ 'https', 'http', 'file' ]` → `[ 'https', 'http' ]`
   - Removed `$file_contents = file_get_contents( $input['filePath'] )` and the conditional `wp_upload_bits` branch that depended on it.
2. `plugins/wp-graphql/tests/wpunit/MediaItemMutationsTest.php`
   - Added `testFileProtocolIsRejected` regression test.

## Test plan

- [x] `MediaItemMutationsTest` runs green locally (28 tests, 41 assertions)
- [x] PHPCS clean on changed lines
- [x] CI matrix passes (WordPress 6.1–trunk × PHP 7.4–8.4 × block/classic × single/multisite)

## Follow-up (not in this PR)

Adjacent `createMediaItem` test coverage is fragile and SSRF-thin. A separate PR will:
- mock the HTTP layer so happy-path tests don't depend on `raw.githubusercontent.com`,
- add SSRF regression tests for `169.254.169.254`, RFC1918 ranges, IPv6 loopback, and redirects to private IPs,
- add a `wp_handle_sideload` failure test.

### DIFF
--- a/plugins/wp-graphql/src/Mutation/MediaItemCreate.php
+++ b/plugins/wp-graphql/src/Mutation/MediaItemCreate.php
@@ -228,16 +228,6 @@ class MediaItemCreate {
 			 */
 			require_once ABSPATH . 'wp-admin/includes/file.php';
 
-			$file_contents = file_get_contents( $input['filePath'] );
-
-			/**
-			 * If the mediaItem file is from a local server, use wp_upload_bits before saving it to the uploads folder
-			 */
-			if ( 'file' === wp_parse_url( $input['filePath'], PHP_URL_SCHEME ) && ! empty( $file_contents ) ) {
-				$uploaded_file     = wp_upload_bits( $file_name, null, $file_contents );
-				$uploaded_file_url = ( empty( $uploaded_file['error'] ) ? $uploaded_file['url'] : null );
-			}
-
 			/**
 			 * Ensure we have a valid URL before attempting download
 			 */
--- a/plugins/wp-graphql/src/Mutation/MediaItemCreate.php
+++ b/plugins/wp-graphql/src/Mutation/MediaItemCreate.php
@@ -196,7 +196,7 @@ class MediaItemCreate {
 			$protocol = wp_parse_url( $input['filePath'], PHP_URL_SCHEME );
 
 			// prevent the filePath from being submitted with a non-allowed protocols
-			$allowed_protocols = [ 'https', 'http', 'file' ];
+			$allowed_protocols = [ 'https', 'http' ];
 
 			/**
 			 * Filter the allowed protocols for the mutation
--- a/plugins/wp-graphql/tests/wpunit/MediaItemMutationsTest.php
+++ b/plugins/wp-graphql/tests/wpunit/MediaItemMutationsTest.php
@@ -1335,4 +1335,30 @@ class MediaItemMutationsTest extends \Tests\WPGraphQL\TestCase\WPGraphQLTestCase
 
 		$this->assertQueryError( $actual, $expected );
 	}
+
+	/**
+	 * Asserts that a filePath using the file:// protocol is rejected and no
+	 * media item is created. Regression test for the local file inclusion
+	 * vector previously enabled by including 'file' in the default
+	 * $allowed_protocols list.
+	 */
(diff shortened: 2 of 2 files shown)


jasonbahl:
Heads up — while building the follow-up test hardening on `chore/harden-create-media-item-tests`, I confirmed an additional dimension to this fix worth noting in the PR body.

The `file_get_contents( $input['filePath'] )` call this PR removes was reachable for **any** URL that passes `wp_check_filetype` + `wp_http_validate_url`, not only `file://`. PHP's `file_get_contents` uses its own HTTP stream wrapper, which bypasses WordPress's HTTP API entirely (no `pre_http_request` filter, no `wp_safe_remote_get` IP filtering). So an authenticated `upload_files` user could coerce the server into HTTP GETs against any URL the PHP process can reach — including cloud instance metadata.

Concrete demo on `origin/main` (before this PR), with an SSRF regression test that runs the mutation against `http://169.254.169.254/latest/meta-data/instance-id.jpg`:

```
debugMessage: file_get_contents(http://169.254.169.254/latest/meta-data/instance-id.jpg):
              Failed to open stream: HTTP request failed!
```

The test took ~60s (PHP's default stream-wrapper timeout) because `file_get_contents` was actually attempting the connection. After rebasing onto this PR, the same test runs in 0.13s and surfaces the gap differently: `download_url`/`wp_safe_remote_get` is now the sink, and `wp_http_validate_url()` doesn't reject `169.254/16` (its reject list covers `127/8`, `10/8`, `0/8`, `172.16-31/12`, `192.168/16` — but not link-local).

What this means for #3834:
- The fix here closes the `file_get_contents` SSRF vector entirely (impact reduction is broader than just LFI).
- A residual SSRF surface remains via `download_url`'s HTTP path against `169.254.169.254`. That's a separate hardening — adding a plugin-level guard for `169.254/16` — which I'll bring in the follow-up tests PR alongside the regression test that surfaced it.

Suggested addition to the PR description (paraphrased): *"This also closes an SSRF vector reachable via the same `file_get_contents()` call: PHP's HTTP stream wrapper bypasses `wp_safe_remote_get`, so any URL that passes `wp_check_filetype` and `wp_http_validate_url` (including `http://169.254.169.254/`) could be fetched by the server. The cloud-metadata vector via `download_url` itself remains and will be addressed in a follow-up."*

PR #3867, opened 2026-05-30
fix(wp-graphql-acf): mock oEmbed in tests and unskip clone-field tests

### LINKED ISSUE #3843: fix(wp-graphql-acf): clone-field tests skipped due to environment-dependent oEmbed expectation
> Plugin lives in this monorepo at `plugins/wp-graphql-acf/`.

## Summary

Two clone-field tests in wp-graphql-acf are currently marked skipped with the note *"Clone field tests need investigation - prefix_name behavior may have changed"*:

- `plugins/wp-graphql-acf/tests/wpunit/FieldTypes/OembedFieldTest.php:45`
- `plugins/wp-graphql-acf/tests/wpunit/FieldTypes/AcfeCountriesFieldTest.php:97`

After running them locally against the current monorepo (ACF Pro 6.7.0.2, this branch's WPGraphQL), the actual `prefix_name` round-trip works fine — the test reaches the assertion and the response shape is correct. The failure mode is **only** in the expected value: `wp_oembed_get()` cannot reach external services from inside the wp-env tests container, so it returns the wrapped-anchor fallback HTML rather than the embed markup the test expects.

## Evidence

Running `OembedFieldTest::testQueryCloneFieldOnPost` un-skipped:

```
data.post.acfTestGroup.clonedTestOembed
  ACTUAL_DATA:    "<a href=\"https://twitter.com/wpgraphql/...\">https://twitter.com/wpgraphql/...</a>"
  WANTED_VALUE:    false
extensions.debug:   []   # no DUPLICATE_FIELD / no schema-build warnings
```

The query itself succeeded end-to-end — schema build, resolver, clone-field traversal, prefix handling, type assignment. The `WANTED_VALUE: false` comes from `wp_oembed_get( 'https://twitter.com/...' , [ 'width' => 550 ] )` returning `false` inside the container (no outbound oEmbed fetch). The test's `get_expected_clone_value()` calls `wp_oembed_get()` at runtime, so when the network call fails, expected becomes `false` while WPGraphQL returns the fallback anchor HTML.

`AcfeCountriesFieldTest::testQueryCloneFieldOnPost` is the same shape; the skip note is identical and almost certainly the same underlying issue.

## Proposed fix

Pick one of:

1. **Mock oEmbed for the test.** Add a `pre_oembed_result` filter (or `oembed_request_post_id` short-circuit) in setUp() that returns a deterministic HTML blob, and update `get_expected_clone_value()` accordingly. This is the most honest fix — the test should exercise data flow, not third-party HTTP.
2. **Match the fallback.** Update `get_expected_clone_value()` to return the wp-supplied fallback anchor (`<a href="…">…</a>`) and explicitly document that the test runs against the WP fallback path. Simpler, but it stops covering "embed markup is preserved through the field's storage/retrieval."
3. **Switch the test fixture to a non-oEmbed clone.** Have the oembed test cover the field-registration/introspection paths but a different test cover the resolved value with a value that doesn't depend on network.

(1) is the preferred direction; (2) is acceptable if mocking is awkward.

## Repro

```
# in this monorepo
npm run wp-env start
npm run -w @wpgraphql/wp-graphql-acf install-acf:pro     # needs ACF_LICENSE_KEY in .env
npm run wp-env -- run tests-cli wp plugin activate advanced-custom-fields-pro wp-graphql wp-graphql-acf
# temporarily replace the markTestSkipped at OembedFieldTest.php:45 with parent::testQueryCloneFieldOnPost();
npm run -w @wpgraphql/wp-graphql-acf test:codecept:wpunit -- tests/wpunit/FieldTypes/OembedFieldTest.php:testQueryCloneFieldOnPost
```

## Related

- Surfaced while investigating whether #3539 unblocks these skipped tests. PR #3539 does **not** fix them — they were already running cleanly past the duplicate-field machinery; the empty `extensions.debug` array is in fact positive evidence that #3539's compatible-override path silenced what would previously have been DUPLICATE_FIELD noise for ACF's `{TypeName}` ↔ `{TypeName}_Fields` interface pattern. This issue is independent of #3539.

priyanshi498:
Closes wp-graphql/wp-graphql#3843
This pull request unskips and fixes two clone-field tests in the wp-graphql-acf plugin that were previously marked skipped: OembedFieldTest and AcfeCountriesFieldTest.

The oEmbed clone test failed because wp_oembed_get cannot make external HTTP requests inside offline test containers like wp-env. It returned false because of the network failure, while the resolver returned the fallback anchor HTML.

To fix this, we:

- Added a pre_oembed_result filter in OembedFieldTest to mock oEmbed resolution and return mock HTML content, bypassing network requests.
- Removed the filter in tearDown to avoid affecting other tests.
- Updated the expected clone value to match the mocked HTML.
- Removed the skipped overrides in both test files to inherit and run the base clone-field tests.
- This resolves the network dependency and makes the tests fast and reliable.


jasonbahl (approved):

jasonbahl:
Thanks for the PR @priyanshi498! 🙏

PR #3936, opened 2026-06-13
feat(ide): query composer reveals the field at the editor cursor

### LINKED ISSUE #3933: IDE: Query Composer should follow the editor cursor and scroll to the field at the cursor
## Problem

The Query Composer in the new WPGraphQL IDE does not visibly follow the cursor position in the document editor. With the cursor on a field (e.g. `title` inside `posts { nodes { ... } }`), the composer stays wherever it was, typically showing the alphabetical root field list, and the user has to scroll and expand manually to find the part of the composer that corresponds to where they are in the document.

The plumbing for cursor tracking exists end to end: the editor publishes cursor changes (`onCursorChange` → `setCursorOffset` in `IDELayout.jsx`), the store exposes `getCursorOffset()`, and the composer consumes it (`QueryComposer.jsx` → `ExplorerView.jsx`). But the only behavior implemented is operation-level: `findOpKeyAtCursor()` expands the operation the cursor is in and collapses the others, which is a visible no-op in a single-operation document, the most common case. There is no field-level reveal or scroll-into-view.

## Desired behavior

The composer passively follows the editor cursor at field granularity:

- When the cursor lands on a field in the document, the composer expands the path to that field and scrolls the corresponding row into view (a brief highlight would help orient the user)
- No new gesture or modifier is needed, this is panel-follows-cursor, the same model as VS Code's Outline view, and it composes cleanly with cmd+click continuing to jump to the Docs
- A possible follow-up, not required for this issue: an explicit "reveal in Query Composer" action (context menu or command palette) that also opens the composer when it is closed

## Implementation notes

- The composer already keeps AST locations for exactly this purpose (`utils.js`, parse with `loc` retained, used by `findOpKeyAtCursor`)
- Extending the existing approach: walk the cursor-matched operation's selection set to find the deepest field whose `loc` spans the cursor offset, derive the AST path, expand ancestors, and `scrollIntoView` the matching row (`FieldView` rows would need refs keyed by path)

## Context

Spun out of #2803, which asked for this against the legacy IDE bundled with core. The legacy IDE is being deprecated and removed from core, so this issue targets the new WPGraphQL IDE (`plugins/wp-graphql-ide/plugins/query-composer-panel/`).

josephfusco:
## What feature does this implement? Explain your changes.

When the cursor lands on a field in the GraphQL editor, the Query Composer panel scrolls to that field's row and briefly highlights it. The panel already received the cursor position, but used it only to expand the operation containing the cursor — invisible in a single-operation document.

Closes #3933.

### Schema Changes

None.

### User Experience

The matching row scrolls into the middle of the panel and highlights with the panel's accent color, then fades. No new shortcut; composes with the existing cmd-click jump-to-Docs.

Tracked follow-ups (#3933):

- An explicit "Reveal in Query Composer" command that opens the panel when it's closed.
- Following the cursor into inline fragments or fragment spreads — the reveal stops at the nearest enclosing field.

### Performance Considerations

The cursor offset updates on every keystroke and arrow press:

- The panel remembers which row was last revealed and skips when nothing changed; typing inside the same field triggers no scroll, highlight, or other DOM work.
- Walking the parsed GraphQL document is O(depth × siblings per level); the parse itself is already memoized upstream.
- The scroll is a no-op when the row is already visible.

### Test Coverage

Unit (Jest) and end-to-end (Playwright).

## Documentation Updates

No public API or user-facing documentation changes. The changelog is generated by release-please from the PR title at the next IDE release.

### Breaking Changes

None.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11799,9 +11799,9 @@
 			}
 		},
 		"node_modules/@wordpress/hooks": {
-			"version": "4.48.0",
-			"resolved": "https://registry.npmjs.org/@wordpress/hooks/-/hooks-4.48.0.tgz",
-			"integrity": "sha512-rU1yGEy0Mb+2oRG5QX/bKIIwKQmYAvATfUQeXIF20/mbR0qutYeVTCIvWEyb4pf71tvnQFiN18RWRXWsvKrDbQ==",
+			"version": "4.48.1",
+			"resolved": "https://registry.npmjs.org/@wordpress/hooks/-/hooks-4.48.1.tgz",
+			"integrity": "sha512-QZh3Cv9TMJkrCQikuZSsDKTgX+6BBzYJe0MFKp7yP8drj/dtRpkqo5+eYmovBq3pk+HoyP7UJ1vTOb3GPJeU6Q==",
 			"license": "GPL-2.0-or-later",
 			"engines": {
 				"node": ">=18.12.0",
--- a/plugins/wp-graphql-ide/plugins/query-composer-panel/src/components/ExplorerView.jsx
+++ b/plugins/wp-graphql-ide/plugins/query-composer-panel/src/components/ExplorerView.jsx
@@ -138,10 +138,22 @@ class ExplorerView extends React.PureComponent {
 		if (!el) {
 			return;
 		}
-		// `center` keeps the row away from the panel's sticky operation
-		// title-bar at the top and the bottom edge of the scroll area —
-		// both clip a row aligned to the nearest edge.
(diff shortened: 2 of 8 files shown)


jasonbahl (approved):

jasonbahl (changes requested):
Image

@josephfusco good work! There is a small bug though. When the smooth scroll kicks in, while the drawer is open, It scrolls the header of the IDE up out of the drawer container.

jasonbahl (approved):

PR #3961, opened 2026-06-18
ci: speed up the integration test workflow

jasonbahl:
## What

Trims removable overhead from the integration test matrix. Four independent changes, each backed by timing/cache data from a recent run.

## Why

A coverage job runs ~7.5 min. Per-step measurement:

| Step | Coverage job | Boundary job |
|---|---|---|
| WPUnit tests | 173s (PCOV) | 94s |
| Start Docker env | 102s | 94s |
| npm ci | 46s | 54s |
| Build JS (all plugins) | 41s | 42s |
| **Post Cache Docker images** | **39s** | 24s |

Some of this is the inherent floor (wp-env boot ~90s, WPUnit ~95s of real work). The rest is removable.

### 1. Drop the wp-env Docker image cache (`ScribeMD/docker-cache`)
The repo's **Actions cache is at 10.9 GB, over GitHub's 10 GB limit**, so caches are LRU-evicted. I verified directly: **0 `docker-*` caches survive** — they're evicted before the next run reads them. So every job spent ~24-39s *saving* an image cache that was never restored, while `Start Docker env` pulled images fresh anyway. Removing it is pure savings and frees budget that was evicting the composer/npm caches that *do* help.

### 2. Add `cleanup-pr-caches` workflow
Deletes a PR's caches when it closes (caches are scoped to `refs/pull/<n>/merge` and otherwise linger until eviction). Keeps the 10 GB budget from thrashing so the base-branch composer/npm caches stay hot.

### 3. Collect PCOV coverage on one core job, not four
PCOV adds ~80s to the WPUnit run (173s vs 94s), and line coverage barely differs across theme/multisite. All four jobs still run to exercise theme × multisite; only the `twentytwentyfive` single-site job collects coverage now. Saves ~80s on three core jobs per run.

### 4. Build only the plugin under test
`--filter='@wpgraphql/<plugin>...'` instead of `./plugins/*`. Smart Cache and ACF have no JS build, so their jobs now build nothing; core's built assets are `file_exists`-guarded (`Admin/GraphiQL`, `Admin/Extensions`), so dependent-plugin jobs that load core don't need it built. Core/IDE jobs still build their own (and internal deps).

Also a minor `npm ci --prefer-offline --no-audit --no-fund`.

## Expected impact
Coverage jobs ~7.5 → ~5 min, boundary jobs ~6 → ~5 min, plus a healthier cache budget. CI-only change; no plugin code is touched.

## Note
Overlaps with #3960 (WP 7.0 matrix) on the core coverage rows. Whichever merges second will need a trivial rebase (keep WP 7.0 from #3960 + the per-row `coverage` flags from this PR).

### DIFF
--- a/.github/workflows/cleanup-pr-caches.yml
+++ b/.github/workflows/cleanup-pr-caches.yml
@@ -1,0 +1,54 @@
+name: Cleanup PR caches
+
+# When a pull request is closed (merged or not), delete the GitHub Actions
+# caches it created. Caches from a PR run are scoped to refs/pull/<number>/merge
+# and otherwise linger until GitHub LRU-evicts them.
+#
+# The integration test suite is cache-heavy (composer, npm, Playwright) and the
+# repo's 10GB Actions cache budget was thrashing — once over the limit, GitHub
+# evicts least-recently-used caches, which meant the caches that actually speed
+# jobs up kept getting dropped. Proactively reclaiming closed-PR caches keeps the
+# useful base-branch caches hot.
+
+on:
+  pull_request:
+    types:
+      - closed
+
+permissions:
+  contents: read
+
+jobs:
+  cleanup:
+    name: Delete caches for closed PR
+    runs-on: ubuntu-latest
+    permissions:
+      # Required to delete Actions caches via the API.
+      actions: write
+      contents: read
+    steps:
+      - name: Delete caches for this PR's ref
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          PR_REF: refs/pull/${{ github.event.pull_request.number }}/merge
+        run: |
+          gh extension install actions/gh-actions-cache || true
+
+          echo "Fetching caches for $PR_REF ..."
+          cache_keys=$(gh actions-cache list -R "$GH_REPO" -B "$PR_REF" --limit 100 | cut -f1)
+
+          if [ -z "$cache_keys" ]; then
+            echo "No caches found for $PR_REF — nothing to clean up."
+            exit 0
+          fi
+
+          echo "Deleting $(echo "$cache_keys" | wc -l | tr -d ' ') cache(s) for $PR_REF:"
+          # Best-effort: don't abort the whole job if a single delete fails
+          # (e.g. the cache was already evicted between list and delete).
+          set +e
+          for key in $cache_keys; do
+            echo "  - $key"
+            gh actions-cache delete "$key" -R "$GH_REPO" -B "$PR_REF" --confirm
+          done
+          echo "Cache cleanup complete."
--- a/.github/workflows/integration-tests-reusable.yml
+++ b/.github/workflows/integration-tests-reusable.yml
@@ -176,16 +176,21 @@ jobs:
         with:
           timeout_minutes: 10
           max_attempts: 3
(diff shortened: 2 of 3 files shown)


jasonbahl:
## 📊 Validation: measured speedup

Compared two full runs of the **same WP 7.0 matrix** — the only difference is this PR's optimizations:

- **Baseline** (WP 7.0 matrix, no optimizations): [run 27788825690](https://github.com/wp-graphql/wp-graphql/actions/runs/27788825690) (the #3960 branch)
- **Optimized** (this PR): [run 27793271726](https://github.com/wp-graphql/wp-graphql/actions/runs/27793271726)

### Headline

**~19% less CI compute per run** — total job execution time across the matrix dropped **9,369s → 7,607s (−1,762s, ~29 min of aggregate compute)**, with **21 of 22 jobs faster**. Measured as per-job execution time (start→finish), so queue time is excluded.

### Representative jobs (execution time)

| Job | Baseline | This PR | Δ |
|---|---:|---:|---:|
| wp-graphql / 7.0 / twentytwentyone | 478s | 283s | **−195s** |
| wp-graphql / 7.0 / twentytwentyfive (Multisite) | 454s | 323s | **−131s** |
| wp-graphql / 7.0 / twentytwentyfive 📊 | 529s | 423s | **−106s** |
| wp-graphql-acf / 6.8 / ACF Pro | 444s | 306s | **−138s** |
| wp-graphql-acf / 7.0 / ACF Pro + Ext Pro 📊 | 568s | 465s | **−103s** |
| wp-graphql-ide / trunk | 345s | 237s | **−108s** |
| wp-graphql-smart-cache / 6.9 | 477s | 406s | **−71s** |

### Where the time went (attribution)

The cleanest example is **smart-cache / WP 6.9** (its coverage flag didn't change, so this isolates the cache + build wins from the PCOV win):

| Step | Baseline | This PR |
|---|---:|---:|
| Build JavaScript assets | 43s | **0s** (smart-cache has no JS build → scoped build skips it) |
| Post Cache Docker images | 20s | **0s** (dead docker cache removed) |
| Start the Docker testing environment | 87s | 91s (unchanged — confirms the image cache was always missing) |
| Run WPUnit Tests | 224s | 227s (unchanged) |

So the three levers all show up as predicted:
1. **Removed the dead `ScribeMD/docker-cache`** → the ~20-40s `Post Cache Docker images` step is gone from every job, and `Start Docker` is unchanged (it was already pulling fresh).
2. **Scoped the JS build** → smart-cache/acf build nothing (−40s); core/IDE build only themselves.
3. **PCOV on one core job instead of four** → the three de-coverage'd core jobs show the biggest drops (−131s to −195s).

### Honest caveats

- This is **one run each**, so normal CI variance applies (runner speed, image-pull, test timing). The direction is consistent across 21/22 jobs.
- The one job that went *up* — smart-cache / 7.0 (+96s) — is variance, not a regression: its `Build JS` (−42s) and `Post Cache` (−28s) both improved as expected, but its WPUnit suite happened to run 165s slower on that runner (410s→575s). Coverage-heavy WPUnit time fluctuates run-to-run.
- Trade-off already noted in the PR: three core jobs no longer collect PCOV (we still collect it on one representative job), in exchange for the speedup.

Plus the `cleanup-pr-caches` workflow should keep the 10GB cache budget from thrashing going forward, which helps the composer/npm caches stay warm.

Issue #3260, opened 2024-12-12
Modify how meta is handled in revisions now that the block editor revises meta


jasonbahl:
### What problem does this address?

For a long time WordPress did not revise meta, so we came up with a bit of a "shim" to have fields that call for meta to resolve from the parent (the published post) so that `null` values were not returned, but rather the published post's values were returned.

Now that [meta revisions are supported by WordPress core](https://make.wordpress.org/core/2023/10/24/framework-for-storing-revisions-of-post-meta-in-6-4/), it would be good to revisit our shim (see: https://github.com/wp-graphql/wp-graphql/blob/develop/src/Utils/Preview.php#L7-L57). 

### What is your proposed solution?

Currently we can use this filter to tell WPGraphQL to use the revision meta instead of the parent (published post) meta. 

```php
add_filter( 'graphql_resolve_revision_meta_from_parent', function( $default, $object_id, $meta_key, $meta_value ) {
	if ( 'footnotes' === $meta_key ) {
		return false;
	}
}, 10, 4);
```

I would like to use the meta registry to automate this, so that any meta field that has opted-in to revisions would automatically work without users having to implement a filter manually like this. 

### What alternatives have you considered?

Allowing users to continue explicitly opt-in/out using the `graphql_resolve_revision_meta_from_parent` filter. 

### Additional Context

related: 

- https://github.com/WordPress/gutenberg/issues/16006#issuecomment-657965028
- see: https://github.com/WordPress/gutenberg/issues/16006#issuecomment-2540137615

traed:
@jasonbahl  I'm having issues showing custom post meta in previews still. `graphql_resolve_revision_meta_from_parent` doesn't seem to do anything. Is there an example of every step needed to make custom post meta work with preivews? And is this supported with the block editor now is that still not working at all?

jasonbahl:
Cross-referencing #2664, which is the same underlying issue from a different angle. The featured image not respecting `asPreview` comes down to the preview reading the parent post's `_thumbnail_id` meta instead of the revision's own, which is the same parent-versus-revision meta resolution this issue is about.

The plan here is two steps. First, a conservative allow-list of the meta keys we already know WordPress core revises, like `footnotes`, shimmed so previews resolve those from the revision itself. Second, a registry-driven way for other meta to opt into that same handling, which I want to design alongside the broader meta mapping work in #3931 so the conventions stay consistent.

jasonbahl:
[...]

PR #3968, opened 2026-06-19
fix: reflect a revision's own featured image in asPreview queries

### LINKED ISSUE #2664: The featuredImage doesn't take the isPreview parameter into account.
### Description

We are using the isPreview attribute on our project to enable users to preview changes they make to posts and pages. 

While the query itself is mostly correct, the featured image does not take the isPreview attribute into account and always displays the last published value instead of the draft one.

### Steps to reproduce

1. Create a post, set a featured image and publish it
2. Edit the post, change title and featured image
3. Click the "Preview" button
4. In the GraphiQL IDE paste the sample query below.

The result has the updated title but the featured image (databaseID and src) are incorrect.

### Additional context

Sample query

```
{
  post(id: 116, idType: DATABASE_ID, asPreview: true) {
    title
    featuredImageDatabaseId
    featuredImage {
      node {
        id
        databaseId
        medium: sourceUrl(size: LARGE)
      }
    }
  }
}
```

Result
```
"post": {
      "title": "Etiam imperdiet imperdiet orci?",
      "featuredImageDatabaseId": 73,
      "featuredImage": {
        "node": {
          "id": "cG9zdDo3Mw==",
          "databaseId": 73,
          "medium": "http://local.test/wp-content/uploads/2022/11/alexander-shatov-PHH_0uw9-Qw-unsplash-scaled.jpg"
        }
      }
    }
```

Expected
```
"post": {
      "title": "Etiam imperdiet imperdiet orci?",
      "featuredImageDatabaseId": 81,
      "featuredImage": {
        "node": {
          "id": "cG9zdDo4MQ==",
          "databaseId": 81,
          "medium": "http://local.test/wp-content/uploads/2022/11/mike-meyers-haAxbjiHds-unsplash-scaled.jpg"
        }
      }
    }
```

### WPGraphQL Version

1.13.6

### WordPress Version

6.1.1

### PHP Version

8.0

_No response_

- [X] Yes

- [X] Yes

jasonbahl:
## What

`featuredImageDatabaseId` (and the `featuredImage` connection and `featuredImageId` that derive from it) now resolve a revision's own featured image when querying a post with `asPreview: true`, instead of always returning the parent post's published featured image.

## Why

When previewing an in-progress change (`asPreview: true`), the model resolved the featured image from the parent (published) post unconditionally. So a draft that changed the featured image still showed the old, published image in the preview, even though the title and other fields reflected the draft. Reported in #2664.

## How

For revisions, read the revision's own `_thumbnail_id` first and fall back to the parent's published featured image only when the revision does not override it. The Preview meta filter (`WPGraphQL\Utils\Preview`) resolves revision meta from the parent by default, so this opts the single `_thumbnail_id` key out via the existing `graphql_resolve_revision_meta_from_parent` filter while reading the revision's own value.

The fallback is conditional, not a blind swap: previews that do not change the featured image continue to show the parent's published image, so the common case is unchanged. The non-revision path is untouched.

## Test

Two regression tests in `PreviewTest`:

- `testFeaturedImageReflectsRevisionOwnThumbnail`: the revision stores a different `_thumbnail_id`; the preview must return the revision's image, not the parent's. Fails before the change, passes after.
- `testFeaturedImageFallsBackToParentWhenRevisionHasNoThumbnail`: the revision has no `_thumbnail_id`; the preview must still return the parent's published featured image. Locks in the conditional fallback.

`composer check-cs` and `composer phpstan` (level 8) pass; the full `PreviewTest` suite is green.

Closes #2664

### DIFF
--- a/plugins/wp-graphql/src/Model/Post.php
+++ b/plugins/wp-graphql/src/Model/Post.php
@@ -542,13 +542,30 @@ class Post extends Model {
 					return $this->html_entity_decode( apply_filters( 'the_excerpt', $excerpt ), 'excerptRendered' );
 				},
 				'featuredImageDatabaseId'   => function () {
-					if ( $this->isRevision ) {
-						$id = $this->parentDatabaseId;
-					} else {
-						$id = $this->data->ID;
+					// For non-revisions, the featured image is simply this post's thumbnail.
+					if ( ! $this->isRevision ) {
+						$thumbnail_id = get_post_thumbnail_id( $this->data->ID );
+
+						return ! empty( $thumbnail_id ) ? absint( $thumbnail_id ) : null;
 					}
 
-					$thumbnail_id = get_post_thumbnail_id( $id );
+					// For revisions (e.g. `asPreview`), prefer the revision's own _thumbnail_id
+					// when it has one, so an in-progress featured image change is reflected in
+					// the preview. The Preview meta filter resolves revision meta from the parent
+					// by default, so opt that single key out to read the revision's own value, and
+					// fall back to the parent's published featured image only when the revision
+					// does not override it.
+					$resolve_own_thumbnail = static function ( $should, $object_id, $meta_key ) {
+						return '_thumbnail_id' === $meta_key ? false : $should;
+					};
+
+					add_filter( 'graphql_resolve_revision_meta_from_parent', $resolve_own_thumbnail, 10, 3 );
+					$thumbnail_id = get_post_meta( $this->data->ID, '_thumbnail_id', true );
+					remove_filter( 'graphql_resolve_revision_meta_from_parent', $resolve_own_thumbnail, 10 );
+
+					if ( empty( $thumbnail_id ) ) {
+						$thumbnail_id = get_post_thumbnail_id( $this->parentDatabaseId );
+					}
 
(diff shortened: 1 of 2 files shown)


jasonbahl:
After digging into how WordPress core actually handles previews, this approach doesn't hold up as a general fix, so I'm closing it.

Core resolves a previewed featured image from the `_thumbnail_id` request parameter on the preview URL (`_wp_preview_post_thumbnail_filter`), and never persists it to the revision or autosave. The block editor sends the featured image as the `featured_media` REST field rather than as revisioned meta, so even the WordPress 6.4 meta revisions framework doesn't capture it (`create_post_autosave` only stores keys present in the request's `meta`). So there's no server side source for the previewed image in the default block editor, and reading the revision's own `_thumbnail_id` only works in narrow cases (classic editor or custom save flows), which made the regression test here somewhat synthetic.

The right fix is to let the client pass the preview context (the previewed thumbnail id, and similar) into the request, mirroring what core does with its request parameters, but expressed through GraphQL. I'm going to pursue that via GraphQL extensions, with preview treated as a request level concern rather than a per field one. More detail to follow.

#2664 stays open and will be addressed by that work. The revisioned meta side of this (footnotes and other keys registered with `revisions_enabled`) is separable and cleanly fixable, tracked in #3260.

PR #3972, opened 2026-06-21
chore: add OpenSSF Scorecard workflow

### LINKED ISSUE #3107: Open SSF Scorecard

My feeling is we do a reasonable job of security. That said I'm no expert and it would be good to have a tool to validate we're staying secure and help communicate those intentions and practices to our users. 

https://securityscorecards.dev/

OSSF scorecard checks your security practices and scores the project based on your use of security tools and best practices. Runs on a GitHub Action and the badge can be added to our readme or site as desired. Here's our current scorecard:

https://securityscorecards.dev/viewer/?uri=github.com/wp-graphql/wp-graphql

none.

### Additional Context

They mention a couple projects who use it: https://github.com/ossf/scorecard?tab=readme-ov-file#prominent-scorecard-users
The CISA seems to recommend it: https://www.cisa.gov/resources-tools/services/openssf-scorecard of

nathanpixodeo:
## Summary

- Fixes #3107
- Adds an OpenSSF Scorecard workflow that runs on pushes to `main` and weekly on Saturdays
- Publishes Scorecard results for the public API/badge flow
- Uploads SARIF results as both a short-lived workflow artifact and GitHub code scanning data
- Pins all actions in the new workflow to immutable SHAs with version comments

## Testing

- `python3` YAML parse check for `.github/workflows/scorecard-analysis.yml`
- `git diff --check -- .github/workflows/scorecard-analysis.yml`

`actionlint` is not installed in my local environment, so I could not run it locally.

### DIFF
--- a/.github/workflows/scorecard-analysis.yml
+++ b/.github/workflows/scorecard-analysis.yml
@@ -1,0 +1,43 @@
+name: Scorecard
+
+on:
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: '30 1 * * 6'
+
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecard analysis
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+      id-token: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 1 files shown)


jasonbahl (approved):

jasonbahl:
Thanks @nathanpixodeo! 🙏

PR #3996, opened 2026-06-25
fix: use the role slug when building UserRoleEnum descriptions

wj-yuta-imai:
## What

`UserRoleEnum::register_type()` switched on `$role` — the role *array* (`['name' => 'Administrator', 'capabilities' => [...]]`) — against string `case` labels like `'administrator'`. An array is never `==` a string, so every role fell through to `default` and all values shared the generic `"User role with specific capabilities"` description. The five role-specific descriptions were effectively dead code.

The role slug is in `$key` (used for the enum `value` and the safe-name fallback). Switching on `$key` restores the per-role descriptions.

## Test

Adds a regression test (`UserRoleEnumTest::testBuiltInRolesHaveRoleSpecificDescriptions`) that introspects `UserRoleEnum` and asserts `ADMINISTRATOR` / `SUBSCRIBER` get their role-specific descriptions. It fails before this change (both return the generic fallback) and passes after.


jasonbahl (approved):
Thanks for this fix @wj-yuta-imai, and for the clear writeup. 

The diagnosis is spot on: `wp_roles()->roles` is keyed by the role slug, so `$role` is the role array and `switch ( $role )` could never match the string cases, which meant every role fell through to the generic description and the five role-specific ones were effectively dead code. Switching on `$key` is the right fix, and it lines up with how `$key` is already used for the enum value.

Confirmed Locally.

### Before

Schema shows the fallback description for each role in the enum.

CleanShot 2026-06-26 at 13 48 48@2x


### After

Each role has the appropriate description for the role.

CleanShot 2026-06-26 at 13 49 59@2x

----

Since you mentioned you couldn't run wp-env locally, I gave it a spin: the new test fails before the change (both roles return the generic fallback) and passes after, and PHPStan is happy. Reverting just the one-line change reproduces the red, so it's a real regression guard. Testing it through introspection is exactly the layer I'd want this covered at.

----

Appreciate the AI-assisted disclosure as well! Approving. 🙏

jasonbahl (approved):

jasonbahl (approved):

jasonbahl (approved):

jasonbahl (approved):

PR #3997, opened 2026-06-25
fix: read use_featured_image for the PostType useFeaturedImage label

wj-yuta-imai:
## What

The `useFeaturedImage` resolver in `PostTypeLabelDetails` read `$labels->use_featured_item`, which does not exist in WordPress core — the post-type label property is `use_featured_image`. The `! empty()` guard therefore always failed and the field returned `null` for every post type. The sibling resolvers (`featuredImage`, `setFeaturedImage`, `removeFeaturedImage`) all use the correct property names; this was the lone typo.

## Test

`PostTypeObjectQueriesTest::testPostTypeQueryForPosts` previously asserted `useFeaturedImage => null` (encoding the bug). It now asserts the real label (`$post_type_object->labels->use_featured_image`), which fails before this change (resolver returned null) and passes after.


jasonbahl (approved):

jasonbahl:
Thanks for the PR! Appreciate the test update to go with it. 🙏

PR #3998, opened 2026-06-25
fix: keep deprecationReason for deprecated fields without a description

wj-yuta-imai:
## What

`InstrumentSchema::wrap_fields()` sanitizes each field's `description` and `deprecationReason`. The `deprecationReason` line guarded on `is_string( $field->description )` — the wrong property (copy-paste from the line above). When a field is deprecated but has **no description**, `is_string( null )` is `false`, so its `deprecationReason` was dropped to `null` in introspection (and `isDeprecated` became `false`). The fix guards on `$field->deprecationReason` itself. Fields that have a description were unaffected, which is why this slipped through.

## Test

Adds `InstrumentSchemaTest::testDeprecatedFieldWithoutDescriptionKeepsDeprecationReason`: registers a deprecated `RootQuery` field with no description and asserts, via introspection, that its `deprecationReason` is preserved. Fails before this change (reason dropped to null), passes after.

### DIFF
--- a/plugins/wp-graphql/src/Utils/InstrumentSchema.php
+++ b/plugins/wp-graphql/src/Utils/InstrumentSchema.php
@@ -83,7 +83,7 @@ class InstrumentSchema {
 			 * Sanitize the description and deprecation reason
 			 */
 			$field->description       = ! empty( $field->description ) && is_string( $field->description ) ? esc_html( $field->description ) : '';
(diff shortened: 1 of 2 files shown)


wj-yuta-imai:
Closing this — on a closer look it's a no-op. The line above normalizes `$field->description` to a string (`esc_html(...)` or `''`) before this line runs, so `is_string( $field->description )` is always true here and the `deprecationReason` is in fact preserved. There's no observable bug, so the guard change makes no difference in practice. Sorry for the noise!

Issue #2459, opened 2022-07-26
Options should be treated as Models / Nodes


jasonbahl:
Currently most data in WPGraphQL uses the DataLoader pattern and treats data as Models / Nodes. 

Options (data stored in `wp_options` table) do not currently follow this pattern.

I believe they should follow this pattern. 

Each Settings Group (ex: `general`, `discussion`, `reading`, `writing`, etc) could be considered a Node. The Node ID could be the name of the group. For example: `Relay::toGlobalId( 'option_group', 'general' );` 

Then, whenever an option is queried, a DataLoader should be used to load the Model. 

For example, a query like so: 

```graphql
{
  generalSettings {
    title
    url
  }
}
```

Should call the DataLoader for options, asking for the `Relay::toGlobalId( 'option_group', 'general' );`  node, which should return an instance of an `Options` Model, and determine which fields of said settings group should be allowed to be returned. Fields, such as `generalSettings.email` should be restricted by the Option model layer. 

### Centralized Actions / Filters

A lot of documentation in WPGraphQL points users to centralized hooks/filters, and options don't currently benefit from these actions/filters. 

For example, the filter [graphql_data_is_private](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Model/Model.php#L260) can be used to restrict an entire model. If someone wants to restrict a specific Post by ID, etc they can do this with this filter. Since options don't pass through this filter, users cannot use this filter to restrict specific options or option groups. Refactoring will allow central hooks/filters like this to be used with options like they're used with other data loaded by WPGraphQL. 

### WPGraphQL Smart Cache (Cache Mapping / Evictions)
By refactoring options to use the DataLoader / Model pattern, data will pass through central hooks in WPGraphQL that can be used by plugins such as WPGraphQL Smart Cache. 

Currently, the WPGraphQL Smart Cache plugin listens for data to be loaded by DataLoaders and tracks the Model that's used to return nodes, and stores a map of said nodes along with the query, then can identify when said Models/nodes are modified and can then evict the caches. 

Because Options don't currently use the Model/Node/Loader pattern, WPGraphQL Smart Cache is not currently properly able to track/evict when changes to options occur. 

## Breaking Changes?

I believe we should be able to do this refactor without making any breaking changes to the Schema.  

The difference should be how data is loaded, but the schema should remain the same. Some additions will be made to the Schema, such as adding an `id` field to all the setting groups and applying the `node` interface to each setting group.

[...]

PR #4021, opened 2026-06-30
feat: expose the Site Address as generalSettings.homeUrl

### LINKED ISSUE #2520: generalSettings returns "Wordpress Address" but not "Site Address"

In the general settings of wordpress you can set the URL to access your wordpress install from and the URL to access your site from.
Screen Shot 2022-09-17 at 14 05 24

Often these would be the same but not always and unfortunately wp-graphql only appears to pull in "WordPress Address" (simply labeled "url" in the general settings group) and not "Site Address":
Screen Shot 2022-09-17 at 14 04 56

I'm using wordpress as a headless CMS so my frontend has one domain and wordpress has another. In my frontend I've got a situation where it would be useful to know the authoritative domain name for the frontend. I could feed that in via an environment variable but that's one more thing to keep track of and it would be great to pull it from the wordpress settings like other values such as the site title are currently being pulled.

Ideally wp-graphql would offer two urls fields in the general settings query named something like `wordpressUrl` and `siteUrl`, instead of the `url`. But to avoid breaking existing sites I assume `url` can't be removed/renamed to `wordpressUrl`. But I imagine it shouldn't cause any issue to add a new field for the Site URL called `siteUrl` or something.

_No response_

### Additional Context

_No response_

thisismyurl:
WordPress core registers the `siteurl` option ("WordPress Address") for the REST API, and WPGraphQL surfaces it as `generalSettings.url`. Core never registers the `home` option ("Site Address"), so there is currently no way to read it from the schema. On headless and decoupled installs the Site Address is usually the canonical front-end URL, and it is often different from the WordPress Address, which is exactly the value a front end wants to resolve.

This registers `home` for GraphQL with `show_in_graphql`, hooked to `init_graphql_request` immediately after core's `register_initial_settings()`. Because it runs only on the GraphQL request lifecycle, it adds `generalSettings.home` (and `allSettings.generalSettingsHome`) without touching the REST API or the admin. The field name follows the existing convention: `home` has no `show_in_rest` name, so it derives from the option key, the same way `siteurl` becomes `url`.

To verify, set `home` and `siteurl` to different URLs, then query `generalSettings { url home }` and confirm each returns its own value. The added wpunit test does exactly that on single-site, and on multisite it asserts the field exists and resolves (where `update_option` does not round-trip through GraphQL, mirroring the existing `generalSettingsUrl` assertion in `testAllSettingsQuery`).

Closes #2520

(full disclosure: AI helped me identify the issue and verify my work)


jasonbahl:
@thisismyurl Thanks for finding a good first issue and opening a PR! 

I want to walk through one concern and recommend a change of approach (because there's some context here you'd have no way of knowing without being deeply engrained in this codebase 😆). 

The thing to watch out for: `register_setting` does more than add a read field. It modifies the global settings registry which WPGraphQL reads from, but other things such as the REST API, etc may read from as well.

WPGraphQL builds the settings schema dynamically from WordPress's global settings registry (which is modified in your PRs use or register_setting). Anything registered there with `show_in_graphql` gets picked up in three places at once: the `generalSettings.home` read field (what you want), the `allSettings.generalSettingsHome` read field, and a writable `generalSettingsHome` input on the `updateSettings` mutation. 

That last one is the catch.

It would make home editable over the API for anyone with `manage_options`. We actually consider the site URL dangerous enough to change remotely that there's a hardcoded guard blocking writes to `generalSettingsUrl`, and `home` is just as load-bearing, so this would open a door we deliberately keep closed elsewhere.

There's also a subtler side effect. Because you're hooking `init_graphql_request`, the `register_setting` call runs on every GraphQL request, which feels like it's limiting the modification of the WordPress settings registry to only graphql request, but, it also includes internal requests triggered in-process by the `graphql()` function (block render callbacks, WP-CLI, cron, an admin request that resolves a query inline). In those cases it adds home to WordPress's global settings registry for the rest of that PHP process, after the GraphQL request is done. So the registry's contents end up depending on whether a GraphQL request happened earlier in the same lifecycle, which could be a tricky class of bug to chase down later.

With all that context in mind, the approach I'd suggest instead: register it as a plain, read-only field.

The goal is to add a queryable field to the WPGraphQL schema without modifying the larger settings registry, and that's what `register_graphql_field` is for, and we already do exactly this in a couple of spots. Take a look at TypeRegistry.php (around line 587): there's a url field hand-registered on GeneralSettings as a polyfill (core doesn't register `siteurl` on multisite), and a `siteIconUrl` field right below it. `home` is the same shape, core never registers it, so we could attach it directly:

```php
$this->register_field(
    'GeneralSettings',
    'home',
    [
        'type'        => 'String',
        'description' => static function () {
            return __( 'Site Address (URL). The front-end address visitors use to reach the site, stored in the `home` option. This can differ from the WordPress Address (`url`) on headless or decoupled installs.', 'wp-graphql' );
        },
        'resolve'     => static function () {
            return get_home_url();
        },
    ]
);
```

Registering it right next to the existing `url`/`siteIconUrl` polyfills (rather than in a separate method on `init_graphql_request`) keeps it with its siblings. This approach touches only the schema: no global settings registry changes, no `updateSettings` write surface, so home is read-only by design with no special guard needed, and because the resolver uses `get_home_url()` it works correctly on multisite too, which sidesteps the round-trip caveat you ran into.

One deliberate scoping note: this lands `home` on `generalSettings` only, not in `allSettings`. But, that's intentional. We're moving away from the flat `allSettings.generalSettings*` shape (it's the one corner of the schema that encodes structure in field-name prefixes instead of types), so adding a new field there would be heading the wrong direction. If you're curious about where this is going, I just wrote up the broader plan for mapping settings into the schema as proper nodes here: https://github.com/wp-graphql/wp-graphql/issues/2459#issuecomment-4847592231. Your read-only field is a clean first step toward it, the field name and types stay identical when that lands.

On next steps, and this is your call entirely: 

Would you like to take a stab at this refactor yourself, pair on it with me, or would you rather I push the changes to your branch so it still lands under your name? 

All three are totally fine, I just don't want to assume. Whatever's the best experience for you. 🙏

One other thing we might want to consider. Do we want the field name to be `home`? or something more explicit like `siteAddress` or `siteUrl`?

jasonbahl (changes requested):
see: https://github.com/wp-graphql/wp-graphql/pull/4021#issuecomment-4848139368

thisismyurl:
Thanks for that thorough walk-through — the `updateSettings` mutation risk makes complete sense and I would have had no idea without your context on the hardcoded `generalSettingsUrl` guard.

Happy to take the refactor myself. Plan:

- Add the `home` field via `$this->register_field()` in `TypeRegistry.php`, right alongside the existing `url` / `siteIconUrl` polyfills (~line 587)
- Remove the `register_setting` approach entirely
- Keep the resolver as `get_home_url()` (multisite-correct, no option round-trip)

On the field name: `home` feels right to me — it mirrors the option key, parallels `url`, and `siteAddress` or `siteUrl` would clash with `url` semantically. If you feel differently I'll follow your lead.

I'll push the updated commit shortly.

thisismyurl:
Done — refactored per @jasonbahl's feedback.

**What changed:**

- Removed `add_action( 'init_graphql_request', [ self::class, 'register_extra_settings' ], 11 )` from `WPGraphQL.php`
- Removed the `register_extra_settings()` static method entirely (it called `register_setting()` with `show_in_graphql => true`, which wired `home` into the writable `updateSettings` mutation — exactly the side effect @jasonbahl flagged)
- Added `$this->register_field( 'GeneralSettings', 'home', [...] )` in `TypeRegistry.php` immediately after the `siteIconUrl` field registration, using `get_home_url()` as the resolver (multisite-aware)

**Result:** `generalSettings.home` is now a read-only `String` field. It does not appear in the `updateSettings` mutation. The resolver calls `get_home_url()` rather than reading the raw option, so it honours multisite URL overrides.

The existing test in `SettingsQueriesTest.php` queries the field at the GraphQL layer and remains valid as-is.

jasonbahl:
@thisismyurl Thanks again for the quick turnaround on the refactor, the read-only field approach landed exactly as hoped and CI came back green across the whole matrix.

To save you another round trip, I pushed a small follow-up commit to your branch (d7534fb) with a few tweaks:

- **Renamed the field from `home` to `homeUrl`.** I know I floated the naming question and you made a fair case for `home`, but after sitting with it I want the name to be self-describing to someone with no WordPress context, and `home: String` could read as a page reference or a flag. `homeUrl` says what it is, parallels `siteIconUrl` on the same type, and matches the field name a lot of headless folks already register by hand with `get_home_url()`, so for those sites this lands as a native replacement (a duplicate registration degrades to a debug notice, not an error).
- **Made the description backend-agnostic.** Our convention is that schema descriptions never reference option keys or other WordPress internals, the schema should read the same whether the backend is WordPress or anything else. The old wording ("stored in the `home` option") came from my own suggested snippet, so that one's on me.
- **Removed two stray blank lines** left behind in `WPGraphQL.php` when the hook and method were removed, so that file is now untouched by the PR.
- **Extended your test** with an introspection assertion that `UpdateSettingsInput` has no corresponding input field, so a future change can't quietly reintroduce the write path we worked to keep closed.

One other thing I looked into while reviewing: whether this value needs a capability check. It doesn't. WordPress already exposes `home` to unauthenticated visitors in the REST API index (`/wp-json/`), and it appears in the markup of every front-end page, so it's public by design. That also matches how the other settings fields behave, only `admin_email` is gated.

I also updated the PR title to reflect the new field name. Once CI is green on the new commit this should be good to merge. Thanks for contributing! 🙌

jasonbahl (approved):

jasonbahl (approved):

Issue #3931, opened 2026-06-12
🎯 Tracking: mapping WordPress data registries (settings, meta, options) to the WPGraphQL Schema


jasonbahl:
WordPress has several registries that describe data which wants representation in the WPGraphQL schema, and today each is handled differently (or not at all):

- **`register_setting()` (site settings)**: auto-mapped as flat scalars from `get_registered_settings()`. No richer typing, no way to express "this int is actually a Category" (#1887), no models or access control beyond the group level. Tracked in #2459 (settings as Models / Nodes).
- **User options / preferences**: hand-added fields (#3571 shipped `adminColor`, `hasRichEditingEnabled`, etc. in v2.9.0). A centralized registry for mapping them is the subject of RFC #3562.
- **`register_meta()` (post, term, user, comment meta)**: no bridge at all. The supported paths are `register_graphql_field()` or WPGraphQL for ACF. The long-standing ask is #118, which deserves its own RFC.
- **Revision meta**: the block editor now revises meta, which changes assumptions for any meta mapping. Tracked in #3260.

These are distinct enough that they should be designed, implemented, and tested independently, but they share design questions that should be answered consistently, which is what this tracker is for:

- The opt-in convention (`show_in_graphql` args vs piggybacking `show_in_rest`)
- Mapping stored scalars to richer GraphQL types (enums, object types, connections, e.g. `defaultCategory` as a `Category` rather than an `Int`)
- Naming conventions (declarative field names like `hasCommentShortcutsEnabled`, per the #3571 discussion)
- Model layer involvement and access control (who can see which setting/meta)
- Mutation support and input typing
- How revisions interact with mapped meta (#3260)

## Children

- [ ] #2459, settings as Models / Nodes (`register_setting`)
- [ ] #3562, centralized registry for user options (RFC)
- [ ] #118, `register_meta()` to schema (needs an RFC)
- [ ] #3260, revision meta behavior

Each child should produce its own RFC or implementation plan; consistency across them gets discussed here.

jasonbahl:
Design note from the issue audit (surfaced via #2060): **settings value resolution and access control should be declarative per-setting, not inline special-cases in the settings group resolver.**

[...]

PR #4077, opened 2026-07-15
docs(settings): document graphql_normalized_settings filter and per-setting config keys

jasonbahl:
Documents the new settings-mapping surface for developers, extending the existing [Settings guide](../plugins/wp-graphql/docs/settings.md).

## Adds
- **Exposing Options WordPress Doesn't Register** — the `graphql_normalized_settings` filter for seeding options core never registers (in memory, without `register_setting()` and its global-state side effects). Includes the clarification that entries are keyed by the **globally-unique option name** (not the group), that `group` is metadata routing the field into its setting-group type, and that a filter entry with an existing option name **overrides** that entry.
- **Per-Setting Configuration** — a `Field | Type | Required | Description` table for `graphql_readonly` and `graphql_resolve`, matching the style of the Custom Post Types doc.

Only shipped keys are documented; the planned config keys (`graphql_field_name`, `graphql_capability`, `graphql_description`) will be added as each slice ships.

The `graphql_normalized_settings` filter reference page is generated from its call-site docblock at release; this PR is the narrative guide.

## Sequencing
Documents behavior added in **#4076** (the per-entry config code). Merge after #4076 lands. Refs #2459, #3931.

### DIFF
--- a/plugins/wp-graphql/docs/settings.md
+++ b/plugins/wp-graphql/docs/settings.md
@@ -72,6 +72,40 @@ This could now be queried like so:
 
 > \*\*NOTE: \*\*If a setting is registered without a group defined it will appear under `generalSettings`.
 
+## Exposing Options WordPress Doesn't Register
+
+Some options are never registered through `register_setting()`, for example values a plugin saved with `update_option()`, or core options that WordPress simply doesn't register. WPGraphQL can't see those by default.
+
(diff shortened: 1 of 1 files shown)


jasonbahl:
Folded into #4076 so the docs ship in the same PR as the feature code. Closing this one.

PR #4079, opened 2026-07-15
fix: return null for restricted setting reads

### LINKED ISSUE #4072: admin_email setting throws an authorization error instead of returning null for unauthorized users
When an unauthenticated or under-privileged user queries `generalSettings { adminEmail }` or `allSettings { generalSettingsAdminEmail }`, WPGraphQL throws a `UserError` ("Sorry, you do not have permission to view this setting"). That error surfaces in the top-level `errors` array and, because of null propagation, can knock out sibling fields the user was allowed to see.

This is inconsistent with how the rest of WPGraphQL handles restricted data. Everything that goes through the Model layer returns `null` for data the current user can't access (via `graphql_data_is_private`), it doesn't throw. The `admin_email` check predates that layer and is the one hardcoded exception.

It's also out of step with the wider GraphQL convention. Field-level authorization is generally handled by returning null and hiding the field, not by raising an error (throwing is reserved for operation-level auth, like a mutation the user isn't allowed to run). Our `updateSettings` capability gate should keep throwing; the field read should not.

## Proposed change

- `adminEmail` returns `null` for users without `manage_options`, instead of throwing.
- When it's nulled for capability reasons, emit a `graphql_debug` message naming the field and the required capability, so the reason is discoverable in Debug Mode rather than silent.
- The mutation-side capability check is unchanged.

This is the field-level access pattern the Settings as Models / Nodes work (#2459, tracked in #3931) is standardizing as declarative per-setting config, so this can either land as a small standalone change or as part of that effort.

## Backward compatibility

This is a minor behavior change. Clients that today rely on catching the authorization error for `adminEmail` would instead receive `null`. Given it aligns the field with WPGraphQL's own default behavior for restricted data, that seems like the right direction, but flagging it explicitly.

## References

- Field-level auth convention: [GraphQL.js Authorization Strategies](https://www.graphql-js.org/docs/authorization-strategies/), [graphql.org Authorization](https://graphql.org/learn/authorization/)
- Errors as data vs errors: [Marc-André Giroux, A Guide to GraphQL Errors](https://productionreadygraphql.com/2020-08-01-guide-to-graphql-errors/)

w3lld1:

Restricted `admin_email` reads currently throw a field error for users without `manage_options`. That error reaches the top-level `errors` array and can null out sibling settings that are safe to expose.

I changed the grouped, flat, and experimental email-setting resolvers to return `null` for restricted reads instead. In debug mode, I also record the affected GraphQL field and the required `manage_options` capability so the reason remains discoverable.

Fixes #4072

## Testing Strategy

I added regression coverage for both settings surfaces and the experimental `adminEmail` field. The tests verify that restricted fields resolve to `null`, public sibling fields still resolve, no field error is emitted, and debug entries identify the field and required capability.

### Test Results

- [x] Full WPUnit suite: 1,097 tests, 7,010 assertions; exit 0 (4 skipped, 7 incomplete)
- [x] Focused settings and email experiment suites: 45 tests, 158 assertions
- [x] PHP CodeSniffer
- [x] PHPStan on the changed source files
- [x] `git diff --check`

### Before

```graphql
query {
  generalSettings {
    email
    title
  }
}
```

The restricted email field adds a top-level authorization error and can null out allowed siblings.

### After

I return `null` for the restricted email field while `title` continues to resolve normally. When GraphQL debug logging is enabled, the debug entry names the field and the `manage_options` capability.

### DIFF
--- a/plugins/wp-graphql/src/Experimental/Experiment/EmailAddressScalarFieldsExperiment/EmailAddressScalarFieldsExperiment.php
+++ b/plugins/wp-graphql/src/Experimental/Experiment/EmailAddressScalarFieldsExperiment/EmailAddressScalarFieldsExperiment.php
@@ -118,7 +118,16 @@ class EmailAddressScalarFieldsExperiment extends AbstractExperiment {
 				},
 				'resolve'     => static function ( $root, $args, $context, $info ) {
 					if ( ! current_user_can( 'manage_options' ) ) {
-						throw new UserError( esc_html__( 'Sorry, you do not have permission to view this setting.', 'wp-graphql' ) );
+						graphql_debug(
+							__( 'The "GeneralSettings.adminEmail" field requires the "manage_options" capability and resolved to null.', 'wp-graphql' ),
+							[
+								'type'                => 'RESTRICTED_SETTING',
+								'field'               => 'GeneralSettings.adminEmail',
+								'required_capability' => 'manage_options',
+							]
+						);
+
+						return null;
 					}
 
 					return get_option( 'admin_email' );
@@ -225,7 +234,16 @@ class EmailAddressScalarFieldsExperiment extends AbstractExperiment {
 
 				// Fallback resolver (same logic as original)
 				if ( ! current_user_can( 'manage_options' ) ) {
-					throw new UserError( esc_html__( 'Sorry, you do not have permission to view this setting.', 'wp-graphql' ) );
+					graphql_debug(
+						__( 'The "GeneralSettings.email" field requires the "manage_options" capability and resolved to null.', 'wp-graphql' ),
+						[
+							'type'                => 'RESTRICTED_SETTING',
+							'field'               => 'GeneralSettings.email',
+							'required_capability' => 'manage_options',
+						]
+					);
+
+					return null;
 				}
 
 				return get_option( 'admin_email' );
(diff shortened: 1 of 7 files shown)


jasonbahl:
Hey @w3lld1, thanks so much for this. 

This is exactly what we were after: returning null for restricted reads instead of throwing, so a restricted field doesn't null out the sibling settings that are safe to expose, while the reason stays discoverable in debug mode. The structured debug payload (field name + required capability) is a nice touch.

One heads up: we just merged #4078, which reworks how settings field names are derived. They're now computed once in the normalized settings map and read by each resolver, instead of every surface re-deriving them. That overlapped with the resolvers you touched here, so I rebased your branch onto the latest main to sort it out. Good news is it merged cleanly, your logic is fully intact, and the `$field_key` your debug message uses now flows from that shared derivation, so the field names stay consistent with the rest of the schema. I re-ran the settings suites plus the email-experiment tests, PHPStan, and PHPCS locally and everything is green.

I force-pushed the rebase to your branch (thanks for leaving "allow edits from maintainers" enabled). If you have any local work in progress, you will want to reset to the updated branch before continuing.

A bit of context on where this fits: No action needed from you here, but we are working through a series of slices that map settings more fully into the schema. A later slice adds a declarative per-setting capability config, at which point the `admin_email` / `manage_options` check becomes a single line of config and the resolver applies this exact null + debug behavior generically. Your PR is the behavior we want, so it slots right in.

Thanks again, really appreciate the contribution. 🙌

jasonbahl (dismissed):

jasonbahl (approved):

Issue #3180, opened 2024-08-02
No menu set for the provided location


MKlblangenois:
### Description

When I try to get `menu` node using it's location, I've an error in WPGraphQL `No menu set for the provided location`, and menu return `null`
![image](https://github.com/user-attachments/assets/5443ca7e-75e3-49e0-b3f3-12fd46d60ea4)

### Steps to reproduce

1. Register a menu position
2. Create a new menu and assign it to the position
3. Add item in the menu
4. Go to GraphiQL and make the query to get the menu by it's location
```graphql
query GetMenuByLocation($location: ID!) {
  menu(id: $location, idType: LOCATION) {
    id
    databaseId
  }
}
```

### Additional context

Here's my menu position:
![image](https://github.com/user-attachments/assets/dc47b651-f8eb-4f06-a7a7-f22781253ca4)

And my GraphQL result if I query all menus with locations:
![image](https://github.com/user-attachments/assets/abd8c9e7-061e-4ca0-a28b-e39d16e144ad)

### WPGraphQL Version

1.27.2

### WordPress Version

6.6.1

### PHP Version

8.2

### Please confirm that you have searched existing issues in the repo.

- [X] Yes

### Please confirm that you have disabled ALL plugins except for WPGraphQL.

- [X] Yes
- [ ] My issue is with compatibility with a specific WordPress plugin, and I have listed all my installed plugins (and version info) above.

josephfusco:
Hey @MKlblangenois, how are the menus registered? Can you provide us with the code you used to register the menu location?

MKlblangenois:
Hi @josephfusco! I've a JSON with `key => value` with `position => name` and I use it like that:

```php
<?php

// JSON Exemple:
// {
// 	"menus": {
// 		"primary": "Menu principal",
// 		"topnav": "Menu haut de page"
//              ...
// 	},
// 	...
// }

class NextPressMenu extends NextPress {
	public function __construct() {
		$this->setup();

		if (!isset($this->theme_settings) || !array_key_exists('menus', $this->theme_settings)) {
			return;
		}

		$this->register_activation_hook('after_setup_theme', $this->register_theme_menus());
	}

	public function register_theme_menus() {
		foreach ($this->theme_settings['menus'] as $slug => $label) {
			register_nav_menu($slug, __($label, $this->theme_textdomain));
		}

		return true;
	}
}
```

If I can add details, querying `menuItems` by location is working fine:
[...]

PR #4083, opened 2026-07-17
Solution (#3180): Accept MenuLocationEnum value for `menu(idType: LOCATION)`

TFGSUMIT:
```markdown
## Pull Request: Accept MenuLocationEnum value for `menu(idType: LOCATION)`

This PR addresses the issue of accepting `MenuLocationEnum` values for the `menu(idType: LOCATION)` query by implementing the necessary changes in the `Menu` type and the `MenuQuery` class.

**Changes:**

* Created a new `Menu` type with fields for menu location and its corresponding enum value.
* Updated the `MenuQuery` class to handle the `idType` argument and accept `MenuLocationEnum` values.

**Testing Instructions:**

1. Run the `wp-graphql` tests to ensure the changes did not introduce any regressions.
2. Verify that the `menu(idType: LOCATION)` query correctly handles `MenuLocationEnum` values.

**Reviewers:** @wp-graphql-reviewers

**Assignee:** @your-name
```

### DIFF
--- a/path/to/wp-graphql/src/Query/Menu.php
+++ b/path/to/wp-graphql/src/Query/Menu.php
@@ -1,0 +1,213 @@
+# complete code
+namespace WPGraphQL\Query;
+
+use WPGraphQL\Types\WPObject;
+use WPGraphQL\Type\WPObject\Menu;
+
+class MenuQuery extends WPObject
+{
+    /**
+     * @var string
+     */
+    public $id;
+
+    /**
+     * @var int
+     */
+    public $databaseId;
+
+    /**
+     * @var string
+     */
+    public $name;
+
+    /**
+     * @var string
+     */
+    public $slug;
+
+    /**
+     * @var string
+     */
+    public $description;
+
+    /**
+     * @var string
+     */
+    public $type;
+
+    /**
+     * @var string
+     */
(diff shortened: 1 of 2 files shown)


jasonbahl:
Thanks for taking an interest in this issue @TFGSUMIT.

I'm going to close this PR though, as it doesn't appear to have been tested or run against the codebase. The files are committed under a placeholder path (`path/to/wp-graphql/...`), they're missing `<?php` opening tags, each class declares its properties twice (which is a fatal error in PHP), and they reference classes and functions that don't exist in WPGraphQL or WordPress (`WPGraphQL\Types\WPObject`, `wp_get_menu_object()`). Nothing in the change connects to the code that actually resolves `menu(idType: LOCATION)`, which lives in `plugins/wp-graphql/src/Type/ObjectType/RootQuery.php`, so this wouldn't change the behavior described in #3180.

If you'd like to take another run at it, a fix for #3180 would modify the `location` case in that resolver to also accept `MenuLocationEnum` values, and include a test (see `plugins/wp-graphql/tests/wpunit/` for examples) that fails before the change and passes after. If AI tools are part of your workflow that's fine, but please make sure the code runs and the tests pass before opening a PR. Happy to review a working version.

PR #4161, opened 2026-08-04
fix: remove WordPress internals from schema field descriptions

### LINKED ISSUE #4026: GraphQL schema descriptions leak WordPress implementation details (WP_Post properties, database columns, meta keys)
The WPGraphQL schema should be self-describing and backend-agnostic: a developer introspecting it should be able to understand it with no prior WordPress knowledge. The backend could be WordPress, Supabase, or a spreadsheet, and the schema shouldn't assume the reader knows which. This is the same instinct behind naming a `ContentNode` rather than a "post" and a `ContentType` rather than a "post type." Several field, argument, and type `description` strings currently fall short, either leaking backend plumbing or leaning on WordPress-specific vocabulary that assumes prior knowledge.

## Two levels to audit

**1. Implementation plumbing (hard rule, never in descriptions).** Reword any description that leaks:

- WordPress class properties (`WP_Post->guid`, `WP_Term->…`)
- database table / column names (the "post_objects" table, the `post_mime_type` column)
- post meta keys (`_wp_page_template`, `_thumbnail_id`)
- WP function names (`wp_*`, `get_post_meta`, `WP_Query`) and query-var names

A recurring templated phrasing names all three at once:

- `src/Type/InterfaceType/ContentNode.php` (`guid`): "The global unique identifier for this post. This currently matches the value stored in `WP_Post->guid` and the guid column in the "post_objects" database table."
- `src/Mutation/PostObjectCreate.php` (`mimeType`): "...equivalent to the value of `WP_Post->post_mime_type` and the post_mime_type column in the "post_objects" database table."

This `WP_Post->X` / "database table" pattern appears across roughly 7 files (`User.php`, `Menu.php`, `Comment.php`, `ContentNode.php`, `PostObjectCreate.php`, and siblings).

**2. WordPress-specific vocabulary (prefer declarative language).** Where a description leans on a WordPress term that assumes prior knowledge, prefer the abstracted / domain phrasing the schema already uses elsewhere, describe the *thing* and what it does rather than its WordPress name. Reach for a WordPress-specific term only when there's genuinely no clearer general word, and even then describe what it is.

## Out of scope

- Type/field **renaming** is not part of this pass, this is descriptions only. (Naming already follows the convention, e.g. `ContentNode`; renames are breaking and out of scope here.)
- Code comments and developer-facing deprecation messages may still reference internals, this is about user-facing `description` strings only.

Core (`plugins/wp-graphql`) is the priority; a follow-up pass over `wp-graphql-acf` and `wp-graphql-smart-cache` descriptions is a stretch goal.

## Notes

- Convention: see `CLAUDE.md` (Shared Coding Conventions) / PR #4027.
- Precedent: PR #4025 reworded the `template` where-arg description from "matched against the `_wp_page_template` value" to "matched by the template file name or slug (e.g. `template-full-width.php` or `page-no-title`)."
- Mostly mechanical, low risk, good first issue. Each reworded description is covered by existing schema/snapshot tests; no behavior changes.

AnkithaBeeram:

Some GraphQL schema descriptions leaked WordPress internals — PHP class
properties (`WP_Post->guid`), database table names, and column names —
instead of describing what the field means to someone using the API.

This PR rewrites 26 descriptions across 7 files to remove those leaks,
following the style used in #4025. Only description text changes —
no behavior, types, or resolvers are touched.

closes #4026

## Testing Strategy

Added `tests/wpunit/SchemaDescriptionLeaksTest.php`, which checks the
`guid`/`slug` descriptions directly and also sweeps the whole schema
for the same leak pattern. Verified it fails against the old code and
passes with the fix.

### Test Results
- [x] Failing test added (verified against reverted source, not a separate commit)
- [x] Fix implemented, test passes
- [x] All tests passing (1098 tests, 0 failures)
- [x] Regression test covers the whole schema, not just the two named fields

`composer check-cs`: 0 violations.

### Test Links (if applicable)

N/A — verified locally.

### Before:
```json
"description": "...matches the value stored in WP_Post->guid and the guid column in the \"post_objects\" database table."
```

### After:
```json
"description": "A globally unique identifier for this post, primarily used internally by WordPress and in RSS feeds."
```

## Additional Context

Scoped to removing internal leaks only. A broader pass on WordPress-specific
wording could be a follow-up. No breaking changes.


plugins/wp-graphql/tests/wpunit/SchemaDescriptionLeaksTest.php
```diff
@@ -0,0 +1,144 @@
...
+		// property accessors and references to raw storage (tables/columns).
+		$leak_markers = [
+			'WP_Post->',
+			'WP_Term->',
+			'WP_User->',
+			'WP_Comment->',
+			'database table',
+			'column in SQL',
+			'post_objects',
+		];
```
jasonbahl, line 111: I did an inventory of `WP_*->` references inside description strings, and there are 12 remaining that this PR does not touch:

- `src/Type/ObjectType/Theme.php` has 9 descriptions referencing `WP_Theme->get(...)` / `WP_Theme->get_stylesheet()`
- `src/Type/ObjectType/Taxonomy.php` has 3 referencing `WP_Taxonomy->label`, `->description`, and `->show_tagcloud`

We should address those descriptions in this pass as well. 🙏

For the marker list, rather than enumerating classes (WordPress has 300+ `WP_*` classes and adds more every release), replace the four class-specific markers with the single marker `'WP_'`. Nothing in our conventions allows a `WP_` reference of any kind in a schema description, so the bare prefix catches every class, current and future, with the same `strpos` check. The whole array becomes:

```php
		$leak_markers = [
			'WP_',
			'database table',
			'column in SQL',
			'post_objects',
		];
```

plugins/wp-graphql/tests/wpunit/SchemaDescriptionLeaksTest.php
```diff
@@ -0,0 +1,144 @@
...
+
+	/**
+	 * Enable public introspection before each test and clear the cached schema.
+	 */
+	public function setUp(): void {
+		parent::setUp();
+
+		// Public introspection is required so the same query used to reproduce the
+		// issue (as an unauthenticated consumer would run it) resolves descriptions.
+		$settings                                 = get_option( 'graphql_general_settings' );
```
jasonbahl, line 19:
```suggestion
		$settings = get_option( 'graphql_general_settings', [] );
		if ( ! is_array( $settings ) ) {
			$settings = [];
		}
```

In a fresh test database `graphql_general_settings` option usually does not exist, so `get_option()` returns `false` and the next line is writing an array offset on `false`, which raises a deprecation on PHP 8.1+ and our CI matrix runs up through 8.4.

We should default to an empty array here and make sure we are writing back to an array.

plugins/wp-graphql/src/Type/InterfaceType/ContentNode.php
```diff
@@ -147,7 +147,7 @@ public static function register_type( TypeRegistry $type_registry ) {
 						'slug'                      => [
 							'type'        => 'String',
 							'description' => static function () {
-								return __( 'The uri slug for the post. This is equivalent to the WP_Post->post_name field and the post_name column in the database for the "post_objects" table.', 'wp-graphql' );
+								return __( 'The url friendly identifier for the content node. This field is equivalent to the human-readable slug used to reference the node in its permalink.', 'wp-graphql' );
```
jasonbahl, line 150: This is a bit circular, as it defines the "slug" as the human-readable "slug". Maybe something like:

```suggestion
								return __( 'The URL-friendly, human-readable identifier for the content node, used in its permalink.', 'wp-graphql' );
```

plugins/wp-graphql/src/Mutation/PostObjectCreate.php
```diff
@@ -168,7 +168,7 @@ public static function get_input_fields( $post_type_object ) {
 			$fields['mimeType'] = [
 				'type'        => 'MimeTypeEnum',
 				'description' => static function () {
-					return __( 'If the post is an attachment or a media file, this field will carry the corresponding MIME type. This field is equivalent to the value of WP_Post->post_mime_type and the post_mime_type column in the "post_objects" database table.', 'wp-graphql' );
+					return __( 'If the content node is a media item, this field carries the corresponding MIME type of the underlying file (for example, image/jpeg or application/pdf).', 'wp-graphql' );
```
jasonbahl, line 171: Since this input is only registered when the post type is attachment, the "If the content node is a media item" isn't needed here.

```suggestion
					return __( 'The MIME type of the file for the media item (for example, image/jpeg or application/pdf).', 'wp-graphql' );
```

plugins/wp-graphql/tests/wpunit/SchemaDescriptionLeaksTest.php
```diff
@@ -0,0 +1,144 @@
...
+			$this->assertStringNotContainsString( 'WP_Post', $description, "ContentNode.$field_name description leaks a WP_Post reference." );
+			$this->assertStringNotContainsString( 'database table', $description, "ContentNode.$field_name description leaks a database table reference." );
+			$this->assertNotEmpty( $description, "ContentNode.$field_name should still have a description." );
+		}
+	}
+
+	/**
+	 * Schema-wide guard: walk every type's field and input-field descriptions as an
+	 * introspection consumer sees them, and assert none of them leak WordPress
+	 * internals. This catches the whole class of regression, not just the two fields
```
jasonbahl, line 73: > This catches the whole class of regression

Not entirely true as we're only checking field and input descriptions, and not Enum Value or Type Descriptions. I'm fine with either addressing those as well in this PR to make this claim accurate, or we can soften this claim and circle back on those in a follow-up.

jasonbahl (changes requested):
Hey @AnkithaBeeram, thanks for taking this on! This is exactly the direction #4026 is asking for, and I appreciate that you included a schema-wide regression test instead of just spot-checking the two fields named in the issue. Bonus points for catching the copy/paste bug where the deprecated id field on terms claimed to match WP_Post->ID.

I left some specific feedback in a few places. Let me know if you're able to follow-up on those or if you'd like me to help get this wrapped up.

Thanks again, this is close, and the remaining changes I called out are all in the same spirit as what you've already done. 🙏🙌

AnkithaBeeram:
Hey @jasonbahl, Thank you for the detailed review! I'll work through all the feedback given and push updates soon. Will get back to you if I need more clarification/support on any specific parts. Thanks again :)

jasonbahl:
@AnkithaBeeram wanted to follow up and see if you were interested in pushing changes? Otherwise I can make the suggested changes and get this merged. Let me know! Appreciate your contributions! 🙌

plugins/wp-graphql/tests/wpunit/SchemaDescriptionLeaksTest.php
AnkithaBeeram: Done in 95435b10 (fixes) and cec73c6c (broadened the marker to bare 'WP_' as suggested, so it now catches any WP_* class reference going forward).

While I was in Theme.php, I noticed the `tags` field's description read "URI for the author/company website" (looks like a copy-paste error from the `authorUri` field above it). I traced how the field resolves (Model/Theme.php → WP_Theme->get('Tags') )and rewrote it to:

"A list of tags associated with the theme, typically describing its features."

Let me know if that's accurate/clear, or if you'd phrase it differently. Happy to change it accordingly!

plugins/wp-graphql/tests/wpunit/SchemaDescriptionLeaksTest.php
AnkithaBeeram: You are right! Went with softening the claim for now (in 222c0d0a) to keep this round focused. Happy to open a follow-up PR extending coverage if useful.

AnkithaBeeram:
@jasonbahl  Sorry for the delay! I just pushed all the changes. Please let me know if anything needs adjusting. Happy to work on it!

jasonbahl:
Thanks @AnkithaBeeram! 🙌

PR #4169, opened 2026-08-06
fix(wp-graphql-ide): rank Docs explorer search results by relevance

### LINKED ISSUE #4108: Docs panel search ranks results alphabetically, so exact and prefix matches get buried
Searching the schema in the Docs panel matches on substring and then sorts the results alphabetically, so the relevance of a match plays no part in where it lands.

Searching for `Post` puts the `Post` type fourteenth in the list, below `CategoryToPostConnection`, `CategoryToPostConnectionEdge`, `CategoryToPostConnectionPageInfo`, `CategoryToPostConnectionWhereArgs`, `ContentTypesOfPostFormatEnum`, and every `CreatePost*` / `DeletePost*` mutation input and payload. The type the user almost certainly wanted is the hardest one to find.

There's a second, sharper version of the same problem: results are capped (25 types, 50 fields) and the cap is applied after the alphabetical sort, so on a schema this size an exact match can be cut from the list entirely. A search whose match sorts late in the alphabet returns everything except the thing being searched for.

Current behavior lives in `plugins/wp-graphql-ide/src/components/DocsExplorerPanel.jsx` (the `typeMatches` / `fieldMatches` memo).

Options worth exploring, roughly in order of effort:

- Score matches instead of sorting by name. A simple tiered score gets most of the way there: exact match (case-insensitive), then prefix match, then word-boundary match on a camel-case segment, then plain substring, with name length as a tie-breaker so shorter names win. Sort by score, then alphabetically inside each tier.
- Apply the cap after ranking, so the best matches always survive truncation, and consider a "show more" affordance rather than silently dropping results.
- Weight by type kind or origin. Object types and interfaces a user actually queries could rank above generated connection, edge, input, and payload wrappers, which are numerous and rarely the target of a search.
- Highlight the matched substring in the result rows, so a user can see *why* something matched and scan the list faster.
- Consider fuzzy matching (something like a small scorer, or a dependency such as fuse.js) for typo tolerance, weighed against bundle size. Probably a follow-up rather than part of a first pass.
- Do the same for field results, where `Type.field` string sorting has the identical problem.

Worth deciding as part of this: whether types and fields stay in separate groups (they're easier to reason about) or merge into one relevance-ordered list.

joenermunch:

Docs Explorer search matched names by substring and then sorted them alphabetically. Searching for `Post` put `CategoryToPostConnection` first and the exact `Post` type well down the list. Because results are limited to 25 types and 50 fields, an exact match could also disappear entirely when enough earlier names matched.

This changes both type and field results to rank by relevance:

1. Exact match (case-insensitive)
2. Prefix match
3. GraphQL name-segment match, including camel-case and underscore boundaries
4. Plain substring match

Shorter names win within a tier, followed by alphabetical order. The existing result limits remain unchanged and are applied after ranking.

Closes #4108.

## Testing Strategy

The regression tests were written first. Before the implementation, the new Jest cases failed on type ordering, field ordering, and capped results; the new Playwright case received `CategoryToPostConnection` where it expected `Post`.

### Test Results

- [x] **Failing test added**: The first commit covers the reported behavior before changing the implementation.
- [x] **Fix implemented**: A shared comparator now ranks type and field names before slicing the result lists.
- [x] **All tests passing**: The full IDE unit suite passes (44 suites, 448 tests).
- [x] **Regression tests**: Jest covers each relevance tier and both result caps; Playwright covers the live `Post` search.

### Test Links (if applicable)

Local verification:

- JavaScript lint passed.
- Production build passed.
- Complete Docs Explorer browser spec passed (9 tests).

### Before (Buggy Behavior)

Searching for `Post` showed `CategoryToPostConnection` as the first type result. Exact matches that sorted late could be removed by the result cap.

### After (Fixed Behavior)

The same search shows `Post` first, followed by prefix, GraphQL name-segment, and substring matches. Types and fields remain in their existing separate groups.

## Additional Context

This changes result ordering only. It adds no dependency and does not alter the GraphQL schema, public APIs, or result limits. There are no breaking changes.

Fuzzy matching, match highlighting, type-kind weighting, and a show-more affordance remain possible follow-ups outside this focused fix.

### DIFF
--- a/plugins/wp-graphql-ide/src/components/DocsExplorerPanel.jsx
+++ b/plugins/wp-graphql-ide/src/components/DocsExplorerPanel.jsx
@@ -91,6 +91,58 @@ function getTypeKind(type) {
 }
 
 /**
+ * Rank a schema name against a search term.
+ *
+ * @param {string} name  GraphQL type or field name.
+ * @param {string} query Lowercase search term.
+ * @return {number} Relevance rank, where lower is better.
+ */
+function getSearchRank(name, query) {
+	const normalizedName = name.toLowerCase();
+	if (normalizedName === query) {
+		return 0;
+	}
+	if (normalizedName.startsWith(query)) {
+		return 1;
+	}
+
+	let index = normalizedName.indexOf(query);
+	while (index !== -1) {
+		const previous = name[index - 1] || '';
+		const current = name[index] || '';
+		const next = name[index + 1] || '';
+		if (
+			/[^A-Za-z0-9]/.test(previous) ||
+			(/[a-z0-9]/.test(previous) && /[A-Z]/.test(current)) ||
+			(/[A-Z]/.test(previous) &&
+				/[A-Z]/.test(current) &&
+				/[a-z]/.test(next))
+		) {
+			return 2;
+		}
+		index = normalizedName.indexOf(query, index + 1);
+	}
+
+	return 3;
+}
+
+/**
+ * Compare schema names by relevance, then length, then alphabetically.
+ *
+ * @param {string} a     First GraphQL name.
+ * @param {string} b     Second GraphQL name.
+ * @param {string} query Lowercase search term.
+ * @return {number} Sort comparison value.
+ */
+function compareSearchNames(a, b, query) {
+	return (
+		getSearchRank(a, query) - getSearchRank(b, query) ||
+		a.length - b.length ||
+		a.localeCompare(b)
+	);
+}
+
+/**
  * Docs Explorer panel content.
  *
  * Provides a browsable tree of the GraphQL schema — types, fields,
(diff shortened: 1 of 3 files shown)


jasonbahl:
@joenermunch Thanks for opening this PR! Will review shortly 🙏

jasonbahl:
@joenermunch This is a really well done PR, thank you! 🙏

The tiered ranking matches what was outlined in #4108. 

I appreciate that the regression tests came first. 🙌

I reverted the implementation locally and confirmed all three new Jest tests fail against the old sort, and the full IDE suite passes with the fix in place.

I also tested this in WordPress Playground and searching Post finally puts Post at the top where it belongs. 💪🥳

NOTE: The failing CI in the integration test suite is _unrelated_ to your changes (they're currently allowed failures related to WordPress trunk).

jasonbahl (approved):

PR #4321, opened 2026-09-13
test: add unit test coverage for Utils helper methods

noruzzamans:

This PR adds comprehensive unit test coverage for helper methods in `WPGraphQL\Utils\Utils` (`plugins/wp-graphql/tests/wpunit/UtilsTest.php`), bringing complete test coverage to the utility class:

1. **`format_type_name_for_wp_template()`**:
   - Tests standard template names (e.g., `'Full Width'` -> `'Template_FullWidth'`).
   - Tests template names that already contain the word "template" (ensuring no redundant `Template_` prefix is added).
   - Tests template names starting with a numeric digit (ensuring `Template_` prefix is added).
   - Tests non-ASCII template names with fallback to the template file name.
   - Tests empty string handling.

2. **`format_graphql_name()`**:
   - Tests standard alphanumeric naming.
   - Tests replacement of invalid characters using default and custom replacement characters.
   - Tests collapsing consecutive leading underscores into a single leading underscore (per GraphQL introspection naming rules).
   - Tests empty string handling.
   - Tests the `graphql_pre_format_name` filter hook.

3. **`format_field_name()` & `format_type_name()`**:
   - Tests formatting hyphenated and underscored strings to camelCase.
   - Tests `$allow_underscores = true` behavior preserving underscores.
   - Tests `format_type_name()` producing PascalCase output.

4. **`prepare_date_response()`**:
   - Tests formatting valid MySQL GMT dates to RFC3339.
   - Tests zero-date `'0000-00-00 00:00:00'` returning `null` to avoid invalid datetime outputs.
   - Tests the precedence of the local `$date` parameter when provided.

5. **`get_node_type_from_id()`**:
   - Tests returning `null` for numeric IDs.
   - Tests decoding Relay global IDs and resolving node type.
   - Tests invalid global ID handling.

6. **`get_allowed_wp_kses_html()`**:
   - Tests allowed HTML tags and attribute configurations for admin settings pages.

7. **`get_post_preview_id()`**:
   - Tests returning original post ID when no revisions exist (for both int and `\WP_Post` object inputs).
   - Tests returning the latest revision ID when revisions exist, asserting directly against the resolved preview ID.

8. **`get_database_id_from_id()`**:
   - Tests resolving integer and numeric string database IDs directly.
   - Tests decoding Relay global IDs to extract database IDs.
   - Tests invalid Relay global IDs and non-numeric encoded IDs returning `false`.

9. **`map_input()`**:
   - Tests non-array input returning an empty array.
   - Tests mapping input argument keys according to mapping configurations.
   - Tests skipping keys defined in `$skip`.
   - Tests automatic sanitization of string values and arrays of strings via `sanitize_text_field()`.

Additionally, this adds missing docblocks to existing tests in `UtilsTest.php` to ensure 100% PHPCS compliance (`check-cs`).

N/A

### Local Verification
- **WPUnit (Codeception):** Passed (11 tests, 59 assertions, 0 failures)
  ```
  Wpunit Tests (11) ------------------------------------------
  ✔ UtilsTest: Get query id (0.01s)
  ✔ UtilsTest: Map enum name to value (0.04s)
  ✔ UtilsTest: Format type name for wp template (0.00s)
  ✔ UtilsTest: Format graphql name (0.00s)
  ✔ UtilsTest: Format field name and type name (0.00s)
  ✔ UtilsTest: Prepare date response (0.00s)
  ✔ UtilsTest: Get node type from id (0.00s)
  ✔ UtilsTest: Get allowed wp kses html (0.00s)
  ✔ UtilsTest: Get post preview id (0.01s)
  ✔ UtilsTest: Get database id from id (0.00s)
  ✔ UtilsTest: Map input (0.00s)
  ------------------------------------------------------------
  OK (11 tests, 59 assertions)
  ```
- **PHPCS:** Passed (`0 errors, 0 warnings`)
- **PHPStan:** Level 8 Passed (`[OK] No errors`)


jasonbahl (changes requested):
Thanks for this @noruzzamans, really appreciate the extra coverage on `Utils`. I ran it locally and checked each assertion against the implementation, and they line up nicely, including the non-ASCII template fallback.

One thing to fix before we merge: in `testGetPostPreviewId`, the revision assertions never actually run. `wp_update_post()` already saves a revision (via the `wp_after_insert_post` hook), so the explicit `wp_save_post_revision()` call sees no changes and returns `null`, and the `if ( is_int( $revision_id ) )` block is skipped. That's why your run shows 45 assertions instead of 47. Could you drop the explicit call and the conditional, and assert against the resolved preview ID instead? Something like:

```php
$preview_id = \WPGraphQL\Utils\Utils::get_post_preview_id( $post_id );
$this->assertNotSame( $post_id, $preview_id );
$this->assertSame( $post_id, wp_is_post_revision( $preview_id ) );
$this->assertSame( $preview_id, \WPGraphQL\Utils\Utils::get_post_preview_id( get_post( $post_id ) ) );
```

Small note, not blocking: `get_database_id_from_id()` and `map_input()` are still uncovered, so maybe soften "complete coverage" in the description, or add those too if you're up for it.

noruzzamans:
Thanks so much for the review and great catch on the revision lifecycle, @jasonbahl!

I've addressed both points in https://github.com/wp-graphql/wp-graphql/pull/4321/commits/a74dd84b41972519dc98bdf64ec982e8be752b12:
1. **`testGetPostPreviewId`**: Dropped the redundant `wp_save_post_revision()` call and conditional, asserting directly against the resolved preview ID and verifying that `wp_is_post_revision()` matches.
2. **Added coverage for `get_database_id_from_id()` and `map_input()`**:
   - `testGetDatabaseIdFromId`: Covers numeric IDs, Relay global ID decoding, and invalid/non-numeric global IDs.
   - `testMapInput`: Covers key re-mapping, skipping unmapped/ignored keys, non-array fallback, and string/nested array sanitization via `sanitize_text_field`.

The suite now runs 11 tests with 59 assertions, and passes cleanly with 0 PHPCS errors and 0 PHPStan errors.

jasonbahl (approved):

jasonbahl (approved):

jasonbahl:
Thanks for the quick turnaround! 🙌

Looks good!

noruzzamans:
> Thanks for the quick turnaround! 🙌
> 
> Looks good!

Thanks you

PR #4333, opened 2026-09-15
fix(smart-cache): prevent a fatal error on option updates when WPGraphQL core is older than 2.18.0

### LINKED ISSUE #4297: Fatal error after smart cache plugin auto-updated and core didn't
The wp-graphql does not support auto-update but wpgraphql-acf and wpgraphql-smart-cache do. After wpgraphql-smart-cache updated this triggererd a fatal error.

wp-admin, wp-cron and the front-end all returned "There has been a critical error". WordPress's recovery-mode email also crashed on the same hook, so no recovery link was ever sent and the site could only be rescued via shell access.

Arslan-TR:
**Maintainer update:** The fatal in #4297 came from settings invalidation, not the version constant. Smart Cache 2.3.0 calls core's settings map without an argument that only became optional in WPGraphQL 2.18.0, so every option update failed on older core. This PR now includes that fix alongside the original constant sync.

---

The plugin header and readme.txt for WPGraphQL Smart Cache declare `Requires WPGraphQL: 2.0.0`, but the runtime compatibility gate (`WPGRAPHQL_SMART_CACHE_WPGRAPHQL_REQUIRED_MIN_VERSION` in `wp-graphql-smart-cache.php`) was still hardcoded to the older `1.12.0`.

Because of that drift, `can_load_plugin()` reported a site as compatible whenever its WPGraphQL core version was anywhere between 1.12.0 and 2.0.0, even though the plugin actually requires 2.0.0+. On such a site, Smart Cache would fully initialize (registering its document/cache classes) against a core it was not really compatible with, instead of bailing out and showing the "please update WPGraphQL" admin notice. That produced a site-wide fatal error when Smart Cache auto-updated ahead of an outdated core, as reported in #4297 (wp-admin, wp-cron, the front-end, and even the recovery-mode email hook all failed).

The fix syncs the runtime constant with the declared header (`2.0.0`), so an incompatible core is correctly caught by the existing `can_load_plugin()` gate and the admin notice is shown instead of fataling.

closes #4297

## Testing Strategy

1. **First commit**: Add `PluginRequiredVersionTest` asserting `WPGRAPHQL_SMART_CACHE_WPGRAPHQL_REQUIRED_MIN_VERSION` matches the `Requires WPGraphQL` header declared in the plugin file. This fails against the previous `1.12.0` constant.
2. **Second commit**: Bump the constant to `2.0.0`, making the test pass.

### Test Results

- [x] **Failing test added**: commit adds `test_required_min_version_constant_matches_declared_header`, which fails before the fix
- [x] **Fix implemented**: commit bumps the constant to `2.0.0`
- [x] **All tests passing**: no other logic changed; `php -l` passes on both files
- [x] **Regression tests**: the added test also asserts `can_load_plugin()` still returns `true` against the currently active (compatible) core

## Additional Context

This is a single-constant fix with no behavior change for sites already running a compatible (2.0.0+) core. Sites on an older, incompatible core will now correctly see the existing admin notice instead of a fatal error.


jasonbahl:
Hi @Arslan-TR, thanks for picking this up!

While reviewing I traced the fatal in #4297 to a different spot. Smart Cache 2.3.0 started calling DataSource::get_allowed_settings_by_group() without arguments on every option update, and that argument only became optional in WPGraphQL 2.18.0. So on core 2.17 and older, every option write threw an ArgumentCountError, which is why wp-admin, cron and the recovery email all broke. Syncing the constant was still the right call, since the header already said 2.0.0, but it didn't cover sites on 2.0 through 2.17.

Since we want this in the next Smart Cache release, I pushed a commit to your branch instead of asking for another round. The guard now checks that the method can be called without arguments and never lets an error escape the option update hook. There are regression tests that reproduce the ArgumentCountError, your drift test now reads the header with get_file_data(), and the persisted query loader is skipped when the plugin can't load. Your commits and authorship stay on this PR.

Thanks again for the contribution!

jasonbahl (approved):

PR #4334, opened 2026-09-15
chore: release wp-graphql-smart-cache as 2.3.2

jasonbahl:
Pins the pending Smart Cache release to 2.3.2. It was computed as 2.4.0 because #4231 (a wpgraphql.com feature) touched `plugins/wp-graphql-smart-cache/docs/docs_nav.json`, and release-please attributes commits by path. Nothing in the plugin's behavior changed, and the security fix in #4332 has been communicated to reporters as 2.3.2.

Root-only change, so it adds no changelog lines to any component. The `release-as` key gets removed in a follow-up once 2.3.2 ships.


jasonbahl:
Closing as unnecessary. After #4335 excluded plugin docs directories from release-please, the pending Smart Cache release regenerated as 2.3.2 on its own, since the only feature commit attributed to it was the docs-only #4231. No version pin needed.

